PR #1, opened 2018-09-29
Record dependency information when executing queries

matklad:
Hi!

I really would love to use this inside rust-analyzer prototype, but to do that, I need at least a bare-bones support for incrementality, and I'd love to help with it if I can :) 

As a first step, and just to familiarize myself with existing code, here's a simple dependency tracking implementation!


nikomatsakis:
I am going to close this in favor of https://github.com/nikomatsakis/salsa/pull/4, which offers complete incremental (if bare-bones) support

matklad:
Sure, haven’t seen that branch at first!

PR #4, opened 2018-09-30
add incremental support

nikomatsakis:


nikomatsakis:
@matklad 

> Interesting! I think we can't quite make this assertion here, because the revision can be updated concurrency with query execution. This happens if, for example, user types stuff, we start type checking code, and in the meantime user types more stuff, while type checking is still in progress.

I had planned to have the active queries simply block until active queries complete. I don't think it is worth the complexity to support anything more than that.

nikomatsakis:
Well, let me back that up a bit.

This is what I expecteded to start:

- Query computation runs async from main input (perhaps we'll use a Future setup or something)
- When user input arrives, we process it after current query completes

But I definitely see value in interrupting queries that are producing outdated results. I'm not sure the best way to integrate that.

I've been wondering if the system ought to move towards a `Future`-based system: that would make this "backpressure" relatively easy, but it would force us into async-await land, and I didn't want to do that to start.

matklad:
Yep, that seems reasonable! Though, it seems to me that handling cancellation is important for IDE use-case. Will write some use-cases as an issue to think about later today.

matklad:
>I've been wondering if the system ought to move towards a Future-based system: that would make this "backpressure" relatively easy, but it would force us into async-await land, and I didn't want to do that to start.

I'd rather avoid futurizing the query engine: we probably don't need too many concurrent tasks running, so "lightweight threading" aspect won't be useful to us. On the other hand, asnyc does make debugging and profiling harder. For rust-analyzer, I use crossbeam-channel as a selectable blocking future.

PR #5, opened 2018-09-30
support "inputs" for incremental computation

nikomatsakis:
You can now define an "input" query:

```rust
salsa::query_definition! {
    crate Input1: Map<(), usize>;
}
```

This permits you to use the `set` method (e.g., `query.input1().set(K, V)`) -- but only outside of query evaluation. These things are integrated into incremental support.


nikomatsakis:
Note that I renamed `of` to `get`, too.

Issue #6, opened 2018-10-01
Cycle recovery


nikomatsakis:
So handling cycles has traditionally been a thorny issue. Right now the code panics in the event of a cycle. We probably want this to stay an option, but it's probably not the best choice long term — particularly since it often happens that the order of queries is determined by user input (and hence that cycles are somewhat out of our control).

This issue is focused on the case where a **cycle represents an error**. I'm going to open a sister issue to focus on the case of "succesful cyclic computations".

One challenge with cycles is that they can easily lead to non-determinism. For example, imagine that we return a `Result` — if that is not faithfully propagated up the entire cycle, that implies that `foo.query().get(K)` could sometimes yield `Ok` (the root call) and other times yield `Err` (the cyclic call). This somewhat invalidates all the memoized results in the middle, since they are dependent upon the `Err` result. But if we had executed things in another order, things might have turned out differently. This is undesirable. It's particularly bad if we cache such results.

nikomatsakis:
One option is to walk the stack whenever we encounter a cycle and mark all the participants in some way. We could then avoid caching their return values. Or, even more extreme, we could ignore their results altogether and just propagate back dummy return values. (That is, the query can finish executing as normal, but its return value is ignoring, and some kind of default or cycle value is propagated back. Obviously this is related to the idea of terminating unwanted queries early.)

To just skip caching would be relatively easy. Each active query would carry some `AtomicBool` and we'd flip the flags to true when we detect a cycle. When the query finished, we'd propagate back this result so that the storage cells can know what to do.

If we wanted to propagate back errors, we have to extend the system slightly to know what value to return. You could *imagine* this working like so: in order to recover from a cycle, a query must have a defined return value like `Result<T, CycleError>`. If all the participants in the cycle have such a return value, and a cycle occurs, we'll force them all to be `Err`. You could also imagine that — in order to recover from a cycle — the query's return value must implement some trait (like `Default`, but probably not `Default`), or that the query must write a "cycle hook" defining what return value to produce.

I'm a bit concerned that having to "opt in" to cycle handling will error prone, though. Perhaps just "no cache" is a better starting point.

matklad:
Here's the bit of IntelliJ that deals with this issue: https://github.com/JetBrains/intellij-community/blob/9d69a9c4245956ce063b20b1bf99165422dfefc2/platform/util/src/com/intellij/openapi/util/RecursionManager.java#L17-L37

lpil:
[...]

PR #13, opened 2018-10-01
Detect volatile cycles

### LINKED ISSUE #2: Track cycles via `Runtime` instead of storage?
Currently, we detect cycles by storing `QueryState::InProgress` inside of `MemoizedStorage`.

I think this could be problematic for two reasons:

* we won't detect cycles for transparent queries
* if `MemoizedStorage` is shared by concurrently executing queries, we can get `CycleDetected` if two threads execute the same query simultaneously. 

I think that perhaps we need to move cycle tracking from storage to `Runtime`? The fix I have in mind is

* remove `QueryState` and use `map: RwLock<FxHashMap<Q::Key, Q::Value>>` in the storage,
* add `in_progress: FxHashSet<QC::QueryDescriptor>,` to `Runtime`, which stores the same info as `execution_stack`, but with an O(1) "contains" check.

matklad:
closes #2

### DIFF
--- a/src/volatile.rs
+++ b/src/volatile.rs
@@ -5,8 +5,8 @@ use crate::QueryContext;
 use crate::QueryStorageOps;
 use crate::QueryTable;
 use log::debug;
-use parking_lot::{RwLock, RwLockUpgradableReadGuard};
-use rustc_hash::FxHashMap;
+use parking_lot::Mutex;
+use rustc_hash::FxHashSet;
 use std::any::Any;
 use std::cell::RefCell;
 use std::collections::hash_map::Entry;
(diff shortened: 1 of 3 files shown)


src/volatile.rs
```diff
@@ -31,12 +50,18 @@ where
...
+            return Err(CycleDetected);
+        }
+
         // FIXME: Should we even call `execute_query_implementation`
         // here? Or should we just call `Q::execute`, and maybe
         // separate out the `push`/`pop` operations.
         let (value, _inputs) = query
             .salsa_runtime()
             .execute_query_implementation::<Q>(query, descriptor, key);
+        let was_in_progress = self.in_progress.lock().remove(key);
```
matklad: Note that, if the previous line panics due to cycle, we won't remove the key from the progress, effectively poisoning query forever. I expect that is the part of #6? Memoized storage seems to have a similar problem.

PR #14, opened 2018-10-01
Add support for "dependency only" tracking

nikomatsakis:
This is closer to what Glimmer does.

### DIFF
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -1,0 +1,247 @@
+use crate::runtime::QueryDescriptorSet;
+use crate::runtime::Revision;
+use crate::runtime::StampedValue;
+use crate::CycleDetected;
+use crate::Query;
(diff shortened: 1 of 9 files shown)


matklad:
LGTM! We might also add a mix between dependecy-only and memoized storage, which stores Option<V> and allows to gc only the values.

nikomatsakis:
@matklad interesting thought, maybe file an issue? that might also just be how memoized works, full stop?

nikomatsakis:
I wanted to get this merged, anyhow, because I realized that supporting it would require threading a bit more information through (e.g., now we have to return the changed-at information).

matklad:
>that might also just be how memoized works, full stop?

Yep, and also https://github.com/nikomatsakis/salsa/issues/7#issuecomment-425893534. Filing the issue.

PR #19, opened 2018-10-01
Assert inputs are not modified concurrently

matklad:
This public fn dodged the assertion before: https://github.com/nikomatsakis/salsa/blob/e18f14fca5eefc4f104cca85dba3256e6811f9f2/src/lib.rs#L165

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -62,10 +62,6 @@ where
     /// case, you can wrap the input with a "no-storage" query and
     /// invoke this method from time to time.
     pub fn next_revision(&self) {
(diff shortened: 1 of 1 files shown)


src/runtime.rs
```diff
@@ -78,6 +74,10 @@ where
 
     /// Increments the current revision counter and returns the new value.
     crate fn increment_revision(&self) -> Revision {
+        if !self.local_state.borrow().query_stack.is_empty() {
```
nikomatsakis: D'oh, good catch.

PR #21, opened 2018-10-02
change `query_prototype` to take entire trait; update the docs

nikomatsakis:
Took a first stab at a readable introduction. I realized I think I want to make a little mdbook but that is for another day.


nikomatsakis:
My motivations for changing to take the entire trait are:

- First, and most importantly, you can't actually have any other members besides queries, or else the `query_context_storage!` macro will not work properly. So the current setup seemed misleading.
- Second, I want to start implementing `QueryContext` for `&impl QueryContext` and hence I want the macro to be able to generate similar impls for all of your query context traits. The motivation is that it would be useful to be able to write structs that wrap a query context merely be being generic over `<Q: QueryContext>` and not by having to be generic over `<'q, Q: QueryContext>` (so that they can store a `&'q Q`). But I've not actually tried to do this yet. (In fact, one could even give the query functions a `impl QueryContext + Copy` argument instead of a `&impl QueryContext` argument.)

Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
 name = "salsa"
-version = "0.1.1"
+version = "0.2.0"
 authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 edition = "2018"
 license = "Apache-2.0/MIT"
```
matklad: I think `Apatch-2.0 OR MIT` is the correct syntax.

matklad: # Multiple licenses can be separated with a `/`, although that usage
# is deprecated.  Instead, use a license expression with AND and OR
# operators to get more explicit semantics.

nikomatsakis: Huh, really? I've been using this forever...

nikomatsakis: > Multiple licenses can be separated with a `/`, although that usage is deprecated.  Instead, use a license expression with AND and OR operators to get more explicit semantics.

TIL, ok.

nikomatsakis: Still feels ambiguous to me though -- is it "OR" from the pov of a consumer or a provider. =) Ah well.

README.md
```diff
@@ -4,30 +4,45 @@
...
+  whenever you like.
+- **Functions**: pure functions (no side effects) that transform your
+  inputs into other values. The results of queries is memoized to
+  avoid recomputing them a lot. When you make changes to the inputs,
+  we'll figure out (fairly intelligently) when we can re-use these
+  memoized values and when we have to recompute them.
+
+## How to use Salsa in three easy steps
+
+Using salsa is as easy as 1, 2, 3...
```
matklad: That should have `</sarcasm>` tag perhaps? Unless you already know how it is supposed to work, it's pretty hard to decipher what those macros actually do :)

examples/hello_world/README.md
```diff
@@ -0,0 +1,172 @@
...
+implement the `QueryContext` trait to tell salsa where to find this
+runtime:
+
+```rust
+#[derive(Default)]
+struct QueryContextStruct {
+    runtime: salsa::runtime::Runtime<QueryContextStruct>,
+}
+
+impl salsa::QueryContext for QueryContextImpl {
```
matklad: Should be `QueryContextStruct`?

Issue #18, opened 2018-10-01
Merge DependencyStorage and MemoizedStorage?


matklad:
Currently, we have `MemoizedStorage` which, for each query, remembers a list of dependencies and the result value, and `DependencyStorage`, which remembers only dependencies (to save storage). 

They could be merged into an Storage which remembers deps and "optional" result. That way, we'll be able to implement smart strategies for evicting large unused results (something like LRU).

cc https://github.com/nikomatsakis/salsa/issues/7#issuecomment-425893534 and https://github.com/nikomatsakis/salsa/pull/14

nikomatsakis:
Yeah, the fact that I started one by copying-and-pasting the other is usually a sign as well that a merge would make sense. =)

matklad:
@nikomatsakis should I give it a shot?

The plan is, roughly:

* rename `DependencyStorage` to `WeakMemoizedStorate`
* change impl to store `Option<Value>`, and don't store values by default
* (???) add some kind of gc hook to allow the client to control if a value should be preserved or not.
* (???) dedupe `DependencyStorage` and `WeakMemoizedStorage` by using some value-level or type level parallelization.

nikomatsakis:
@matklad Sure, though I'd leave the gc hook for later =) I had expected though to merge `DependencyStorage` and `MemoizedStorage` ultimately. If we want to statically configure things, we could have them be parameterized by some type of `StoragePolicy` type that configures the various bits.

e.g., when we have invoked the fn and we have a value, we could use `storage_policy.create_memoized_value(&user_value, changed_at)` and it would return a `Option<V>` (`None`, in the case of dependency only).

matklad:
>and it would return a Option<V> (None, in the case of dependency only).

I was thinking about it returning a `Result<V, ()>` or `Result<V, Void>`, depending on the policy, but that is probably over engineering. Will look into this around the weekend then!

PR #33, opened 2018-10-05
add a `set_unchecked` method that can be used to do mocking in tests

nikomatsakis:
This PR adds a `set_unchecked` method available for input and memoized queries. (I would expect that after https://github.com/salsa-rs/salsa/issues/18 is fixed, it would work for all queries.) It permits one to forcibly **set** a query's value without running the code. This is useful for mocking purposes in writing tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "salsa"
-version = "0.4.0"
+version = "0.4.1"
 authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 edition = "2018"
 license = "Apache-2.0 OR MIT"
--- a/src/input.rs
+++ b/src/input.rs
@@ -8,6 +8,7 @@ use crate::Query;
 use crate::QueryDescriptor;
 use crate::QueryStorageOps;
 use crate::QueryTable;
+use crate::UncheckedMutQueryStorageOps;
 use log::debug;
 use parking_lot::{RwLock, RwLockUpgradableReadGuard};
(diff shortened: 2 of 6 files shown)


nikomatsakis:
Impatient, so just gonna merge =)

We can bikeshed the name at some point though. I debated between `mock`, `set_unchecked`, and `mock_unchecked`.

nikomatsakis:
just realized travis as broken, fixed it I think

matklad:
LGTM, but a couple of questions:

Because we use empty inputs, set unchecked is forever, I think? This seems like it could be surprising, but maybe not?

Should we have an issue for the general testing infrastructure? I expect it would be useful to get some sort of the trace of the query, to make sure it does’t eval too much after invalidation, for example?

nikomatsakis:
@matklad 

> Because we use empty inputs, set unchecked is forever, I think? This seems like it could be surprising, but maybe not?

I was wondering about it too. I was imagining that one would use this primarily in a "revision-less" fashion -- you would basically mock up some inputs and then test that you get the output you expect. But I can also imagine wanting to do fancier tests, where you mock up some inputs, test that you get the output you expect, and then *alter* the mocked inputs, and test (e.g.) that the output doesn't wind up being recomputed. This kind of bleeds into your next question...

> Should we have an issue for the general testing infrastructure? I expect it would be useful to get some sort of the trace of the query, to make sure it does’t eval too much after invalidation, for example?

...I agree that could be useful, yes. I don't really know what that would look like or what we will want 100% yet. I would think it's best to see what we find we want while iterating.

PR #36, opened 2018-10-06
Merge Memoized and Dependency storages

### LINKED ISSUE #18: Merge DependencyStorage and MemoizedStorage?

matklad:
closes #18


src/memoized.rs
```diff
@@ -18,16 +18,56 @@ use std::fmt::Debug;
...
 where
     Q: QueryFunction<DB>,
     DB: Database,
+    M: ShouldMemoizeValue<DB, Q>,
 {
     map: RwLock<FxHashMap<Q::Key, QueryState<DB, Q>>>,
+    m: PhantomData<M>,
+}
+
+pub trait ShouldMemoizeValue<DB, Q>
```
nikomatsakis: Nit: I think I would prefer to call this `MemoizationPolicy`

src/memoized.rs
```diff
@@ -18,16 +18,56 @@ use std::fmt::Debug;
...
+/// "Dependency" queries just track their dependencies and not the
+/// actual value (which they produce on demand). This lessens the
+/// storage requirements.
+pub type DependencyStorage<DB, Q> = WeakMemoizedStorage<DB, Q, NeverMemoizeValue>;
+
+pub struct WeakMemoizedStorage<DB, Q, M>
 where
     Q: QueryFunction<DB>,
     DB: Database,
+    M: ShouldMemoizeValue<DB, Q>,
```
nikomatsakis: Then we can rename `M` to `MP` for "memoization policy"

src/memoized.rs
```diff
@@ -105,16 +152,22 @@ where
                             key,
                             m.verified_at,
                         );
-
+                        // We've found that the query is defenitelly up-to-date.
```
nikomatsakis: Nit: "definitely"

src/memoized.rs
```diff
@@ -129,25 +182,29 @@ where
         // first things first, let's walk over each of our previous
         // inputs and check whether they are out of date.
         if let Some(QueryState::Memoized(old_memo)) = &mut old_value {
-            if old_memo.inputs.iter().all(|old_input| {
-                !old_input.maybe_changed_since(db, old_memo.stamped_value.changed_at)
-            }) {
+            if old_memo
```
nikomatsakis: I don't think there is any point to doing this loop unless `old_memo.value.is_some()`.

src/memoized.rs
```diff
@@ -18,16 +18,56 @@ use std::fmt::Debug;
...
 /// none of those inputs have changed.
-pub struct MemoizedStorage<DB, Q>
+pub type MemoizedStorage<DB, Q> = WeakMemoizedStorage<DB, Q, AlwaysMemoizeValue>;
+
+/// "Dependency" queries just track their dependencies and not the
+/// actual value (which they produce on demand). This lessens the
+/// storage requirements.
+pub type DependencyStorage<DB, Q> = WeakMemoizedStorage<DB, Q, NeverMemoizeValue>;
+
+pub struct WeakMemoizedStorage<DB, Q, M>
```
nikomatsakis: I'd probably just call this `MemoizedStorage` and remove the type aliases, then adjust the macro — we don't really expect people to write these names by hand, do we? But these names / aliases are ok. Whichever you prefer I guess.

nikomatsakis: I've been thinking about eventually merging `VolatileStorage` into this struct, such that it will probably want to be called `DerivedStorage` (vs `InputStorage`), with the "policy" determining other details.

PR #43, opened 2018-10-09
Combine memoized and volatile to make "derived storage"

nikomatsakis:
These are a series of refactorings working towards a model of cancellation. I didn't get all the way today, but I still reached a decent stopping point: this PR combines memoized and volatile storage to have the idea of "derived" storage. Volatile storage just becomes a special kind of read -- an **untracked read**. Untracked reads basically indicates that there are dependencies beyond those that we can make sense of.

This refactoring makes one user-visible change (beyond some renamings): Volatile queries now cache their results. This isn't a *necessary* change, but it does ensure *consistency* .

This refactoring also enables queries to introduce "untracked reads" at any time. I plan to use this as part of a cancellation plan (when a query result is *cancelled*, it becomes treated as having untracked inputs, thus ensuring it will be re-executed in the next revision).

I also introduce the `ChangedAt` enum — the intention is to permit inputs to be set to **constant**, which means that you promise they will "never, ever" change. This can propagate to derived queries if they consume only constant inputs, allowing them not to track their inputs. (Example use case: crates.io sources.) This is totally orthogonal but just something I was working towards.


src/derived.rs
```diff
@@ -46,6 +54,8 @@ where
     DB: Database,
 {
     fn should_memoize_value(key: &Q::Key) -> bool;
+
+    fn is_volatile(key: &Q::Key) -> bool;
```
matklad: Let's add a doc-comment explaining what `volatile` means? Or perhaps just rename to `should_track_dependencies`?

src/derived.rs
```diff
@@ -370,16 +410,44 @@ where
...
         );
     }
 }
+
+impl<DB, Q> Memo<DB, Q>
+where
+    Q: QueryFunction<DB>,
+    DB: Database,
+{
+    fn verify_memoized_value(&self, db: &DB) -> bool {
```
matklad: For a tiny-bit less unwrappy API we can return an `Option<Q::Value>` instead of `bool` here.

nikomatsakis: I considered that — I did not want to "consume" the memo entirely, but maybe that's silly. I could play around with it.

src/derived.rs
nikomatsakis: Sounds good..

src/derived.rs
matklad: We are cloning the value if it is valid, so we can clone it inside this method.

nikomatsakis: OK, done

PR #45, opened 2018-10-09
Constants

nikomatsakis:
This branch is incomplete -- it has no tests -- but I wanted to sketch out how I thought constant input would work.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -91,8 +91,26 @@ where
         // (Otherwise, someone else might write a *newer* revision
         // into the same cell while we block on the lock.)
         let changed_at = changed_at();
+        let stamped_value = StampedValue { value, changed_at };
 
-        map.insert(key, StampedValue { value, changed_at });
+        match map.entry(key) {
+            Entry::Occupied(mut entry) => {
+                assert!(
+                    !entry.get().changed_at.is_constant(),
+                    "modifying `{:?}({:?})`, which was previously marked as constant (old value `{:?}`, new value `{:?}`)",
+                    Q::default(),
+                    entry.key(),
+                    entry.get().value,
+                    stamped_value.value,
+                );
+
+                entry.insert(stamped_value);
+            }
+
+            Entry::Vacant(entry) => {
+                entry.insert(stamped_value);
+            }
+        }
     }
 }
 
@@ -151,6 +169,13 @@ where
         self.set_common(key, value, || {
             ChangedAt::Revision(db.salsa_runtime().increment_revision())
         });
+    }
+
+    fn set_constant(&self, _db: &DB, key: &Q::Key, value: Q::Value) {
+        // FIXME. One weirdness: if this previously had a value, but
+        // was not marked as constant, we will not "lower" the
+        // `ChangedAt` rating with the code as it is.
+        self.set_common(key, value, || ChangedAt::Constant);
     }
 }
 
--- a/src/input.rs
+++ b/src/input.rs
@@ -70,6 +70,30 @@ where
             changed_at: ChangedAt::Revision(Revision::ZERO),
         })
     }
+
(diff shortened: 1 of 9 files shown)


src/input.rs
```diff
@@ -70,6 +70,48 @@ where
             changed_at: ChangedAt::Revision(Revision::ZERO),
         })
     }
+
+    fn set_common(&self, key: &Q::Key, value: Q::Value, changed_at: impl FnOnce() -> ChangedAt) {
+        let map = self.map.upgradable_read();
```
matklad: Why do we need two-phase read/write locking here? I think that setting has to be exclusive, so we can just grab the write lock?

src/input.rs
```diff
@@ -70,6 +70,48 @@ where
...
+        // (Otherwise, someone else might write a *newer* revision
+        // into the same cell while we block on the lock.)
+        let changed_at = changed_at();
+        let stamped_value = StampedValue { value, changed_at };
+
+        match map.entry(key) {
+            Entry::Occupied(mut entry) => {
+                assert!(
+                    !entry.get().changed_at.is_constant(),
+                    "modifying `{:?}({:?})`, which was previously marked as constant (old value `{:?}`, new value `{:?}`)",
```
matklad: So, if want to change a constant (say, the user has run a `cargo update`), we need to tear-down runtime completely and re-insert all the inputs, right? An alternative would be to just clean all the derived storages.

src/input.rs
nikomatsakis: Revised. In the process, I also fixed some latent bugs. See latest commit.

src/input.rs
matklad: I wonder if we’ll need some kind of batch modification API for that, to set several constants in one go, bit clean the store only once.

nikomatsakis: Hmm, interesting. It feels like it'd be nice in general to make it so that

```rust
input.set(k1, v1);
input.set(k2, v2);
```

was smart enough not to increment the revision twice. The idea would be to track something like whether the revision is "unobserved" (no derived queries are based on it).

nikomatsakis: Seems like we should file an issue on it.

nikomatsakis: Thinking more about this, seems silly. We should just handle overflow some day if it ever happens. Until then, increments are free =)

nikomatsakis:
Added a first rudimentary test. Have to think how to test this though — I may have to add some sort of accessors =)

(Is there a way for your tests to request the library to be built with non-default-features enabled?)

nikomatsakis:
Perhaps some way to query when a given query last changed would be an ok thing to offer, though

matklad:
>Have to think how to test this though — I may have to add some sort of accessors =)

I think clients of the library will also want to write tests like "asking this query after this modification won't trigger bazilion of over queries". Perhaps we need to give access, for tests, to some kind of "query trace" API? That is, a `.get` method, which returns a pair of `(Q::Value, QueryTrace)`, where trace is a list of queries which were evaluated and/or freshened during the evaluation of top-level query.

nikomatsakis:
@matklad perhaps, but I think that is not sufficient to debug what is happening here (e.g., you can readily implement such introspection by hand, and indeed I have done so, but you can only find out if the query actually *executed*). With constants, I really want to be able to ask for a bit more detailed state (e.g., do you think that this query depends on that one?). It seems like some sort of debugging API is the way to go though — I'm just not 100% sure what it should tell you back.

I guess, to start, that if the "trace" included which queries had to be "reverified" that would be good.

nikomatsakis:
OK, I added `is_constant` and wrote some tests. Sure enough, I found a bug — even for constants, we still need to track a revision, so that we can tell *when* the value became a constant. Fixed that and I think things are working properly now.

PR #48, opened 2018-10-12
Cancellation and parallel queries

nikomatsakis:
This is a WIP branch that implements *most* of cancellation and parallel queries. The history is still a bit messy and I've not implemented cycle detection yet, but hopefully you get the basic idea.

The branch introduces two key public APIs:

- You can `fork` a salsa runtime -- and your database can implement a `ParallelDatabase` trait that offers a `fork` method.  The intention is that the other forked database should be sent to another thread (not doing so can result in deadlock).
- We now support concurrent queries and operations on inputs. If you execute `set` while some other thread is actively executing a query, you will block waiting for them to finish, but you will also signal to them that they should terminate by setting an internal flag.
    - Queries can then poll the `is_current_revision_cancelled` method. If they observe a true value, then they know that their results no longer matter, and they can go ahead and return whatever value they deem appropriate. Those results will be recomputed in the next revision.

My intention is that these are the basic primitives and that atop them we would add convenient ways to do parallel execution and cancellation:

- Cancellable queries, that automatically return a `Cancallable::cancel` value of some kind.
- Ways to apply a query to many keys in parallel.
- Futures to let a query execute concurrently and later demand its value.

The last two will require figuring out some sort of thread-pool story -- I'd prefer if we don't bake one into salsa, which is why the current building blocks don't actually manage the threads themselves.


nikomatsakis:
Oh, I also expected to add some sort of wrapper for convenient cancellable queries. I imagined that we'd have a trait (`CancelledValue`) sort of like `Default` -- if you tag a query as "cancellable", then we will check for cancellation on entry and return `CancelledValue::default`, something like that. 

We would implement `CancelledValue` for `Result<T, Cancelled>` as well.

src/runtime.rs
```diff
@@ -66,21 +80,98 @@ where
...
+    /// ignored) your query is free to shortcircuit and return
+    /// whatever it likes.
+    #[inline]
+    pub fn is_current_revision_canceled(&self) -> bool {
+        let pending_revision_increments = self
+            .shared_state
+            .pending_revision_increments
+            .load(Ordering::SeqCst);
+        if pending_revision_increments > 0 {
+            self.report_untracked_read();
```
matklad: Hm, will this be "transitive"?

I am thinking about a scenario where we try to revalidate query A, which executes (for validation) query B. Query B is then checks `is_current_revision_canceled`, sees that it is indeed cancelled and returns a dummy value, *which happens to match its previous value*. We then set `verified_at` to the current revision for `A`, which is not entirely correct. Maybe it's OK though?

nikomatsakis: I think that's OK, but let's talk through it. 

So cancellation occurs in revision R1. Query A will be marked as verified-at for the current revision R1 (as will Query B). Then we will increment the revision to go to R2. Now we fetch Query A again -- it will check its inputs. Query B (which has an untracked read) will see that it had an untracked read and report itself as out of date. Query B will thus re-execute. *If it produces the same (dummy) value again even without cancellation*, then Query A will indeed not re-execute: but that's ok, it would have produced the same value. If it produces something different, then Query A should re-execute as normal.

nikomatsakis:
OK, so, this PR now correctly (fingers crossed) handles parallel execution, with a blocking semantics. Things got a bit hairier than I imagined, and I'm not entirely sure if this is the best overall setup. I *know* it's not good for dependency-only queries.

This is the general strategy right now:

- We use a placeholder to prevent the same query/key from being evaluated more than once at a time, and we detect cycles across threads.
- We allow two threads to independent verify whether a query has changed without blocking.

For dependency-only queries, this setup isn't great. Blocking doesn't make any sense because after you are done blocking there is no memoized value to reuse. In that scenario, you really want to just start executing a second copy which (in theory, anyway) should be fine.

That said, it seems like having the **ability** to block is important. One can imagine there being semi-expensive queries where it doesn't really make sense to do the work twice. (Of course, side-effects are also possible, but I'd like to offer some principled mechanism for side-effects where you create a struct that encapsulates the action and you hand it over to the runtime; in this case, the runtime will discard values from the second fork.)

Overall, I think the pieces in this PR are ultimately probably needed. I'd be happier probably landing it in smaller pieces, but I'm not sure yet how best to carve it up. It all feels interdependent.

src/runtime.rs
matklad: I was thinking about a slightly different scenario, when B is canceled during verification (that is, when cancellation happens at R2).

Let's say we verified both A and B at revision R1. Then, revision was advanced to R2, and we started to verify A again. When we were re-evaluating B at R2, cancellation occurred, so B returned a dummy value which happened to match that of R1 revision. Then, we mark A as verified at R2.

nikomatsakis:
Something else I don't like: in derived queries, the `verified_at`, `changed_at`, and `inputs` all feel like they are storing semi-overlapping information that I'd like to setup more cleanly. For constants in particular they all become semi-irrelevant somehow (though not entirely). But there is some subtlety to capture there (e.g., we need to track *when* something became constant, and we may have a "changed-at" that is older than our inputs, so we need to track "verified-at" separately).

src/runtime.rs
nikomatsakis: OK, seems fine, since we don't care about the value of A at R2 (we cancelled R2).

nikomatsakis: The question is: what happens to A in *R3*?

nikomatsakis: In that case, when we re-demand A, it will go check its inputs, and B will get re-executed, etc.

nikomatsakis: Put another way, I think it is correct not to re-execute A in R2 in that scenario, because its inputs did not change. But we ought to be sure that we re-execute A in R3 (presuming the result of B changes this time).

src/derived.rs
```diff
@@ -46,6 +44,18 @@ where
 {
     map: RwLock<FxHashMap<Q::Key, QueryState<DB, Q>>>,
     policy: PhantomData<MP>,
+
+    /// This cond var is used when one thread is waiting on another to
+    /// produce some specific key. In that case, the thread producing
+    /// the key will signal the cond-var. The threads awaiting the key
+    /// will check in `map` to see if their key is present and (if
+    /// not) await the cond-var.
+    signal_cond_var: Condvar,
```
matklad: This is a per-table cond-var, so we potentially wake to many threads, right?

Curious, would it be easier/more efficient to implement per-key notification? That is, could we add some struct which a thread can wait on to the `InProgress` variant?

For example, I think we can use crossbeam-channels for this sort of notification: we store an `Receiver<Void>` in the `InProgress` struct:

```
InProgress {
  id: RuntimeId,
  ready: Receiver<Void>,
}
```

When we store the in-progress placeholder, we create a Sender/Receiver pair and make sure to drop the Sender only after we've overwritten the placeholder with value.

When we read a placeholder, we do `ready.recv()` to block until the results are ready. 

That way, all subtle notification and waiting logic is delegated to crossbeam channel. 


In other words, these signals, mutexes and atomic variables add an entirely new dimension of subtleness, and it would be cool to abstract them away with something more high-level :)

nikomatsakis: However, I don't think that the setup you described will work: you would need a channel per receiver, from what I can tell. 

What I was considering was having a `Mutex<Vec<Receiver<T>>>` in the `InProgress`, initially empty (so cheap to construct). Receivers would grab the mutex and insert themselves, then release all locks and wait for a message. 

After completing, senders would unwrap the mutex, extract the vector, and send clones to the value. 

This also handles the "no memoized value" case.

matklad: >you would need a channel per receiver, from what I can tell.

If we want to use `Receiver<T>`, that is, to send an actual value, then yes, we need several channels, because they don't have broadcast. 

However, I was suggesting using channels *only* for synchronization, using `!` as a message type. That way, you'd have n receivers, one sender, and, when the sender is dropped, all receivers are notified, but have to look up the actual value in the map themselves. 

,
Agree though that this approach would have to allocate for every re computation of a query which, on one hand, seems sub optimal, but on the other hand, we probably will allocate an Arc for the final result anyway.

matklad: Hm, I think we can tweak the channel approach so that we allocate the channel only when the blocking actually happens? That is, we can change `InProgress` to something like

```
InProgress {
    id: RuntimeId,
    ready: Option<(Sender<!>, Receiver<!>)>,
}
```

The first thread to see `InProgress` then allocates the channels, subsequent threads then clone the receiver, and the notification happens naturally when the `InProgress` value is dropped. Not sure that these additional logic makes this solution more complex than the current one.

nikomatsakis: OK, had a different (but similar) idea that I will experiment with later

nikomatsakis:
@matklad I pushed an update -- I wound up just using a `Mutex<SmallVec<>>`. I think it's a decent approach personally -- you pay nothing until somebody is waiting, and they just pay to allocate a single channel.  In any case, the code is definitely nicer this way.

I am now reminded that I never 100% finished with the `maybe_changed_since` impl -- in particular, it is overly conservative with an `InProgress` placeholder. It ought to register as a blocker and wait. I was hoping to re-use `probe` for this, but poking at it I found it wasn't as easy as I'd hoped.

nikomatsakis:
I think not worth it, but I ran out of time to cleanup a bit more.

nikomatsakis:
@matklad I've debated about it, though they're somewhat different now. For example, `maybe_changed_since` doesn't insert placeholders (though it winds up invoking `read` for memoized queries, which does). 

Regardless, it definitely feels like there is some cleanup possible -- this is what I was complaining about earlier with the `verified_at` etc too, I think.

Another example is that `maybe_changed_since` currently gets sent a clone of the value when it blocks on another thread, but it only really needs the revision. That'd be easy enough to fix by having two vectors, one for clients who need the whole value, and one that just needs the revision, but I didn't have time to do those edits.

nikomatsakis:
OK, I *think* the code is landable now, though it's not got nearly as many tests as I would like, and I agree it could use more cleanup. 

Thoughts?

matklad:
I don't have time for a super thorough review right now, but this looks good to merge. In general, I think it's better to merge early and fix up if necessary, rather than to keep PRs in the queue.

nikomatsakis:
Guess I'll merge for now.

PR #57, opened 2018-10-18
Issue 53 frozen revision

### LINKED ISSUE #53: Freezing revision for several consequitive queries
Currently, we track that revision is frozen while a single query is in progress, but we don't have a way to freeze the revision for consecutive queries. I think this might be important for correctness for clients which issue multiple queries for a single high-level op.

For example, IDE might ask for types (one query) and control flow info (other query) for syntax highlighting. If a revision bump happens in-between, the end result might be an inconsistent highlighting. 

One way to solve this is to turn "highlighting" itself into a query.

Another way is to place onus on consumers to check that revision haven't advanced before merging results of the consecutive queries.

The third way is to provide some explicit API for freezing revisions. This could work with a drop-guard style API, and means turning [this dynamic choice](https://github.com/salsa-rs/salsa/blob/a128861a2cdf28145e1dd365dac408d76f34b6b6/src/runtime.rs#L97-L99) of `None` vs `Some` into a static one.

nikomatsakis:
Creates a `with_frozen_revision` method.

Fixes #53.


nikomatsakis:
Some questions: 

- Use closure or expose the RAII? See comment on #53 for notes
- Different name?
- Expose just on the runtime? Also on the database trait? Just on the database trait (it can use crate-private things in its default methods...)?

matklad:
Have thoughts, but don't have time :) Perhaps something lower-level like `fn freeze_revision() -> ()`, `unfreeze_revision() -> ()` would be even better? I fear that closures/RAII with lifetime params would make difficult to abstract over the API.

Ideally, on the high-level in the LSP, I'd love to do 

```
let ctx = state.current_snapshot()
```

and then send the ctx to another thread and do queries against it. It'd be cool if ctx was `'static` and not bound to a scope. With freeze/unfreeze I can implement RAII for ctx myself. WIth closures, not sure

matklad:
SGTM!

nikomatsakis:
I'm going to call it `lock_revision`, since it acts exactly like a read-write lock.

One (relatively minor) thing: the value returned by `lock_revision` would not mark us as having a "query in progress" (because that is thread-local state). That means that if you invoke `set` on a thread that holds this , you will get a deadlock (vs the current API, which would `panic`). Seems obvious enough though (which is partly why I wanted to choose the name `lock`.

nikomatsakis:
@matklad done

nikomatsakis:
You can basically ignore the 1st commit. I should probably edit it out of the history...

Issue #56, opened 2018-10-18
support volatile queries without equality comparisons


nikomatsakis:
Currently, every "value" must implement `Eq`, so that we can compare if the value produced in revision R+1 differs from the value in revision R. For volatile queries, though, this isn't strictly needed -- it might be nice to offer a variant on volatile queries that does not require the value to implement `Eq`. Such a variant would always be assumed to change, even if it produces an "equal" value.

To implement this, we would move the "equality comparison" into the `MemoizationPolicy`, yielding something like this:

```rust
pub trait MemoizationPolicy<DB, Q>
where
    Q: QueryFunction<DB>,
    DB: Database,
{
    fn should_memoize_value(key: &Q::Key) -> bool;

    fn memoized_value_eq(old_value: &Q::Value, new_value: &Q::Value) -> bool;

    fn should_track_inputs(key: &Q::Key) -> bool;
}
```

Instead of doing `old_value == new_value`, we would invoke `memoized_value_eq`. Next, we remove the `Eq` (and `Hash` etc) bounds from the `Query` trait. Those bounds would rather move into the `MemoizationPolicy`.

Then we can offer a `untracked` memoization policy (which requires `Eq` and does comparisons) and a `volatile` (which is always assumed to change). Or something like that.

nikomatsakis:
I created a [Zulip topic `#56 remove Eq bounds`][zulip] for discussing this.

[zulip]: https://salsa.zulipchat.com/#narrow/stream/146365-good-first-issue/topic/.2356.20remove.20.60Eq.60.20bounds

matklad:
Another case I've hit when equality is not necessary is for "top-level" queries, which can't be a dependency of any other query. I don't think it's worth special-casing this case, it's just smth to keep in mind.

nikomatsakis:
I *think* this issue can be marked closed. There were a few questions that came up in the implementation:

- Do we want a version of `input` that doesn't do "equality comparisons" (e.g., it just assumes that each call to `set` has a new value and updates the revision accodingly)? Maybe that should be the *only* kind of `input`? :)
- Similarly, do we want a version of "volatile" that *does* assume `Eq` (I was tentatively calling it "untracked")? You can always model this by wrapping a volatile with a memoized.

It *feels* like there is something a bit orthogonal here -- like there might be some orthogonal flag we can pull out about whether to compare values for equality. 

[...]

PR #58, opened 2018-10-18
[WIP] Removal of Eq bound on `Q::Val`

kleimkuhler:
WIP for #56

### DIFF
--- a/examples/compiler/class_table.rs
+++ b/examples/compiler/class_table.rs
@@ -39,6 +39,7 @@ fn all_fields(db: &impl ClassTableDatabase, (): ()) -> Arc<Vec<DefId>> {
             .flat_map(|def_id| {
                 let fields = db.fields(def_id);
                 (0..fields.len()).map(move |i| fields[i])
-            }).collect(),
+            })
+            .collect(),
     )
 }
--- a/src/derived.rs
+++ b/src/derived.rs
@@ -53,6 +53,8 @@ where
 {
     fn should_memoize_value(key: &Q::Key) -> bool;
 
+    fn memoized_value_eq(old_value: &Q::Value, new_value: &Q::Value) -> bool;
+
     fn should_track_inputs(key: &Q::Key) -> bool;
 }
 
@@ -60,10 +62,15 @@ pub enum AlwaysMemoizeValue {}
 impl<DB, Q> MemoizationPolicy<DB, Q> for AlwaysMemoizeValue
 where
     Q: QueryFunction<DB>,
+    Q::Value: Eq,
     DB: Database,
 {
     fn should_memoize_value(_key: &Q::Key) -> bool {
         true
+    }
+
+    fn memoized_value_eq(old_value: &Q::Value, new_value: &Q::Value) -> bool {
+        old_value == new_value
     }
 
     fn should_track_inputs(_key: &Q::Key) -> bool {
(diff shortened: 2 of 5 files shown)


src/derived.rs
```diff
@@ -101,6 +112,10 @@ where
         true
     }
 
+    fn memoized_value_eq(_old_value: &Q::Value, _new_value: &Q::Value) -> bool {
+        false
```
kleimkuhler: `VolatileValue`s now always return `false` (and therefore do not need an `Eq` bound), but you mentioned originally:

> maybe we want to have a new policy ("untracked") where we do the equality comparisons

I'd like to have a little more discussion on what you mean by that - perhaps in the [Zulip](https://salsa.zulipchat.com/#narrow/stream/146365-good-first-issue/subject/.2356.20remove.20.60Eq.60.20bounds) topic?

src/input.rs
```diff
@@ -42,6 +42,7 @@ struct IsConstant(bool);
 impl<DB, Q> InputStorage<DB, Q>
 where
     Q: Query<DB>,
+    Q::Value: Eq,
```
kleimkuhler: Would this `untracked` policy replace one of the following enums: `AlwaysMemoizeValue`, `NeverMemoizeValue`, `VolatileValue`? They are all unique in with respect to memoizing and tracking inputs, so I did not follow what you mean by that.

I needed to move the `Eq` bounds into the input queries as well. You mentioned:

> we might want to keep them, or else to remove the logic that tries to ensure that ...

Again this is maybe something we could discuss a little more. I'll continue to familiarize myself more with the utility of having/not having the bound so that maybe I can offer an opinion instead of just asking what to do.

src/input.rs
```diff
@@ -137,6 +138,7 @@ where
 impl<DB, Q> QueryStorageOps<DB, Q> for InputStorage<DB, Q>
 where
     Q: Query<DB>,
+    Q::Value: Eq,
```
kleimkuhler: Same as above.

src/input.rs
```diff
@@ -197,6 +199,7 @@ where
 impl<DB, Q> InputQueryStorageOps<DB, Q> for InputStorage<DB, Q>
 where
     Q: Query<DB>,
+    Q::Value: Eq,
```
kleimkuhler: Same as above.

src/derived.rs
nikomatsakis: Yeah, we can defer that for the purposes of this PR I think anyway.

examples/compiler/class_table.rs
```diff
@@ -39,6 +39,7 @@ fn all_fields(db: &impl ClassTableDatabase, (): ()) -> Arc<Vec<DefId>> {
             .flat_map(|def_id| {
                 let fields = db.fields(def_id);
                 (0..fields.len()).map(move |i| fields[i])
-            }).collect(),
+            })
+            .collect(),
```
nikomatsakis: btw, I should add that I always edit the salsa code with rustfmt-on-save enabled; I would appreciate it if you did the same. (Not sure if this edit came from rustfmt? Could be, if we are running slightly different versions...)

nikomatsakis:
Looks right to me!

examples/compiler/class_table.rs
kleimkuhler: Yep I have rustfmt-on-save enabled! This may have come from when I had the file open for understanding the example; I won't push single file formatting changes like this in the future.

As for the versioning, I was originally running this in nightly as I mentioned. I'm not sure if that would make the difference, but it could explain it.

Issue #24, opened 2018-10-02
Ensure panic safety


matklad:
When a query panics due to a bug during execution, the panic should be propagated to the caller, but the runtime should be left in a consistent state. 

This is important for IDE use-case: language server is a long-lived process, and it is being exposed to incomplete code comparatively more often than a command-line compiler. As a result, panics due to bugs are expected to be both more numerous and more annoying.

Currently, I believe we leave `InProgress` marks in the storage in the case of a panic, which effectively poisons the query, and can't be fixed by changing inputs.

nikomatsakis:
Hmm. Ordinarily, I'm not a big fan of trying to be panic safe, and would rather that we just "tear it down". But I suppose that in the case of salsa -- since we are controlling all the mutable state -- so long as we propagate the panic cleanly, we can be sure that the system is in a consistent state afterwards.

nikomatsakis:
So I think what we need to do here... 

The current strategy when executing a query is that we lock the map and we swap in a placeholder during execution. That is done *here*, and the value `old_memo` contains the value we swapped out from the map:

https://github.com/salsa-rs/salsa/blob/4a6a626f5d140e243a8508ce18e1ecbe7cb2bb42/src/derived.rs#L248-L264

Under ordinary circumstances, we eventually invoke `overwrite_placeholder` to remove the placeholder and replace it with a fresh value:

https://github.com/salsa-rs/salsa/blob/4a6a626f5d140e243a8508ce18e1ecbe7cb2bb42/src/derived.rs#L456-L466

You can see that this function also has the job of waking up other threads that may be blocked on us.

The danger is that a panic could occur and we would wind up with the placeholder being left in there. That is bad for many reasons: future calls from this thread will error out as if a cycle occurred, and other threads that are blocked waiting for a message will hang indefinitely.

I think that what we want to do is to install an "RAII guard" right here, basically after the `InProgress` placeholder has been installed, but before we've done anything else:

https://github.com/salsa-rs/salsa/blob/4a6a626f5d140e243a8508ce18e1ecbe7cb2bb42/src/derived.rs#L265

This guard would hold on to a reference to the key and looks something like this:

```rust
struct PanicGuard<'me, DB, Q> {
    map: &'me RwLock<FxHashMap<Q::Key, QueryState<DB, Q>>>,
    key: &'me Q::Key,
}

impl<'me, DB, Q> Drop for PanicGuard<'me, DB, Q> {..}
```

[...]

PR #62, opened 2018-10-23
Panic safely in a single threaded context

kleimkuhler:
WIP for #24

### DIFF
--- a/src/derived.rs
+++ b/src/derived.rs
@@ -263,6 +263,8 @@ where
             }
         };
 
+        let panic_guard = PanicGuard::new(&self.map, key);
+
         // If we have an old-value, it *may* now be stale, since there
         // has been a new revision since the last time we checked. So,
         // first things first, let's walk over each of our previous
@@ -277,7 +279,14 @@ where
                 let changed_at = memo.changed_at;
 
                 let new_value = StampedValue { value, changed_at };
-                self.overwrite_placeholder(runtime, descriptor, key, old_memo.unwrap(), &new_value);
+                self.overwrite_placeholder(
+                    runtime,
+                    descriptor,
+                    key,
+                    old_memo.unwrap(),
+                    &new_value,
+                    panic_guard,
+                );
                 return Ok(new_value);
             }
         }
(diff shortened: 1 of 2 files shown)


src/derived.rs
```diff
@@ -501,6 +515,37 @@ where
...
+    }
+}
+
+impl<'db, DB, Q> Drop for PanicGuard<'db, DB, Q>
+where
+    DB: Database + 'db,
+    Q: QueryFunction<DB>,
+{
+    fn drop(&mut self) {
+        let map = self.map.upgradable_read();
```
nikomatsakis: Here, you might as well just do `let map = self.map.write()` and not bother with the "upgradable read" and the "upgrade". (A read-write lock has two modes: read (like shared) and write (one write, no readers) -- an "upgradable read" allows readers but permits you to "upgrade" to a write-lock atomically with respect to any writes -- i.e., there can only be one "upgradable reader" at a time.)

src/derived.rs
```diff
@@ -501,6 +515,37 @@ where
...
+        let mut map = RwLockUpgradableReadGuard::upgrade(map);
+        let _ = map.remove(self.key);
```
nikomatsakis: To handle the parallel case, we would basically just *inspect* this returned value, much as we do in `overwrite_placeholder`:

https://github.com/salsa-rs/salsa/blob/bfc56d25919c5df7271f2c6421c082d6b93b003e/src/derived.rs#L459

This consists of two steps:

First, with the lock held, we check if there are any waiters and invoke `unblock_queries_blocked_on_self` to update salsa's internal data structures:

https://github.com/salsa-rs/salsa/blob/bfc56d25919c5df7271f2c6421c082d6b93b003e/src/derived.rs#L474-L481

Next, once lock is released, we sent messages:

```rust
https://github.com/salsa-rs/salsa/blob/bfc56d25919c5df7271f2c6421c082d6b93b003e/src/derived.rs#L490-L492
```

Before you tackle this, though, it's a good idea to add some test cases that (ought to...) deadlock, to make sure we are testing what we think we are.

PR #63, opened 2018-10-23
Simple logging methods

nikomatsakis:
This is an experimental branch.  It introduces a method `salsa_event` into the `Database` trait that defaults to a no-op. This event is given a closure (`event_fn`) which, when called, will produce information about what is happening -- the purpose of the closure is to defer any real work (notably, cloning the descriptor) if the `salsa_event` is not enabled.

This could be used for various purposes. For example, tests can override `salsa_event` and log the events somewhere, using them to reconstruct what happened. I was imagining though that we might provide some higher-level consumers at some point.

This branch uses `salsa_event` in the parallel test to force one thread blocks for another, which we could not do before.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,11 +17,12 @@ use crate::plumbing::QueryStorageMassOps;
 use crate::plumbing::QueryStorageOps;
 use crate::plumbing::UncheckedMutQueryStorageOps;
 use derive_new::new;
-use std::fmt::Debug;
+use std::fmt::{self, Debug};
 use std::hash::Hash;
 
-pub use crate::runtime::Runtime;
 pub use crate::runtime::RevisionGuard;
+pub use crate::runtime::Runtime;
+pub use crate::runtime::RuntimeId;
 
 /// The base trait which your "query context" must implement. Gives
 /// access to the salsa runtime, which you must embed into your query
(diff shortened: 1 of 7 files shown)


nikomatsakis:
One thing I did not like about this interface: there is no way to inspect the *kind* of the event before you invoke `event_fn`. An earlier draft had multiple methods (one per callback kind) -- maybe that is better, and they should have default implements that invoke the "default" `salsa_event` callback. Alternatively, `salsa::Event` could have a lifetime associated with it, or use `Cow` values, but those options seem kind of annoying.

src/lib.rs
```diff
@@ -38,6 +39,45 @@ pub trait Database: plumbing::DatabaseStorageTypes {
...
+
+    /// This function is invoked at key points in the salsa
+    /// runtime. It permits the database to be customized and to
+    /// inject logging or other custom behavior.
+    fn salsa_event(&self, event_fn: impl Fn() -> Event<Self>) {
+        #![allow(unused_variables)]
+    }
+}
+
+pub struct Event<DB: Database> {
```
matklad: #[derive(Debug)] will be useful here

src/runtime.rs
```diff
@@ -137,11 +137,23 @@ where
         RevisionGuard::new(&self.shared_state)
     }
 
+    /// The unique identifier attached to this `SalsaRuntime`. Each
+    /// forked runtime has a distinct identifier.
     #[inline]
-    pub(crate) fn id(&self) -> RuntimeId {
+    pub fn id(&self) -> RuntimeId {
```
matklad: I think we need to re-export `RuntimeId` in root as well.

src/lib.rs
nikomatsakis: Heh, yes, very much so. Also the "other things" (e.g., `Eq`, `Hash`).

src/runtime.rs
nikomatsakis: Yes....I thought I did that for some reason.

src/lib.rs
nikomatsakis: Ah, I can't really *derive* because of the `DB` parameter. Have to manually implement.

tests/parallel/true_parallel.rs
```diff
@@ -66,13 +66,15 @@ fn true_parallel_same_keys() {
         }
     });
 
-    // Thread 2 will sync barrier *just* before calling `sum`. Doesn't
-    // guarantee the race we want but makes it highly likely.
+    // Thread 2 will wait until Thread 1 has entered sum and then --
+    // once it has set tself to block -- signal Thread 1 to
```
kleimkuhler: Small typo here: "tself"

tests/parallel/true_parallel.rs
```diff
@@ -66,13 +66,15 @@ fn true_parallel_same_keys() {
         }
     });
 
-    // Thread 2 will sync barrier *just* before calling `sum`. Doesn't
-    // guarantee the race we want but makes it highly likely.
+    // Thread 2 will wait until Thread 1 has entered sum and then --
```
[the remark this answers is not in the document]
nikomatsakis: > From reading the function names and comments, I would assume thread1 signals 1 upon entry, and then waits for 2 to begin summing. 

Both of these things are true -- it does those **immediately upon entering** the query function, and before the query has really done anything yet. Is this the part that is confusing? It sounds like it will be signalling/waiting *before* entering the query?

Do you have a suggestion for a clearer name?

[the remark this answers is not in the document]
kleimkuhler: Okay I've taken time to step through line by line and understand where I was confused. I'll post my explanation here.

My understanding of how an `InProgress` marker appears in `db` was incorrect. I was following the control flow with the thought that an `InProgress` marker could only be set here:
```
for ch in key.chars() {
    sum += db.input(ch);
}
```
What I followed from this, is that `key` would still be the individual letters, and _not_ "abc". Therefore, the only way that `thread2` would find an `InProgress` is if `thread1` was executing those three lines at the same time and they happened to check the _same_ letter at the _same_ time.

What finally clicked was that `key` is not an individual letter, but the string "abc". That key has therefore already been updated to `InProgress` when `thread1` entered the query, and `thread2` therefore blocks right away.

In summary, what I was confused by was how `thread1` became unblocked because `thread2` would not find `InProgress` for any specific letter. That would have led to `thread2` never advancing past `wait_for` and thus, the test not passing.

src/lib.rs
nikomatsakis: Done.

nikomatsakis:
OK, so, I added `Debug` impls and also added per-event methods (`salsa_event_foo`) so that you can easily intercept just the event kinds that you like. I also refactored the `Event` structure to move some more common data into one place (e.g., the descriptor that appears in every event).

I'm feeling pretty "ok" about this change though mildly nervous about some of the details (e.g., just when do events fire etc). 

I'd like to have a "gc collected" event but that's a *bit* of a pain to thread through (doable, just takes a bit more thought).

I'd prefer if the "event delivery" methods were in a separate trait, but lacking specialization there is no easy way to do that without all database implementations having to add an `impl` for it, which  don't really want.

I guess overall I'm inclined to merge but this seems like the sort of thing we might want to revisit before 1.0 and reconsider.

nikomatsakis:
OK, I've removed the "per event kind" methods. That seemed like a lot more surface area than was warranted.

kleimkuhler:
I like these changes. I agree that the utility of a "gc collected" event would be nice. I have a few questions about where the complexity occurs specifically for that, but I'm sure it will become a little more clear as I use the events.

PR #67, opened 2018-10-25
Add one more cancellation test

### LINKED ISSUE #66: Canceled queries are not removed from the database after revision bump

matklad:
closes #66


matklad:
Let's make a relase after this? 0.6.1 has the bug

Issue #37, opened 2018-10-07
Add `remove` to `MutQueryStorageOps`?


matklad:
We have a `set` method to associate a value with a key, but we don't have a corresponding `remove` method. I need this because I try to model a set of files, known to rust-analyzer, as two queries:

```
salsa::query_group! {
    trait FilesDatabase: salsa::Database {
        fn file_text(key: FileId) -> Arc<String> {
            type FileText;
            storage input;
        }
        fn file_set(key: ()) -> Arc<Vec<File>> {
            type FileSet;
            storage input;
        }
    }
}
```

One query returns text, given file id. Another query returns a set of file ids. Currently, I can add and change files, but removing files is not really possible: a file can be removed from `FileSet`, but not from `FileText`. 

Perhaps this all should be left to GC though?

nikomatsakis:
I think `remove` makes sense for inputs but not derived queries. I hadn't expected the GC to prune the inputs.

nikomatsakis:
So now that inputs panic if there is no value, I guess a `remove` would be fairly straightforward to implement. It would be "on you" though to ensure that when you remove something, no derived query will access it, right?

nikomatsakis:
I imagine in that case we can literally just remove the key from the map and trigger the next revision, full stop.

matklad:
I guess so!

We probably need to forbid removal of constants, and we need to review `maybe_changed_since` logic, though it seems correct.

samestep:
Would a PR for this be welcome?

PR #69, opened 2018-10-26
add input policies to permit a wider range of inputs

nikomatsakis:
I was getting annoyed that input values must implement `Eq` so I added the option to specify different policies. This also removes the `Default` bound by default.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -13,141 +13,36 @@ use log::debug;
 use parking_lot::{RwLock, RwLockUpgradableReadGuard};
 use rustc_hash::FxHashMap;
 use std::collections::hash_map::Entry;
-use std::marker::PhantomData;
 
 /// Input queries store the result plus a list of the other queries
 /// that they invoked. This means we can avoid recomputing them when
 /// none of those inputs have changed.
-pub struct InputStorage<DB, Q, IP>
+pub struct InputStorage<DB, Q>
 where
     Q: Query<DB>,
     DB: Database,
-    IP: InputPolicy<DB, Q>,
 {
     map: RwLock<FxHashMap<Q::Key, StampedValue<Q::Value>>>,
-    input_policy: PhantomData<IP>,
 }
 
-pub trait InputPolicy<DB, Q>
+impl<DB, Q> Default for InputStorage<DB, Q>
 where
     Q: Query<DB>,
     DB: Database,
-{
-    fn compare_values() -> bool {
-        false
-    }
-
-    fn compare_value(_old_value: &Q::Value, _new_value: &Q::Value) -> bool {
-        panic!("should never be asked to compare values")
-    }
-
-    fn missing_value(key: &Q::Key) -> Q::Value {
-        panic!("no value set for {:?}({:?})", Q::default(), key)
-    }
-}
-
-/// The default policy for inputs:
-///
-/// - Each time a new value is set, trigger a new revision.
-/// - On an attempt access a value that is not yet set, panic.
-pub enum ExplicitInputPolicy {}
-impl<DB, Q> InputPolicy<DB, Q> for ExplicitInputPolicy
-where
-    Q: Query<DB>,
-    DB: Database,
-{
-}
-
-/// Alternative policy for inputs:
-///
-/// - Each time a new value is set, trigger a new revision.
-/// - On an attempt access a value that is not yet set, use `Default::default` to find
-///   the value.
-///
-/// Requires that `Q::Value` implements the `Default` trait.
-pub enum DefaultValueInputPolicy {}
-impl<DB, Q> InputPolicy<DB, Q> for DefaultValueInputPolicy
-where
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -224,6 +224,57 @@ where
...
+///   which can significantly reduce the amount of recomputation
+///   required in new revisions. This does require that the value
+///   implements `Eq`.
+/// - `storage volatile` -- indicates that the inputs are not fully
+///   captured by salsa. The result will be recomputed once per revision.
+/// - `storage dependencies` -- does not cache the value, so it will
+///   be recomputed every time it is needed. We do track the inputs, however,
+///   so if they have not changed, then things that rely on this query
+///   may be known not to have changed.
+/// - `storage input` -- does not cache the value, so it will
```
kleimkuhler: What is the main difference between `storage dependencies` and `storage input` for derived queries? It seems `storage dependencies` requires a `use fn` (`dep_derived1` in _memoized_dep_inputs.rs_), but wouldn't a derived query with `storage input` just make it an input query?

src/input.rs
```diff
@@ -13,40 +13,141 @@ use log::debug;
...
+    fn compare_values() -> bool {
+        false
+    }
+
+    fn compare_value(_old_value: &Q::Value, _new_value: &Q::Value) -> bool {
+        panic!("should never be asked to compare values")
+    }
+
+    fn missing_value(key: &Q::Key) -> Q::Value {
+        panic!(format!("no value set for {:?}({:?})", Q::default(), key))
```
matklad: replace `panic!(format!` with just `panic!`?

src/input.rs
```diff
@@ -13,40 +13,141 @@ use log::debug;
...
     Q: Query<DB>,
     DB: Database,
-    Q::Value: Default,
+    IP: InputPolicy<DB, Q>,
 {
     map: RwLock<FxHashMap<Q::Key, StampedValue<Q::Value>>>,
+    input_policy: PhantomData<IP>,
+}
+
+pub trait InputPolicy<DB, Q>
```
matklad: Note that we have

Memoization**Policy**
Sweep**Strategy**
Input**Policy**

Should we rename strategy?

nikomatsakis:
I was thinking that maybe I should remove the different input policies altogether and instead we could have `db.query(Input).get(key)` return an `Option` -- somewhat analogous to the way to  `foo[30]` vs `foo.get(30)`. Then you can get the `input default` effect by doing the long-hand way (hmm, admittedly, far less convenient if you want it ubiquitously).

The "eq" policy seems like something you can quite easily do yourself as well, or which we could model with an additional helper `set_if_not_eq` or something -- after all, it's just:

```rust
if let Some(v) = db.query(Q).get(key) {
    if v == new_value { 
        return;
    }
}

db.query(Q).set(key, value);
```

src/input.rs
nikomatsakis: Probably, yes.

nikomatsakis: They seem different to me but not for any good reason I guess.

src/lib.rs
nikomatsakis: I think that's a typo maybe..?

nikomatsakis: Er, yes, just a mistake.

src/input.rs
nikomatsakis: Seems like it should go in another PR though

matklad:
I wonder if it has interactions with https://github.com/salsa-rs/salsa/issues/37:

It seems like if we have an `Option`-returning method, then `.remove(key)` can't simply delete the key from the store, because we'll lose info about when the change was made then.

It's also interesting that, in case of rust-analyzer at least, you never actually ask for missing keys, so panicing policy works well. At the same time, if getting a non-existent key always panics, then trivial remove strategy seems to work.

nikomatsakis:
@matklad 

> I suggest switching to panic.

This is what I meant. Just remove the `InputPolicy` for now and only support the simplest behavior.

src/lib.rs
```diff
@@ -224,6 +224,53 @@ where
...
+/// You can change either of these two things by choosing one of the following
+/// "extended" options:
+///
+/// - `storage (input default)` -- Attempting to fetch a key that has
+///   not been explicitly `set` will use the `Default` trait to
+///   provide a value. Requires that the value implements `Default`.
+/// - `storage (input eq)` -- Compare new values against the old
+///   values and avoid triggering a new revision if they are
+///   equal. Requires value types to implement the `Eq` trait.
+/// - `storage (input default eq)` -- Both of the above.
```
matklad: these docs should be removed :crying_cat_face:

kleimkuhler:
Does this give any closure for the open conversation in #56 ?

nikomatsakis:
@kleimkuhler the two questions seem "related" but I don't think this PR provides closure per se

PR #71, opened 2018-10-29
Panic in PanicGuard if it's not explicitly forgotten

matklad:
My understanding is that `PanicGuard` should be forgotten for every non-panicing execution. 

Let's enforce it with a panic, so that, for example, we won't silently overwrite the keys by adding a `return`.


nikomatsakis:
cc @kleimkuhler

nikomatsakis:
This seems good.

PR #72, opened 2018-10-30
be more careful when updating after a dependency check

nikomatsakis:
@matklad I suspect this fixes your problem, though I've not yet tried to reproduce — we should probably get some sort of test. Still, might be good to verify the hypothesis.

Fixes #70

PR #74, opened 2018-10-30
Add a stress test

matklad:


matklad:
This currently deadlocks.

matklad:
I've check that even with #72 **and** without GC, this stress test still triggers a deadlock. I've never seen the assertion error tough, with or without 72. 

Looks like an unrelated bug

matklad:
This is in master now

PR #75, opened 2018-10-30
remove use of upgradable reads from derived queries

nikomatsakis:


nikomatsakis:
I'm now thinking that this PR is not necessary. But I'm trying to write out a comment clearly explaining what went wrong and why I think we're ok and dang is it hard. =)

The TL;DR of why I think we're ok *now* is once we acquire this particular upgradable-read, we will always release it without blocking on any other locks (apart from the mutex guarding the dependency graph, but that is always a "leaf lock" so we should be ok).

matklad:
I feel maybe we should stick to read/write access for the time being, just to make debugging deadlocks, should they reappear, easier. That is, I'd like to avoid thinking "oh, *maybe* its that upgrade lock again" when debugging.

We probably can put upgradable reads in back when parking_lot is fixed.

matklad:
>ut I'm trying to write out a comment clearly explaining what went wrong and why I think we're ok

I think the main problem here is that parking_lot's behavior does not match the docs, so it's difficult to reason about **what** are you trying to prove in the first place.

matklad:
@nikomatsakis see https://github.com/salsa-rs/salsa/issues/99 for a deadlock which happens on those `upgradable_read` calls :-)

nikomatsakis:
I'd like to understand this deadlock, but I rebased.

matklad:
>I'd like to understand this deadlock, but I rebased.

I've mapped where each thread is blocked over here:
https://github.com/matklad/salsa/commit/330b27d25c9e2858057631d9d3713234a4766f45

This doesn't look like the "upgradable reads block reads" situation we've got before, because only writes and upgradable reads are involved, no plain reads. The next step would probably be to write a parking-lot stress test, [like the one for reads](https://github.com/matklad/parking-lot-deadlock/blob/95989b3da40ed42e1ed7acba829d20b8b4319845/src/main.rs). I sort of expect an "interesting" behavior of upgradable read here :)

matklad:
@nikomatsakis see https://github.com/Amanieu/parking_lot/issues/101#issuecomment-450146932. Looks like it's a pl bug.

nikomatsakis:
Fascinating. OK, merging and publishing!

matklad:
This bug is fixed in parking lot 0.7.1 so in theory we can perhaps revert this now? Though, I'd rather avoid using upgradable reads: even if they are bug free now, I'll still will be blaiming/testing them when facing with further salsa bugs :-) 

And in general, it might be a good idea to improve locking infra once with have benchmarks to quantify improvements (FWIW, salsa is fast enough for everything I through at at in rust-analyzer so far) :-)

PR #78, opened 2018-10-31
make `fork` give you a `Frozen<DB>`

nikomatsakis:
This is the first step towards the "fewer footguns" API.

Opening for discussion.

Some to do items:

- [x] Move `set` to a `query_mut` method that requires `&mut self` (https://github.com/salsa-rs/salsa/issues/79)
- [x] Add tests for the "recursive read lock" case; I think i understand well enough now to be able to trigger the deadlock deterministically (and show that we are avoiding it)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -202,7 +202,7 @@ pub trait ParallelDatabase: Database + Send {
     /// cloning over each of the fields from `self` into this new
     /// copy. For the field that stores the salsa runtime, you should
     /// use [the `Runtime::snapshot` method][rfm] to create a snapshot of the
-    /// runtime. Finally, package up the result using `Frozen::new`,
+    /// runtime. Finally, package up the result using `Snapshot::new`,
     /// which is a simple wrapper type that only gives `&self` access
     /// to the database within (thus preventing the use of methods
     /// that may mutate the inputs):
@@ -211,8 +211,8 @@ pub trait ParallelDatabase: Database + Send {
     ///
     /// ```rust,ignore
     /// impl ParallelDatabase for MyDatabaseType {
-    ///     fn snapshot(&self) -> Frozen<Self> {
-    ///         Frozen::new(
+    ///     fn snapshot(&self) -> Snapshot<Self> {
+    ///         Snapshot::new(
     ///             MyDatabaseType {
     ///                 runtime: self.runtime.snapshot(self),
     ///                 other_field: self.other_field.clone(),
@@ -227,7 +227,7 @@ pub trait ParallelDatabase: Database + Send {
     /// This second handle is intended to be used from a separate
     /// thread. Using two database handles from the **same thread**
     /// can lead to deadlock.
-    fn snapshot(&self) -> Frozen<Self>;
+    fn snapshot(&self) -> Snapshot<Self>;
 }
 
 /// Simple wrapper struct that takes ownership of a database `DB`
@@ -235,23 +235,23 @@ pub trait ParallelDatabase: Database + Send {
 /// more details.
 ///
 /// [fm]: trait.ParallelDatabase#method.snapshot
-pub struct Frozen<DB>
+pub struct Snapshot<DB>
 where
     DB: ParallelDatabase,
 {
     db: DB,
 }
 
-impl<DB> Frozen<DB>
+impl<DB> Snapshot<DB>
 where
     DB: ParallelDatabase,
 {
     pub fn new(db: DB) -> Self {
-        Frozen { db }
+        Snapshot { db }
     }
 }
 
-impl<DB> std::ops::Deref for Frozen<DB>
+impl<DB> std::ops::Deref for Snapshot<DB>
 where
     DB: ParallelDatabase,
 {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,7 +20,6 @@ use derive_new::new;
 use std::fmt::{self, Debug};
 use std::hash::Hash;
 
-pub use crate::runtime::Frozen;
 pub use crate::runtime::Runtime;
 pub use crate::runtime::RuntimeId;
 
@@ -196,34 +195,71 @@ pub trait ParallelDatabase: Database + Send {
     ///
     /// [`is_current_revision_canceled`]: struct.Runtime.html#method.is_current_revision_canceled
     ///
+    /// # How to implement this
+    ///
+    /// Typically, this method will create a second copy of your
+    /// database type (`MyDatabaseType`, in the example below),
+    /// cloning over each of the fields from `self` into this new
+    /// copy. For the field that stores the salsa runtime, you should
+    /// use [the `Runtime::fork` method][rfm] to create a fork of the
+    /// runtime. Finally, package up the result using `Frozen::new`,
+    /// which is a simple wrapper type that only gives `&self` access
+    /// to the database within (thus preventing the use of methods
+    /// that may mutate the inputs):
+    ///
+    /// [rfm]: struct.Runtime.html#method.fork
+    ///
+    /// ```rust,ignore
+    /// impl ParallelDatabase for MyDatabaseType {
+    ///     fn fork(&self) -> Frozen<Self> {
+    ///         Frozen::new(
+    ///             MyDatabaseType {
+    ///                 runtime: self.runtime.fork(self),
+    ///                 other_field: self.other_field.clone(),
+    ///             }
+    ///         )
+    ///     }
+    /// }
+    /// ```
+    ///
     /// # Deadlock warning
     ///
     /// This second handle is intended to be used from a separate
     /// thread. Using two database handles from the **same thread**
     /// can lead to deadlock.
-    fn fork(&self) -> Frozen<Self> {
-        Frozen::new(self)
-    }
+    fn fork(&self) -> Frozen<Self>;
+}
 
-    /// Creates another handle to this database destined for another
-    /// thread. You are meant to implement this by cloning a second
-    /// handle to all all the state in the database; to clone the
-    /// salsa runtime, use the [`Runtime::fork_mut`] method.
-    ///
-    /// Note to users: you only need this method if you plan to be
-    /// setting inputs on the other thread. Otherwise, it is preferred
-    /// to use [`fork`], which will given back a frozen database fixed
-    /// at the current revision.
-    ///
-    /// # Deadlock warning
-    ///
-    /// This second handle is intended to be used from a
-    /// separate thread. Using two database handles from the **same
-    /// thread** can lead to deadlock.
-    ///
-    /// [`Runtime::fork_mut`]: https://docs.rs/salsa/0.7.0/salsa/struct.Runtime.html#method.fork_mut
-    /// [`fork`]: trait.ParallelDatabase.html#tymethod.fork
-    fn fork_mut(&self) -> Self;
+/// Simple wrapper struct that takes ownership of a database `DB`
+/// and only gives `&self` access to it. See [the `fork` method][fm] for
+/// more details.
+///
+/// [fm]: trait.ParallelDatabase#method.fork
+pub struct Frozen<DB>
+where
+    DB: ParallelDatabase,
+{
+    db: DB,
+}
+
+impl<DB> Frozen<DB>
+where
+    DB: ParallelDatabase,
+{
+    pub fn new(db: DB) -> Self {
+        Frozen { db }
+    }
+}
+
+impl<DB> std::ops::Deref for Frozen<DB>
+where
+    DB: ParallelDatabase,
+{
+    type Target = DB;
+
+    fn deref(&self) -> &DB {
+        &self.db
+    }
 }
 
 pub trait Query<DB: Database>: Debug + Default + Sized + 'static {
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -177,13 +177,53 @@ impl Default for SweepStrategy {
...
+    /// # Deadlock warning
+    ///
+    /// This second handle is intended to be used from a
     /// separate thread. Using two database handles from the **same
     /// thread** can lead to deadlock.
-    fn fork(&self) -> Self;
+    ///
+    /// [`Runtime::fork_mut`]: https://docs.rs/salsa/0.7.0/salsa/struct.Runtime.html#method.fork_mut
+    /// [`fork`]: trait.ParallelDatabase.html#tymethod.fork
+    fn fork_mut(&self) -> Self;
```
matklad: Do we have a use-case for setting from different threads?

I see that we *could* do it, but *should* we? Perhaps it makes sense to keep API minimal until someone comes along with a feature request?

nikomatsakis: I don't have a particular use case but I don't see a way to make `fork` work without this method.

nikomatsakis: In particular, I'm trying to model this API on the usual "rust mode", in which you have:

- `T` -- a type with full privileges
- `&T` -- here, `Frozen<T>`, a view on `T` with reduced privileges

And the point is that to implement `fork` which gives you the `Frozen<T>`, I  need a way to get another `T`

nikomatsakis: Another question is: does supporting `fork_mut` cause a *problem* for some use case we want to support? I could *imagine* that it might at least make some things more awkward: it requires that the main `DB` type be "forkable" to other threads. So if you do have the model where there is a "main" thread with full access and "helper" threads with frozen access, maybe you'd like this 'main thread" to have some fields that they don't have.

This kind of corresponds to `String` (which has a capacity) and `str` (which does not), or the usual `PathBuf`/`Path` naming. So basically you could imagine a `DbBuf` type and a `Db` type. But I am wary because -- as those analogies may suggest -- that is another "Level up" in complexity for people to deal with.

I'd like to have a good use case for introducing a "buf-split". I am *mildly* skeptical that there is one -- you could always have a `Full<DB>` wrapper that adds the extra fields, after all. Moreover, you ought not to be adding shared, mutable state that *queries* can access since it would bypass the revision system and just be a footgun (and thus I'm not inclined to make it any easier).

So, put another way, does the `fork_mut` method do any **harm**?

matklad: >And the point is that to implement fork which gives you the Frozen<T>, I need a way to get another T

That's a good point why we really need it. And no, I don't see any obvious problems with fork_mut.

src/runtime.rs
```diff
@@ -650,3 +594,120 @@ impl<DB: Database> DependencyGraph<DB> {
...
+        //
+        // We could just use recursive reads *always*, but that could
+        // lead to starvation in the case where you have various
+        // threads invoking get and set willy nilly. Not sure how
+        // important it is to ensure that case works -- it's not a
+        // recommended pattern -- but it seems (for now at least) easy
+        // enough to do so.
+        //
+        // [may deadlock]: https://docs.rs/lock_api/0.1.3/lock_api/struct.RwLock.html#method.read
+        let use_recursive_lock = source_runtime.query_in_progress();
```
matklad: This indeed looks subtle and too elaborate for a thing which is basically a counter for "how many forks are there". A particular funny point is that, if `use_recursive_lock` is `true`, then we **know** that we already hold the lock, so we don't really need to lock it, we need to upgrade the locked counter. That is, we know that we can't block.

I feel like there's perhaps a higher-level synchronization primitive we can use here, but I can't think of one. Perhaps a frozen runtime can hold an `Arc<Thing>`, such that forking a frozen runtime just bumps ref count on the `thing`, and, when the last runtime dies, the thing's drop unlocks some lock?

src/lib.rs
nikomatsakis: One thing I was wondering is whether there might be a better name than `fork_mut`. It feels like a qualitatively different operation than `fork`, in some sense, in that what you get out not only permits mutation but lacks the "single, consistent revision" property. (Of course, those are linked, but still.)

src/runtime.rs
nikomatsakis: >  A particular funny point is that, if `use_recursive_lock` is true, then we know that we already hold the lock, so we don't really need to lock it, we need to upgrade the locked counter

The intuition *here* I think is that -- if this flag is true -- we are "splitting" our existing read permission into two (hence we acquire `read_recursive`, to do that). The main use case I see for this is that we will want to permit one query to fork off new (temporary) handles in order to do things like "do-all-queries-in-parallel", so it may want to invoke `fork()` (it is interesting that it can also invoke `fork_mut`, although it doesn't violate any of the nice properties -- kind of a foot gun though).

> I feel like there's perhaps a higher-level synchronization primitive we can use here, but I can't think of one.

I know what you mean, but I too couldn't think of one. I did consider making our own primitive here -- basically wrapping a `Mutex<usize>` and a condition variable -- but it seems like a reasonable amount of logic to reproduce for little reason.

An alternative is just to **always** use `read_recursive`. This is not terrible. The main problem is that if you have two threads just doing random gets/sets, the `set` operations may starve. This wouldn't really arise in the "expected usage pattern", though, because there we have one master thread doing the sets *and* giving rise to the read-only queries, and those read-only queries always execute with a frozen DB anyway.

src/lib.rs
nikomatsakis: Thinking more about it, I am wondering if it *is* worth trying to expose the "buf/view" split somehow. I think one of the main effects would be that the `salsa::Database` trait would represent *views* and we would add a separate `salsa::DatabaseBuf: salsa::Database` trait for the "backing buffer".. maybe? I have to think on it a bit. 

The main thing we would want, I think, is that:

- `Database` and `ParallelDatabase` support get queries and `fork`
- `DatabaseBuf` and `ParallelDatabaseBuf` add in the `query_mut(&mut self, query: Q)` and `fork_mut(&mut self)` operations

I think that the way this would work is that the runtime would also offer `fork(&self, from_db: &DB) -> Self` and `fork_mut(&mut self, from_db: &mut DB) -> Self` operations. 

Importantly, you would **not** have to implement `fork_mut` at all -- maybe this means we can even cut it out? But you would have to implement the `fork`, which -- since we want it to return a `Frozen<DB>` type I think -- is a *touch* awkward. You would do something like:

```rust
fn fork(&self) -> Frozen<Self> {
    Frozen::new(MyDatabaseType {
        runtime: self.runtime.fork(self),
        other_fields: self.other_fields.clone(),
    })
}
```

The advantages of this are:

- We do not need `fork_mut`, we can cut that for now altogether
- This implies we can **always** use `read_recursive` for the "query lock", because there is only one thread that can both get/set -- all other threads are using forked databases, which are already holding the read lock

The main disadvantage is that the user has to write the call to `Frozen::new` themselves.

nikomatsakis: Er, that wound up in a slightly different place than I originally imagined, in that it doesn't have the buf/view split -- you still have `DB` and `Frozen<DB>` as the buf/view types. But it seems... good? Still not sure what a nice Buf/View split would look like of it that might feel slicker, seems like it must be strictly more work.

nikomatsakis: (Maybe rename `Frozen<DB>` to `Forked<DB>`? Or `fork` to `freeze`? Or `fork_view` and `View`?)

matklad: Hm, so I see various designs here, but neither strikes me as significantly better/worse than others. In Rust analyzer, I have a design where get/set parts are completely separate: you can't `get` from a "mutable" struct: I have an `AnalysisHost` with three methods, `new`, `apply_changes`, `snpahost`, and `Analysis` (returned by a `snapshot`), which has a bazillion of `get_foo` methods.  However, in this setup `apply_change` internally uses `.get` methods (change is a diff, so you need to know the current state to apply it). 


Unrelated, but maybe `Readonly<DB>` or `Snapshot<DB>`?

nikomatsakis:
OK, I could not stop myself from implementing the new strategy. We now **enforce** that there is a single mutator thread (or else we get deadlock -- we can move this to static enforcement as a next step). There are two "downsides" to this:

- It is not visible from the type of a single `Runtime` whether it is frozen or not -- but that doesn't leak out to the user, who either interacts with either a `DB` or a `Frozen<DB>`. Seems ok.
- The user has to invoke `Frozen::new` themselves as part of the `fork` operation. Seems ok.
- I had to rewrite all the stress tests because the scenarios we support are fundamentally simpler. Um, seems ok.

nikomatsakis:
Er, @matklad, I missed this but let me respond here:

> Hm, so I see various designs here, but neither strikes me as significantly better/worse than others. [...]

I actually don't 100% understand what you are talking about, I'll have to read into the analyzer code a bit, but I think the new design is better in that it guides you to the pattern we want to support -- and, moreover, supporting only that pattern is definitely simpler from the impl POV since we don't have to worry as much about deadlock and starvation (we kind of push that off in the form of cancellation). And yet from the end-user POV the complex basically boils down to the fact that you have to add `Frozen::new` when you implement `fork`, which seems "ok".

> Unrelated, but maybe Readonly<DB> or Snapshot<DB>?

I like those names; I would like the name of `fork` and the name of `Frozen` to line up. Perhaps `fork_snapshot` and `Snapshot`?

nikomatsakis:
I suppose that `db.snapshot()` and `Snapshot<DB>` would also be very good.

nikomatsakis:
I am reminded that I wanted to see if I could rework the dependency tracking to operate not based on `RuntimeId` but based on `ThreadId` -- this...might actually be very easy -- in which case creating a snapshot and then interleaving operations between two snapshots (which is actually quite hard to do, but I guess it can be arranged) wouldn't really have any negative effects in particular.

matklad:
> I actually don't 100% understand what you are talking about, 

What we have in this PR is, basically,

```
struct Db {
   fn get(&self);
   fn set(&mut self);
}
```

What I have in Rust analyzer is

```
struct Db {
    fn set(&mut self);
    fn snapshot(&self) -> Snapshot;
}

struct Snapshot {
    fn get(&self);
}
```

That is, read and write accesses are independent, they are not "subsets".

nikomatsakis:
Ah, another thing I remember: in my code, when you start a query, I now have this logic:

```rust
    pub(crate) fn start_query(&self) -> Option<QueryGuard<'_, DB>> {
        let mut local_state = self.local_state.borrow_mut();
        if !local_state.query_in_progress {
            local_state.query_in_progress = true;
            let guard = self.shared_state.query_lock.read();

            Some(QueryGuard::new(self, guard))
        } else {
            None
        }
    }
```

This should probably use `read_recursive` too but in fact I think we may not need a lock here **at all**. Once we implement the `query_mut` change (so that you can't trigger a new revision without `&mut` access), I think this logic will be basically entirely unnecessary. Even now it...may well be unnecessary.

Reasoning:

- The purpose of this lock is to prevent a call to `set` from proceeding until all queries have completed. But who launched that query?
    - In the case of a snapshot, the lock is already held.
    - In the case of the master, then the `set` would be occurring from the same thread as the query, in which case just checking for an active query on the local stack should suffice.

So I think we can just remove this logic.

nikomatsakis:
@matklad ah, right, ok. Kind of a "refcell" or "rwlock" setup. The appeal of that of course is that if you perform consecutive operations, they must be within a snapshot, so the atomicity lines are clear.

On the other hand, if you have only one mutator thread -- as this setup does -- then there is no particular concern, because you have a single owner. (Basically it acts the same as any other data structure in Rust: you need to pass around an `&mut` to do mutation, and that avoids aliasing. Well, once I add the `query_mut` stuff.)

matklad:
This setup relates to that `is_query_in_progess` field: if you always need to create snapshot to do a query, then snapshot creation can be the sole place where you lock the lock (as opposed to checking on every .get). Though yeah, with a single mutator, I think we can remove that just fine as well.

matklad:
concern: `query_in_progress` field is now useless and doesn't do what the name implies. Should we remove the field?

matklad:
r=me

It seems slightly amusing that we don't need RevisionGuard on the main thread, but I think that does work actually! Nifty!

nikomatsakis:
I think last remaining thing might be applying this rename:

> I suppose that db.snapshot() and Snapshot<DB> would also be very good.

tests/panic_safely.rs
```diff
@@ -44,7 +52,10 @@ fn should_panic_safely() {
 
     // Invoke `db.panic_safely() without having set `db.one`. `db.one` will
     // default to 0 and we should catch the panic.
-    let result = panic::catch_unwind(AssertUnwindSafe(|| db.panic_safely()));
+    let result = panic::catch_unwind(AssertUnwindSafe({
```
kleimkuhler: I mentioned this in Zulip, but I did not originally intend for this test case to test parallel behavior. I'm okay with the change as is, but this will be covered by the new test in _tests/parallel/true_parallel.rs_ after these branches merge anyways.

This was meant to be the minimal working example of a safe panic.

nikomatsakis: @kleimkuhler yeah — makes sense. The question is whether we intend to the database to be "unwind safe" (that is, recoverable within a single thread). I don't normally like to promise that, so in a way the test is testing something we don't necessarily expect to work. *On the other hand,* I can't see any reason *not* to support this scenario (but we don't presently).

nikomatsakis: I'll merge with this test but let's update it =)

PR #81, opened 2018-10-31
Ensure parallel panic safety

### LINKED ISSUE #24: Ensure panic safety

kleimkuhler:
Fixes #24


nikomatsakis:
@kleimkuhler 

>  I think the best way to handle that is up for some discussion. 

Indeed. And, it's not something we *have* to do -- it seems "ok" to say that we recover from panics at the level of snapshots but not within a thread. In fact, this makes some sense -- if the thread with the mutable handle panics, we don't know if all of the input queries etc are in a coherent state.

For example, you might be doing:

```rust
db.query_mut(Input1).set(...);
let old_value = db.bar(); // read bar query, which panics unexpectedly
db.query_mut(Input2).set(...); // this will never execute!
```

However, if we are not going to recover from panics -- and I am now thinking we should not -- then perhaps we want to at least *poison* the database so we "fail fast". The idea would be to have some panic-guard that sets a "poisoned" flag in the runtime to true when unwinding, and any attempt to use that database handle again will panic quickly with "poisoned".

PR #82, opened 2018-11-01
introduce `query_mut` which you must use to get `set` methods

### LINKED ISSUE #79: migrate `set` to `query_mut`
Today, to set the value of a query, you do `db.query(Q).set(...)`. 

To build on https://github.com/salsa-rs/salsa/pull/78, we would like to change this to `db.query_mut(Q).set(...)`. This would then require an `&mut self` on the database.

Together with https://github.com/salsa-rs/salsa/pull/78, this moves us completely to the model of "one mutator thread" (which owns the main database handle, of type `DB`) and various "reader threads" (which own `Frozen<DB>` handles). Only the mutator thread can get `&mut self` access to the DB, so only it can invoke `set`. 

I'll follow up with mentoring instructions perhaps.

nikomatsakis:
This makes it impossible to try and invoke `set` from a snapshot.

Interestingly, at least the way the library is currently setup, the `&mut` we acquire on the database is only "skin deep" -- this because the `QueryTable` needs to store both a `&Storage` (which is owned by the `db`) and a reference to the db itself. This means that users who "hack up something" against the plumbing APIs directly could -- if they *truly wanted* -- get themselves into deadlocks. I think they have no one to blame but themselves. It might (maybe?) be interesting to think about refactoring things to avoid this, but it doesn't seem very important. I guess the way we would do it using traits so that we only pass down the `&mut DB` but we can extract the storage or other things we need later (vs the current scheme, which creates a `QueryTableMut` containing all the pieces we are going to need in one place).

Fixes #79

### DIFF
--- a/examples/hello_world/main.rs
+++ b/examples/hello_world/main.rs
@@ -93,11 +93,11 @@ salsa::database_storage! {
 
 // This shows how to use a query.
 fn main() {
-    let db = DatabaseStruct::default();
+    let mut db = DatabaseStruct::default();
 
     println!("Initially, the length is {}.", db.length(()));
 
-    db.query(InputString)
+    db.query_mut(InputString)
         .set((), Arc::new(format!("Hello, world")));
 
     println!("Now, the length is {}.", db.length(()));
--- a/src/debug.rs
+++ b/src/debug.rs
@@ -7,7 +7,11 @@ use crate::Query;
 use crate::QueryTable;
 use std::iter::FromIterator;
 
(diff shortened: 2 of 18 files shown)


matklad:
>It might (maybe?) be interesting to think about refactoring things to avoid this, but it doesn't seem very important.

Agree with a general sentiment:

- if we can refactor to eliminate "you did it to yourself" style footgans with refactoring, we should do that.
- if we need to write substantial amount of code for protection, it's better to stick with blaming users :)

matklad:
LGTM!

PR #89, opened 2018-11-05
Add initial impl of poison handles

### LINKED ISSUE #88: poison database handles on panics
Continuing the conversation with @kleimkuhler from https://github.com/salsa-rs/salsa/pull/81, I think we should **not** attempt to make an individual database handle panic safe, but we should instead **poision** the handles so that -- if they are used again -- you get an immediate panic. You can still recover from panics by using snapshots.

What I wrote in #81 was as follows: 

> It seems "ok" to say that we recover from panics at the level of snapshots but not within a thread. In fact, this makes some sense -- if the thread with the mutable handle panics, we don't know if all of the input queries etc are in a coherent state.
> 
> For example, you might be doing:
> 
> ```rust
> db.query_mut(Input1).set(...);
> let old_value = db.bar(); // read bar query, which panics unexpectedly
> db.query_mut(Input2).set(...); // this will never execute!
> ```
> 
> However, if we are not going to recover from panics -- and I am now thinking we should not -- then perhaps we want to at least _poison_ the database so we "fail fast". The idea would be to have some panic-guard that sets a "poisoned" flag in the runtime to true when unwinding, and any attempt to use that database handle again will panic quickly with "poisoned".

kleimkuhler:
Fixes #88

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -192,6 +192,10 @@ where
     pub(crate) fn with_incremented_revision<R>(&self, op: impl FnOnce(Revision) -> R) -> R {
         log::debug!("increment_revision()");
 
+        if self.local_state.borrow().poisoned {
+            panic!("attempted to use a poisoned database")
+        }
+
         if !self.permits_increment() {
             panic!("increment_revision invoked during a query computation");
         }
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -243,6 +243,10 @@ where
     ) -> ComputedQueryResult<DB, V> {
         debug!("{:?}: execute_query_implementation invoked", descriptor);
 
+        if self.local_state.borrow().poisoned {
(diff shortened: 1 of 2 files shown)


nikomatsakis:
> Based off my initial implementation of the panic guard, using set with a poisoned database handle will not panic because the handle is poisoned; it will panic because the query_stack is not empty here

It seems like it would yield a better user experience if we panic because of poisoning -- can we add a test, too?

kleimkuhler:
> It seems like it would yield a better user experience if we panic because of poisoning

I can add in a poison check to `with_incremented_revision`! Should be fairly straight forward to check that before the `query_stack` and add a test as well. I'll get that in now before merging it should be quick.

nikomatsakis:
Obviated by https://github.com/salsa-rs/salsa/pull/114 -- if we claim that database handles are panic safe, then they should not be poisoned

PR #95, opened 2018-12-13
No dynamic dispatch for constructing query descriptors

nikomatsakis:
A small tweak to the trait setup: don't use a fn pointer to construct query keys anymore.

This mostly arose from playing around with the trait setup to try and think how to avoid listing out all queries when defining the query storage.

(On a related note, after coming back to them from a short break, I have the sense that the salsa traits are a bit tangled and could be clarified some. This PR doesn't really take any steps in that direction, but it adds some comments to `GetQueryTable` to try and get at the "essence" of what that trait is about.)


nikomatsakis:
Strictly speaking, this is a **breaking change**, because it adds a requirement to the (non-plumbing) `QueryTable` struct, but I can't imagine any clients being affected.

matklad:
lgtm

PR #97, opened 2018-12-18
WIP: Make GC API more orthogonal and flexible

matklad:
WIP: will add tests later.

The idea here is to make gc API more explicit and orthogonal (you need to explicitly say what you are collecting, by default gc is a no-op), as well as more flexible (we can now collect all keys, not only "unused" ones). Does this sound reasonable? :)

### DIFF
--- a/src/debug.rs
+++ b/src/debug.rs
@@ -3,7 +3,6 @@
 
 use crate::plumbing;
 use crate::plumbing::QueryStorageOps;
-use crate::Database;
 use crate::Query;
 use crate::QueryTable;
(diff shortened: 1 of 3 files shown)


matklad:
In rust-analyzer, I plan to use 

`SweepStragegy::default().discard_values().discard_used()` to collect concrete syntax trees.

matklad:
Let's close this for now: I still *think* that this API will fit me better in Rust Analyzer, but it's better to get an answer to "how much garbage is there" before attempting to tweak garbage collection.

PR #105, opened 2019-01-04
Issue 66 transitive cancelation

### LINKED ISSUE #66: Canceled queries are not removed from the database after revision bump

nikomatsakis:
An alternate approach to fix #66. The basic problem is that `is_current_revision_canceled` was treated as observing *nothing* when it returned false, but that is only valid if there has never been a value of *true* observed. This PR revises how we track cancellation, which can actually be somewhat simpler since we no longer have multiple active writers to the database. We now track the revisions where cancellation occurs -- so, if revision R2 was canceled, then when you get back `false` from `is_current_revision_canceled`, that is treated as a read from R3.

Essentially, the 'next uncanceled revision' acts like an ordinary input -- it records the last time it was changed, and when you read from it, you see the corresponding dependency. There are however two differences between this and an ordinary input:

- It can change value asynchronously.
- We don't record a full dependency: so if you observed false in R1, and then R2 is canceled, and then R3 checks your value, you are not considered invalidated, as you would be with a regular input (since that input would've changed in R2).

r? @matklad 

Fixes #66

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -276,6 +276,10 @@ where
 
             local_state.query_stack.pop().unwrap()
         };
+        // Transitively propagete untracked.
+        if subqueries.is_none() {
+            self.report_untracked_read()
+        }
 
         ComputedQueryResult {
             value,
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -414,7 +414,10 @@ where
         fmt.debug_struct("SharedState")
             .field("query_lock", &query_lock)
             .field("revision", &self.revision)
-            .field("pending_revision_increments", &self.pending_revision_increments)
+            .field(
+                "pending_revision_increments",
+                &self.pending_revision_increments,
+            )
             .finish()
     }
 }
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -276,10 +276,6 @@ where
 
             local_state.query_stack.pop().unwrap()
         };
-        // Transitively propagete untracked.
-        if subqueries.is_none() {
-            self.report_untracked_read()
-        }
 
         ComputedQueryResult {
             value,
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -159,6 +159,14 @@ where
         }
     }
 
+    /// Read current value of the revision counter.
+    #[inline]
+    fn pending_revision(&self) -> Revision {
+        Revision {
+            generation: self.shared_state.pending_revision.load(Ordering::SeqCst) as u64,
+        }
+    }
+
     /// Check if the current revision is canceled. If this method ever
     /// returns true, the currently executing query is also marked as
     /// having an *untracked read* -- this means that, in the next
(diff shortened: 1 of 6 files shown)


src/runtime.rs
```diff
@@ -374,12 +421,9 @@ struct SharedState<DB: Database> {
...
 
-    /// Counts the number of pending increments to the revision
-    /// counter. If this is non-zero, it means that the current
-    /// revision is out of date, and hence queries are free to
-    /// "short-circuit" their results if they learn that. See
-    /// `is_current_revision_canceled` for more information.
-    pending_revision_increments: AtomicUsize,
+    /// If revision N is the last revision to be cancelled, then this
+    /// stores N+1. Stores 0 if there have been no cancellations.
+    next_uncanceled_revision: AtomicUsize,
```
matklad: Hm, I am slightly confused here... Am I correct that `next_uncanceled_revision` may be *earlier* then revision? The `next_` word sort-of makes me think that this is necessary a revision in the future. Perhaps flip this to `last_cancelled_revision`?

src/runtime.rs
```diff
@@ -501,6 +545,10 @@ impl<DB: Database> ActiveQuery<DB> {
         self.changed_at.is_constant = false;
         self.changed_at.revision = changed_at;
     }
+
+    fn add_anon_read(&mut self, changed_at: Revision) {
+        self.changed_at.revision = changed_at;
```
matklad: Why this is not a `self.changed_at.revision = self.changed_at.revision.max(changed_at);`?

This seems like a bug, if a query checks for cancellation not on entry, but in the middle.

matklad:
Verified that this works with rust-analyzer! Though, https://github.com/salsa-rs/salsa/pull/105#discussion_r245332436 seems like a bug which is dodged by ra :)

src/runtime.rs
```diff
@@ -190,46 +234,35 @@ where
     /// method will also increment `pending_revision_increments`, thus
     /// signalling to queries that their results are "canceled" and
     /// they should abort as expeditiously as possible.
+    ///
+    /// Note that, given our writer model, we can assume that only one
+    /// thread is attempting to increment the global revision at a
+    /// time.
```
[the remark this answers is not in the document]
nikomatsakis: I was wondering how to do that, yeah. We can't use `&mut self` to prove it, unfortunately.

src/runtime.rs
nikomatsakis: I debated a lot about what to name it. `last_canceled_revision` isn't quite right, because it stores N+1 (where N is the last canceled revision). That's important because otherwise the initial value would suggest that R0 was canceled. I could make R0 not a valid revision (and just start the revisions at R1). Maybe I should just do that.

src/runtime.rs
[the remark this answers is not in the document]
matklad: assert!(`lock.try_read().is_some()`)?

src/runtime.rs
matklad: Making initial revision 1 seems natural.

src/runtime.rs
nikomatsakis: um yeah that is wrong

nikomatsakis: copy and paste error, good catch

nikomatsakis: I should make a test that exposes it :)

nikomatsakis: actually, I think I already did, I was wondering why this local test I have wasn't passing

nikomatsakis: Ah, no, the code is wrong, but I don't think it's **observably** wrong. This is because `next_uncanceled_revision`, in fact, is always equal to either the current revision or the next one -- it is updated on every revision bump, since we don't know whether anyone is listening or not. I don't actually think this is harmful, although it leads to higher "changed at" numbers (or would, without "backdating").

I'll maybe rename it to `pending_revision` though

src/runtime.rs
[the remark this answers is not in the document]
nikomatsakis: the new version has some asserts that would likely fail if there were concurrent threads invoking `with_incremented_revision`

PR #107, opened 2019-01-10
Unwinding-based cancelation via panic hook

matklad:
This adds unwinding based cancellation (user is responsible for both panicking and catching). The current `anon_read` infrastructure is not deleted yet. We probably should tear it down though?

Tests are rewritten to use panics, except for the one which does not make sense with panicking. This test still uses the old-style cancelation.

Future work:

* either remove anon-read, or make sure we test both anon-reads and unwinding
* improve PanicGuard such that it restores old value, thus achieving strong exception safety guarantee.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -104,6 +104,12 @@ pub trait Database: plumbing::DatabaseStorageTypes + plumbing::DatabaseOps {
     fn salsa_event(&self, event_fn: impl Fn() -> Event<Self>) {
         #![allow(unused_variables)]
     }
+
+    /// This function is invoked when a depndent query is being computed by the
+    /// other thread, and that thread panics.
+    fn on_propagated_panic(&self) -> ! {
+        panic!("concurrent salsa query paniced")
+    }
 }
 
 /// The `Event` struct identifies various notable things that can
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -105,10 +105,10 @@ pub trait Database: plumbing::DatabaseStorageTypes + plumbing::DatabaseOps {
         #![allow(unused_variables)]
     }
 
-    /// This function is invoked when a depndent query is being computed by the
+    /// This function is invoked when a dependent query is being computed by the
     /// other thread, and that thread panics.
     fn on_propagated_panic(&self) -> ! {
-        panic!("concurrent salsa query paniced")
+        panic!("concurrent salsa query panicked")
     }
 }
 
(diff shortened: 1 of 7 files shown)


matklad:
Corresponding PR to rust-analyzer: https://github.com/rust-analyzer/rust-analyzer/pull/478

src/lib.rs
```diff
@@ -104,6 +104,12 @@ pub trait Database: plumbing::DatabaseStorageTypes + plumbing::DatabaseOps {
     fn salsa_event(&self, event_fn: impl Fn() -> Event<Self>) {
         #![allow(unused_variables)]
     }
+
+    /// This function is invoked when a depndent query is being computed by the
```
nikomatsakis: Nit: "depndent" => dependent

src/lib.rs
```diff
@@ -104,6 +104,12 @@ pub trait Database: plumbing::DatabaseStorageTypes + plumbing::DatabaseOps {
...
+    /// other thread, and that thread panics.
+    fn on_propagated_panic(&self) -> ! {
+        panic!("concurrent salsa query paniced")
```
nikomatsakis: Nit: paniced => panicked

src/runtime.rs
```diff
@@ -223,6 +225,16 @@ where
         }
     }
 
+    /// Calls the callback if the current revision is cancelled.
+    ///
+    /// Observing cancellation may lead to inconsistencies in database storage,
+    /// so the callback must panic.
+    pub fn if_current_revision_is_canceled(&self, cb: fn() -> !) {
```
nikomatsakis: I'm not sure this method is really *needed* -- I had sort of expected to just do something like this on the client side

```
if self.is_current_revision_canceled() {
  panic!()
}
```

But I guess the point of view this is to avoid adding any reads etc?

I'd still (I think) like to keep the `is_current_revision_canceled` infrastructure, since it's really not much complexity and people might not want to `panic!`.

src/runtime.rs
```diff
@@ -35,6 +35,8 @@ pub struct Runtime<DB: Database> {
     shared_state: Arc<SharedState<DB>>,
 }
 
+impl<DB> std::panic::RefUnwindSafe for Runtime<DB> where DB: Database {}
```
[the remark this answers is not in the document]
matklad: Ow, right, storage contains user-defined types, and they are not necessarily unwind safe. I can’t implement this *just* for storage(runtime has refcell), but conditional impl would word.

src/runtime.rs
matklad: Yeah, if we keep the current infra, `is_current_revision_canceled` will be enough. If we *remove* current infra, than this API will be better, because it prevents you from being accidentally non-determentistic.

matklad:
Fixed! Will be cool to have a release with this API

PR #108, opened 2019-01-12
Implement `query_group` as procedural macro

fabianschuiki:
Replace the `query_group` macro with a procedural macro implementation. This changes the way query groups are declared from
```rust
salsa::query_group! {
    trait MyGroup { ... }
}
```
to
```rust
#[salsa::query_group]
trait MyGroup { ... }
```

Within the trait being converted into a query group, the following additional attributes on methods are supported:

- `#[salsa::{memoized, volatile, dependencies, input}]` to replace the old `storage ...;`
- `#[salsa::invoke(...)]` to replace the old `use fn ...;`
- `#[salsa::query_type(...)]` to replace the old `type ...;`

All these are optional. Query type defaults to the camel-cased method name with an appended `Query`.

**Todo:**
- [x] ~~I'm not convinced by the last commit though. Maybe we should rather add `#[salsa::query_type(...)]` everywhere to maintain the old query type names and have less replacements in the examples?~~
- [x] ~~Add generics?~~
- [x] ~~Make database storage macro procedural as well?~~
- [x] Drop `&mut self` methods
- [x] ~~Add negative doctest that checks for compilation error on missing `&self`~~ Haven't found a way to ensure compilation failure; just `should_panic`. Skip this for now.
- [x] Remove `tests/proc_macros.rs`
- [x] Error if `#[salsa::invoke]` and `#[salsa::input]` are set on the same method


nikomatsakis:
@fabianschuiki very cool! Sorry I've no been able to comment yet

tests/proc_macros.rs
```diff
@@ -0,0 +1,50 @@
+use salsa::Database;
+use std::sync::Arc;
+
+/// My awesome db trait.
```
nikomatsakis: Seems like at this point this doesn't test anything that isn't covered by other tests, right?

nikomatsakis:
Reading your "to do" list, my thoughts:

> I'm not convinced by the last commit though. Maybe we should rather add `#[salsa::query_type(...)]` everywhere to maintain the old query type names and have less replacements in the examples?

I'd rather we adopt the "new convention", as you did. It's good that we have some tests though (as you did).

> Add generics?

I'd rather leave that for later. I think we should transition to the "group-based model" first.

> Make database storage macro procedural as well?

Again, seems like it should be a separate PR to me =)

This version stands alone.

tests/proc_macros.rs
fabianschuiki: Good point, I'll remove this.

fabianschuiki:
> This version stands alone.
Okay that sounds like a plan. So let's defer further additions like generics and other proc macros and get this merged.

PR #114, opened 2019-01-18
Clear query stack on panic

nikomatsakis:
Previously, the local query stack was not recovered if there was a panic. Fix that.

cc @kleimkuhler

### DIFF
--- a/src/runtime/local_state.rs
+++ b/src/runtime/local_state.rs
@@ -28,19 +28,13 @@ impl<DB: Database> Default for LocalState<DB> {
 }
 
 impl<DB: Database> LocalState<DB> {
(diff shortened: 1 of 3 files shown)


src/runtime/local_state.rs
```diff
@@ -0,0 +1,120 @@
...
+        query
+    }
+}
+
+impl<'me, DB> Drop for ActiveQueryGuard<'me, DB>
+where
+    DB: Database,
+{
+    fn drop(&mut self) {
+        self.pop_helper();
```
matklad: Probably an overkill, but we can add an assertion here:

```
if !std::thread::panking() {
    panic!("active query was not completed")
}
```

PR #115, opened 2019-01-18
move the RefUnwindSafe impls to shared/local state structs

nikomatsakis:
in general, the more narrowly we can make the claim, the easier it is to verify

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -37,13 +37,6 @@ pub struct Runtime<DB: Database> {
     shared_state: Arc<SharedState<DB>>,
 }
 
-impl<DB> std::panic::RefUnwindSafe for Runtime<DB>
-where
-    DB: Database,
-    DB::DatabaseStorage: std::panic::RefUnwindSafe,
-{
-}
-
 impl<DB> Default for Runtime<DB>
 where
     DB: Database,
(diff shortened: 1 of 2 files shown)


src/runtime.rs
```diff
@@ -446,6 +439,8 @@ struct SharedState<DB: Database> {
     dependency_graph: Mutex<DependencyGraph<DB>>,
 }
 
+impl<DB> std::panic::RefUnwindSafe for SharedState<DB> where DB: Database {}
```
matklad: This is missing `DB::DatabaseStorage: std::panic::RefUnwindSafe,`?

src/runtime/local_state.rs
```diff
@@ -80,6 +80,8 @@ impl<DB: Database> LocalState<DB> {
     }
 }
 
+impl<DB> std::panic::RefUnwindSafe for LocalState<DB> where DB: Database {}
```
matklad: And this I think wants `DB::QueryDescriptor: RefUnwindSafe`

src/runtime.rs
nikomatsakis: Ah, yeah, you're right.

nikomatsakis: I forgot that this was stored in the shared-state

matklad: Should we have a compile-fail test for non refunwind safe data?

Or perhaps we should just bound keys & values by `RefUnwindSafe`? Doesn't "keys&values must not contain observable interior mutability" implies unwind safety?

nikomatsakis: Debatable. We always clone the values that we give back to you etc. I suppose you could put "mutable shared state" in there -- it'd just be very, very wrong.

nikomatsakis: So yeah this:

> Doesn't "keys&values must not contain observable interior mutability" implies unwind safety?

I think this is true

PR #116, opened 2019-01-21
add #[doc(hidden)] to __SalsaQueryDescriptor

cormacrelf:
The diff says it all. Currently it sits around polluting docs listings, but it could just be hidden.


nikomatsakis:
Seems good, thanks!

PR #119, opened 2019-01-21
Make `database_storage` a proc macro

### LINKED ISSUE #118: Convert `database_storage` macro to a procedural macro
As part of https://github.com/salsa-rs/salsa-rfcs/pull/1, we need to convert the `salsa::database_storage` macro to a procedural macro, presumably living alongside the existing `salsa::query_group` macro:

https://github.com/salsa-rs/salsa/blob/af48c409e37c257a8fec9caa73e68f9de27719d1/components/salsa-macros/src/lib.rs#L124-L125

The idea here is to retain roughly the current input format (in particular, we will still list out individual queries). We can make some changes however. Currently `salsa::database_storage` looks like this:

https://github.com/salsa-rs/salsa/blob/af48c409e37c257a8fec9caa73e68f9de27719d1/examples/hello_world/main.rs#L75-L82

The `DatabaseStorage` struct name is never explicitly used by the user, though, so I would propose that we auto-generate it. The rest could stay as is. One challenge however is that we can't use `syn` to readily parse such a setup, since it doesn't match Rust syntax. What to do about this? I'm not sure how easy it is to parse alternative structures in syn, but we should feel free to adapt the syntax to make it easier to parse. 

Discuss in the [#rfcs/salsa-query-group](https://salsa.zulipchat.com/#narrow/stream/181542-rfcs.2Fsalsa-query-group) stream on Zulip, topic is [`database_storage` macro (#118)](https://salsa.zulipchat.com/#narrow/stream/181542-rfcs.2Fsalsa-query-group/topic/.60database_storage.60.20macro.20(.23118))

nikomatsakis:
Joint work with @cormacrelf 

Fixes #118


nikomatsakis:
cc @fabianschuiki

PR #122, opened 2019-01-22
allow to peek at values via debug query interface

matklad:
I'd love to answer the question "how much syntax trees are retained in the salsa database", and for that, peeking at values seems useful.

### DIFF
--- a/src/debug.rs
+++ b/src/debug.rs
@@ -14,6 +14,9 @@ pub trait DebugQueryTable {
     /// Key of this query.
     type Key;
 
+    /// Value of this query.
+    type Value;
+
     /// True if salsa thinks that the value for `key` is a
     /// **constant**, meaning that it can never change, no matter what
     /// values the inputs take on from this point.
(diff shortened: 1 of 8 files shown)


src/debug.rs
```diff
@@ -23,6 +26,30 @@ pub trait DebugQueryTable {
...
+        C: FromIterator<TableEntry<Self::Key, Self::Value>>;
+}
+
+/// An entry from a query table, for debugging and inspecting the table state.
+pub struct TableEntry<K, V> {
+    /// key of the query
+    pub key: K,
+    /// value of the query, if it is stored
+    pub value: Option<V>,
+    _for_future_use: (),
```
matklad: We probably want to expose `is_const`, `verified_at`, etc here, but, as I don't have immediate need for those, I've ommited them.

src/debug.rs
```diff
@@ -23,6 +26,30 @@ pub trait DebugQueryTable {
...
+    where
+        C: FromIterator<TableEntry<Self::Key, Self::Value>>;
+}
+
+/// An entry from a query table, for debugging and inspecting the table state.
+pub struct TableEntry<K, V> {
+    /// key of the query
+    pub key: K,
+    /// value of the query, if it is stored
+    pub value: Option<V>,
```
nikomatsakis: At first I was surprised not to see references here, but given that salsa really encourages your key/value types to be clone-able, makes sense.

matklad:
Hm, should we remove `keys` as well? Entries are strictly more powerful.

src/debug.rs
matklad: Yep, this whole setup with `FromIterator` instead of returning an iterator of references is unusual, but that does look like the best solution, given internal mutexes and the fact that this is `debug`.

matklad:
rewritten tests to just use entries

Issue #120, opened 2019-01-21
introduce "group" structs in place of query structs


nikomatsakis:
Now that we have converted from macro-rules macros to procedural macros (#108, #118), the next step in implementing https://github.com/salsa-rs/salsa-rfcs/pull/1 is probably to try and convert from the current setup, where we have a "struct per query", to the setup of a "struct per query **group** that is described in the RFC.

I'll try to update the issue with a vague plan in a bit.

nikomatsakis:
As a first step we can push more of the details into the query group without making the full transition:

- [ ] We need to create the [**group descriptor enum**](https://github.com/nikomatsakis/salsa-rfcs/blob/query-group-trait/RFC0001-Query-Group-Traits.md#the-group-descriptor) from `salsa::query_group`
- [ ] We need to create the [**group descriptor storage struct**](https://github.com/nikomatsakis/salsa-rfcs/blob/query-group-trait/RFC0001-Query-Group-Traits.md#the-group-descriptor) from `salsa::query_group`
- [ ] Modify the [`DatabaseStorageTypes` impl](https://github.com/nikomatsakis/salsa-rfcs/blob/query-group-trait/RFC0001-Query-Group-Traits.md#database-storage-types-impl) so that the storage type is a tuple of the group storage
- [ ] Modify the [existing `GetStorageTable` implementations](https://github.com/salsa-rs/salsa/blob/3e81f12dd28bc35819db47cd50b31a972bb6a487/components/salsa-macros/src/database_storage.rs#L165-L199) so that they access the query data using `storage().N.#query` where `N` is the index of the group in the list, instead of the [`storage().#query`](https://github.com/salsa-rs/salsa/blob/3e81f12dd28bc35819db47cd50b31a972bb6a487/components/salsa-macros/src/database_storage.rs#L172-L174) that they use today.

I think if we make those changes, everything should still work as it does today but we are closer to the new design.

nikomatsakis:
I started work on this first chunk. Currently in https://github.com/nikomatsakis/salsa/tree/group-struct branch

nikomatsakis:
Next steps once #123 lands is probably to refactor the group descriptor enum (again, with goal of keeping everything as it works today as far as user is concerned):

- [ ] Uncomment the ["group descriptor" enum](https://github.com/nikomatsakis/salsa-rfcs/blob/query-group-trait/RFC0001-Query-Group-Traits.md#the-group-descriptor) code I added in #123 =)
[...]

PR #124, opened 2019-01-24
remove the need to list individual queries in group storage struct

nikomatsakis:
By the end of this PR, the `database_storage` macro format just lists out the query group *traits*, not the queries within:

```rust
salsa::database_storage! {
    pub MyDatabaseStruct {
        impl MyQueryGroup;
    }
}
```

I would like to move to a `#[salsa::database(...)]` annotation on the struct itself but that is left for a future PR. =) (Other bits of future work include generating the `set_foo` methods and so forth and (hopefully) removing the `query` and `query_mut` methods in the `Database` trait).

I had to do things just a *hair* differently than anticipated, so I need to update the RFC. The salsa `Query` trait grew "group storage" and "group descriptor" associated types and so forth. I think that trait really ought to wind up in `plumbing`.

cc #120

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -141,12 +141,12 @@ pub enum EventKind<DB: Database> {
     ///
     /// Executes before the "re-used" value is returned.
     DidValidateMemoizedValue {
-        /// The descriptor for the affected value. Implements `Debug`.
-        descriptor: DB::QueryDescriptor,
+        /// The database-key for the affected value. Implements `Debug`.
+        database_key: DB::DatabaseKey,
     },
 
     /// Indicates that another thread (with id `other_runtime_id`) is processing the
-    /// given query (`descriptor`), so we will block until they
+    /// given query (`database_key`), so we will block until they
     /// finish.
     ///
     /// Executes after we have registered with the other thread but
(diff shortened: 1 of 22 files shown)


src/lib.rs
```diff
@@ -348,6 +348,18 @@ pub trait Query<DB: Database>: Debug + Default + Sized + 'static {
...
+    type GroupStorage;
+
+    /// Generated descriptor for all queries in a group.
+    type GroupDescriptor;
+
+    /// Extact storage for this query from the storage for its group.
+    fn storage(group_storage: &Self::GroupStorage) -> &Self::Storage;
+
+    /// Create group descriptor for this query.
+    fn descriptor(key: Self::Key) -> Self::GroupDescriptor;
```
matklad: `fn group_descriptor`? I'd expect `Query::descriptor` to return `QureyDescriptor`

matklad: Oh, the fact that `QueryDescriptor` is **larger** than `GroupDescriptor` is also confusing :)

nikomatsakis: Heh, true. I suspect the names could use some work. OTOH, `QueryDescriptor` is a type end-users may actually interact with at some point, if only marginally.

nikomatsakis: Maybe the names should be `GroupKey` and `DatabaseKey`

nikomatsakis: these are all basically "upcasts" of key from a wider context -- a single query, to a group of queries, to the full database.

matklad: I am fine with punting on this until we completely migrate to the new design, and *then* doing a one off run through all of the API to adjust trait names, visibility, etc.

PR #125, opened 2019-01-25
adopt `salsa::database` attribute macro

nikomatsakis:
With this change, you write

```rust
#[salsa::database(QueryGroup1, QueryGroup2)]
struct MyDatabase { .. }
```

instead of `salsa::database_storage! { ... }`.

cc https://github.com/salsa-rs/salsa/issues/120

### DIFF
--- a/examples/hello_world/main.rs
+++ b/examples/hello_world/main.rs
@@ -56,6 +56,7 @@ fn length(db: &impl HelloWorldDatabase, (): ()) -> usize {
 
 // Define the actual database struct. This must contain a salsa
 // runtime but can also contain anything else you need.
+#[salsa::database(HelloWorldDatabase)]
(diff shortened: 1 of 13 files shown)


examples/hello_world/main.rs
```diff
@@ -56,6 +56,7 @@ fn length(db: &impl HelloWorldDatabase, (): ()) -> usize {
 
 // Define the actual database struct. This must contain a salsa
 // runtime but can also contain anything else you need.
+#[salsa::database(HelloWorldDatabase)]
 #[derive(Default)]
```
matklad: This is an example, so short comment explaining the attribute would be helpful

nikomatsakis: good catch

PR #130, opened 2019-01-25
Introduce "query group struct"

nikomatsakis:
As @matklad noted, having the `#[salsa::database]` macro find the associated group storage / group key structs by manipulating the path proved to be too fragile, thus making the salsa abstraction more leaky than we might like. To fix this, we introduce the idea of a **query group struct** -- this struct now defines the query group, rather than the trait. You specify the struct name in the `#[salsa::query_group]` attribute macro:

```rust
#[salsa::query_group(MyQueryGroup)]
trait MyQueryGroupDatabase {
    ...
}
```

As you can see, by convention the name of the struct is the name of the query group (`MyQueryGroup`) and the trait has the word `Database` attached to it (i.e., it is implemented by the database which supports this query group).

Now, the `#[salsa::database]` struct refers to the *query group struct*, not the trait:

```rust
#[salsa::database(MyQueryGroup)]
```

This does require a bit more boilerplate, but it's probably better than relying on the terrible string munging we were doing.

### DIFF
--- a/components/salsa-macros/src/database_storage.rs
+++ b/components/salsa-macros/src/database_storage.rs
@@ -1,6 +1,5 @@
 use heck::SnakeCase;
 use proc_macro::TokenStream;
-use proc_macro2::Span;
 use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
 use syn::{Ident, ItemStruct, Path, Token};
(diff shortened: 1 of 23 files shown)


matklad:
Looks like the tests actually use the wrong convention? I would prefer the struct to be longer than a trait. What about

HirDatabase 
HirDatabaseStorage

? It makes sense to add storage to the struct, even though it does a more than just store things

nikomatsakis:
I wasn't sure what convention to use. I agree you will use the trait name more often, so it makes some sense for it to be shorter. Hmm.

nikomatsakis:
ok, @matklad, I tried out a sample explanation above and it felt pretty ok (see final commit).

nikomatsakis:
I didn't update all the tests and don't have time for that now, but we can presumably merge without that and file an issue for later.

matklad:
LGTM!

PR #132, opened 2019-01-25
Release salsa-0.10.0-alpha4

nikomatsakis:
- Introduce requirement for group storage structs (#130)
- Manually implement Default (#131)

Contributors to this release:

- @cormacrelf
- @nikomatsakis

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "salsa"
-version = "0.10.0-alpha3"
+version = "0.10.0-alpha4"
 authors = ["Niko Matsakis <niko@alum.mit.edu>"]
 edition = "2018"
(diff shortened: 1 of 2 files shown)


cormacrelf:
Hm. Docs.rs failed to build. https://docs.rs/crate/salsa/0.10.0-alpha4/builds/139121

nikomatsakis:
It works on stable for me

cormacrelf:
Docs.rs runs nightly, but seems to be failing on uniform paths, which afaik were stable as of 1.32. Possible nightly regression.

nikomatsakis:
So, I've had this problem on nightly, but only on older nightlies. I was however stuck on an older nightly because RLS wasn't building.

PR #135, opened 2019-01-26
apply `non_camel_case_types` to the enum variant

nikomatsakis:
Not sure why this isn't signaling warnings all over the place, but I see them in other projects.


cormacrelf:
It was showing warnings for me using cargo-expand, but not check/build/test/etc.

PR #146, opened 2019-01-30
introduce the basic structure of a book (very much a WIP)

nikomatsakis:


matklad:
:+1: 

Though, I am worried here that the book might quickly bitrot, given the pace of change.

A possible strategy is to have a sort-of "minimal" book, and defer the discussion of specific APIs to doc-comments. 

Doc-comments are less susceptible to bit-rot, and having a book setup in place is an important capital cost, which makes actually adding the content much easier in the future.

nikomatsakis:
It's true, it may be a bit early, but I suspect a lot of the content I have in mind will still be relevant no matter what we do. (Also, one nice feature of the mdbook setup is that we can test the code samples, so that should help with bitrot)

nikomatsakis:
I should push some commits to do more travis integration I guess

PR #147, opened 2019-02-03
feat: Allow queries to avoid panics on cycles

Marwes:
The default behavior when salsa detects a cycle is to panic. This is fine behavior when the cycle occurs as a programmer error but in some instances cycles can be expected to occur. To allow this case while still keeping "programmer error" as the default assumption on cycles a `salsa::cycle` is added which can be attached to any function which could naturally appear in a cycle.

The `salsa::cycle` takes a function which takes the keys of the query as well as description of all the queries that were part of the cycle (currently a slice of `Debug` formatted strings for simplicity's sake). When a cycle is detected salsa invokes the cycle function to produce the result for the query instead of panicking. Furthermore, every other query in the cycle will discard whatever value they produced normally and invoke their own cycle function (if a function in the cycle lacks the attribute salsa will panic as usual). All values produced from a cycle function is memoized in the same way as if the query executed normally.

This has been tested to work well in gluon's import system which needs to error if modules import each other in a cycle (a programmer error for the one writing gluon programs, but not in the salsa query usage in gluon) https://github.com/gluon-lang/gluon/pull/683 . The only nasty part is that the cycle information needs to be parsed out of the `Debug` formatted string but currently I don't have a good abstraction for solving that.

cc #6

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -1,15 +1,15 @@
 use crate::dependency::DatabaseSlot;
 use crate::dependency::Dependency;
 use crate::durability::Durability;
+use crate::plumbing::CycleDetected;
 use crate::revision::{AtomicRevision, Revision};
-use crate::{Database, Event, EventKind, SweepStrategy};
+use crate::{CycleError, Database, Event, EventKind, SweepStrategy};
 use log::debug;
-use parking_lot::{Mutex, RwLock};
 use parking_lot::lock_api::{RawRwLock, RawRwLockRecursive};
+use parking_lot::{Mutex, RwLock};
 use rustc_hash::{FxHashMap, FxHasher};
 use smallvec::SmallVec;
-use std::fmt::Write;
-use std::hash::BuildHasherDefault;
+use std::hash::{BuildHasherDefault, Hash};
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 
@@ -355,6 +355,7 @@ where
             dependencies,
             changed_at,
             durability,
+            cycle,
             ..
         } = active_query.complete();
 
@@ -363,6 +364,7 @@ where
             durability,
             changed_at,
             dependencies,
+            cycle,
         }
     }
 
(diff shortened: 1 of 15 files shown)


nikomatsakis:
Argh, @Marwes, you opened this PR at just the wrong time to get timely feedback. Sorry about that. Maybe we should discuss over on Zulip -- [I just created a topic here](https://salsa.zulipchat.com/#narrow/stream/145099-general/topic/cyclic.20queries). I find that better for conversation (and far easier to monitor than GitHub!).

Marwes:
@nikomatsakis  Updated the PR description with an explanation and motivation.

Marwes:
Have you had a chance to look at this @nikomatsakis ? Any thoughts?

Marwes:
Rebased. For the most part I only needed to move the changes from derived.rs to derived/slot.rs .

bjorn3:
I think you did something wrong during the rebase. This contains unrelated commits.

Marwes:
Yeah maybe, but no idea what. The most recent unrelated commit points to the current master https://github.com/salsa-rs/salsa/pull/147/commits/4815e147961c2a17e165e40c226ae048aaf874af so I don't really know what is going on :/

Marwes:
Squashed the commits so I could force push and they are now gone 🤷‍♂

Marwes:
@nikomatsakis Have you had a chance to look at this again?

nikomatsakis:
@Marwes 

Argh, yeah I actually did spend some time on this. I got a bit distracted trying to think what the "right" semantics are for cycles and if we can do better than "controlled errors" of this kind, but I suppose it would be good to move forward with *something*.

Anyway, I also started a branch that was implemented a bit differently. What I was trying to do was to have each active query have an associated `Arc` with a pointer to the next active query on the stack (i.e., what it is blocked on). This way you could follow a chain of queries forward across threads pretty cleanly and lazily. I seem to recall that the code in this PR was a bit more complex -- but then I don't think I've looked at it since you rebased. 

I will try to put some time into it this weekend or early next week!

Marwes:
Yeah the cycle reconstruction isn't that simple. The cost of it is mostly paid when an cycle happens however so it may be faster than a list of `Arc`s.

nikomatsakis:
OK, so obviously I don't have time to follow up here in a timely fashion. I think at this point we should  merge this PR and I can do whatever experimentation I want to do (or not do, as the case may be) on master. 

I think my reluctance is a few things:

- I'm curious to explore an alternate impl strategy, mostly for simplicity/perf, though I can't remember all the things I had in mind there.
- I want to find a better approach to cycle handling.

But I don't see how landing this branch really impedes either of those things. It probably just helps.

src/runtime.rs
```diff
@@ -634,16 +713,26 @@ pub(crate) struct StampedValue<V> {
     pub(crate) changed_at: Revision,
 }
 
-struct DependencyGraph<DB: Database> {
+#[derive(Debug)]
+struct Edge<K> {
+    id: RuntimeId,
+    path: Vec<K>,
```
nikomatsakis: OK, so we are creating this vector eagerly (and repeatedly) for each cross-thread call. I'm *still* inclined to merge the PR, but I suspect this was one of the things that motivated me to want to examine an alternative impl strategy.

nikomatsakis: For that matter, though, I still think the "parallel" support is kind of weak in salsa and an area for further development. =)

Marwes: True, been a while since I wrote it but since it takes the path from the local state (which is not thread safe) I couldn't think of a way to handle it lazily.

Issue #152, opened 2019-03-26
Query values still require `Send + Sync`, despite PR #42


regexident:
Despite https://github.com/salsa-rs/salsa/pull/42 it seems like salsa still required query values of type `T` to be `T: Send + Sync`.

Minimal failing example:

```rust
#![feature(optin_builtin_traits)]

#[macro_use]
extern crate salsa;

#[derive(Clone, PartialEq, Eq, Debug)]
struct NoSend;

impl !Send for NoSend {}
impl !Sync for NoSend {}

#[salsa::query_group(NoSendStorage)]
trait NoSendDatabase: salsa::Database {
    fn no_send(&self, key: ()) -> NoSend;
}

fn no_send(db: &impl NoSendDatabase, (): ()) -> NoSend {
    NoSend
}

#[salsa::database(NoSendStorage)]
#[derive(Default)]
struct DatabaseStruct {
    runtime: salsa::Runtime<DatabaseStruct>,
}

impl salsa::Database for DatabaseStruct {
    fn salsa_runtime(&self) -> &salsa::Runtime<DatabaseStruct> {
        &self.runtime
    }
}

fn main() {
    let mut db = DatabaseStruct::default();

    println!("{:?}", db.no_send(()));
}
```

Errors:

```
error[E0277]: `NoSend` cannot be sent between threads safely
  --> src/main.rs:14:1
   |
14 | #[salsa::query_group(NoSendStorage)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `NoSend` cannot be sent between threads safely
```

```
error[E0277]: `NoSend` cannot be shared between threads safely
  --> src/main.rs:14:1
   |
14 | #[salsa::query_group(NoSendStorage)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `NoSend` cannot be shared between threads safely
```

<details>

```
error[E0277]: `NoSend` cannot be sent between threads safely
  --> src/main.rs:14:1
   |
14 | #[salsa::query_group(NoSendStorage)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `NoSend` cannot be sent between threads safely
   |
[...]

PR #153, opened 2019-03-26
Removed remaining stray bounds for `Send + Sync` that still survived PR #42 

regexident:
This removes the last remaining uses of `: Send + Sync`, fixing https://github.com/salsa-rs/salsa/issues/152.


regexident:
Fair point. Done.

regexident:
Now it works with `!Send + !Sync` keys and values. 🥳

PR #154, opened 2019-03-27
release 0.11.0

nikomatsakis:
Changes:

- Add interned keys to salsa (#150) -- see salsa-rs/salsa-rfcs#2
  for more details.
- Fix bugs re: GC and volatile queries
- Optimization for GC (#144), though I think this later got undone
  as part of the bug fixes =)


nikomatsakis:
Released.

PR #155, opened 2019-03-27
make sure interned state is send

matklad:


tests/interned.rs
```diff
@@ -12,6 +12,14 @@ impl salsa::Database for Database {
     }
 }
 
+impl salsa::ParallelDatabase for Database {
```
matklad: This doesn't work without a patch, because rustc thinks that `Database` contains an `Runtime<Database>`, which has `RefCell`

PR #157, opened 2019-03-30
adopt raw-id for interned keys

nikomatsakis:
Per @matklad's suggestion, create a `RawId` type and use it in place of `u32` for interned keys.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -3,6 +3,7 @@ use crate::plumbing::CycleDetected;
 use crate::plumbing::HasQueryGroup;
 use crate::plumbing::QueryStorageMassOps;
 use crate::plumbing::QueryStorageOps;
+use crate::raw_id::RawId;
 use crate::runtime::ChangedAt;
 use crate::runtime::Revision;
 use crate::runtime::StampedValue;
@@ -46,7 +47,7 @@ where
 
 struct InternTables<K> {
     /// Map from the key to the corresponding intern-index.
-    map: FxHashMap<K, InternIndex>,
+    map: FxHashMap<K, RawId>,
 
     /// For each valid intern-index, stores the interned value. When
     /// an interned value is GC'd, the entry is set to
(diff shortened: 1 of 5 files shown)


matklad:
LGTM

nikomatsakis:
I added two changes:

- we do not permit `RawId` to be converted to a `u32`, only `usize` -- seems more forward compat?
- adopt `NonZeroU32` internally -- why not?

matklad:
>we do not permit RawId to be converted to a u32, only usize -- seems more forward compat?

Is there a reason why we might want something bigger than `u32` in the future? If not, I'd stick with exposing `u32`: otherwise we might be just pushing unchecked `as` casts to the clients.

nikomatsakis:
I was just debating about the u32 vs usize distinction myself, since I noticed I had forgotten to change the`MAX` constant to `usize`. I mean, I can't foresee a reason to use `usize` apart from the obvious one (lots and lots of data). I'll stick with u32 for now under the YAGNI principle.

matklad:
>(lots and lots of data) 

that would be16GB to store IDs themselves, since they are dense :)

nikomatsakis:
@matklad no, I mean, the point is not the amount of space the *indices* take, but the max number of values *you can intern*. I was imagining someday we might want a feature that uses `NonZeroUsize` instead of `NonZeroU32` as the representation of a `RawId`, for example.

nikomatsakis:
Or maybe I misunderstood your comment...

nikomatsakis:
Regardless, I agree that it's pretty unlikely to be a problem in practice. =)

nikomatsakis:
(Also, I suppose that, *in that unlikely event*, we could have a `#[salsa::usize_interned]` thing or something, rather than a feature.)

nikomatsakis:
OK, reverted =)

matklad:
I mean that, in order to run out of indices u32, we must, at some point, have 2**32 objects interned. That's indeed lots of data: at least 2**32 * size_of::<u32> to store the indices.

PR #162, opened 2019-04-12
Remove mention of unsupported '&mut self' in a query group panic

### LINKED ISSUE #161: salsa::query_group panicked regarding 'self' parameter
```toml
[dependencies]
salsa = "0.12.0"
```

Given the following:

```rust
#[salsa::query_group(ParserStorage)]
pub trait Parser: salsa::Database {
    #[salsa::input]
    fn file_text(&self, file_id: FileId) -> Text;

    #[salsa::input]
    fn file_names(&self) -> Vec<String>;

    #[salsa::input]
    fn diagnostics(&self, file_id: FileId) -> Vec<Diagnostic<FileSpan>>;

    fn file_tokens(&mut self, file_id: FileId) -> Vec<Token>;
    fn file_nodes(&mut self, file_id: FileId) -> Vec<Node>;
}

pub trait ParserExt: Parser {
    fn init(&mut self) {
        self.set_file_names(Default::default());
    }

    fn add_file(&mut self, file_name: impl Into<String>, file_contents: impl Into<Text>) -> FileId {
        let file_name = file_name.into();

        let mut file_names = self.file_names();
        let file_id = FileId(file_names.len());
        file_names.extend(Some(file_name));

        self.set_file_names(file_names);
        self.set_file_text(file_id, file_contents.into());

        file_id
    }
}

impl ParserExt for OraideDatabase {}

fn file_tokens(db: &mut impl Parser, file_id: FileId) -> Vec<Token> {
    let text = db.file_text(file_id);

    let mut tokenizer = parser::Tokenizer::new(file_id, &text);
    let tokens = tokenizer.by_ref().collect();

    let diags = tokenizer.take_diagnostics();
    db.set_diagnostics(file_id, diags);

    tokens
}

fn file_nodes(db: &mut impl Parser, file_id: FileId) -> Vec<Node> {
    let tokens = db.file_tokens(file_id);

    let mut parser = parser::Parser::new(tokens.into_iter());
    let nodes = parser.by_ref().collect();

    let diags = parser.take_diagnostics();
    db.set_diagnostics(file_id, diags);

    nodes
}

#[salsa::database(ParserStorage)]
pub struct OraideDatabase {
    rt: salsa::Runtime<Self>,
}

impl salsa::Database for OraideDatabase {
    fn salsa_runtime(&self) -> &salsa::Runtime<Self> {
        &self.rt
    }
}

impl Default for OraideDatabase {
    fn default() -> Self {
        let mut db = Self {
            rt: salsa::Runtime::default(),
        };

        db.init();
        db
    }
}
```

I get the following output after a `cargo check`:

```
error: custom attribute panicked
   --> components/oraide_miniyaml_salsafied/src/lib.rs:287:1
    |
287 | #[salsa::query_group(ParserStorage)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: message: first argument of query `file_tokens` must be `&self` or `&mut self`

error: aborting due to previous error
```

But, as you can see, `file_tokens` in the `Parser` trait has the following signature:

```rust
fn file_tokens(&mut self, file_id: FileId) -> Vec<Token>
```

The free-standing function has the following signature:

```rust
fn file_tokens(db: &mut impl Parser, file_id: FileId) -> Vec<Token>
```

Is this a type or logic error in my code or perhaps a logic error in `query_group`?

Phrohdoh:
Closes #161

I am not sure if this is entirely correct only because the code doesn't "read right" to me.

If the `SelfRef`'s `mutability` is `None` (so it is `&self`, I _assume_) panic saying you must use `&self`.

Is this correct?

### DIFF
--- a/components/salsa-macros/src/query_group.rs
+++ b/components/salsa-macros/src/query_group.rs
@@ -87,7 +87,7 @@ pub(crate) fn query_group(args: TokenStream, input: TokenStream) -> TokenStream 
                 match iter.next() {
                     Some(FnArg::SelfRef(sr)) if sr.mutability.is_none() => (),
                     _ => panic!(
-                        "first argument of query `{}` must be `&self` or `&mut self`",
+                        "first argument of query `{}` must be `&self`",
                         method.sig.ident
                     ),
                 }


components/salsa-macros/src/query_group.rs
```diff
@@ -87,7 +87,7 @@ pub(crate) fn query_group(args: TokenStream, input: TokenStream) -> TokenStream
                 match iter.next() {
                     Some(FnArg::SelfRef(sr)) if sr.mutability.is_none() => (),
                     _ => panic!(
```
nikomatsakis: Isn't there a way to give errors that are specific to a particular span? maybe not. It'd be nice if we could highlight the query or the `mut` keyword.

Ah, well, [I guess we could use this trick](https://internals.rust-lang.org/t/custom-error-diagnostics-with-procedural-macros-on-almost-stable-rust/8113), but that seems like a lot of work, and I wouldn't want to burden this PR with it.

PR #163, opened 2019-04-30
Add transparent query type

matklad:
Transparent queries are not really queries: they are just plain
uncached functions without any backing storage.

Making a query transparent can be useful to figure out if caching it
at all is a win.

Context:

I need this in rust-analyzer. I want to change the input to the system from files with `String` contents to a roughly pair of `(String, Mutex<Option<CachedSyntaxTree>>)`, such that I can implement LRU eviction of syntax trees outside of salsa. Currently `parse(FIleId) -> SyntaxTree` is a query, and I'd love to preserve this API. However, if `parse` will be caching trees in any way, that would completely undo any eviction I implement :-)

### DIFF
--- a/components/salsa-macros/src/query_group.rs
+++ b/components/salsa-macros/src/query_group.rs
@@ -35,7 +35,7 @@ pub(crate) fn query_group(args: TokenStream, input: TokenStream) -> TokenStream 
                     // Leave non-salsa attributes untouched. These are
                     // attributes that don't start with `salsa::` or don't have
                     // exactly two segments in their path.
-                    if is_salsa_attr_path(&attr.path) {
+                    if is_not_salsa_attr_path(&attr.path) {
                         attrs.push(attr);
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


matklad:
By the way, I am not exactly thrilled about "transparent" name, but I like it more than alternatives I came up with:

* `salsa::uncached`
* `salsa::zero`
* `salsa::epsilon`
* `salsa::passthru`
* `salsa::computed`

nikomatsakis:
I've definitely wanted this. I liked the ideas you were kicking about of trying to change the overall names to be more 'consistent' somehow.

matklad:
@nikomatsakis hm, yeah, it's true that almost all query names could be a touch more poetic...

Let's perhaps merge with this naming and do renaming separately, together with other queries, once we come up with a good *set* of names?

nikomatsakis:
In principle, I would like a simple RFC for this describing the motivations, but it seems simple enough I'm not inclined to press for it.

PR #165, opened 2019-05-05
allow private requirements in query groups

matklad:


tests/requires.rs
```diff
@@ -0,0 +1,66 @@
+//! Test that transparent (uncached) queries work
```
nikomatsakis: This comment looks out of date

tests/requires.rs
```diff
@@ -0,0 +1,66 @@
...
+        &self.runtime
+    }
+}
+
+#[test]
+fn require_clauses_work() {
+    use queries::{InputGroup, PubGroup};
+    let mut db = Database::default();
+
+    db.set_input(1, 10);
```
nikomatsakis: Of course, this isn't really testing that `db.private_a(x)` gets a compilation error -- but that's kind of hard to do right now without some form of "compiletest" like thing. I can't think of way to do that so I guess I'll take your word for it =)

matklad: Wait a second, I guess we can use name clash for this? like, define `fn foo` for both private and public, and check that the public one is called.

matklad: Ah, that won't work, because we'd get a clash of query storage structs...

PR #169, opened 2019-06-01
implement strong panic safety

matklad:
Previosly, panicking query would remove in-progress memos from the
table.

However, we use panic for cancellation, so it's reasonable to **not**
remove the old result after panic. This is also known as string
exception safety guarantee: not only database is in *some* consistent
state after a panic, it is in the same state it was before the panic!

### DIFF
--- a/src/derived.rs
+++ b/src/derived.rs
@@ -319,7 +319,7 @@ where
         // FIXME(Amanieu/parking_lot#101) -- we are using a write-lock
         // and not an upgradable read here because upgradable reads
         // can sometimes encounter deadlocks.
-        let mut old_memo = match self.probe(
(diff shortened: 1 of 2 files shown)


matklad:
Context: in rust-analyzer, I was surprised a couple of times, when I saw query recomputations which really shouldn't be happening. Turns out, they were required because panic currently clears the database.

matklad:
I've switched rust-analyzer to this PR, as ra hits this issue a lot, after we've started to check for cancellation during validation: https://github.com/rust-analyzer/rust-analyzer/pull/1375#issuecomment-498619426

I would make a point release if/after the PR is merged :)

PR #170, opened 2019-06-06
publish v0.12.2

matklad:
- Don't remove query from storage if its dependency panics (#169)


matklad:
CI is red due to https://github.com/rust-random/rand/issues/818

PR #171, opened 2019-06-06
Add LRU to derived storage

matklad:
LRU allows to bound the maximum number of *values* that are present in
the table.

### DIFF
--- a/src/derived.rs
+++ b/src/derived.rs
@@ -1,6 +1,7 @@
 use crate::debug::TableEntry;
 use crate::plumbing::CycleDetected;
 use crate::plumbing::DatabaseKey;
+use crate::plumbing::LruQueryStorageOps;
 use crate::plumbing::QueryFunction;
 use crate::plumbing::QueryStorageMassOps;
 use crate::plumbing::QueryStorageOps;
@@ -11,13 +12,16 @@ use crate::runtime::Runtime;
 use crate::runtime::RuntimeId;
 use crate::runtime::StampedValue;
 use crate::{Database, DiscardIf, DiscardWhat, Event, EventKind, SweepStrategy};
+use linked_hash_map::LinkedHashMap;
 use log::{debug, info};
 use parking_lot::Mutex;
 use parking_lot::RwLock;
-use rustc_hash::FxHashMap;
+use rustc_hash::{FxHashMap, FxHasher};
 use smallvec::SmallVec;
+use std::hash::BuildHasherDefault;
 use std::marker::PhantomData;
 use std::ops::Deref;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::mpsc::{self, Receiver, Sender};
 use std::sync::Arc;
 
@@ -36,6 +40,8 @@ pub type DependencyStorage<DB, Q> = DerivedStorage<DB, Q, NeverMemoizeValue>;
 /// storage requirements.
 pub type VolatileStorage<DB, Q> = DerivedStorage<DB, Q, VolatileValue>;
 
+type LinkedHashSet<T> = LinkedHashMap<T, (), BuildHasherDefault<FxHasher>>;
+
 /// Handles storage where the value is 'derived' by executing a
 /// function (in contrast to "inputs").
 pub struct DerivedStorage<DB, Q, MP>
@@ -44,6 +50,10 @@ where
     DB: Database,
     MP: MemoizationPolicy<DB, Q>,
 {
+    lru_cap: AtomicUsize,
+    // if `lru_keys` and `map` are locked togeter,
+    // `lru_keys` is locked first, to prevent deadlocks.
+    lru_keys: Mutex<LinkedHashSet<Q::Key>>,
     map: RwLock<FxHashMap<Q::Key, QueryState<DB, Q>>>,
     policy: PhantomData<MP>,
 }
@@ -237,6 +247,8 @@ where
     fn default() -> Self {
         DerivedStorage {
             map: RwLock::new(FxHashMap::default()),
+            lru_cap: AtomicUsize::new(0),
+            lru_keys: Mutex::new(LinkedHashSet::with_hasher(Default::default())),
             policy: PhantomData,
         }
     }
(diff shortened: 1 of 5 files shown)


matklad:
I've tried rust-analyzer with bound of 128 syntax trees. The results for from scratch analysis are as follows:

```
Summary
  './target/ra_cli-old analysis-stats ../chalk' ran
    1.44 ± 0.04 times faster than './target/ra_cli-new analysis-stats ../chalk'
```

old is without lru, new is with lru

RSS for old is 1_232_324k
RSS for new is 2_883_824k

matklad:
Another interesting figure: if I set the LRU size to `128_000_000` (So, bigger than I'd ever need), then

```
Summary
  './target/ra_cli-old analysis-stats ../chalk' ran
    1.10 ± 0.02 times faster than './target/ra_cli-new2 analysis-stats ../chalk'
```

That is, LRU gives us 10% penalty just for maininig LRU, and another 35% due to doing the work twice

matklad:
>That is, LRU gives us 10% penalty just for maininig LRU, and another 35% due to doing the work twice

Hm, that's not entirely correct, the LRU overhead should be more than 10% in general. The test measures the case when the cahce is never full. If the cache is full, and we need to evict, we do more linked hash map operations.

src/derived.rs
```diff
@@ -44,6 +50,10 @@ where
     DB: Database,
     MP: MemoizationPolicy<DB, Q>,
 {
+    lru_cap: AtomicUsize,
+    // if `lru_keys` and `map` are locked togeter,
+    // `lru_keys` is locked first, to prevent deadlocks.
+    lru_keys: Mutex<LinkedHashSet<Q::Key>>,
     map: RwLock<FxHashMap<Q::Key, QueryState<DB, Q>>>,
```
matklad: The alternative here, instead of "lock lru first", is to group both collections under the same lock:

```
map: RwLock<(FxHashMap<Q::Key, QueryState<DB, Q>>, LinkedHashSet<Q::Key>)>
```

This won't be as ergonomic, because, we'd have to map lock guards go get access to `FxHashMap` only. 

OTOH, now that I think about it, in the steady sate the LRU is full, so we have to lock both lru and map anyway, and collapsing that to just one lock seems like a win.

matklad:
The second commit groups both LRU and the original map under  a single lock. I slightly perefer the new version

src/derived.rs
```diff
@@ -487,9 +549,9 @@ where
         key: &Q::Key,
     ) -> ProbeState<StampedValue<Q::Value>, MapGuard>
     where
-        MapGuard: Deref<Target = FxHashMap<Q::Key, QueryState<DB, Q>>>,
+        MapGuard: Deref<Target = QueryMap<DB, Q>>,
     {
-        match map.get(key) {
+        match map.data.get(key) {
```
matklad: Instead of reaching into the `.data` field, we can add map api methods to `QueryMap`. However, we use all of `get`, `get_mut`, `insert`, `remove`, `iter`, `retain`, so I figure that that's to much delegation boilerplate

src/derived.rs
```diff
@@ -44,7 +48,8 @@ where
     DB: Database,
     MP: MemoizationPolicy<DB, Q>,
 {
-    map: RwLock<FxHashMap<Q::Key, QueryState<DB, Q>>>,
+    lru_cap: AtomicUsize,
```
matklad: We deliberately store `lru_cap` outside of `QueryMap` (although it logically belongs there!) such that we can read it wihtout locking

nikomatsakis: can you put a comment about why the lru-cap is stores outside of map?

src/derived.rs
```diff
@@ -228,6 +233,59 @@ impl<DB: Database> std::fmt::Debug for MemoInputs<DB> {
...
+        }
+    }
+}
+
+impl<DB, Q> QueryMap<DB, Q>
+where
+    Q: QueryFunction<DB>,
+    DB: Database,
+{
+    fn set_lru_capacity(&mut self, new_capacity: usize) {
```
nikomatsakis: what happens if this gets invoked with `set_lru_capacity(0)`? seems like it will remove all of the keys, which probably isn't what we want?

tests/lru.rs
```diff
@@ -0,0 +1,68 @@
...
+    runtime: salsa::Runtime<Database>,
+}
+
+impl salsa::Database for Database {
+    fn salsa_runtime(&self) -> &salsa::Runtime<Database> {
+        &self.runtime
+    }
+}
+
+#[test]
```
nikomatsakis: If it's not too much trouble, can we add a test that modifies `set_lru_capacity` after the fact, too? (Ideally, one to some intermediate value (e.g., 22), one to 0, and one to something bigger (64))

src/derived.rs
matklad: Ah, good catch!

matklad: So, let's talk through why we use `usize` here and not an `Option<usize>`. I debated where I should go with `Option`, but decided to use special-cased zero instead. It seems like using 0 to mean "unlimited" is a common pattern in various configurations, and, for example, in rust-analyzer we can use `RA_LRU_CAPACITY=0` to disable LRU, instead of inventing some ad-hoc syntax like `RA_LRU_CAPACITY=none`. Note that this *could* be done on the UI level in rust-analyzer, by translating `0` into `None`, but I figured that just using `0` as a sentinel value internally will also work

matklad:
Fixed the comments, thanks!

As for documentation, I can write a quick RFC here, but, if we are going to write a book, I can just write a short guide there. 

As for a book, well, I am more and more a fan of AsciiDoctor (even [blogged](https://matklad.github.io/2019/05/19/consider-using-asciidoctor-for-your-next-presentation.html) about it), but we definitely should just use mdBook, as that's more familiar to other folks in the ecosystem.

matklad:
RFC: https://github.com/salsa-rs/salsa-rfcs/pull/5

src/derived.rs
nikomatsakis: Yeah, it seemed fine to me. There's no useful interpretation to `0`.

nikomatsakis: Well, I guess you could interpret it as "storage mode = deps only". But still I think using `0` as a sentinel value is ok.

src/derived.rs
```diff
@@ -228,6 +235,70 @@ impl<DB: Database> std::fmt::Debug for MemoInputs<DB> {
...
+        }
+    }
+
+    fn remove_lru(&mut self) {
+        if let Some((evicted, ())) = self.lru_keys.pop_front() {
+            if let Some(QueryState::Memoized(memo)) = self.data.get_mut(&evicted) {
+                // Similar to GC, evicting a value with an untracked input could
+                // lead to inconsistencies. Note that we can't check
+                // `has_untracked_input` when we add the value to the cache,
+                // because inputs can become untracked after the fact.
```
nikomatsakis: Can you elaborate on what you mean by "after the fact"? By the time we record the value, we *do* know its inputs, but I guess you're saying that's too late? Maybe we should loop in the event that `has_untracked_input` is true (i.e., keep trying to remove things?)

nikomatsakis: I guess you could say the same for the other conditions.

PR #174, opened 2019-06-19
Replace volatile query type with report_untracked_read fn

matklad:


src/derived.rs
```diff
@@ -116,30 +111,6 @@ where
...
-        // change -- otherwise the system gets into an inconsistent
-        // state where the same query reports back different values.
-        true
-    }
-
-    fn memoized_value_eq(_old_value: &Q::Value, _new_value: &Q::Value) -> bool {
-        false
-    }
-
-    fn should_track_inputs(_key: &Q::Key) -> bool {
```
nikomatsakis: I think we can remove the `should_track_inputs` method from the trait entirely.

matklad: indeed, fixed!

src/runtime.rs
```diff
@@ -344,7 +344,11 @@ where
         self.local_state.report_query_read(database_key, changed_at);
     }
 
-    pub(crate) fn report_untracked_read(&self) {
+    /// Reports that the query depends on some state unknown to salsa.
+    ///
+    /// If query reports untracked read, it will be reexuted in the next
```
nikomatsakis:
```suggestion
    /// Queries which report untracked reads will be re-executed in the next
```

PR #176, opened 2019-06-21
introduce slots

nikomatsakis:
As discussed on Zulip this PR series optimizes our internal dependency data structures to avoid hashing of keys/values on query revalidation. My measurements found this to be approximately a ~35% win, though @matklad saw results up to 50%. 

This PR *does* introduce some subtle logic around Send+Sync owing to the use of trait objects. It requires a bit of a hack to ensure that the database is Send+Sync iff the keys/values are Send+Sync. I found what feels like a decent solution to this problem by transmuting a `Arc<dyn Slot>` to a `Arc<dyn Slot + Send + Sync>` and using phantom-data to ensure that the composite is only send+sync if the keys/values are send+sync. The truth is that this is kind of overkill since slots are not exposed outside of the database, and if the keys/values are not send+sync, the database as a whole won't be -- but it still feels good to try and be precise here.

| version | from scratch | trivial change x 10 |
| --- | --- | --- |
| master | ~4.5s | 815-820ms |
| approx. LRU | ~4.5s | 537ms |
| precise LRU | 4.5-4.8s | 540-550ms |

### DIFF
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -1,0 +1,58 @@
+use crate::runtime::Revision;
+use crate::Database;
+use std::fmt::Debug;
+use std::hash::Hasher;
+use std::sync::Arc;
+
+/// Each kind of query exports a "slot".
+pub(crate) trait DatabaseSlot<DB: Database>: Debug {
+    /// Returns true if the value of this query may have changed since
+    /// the given revision.
+    fn maybe_changed_since(&self, db: &DB, revision: Revision) -> bool;
+}
+
+pub(crate) struct Dependency<DB: Database> {
+    slot: Arc<dyn DatabaseSlot<DB> + Send + Sync>,
+}
+
+impl<DB: Database> Dependency<DB> {
+    pub(crate) fn new(slot: Arc<dyn DatabaseSlot<DB> + '_>) -> Self {
+        // Forgive me for I am a horrible monster and pray for death:
+        //
+        // Hiding these bounds behind a trait object is a total hack
+        // but I just want to see how well this works. -nikomatsakis
+        let slot: Arc<dyn DatabaseSlot<DB> + Send + Sync> = unsafe { std::mem::transmute(slot) };
+        Self { slot }
+    }
+
+    fn raw_slot(&self) -> *const dyn DatabaseSlot<DB> {
+        &*self.slot
+    }
+
+    pub(crate) fn maybe_changed_since(&self, db: &DB, revision: Revision) -> bool {
+        self.slot.maybe_changed_since(db, revision)
+    }
+}
+
+impl<DB: Database> std::hash::Hash for Dependency<DB> {
+    fn hash<H>(&self, state: &mut H)
+    where
+        H: Hasher,
+    {
+        self.raw_slot().hash(state)
+    }
+}
+
+impl<DB: Database> std::cmp::PartialEq for Dependency<DB> {
+    fn eq(&self, other: &Self) -> bool {
+        self.raw_slot() == other.raw_slot()
+    }
+}
+
+impl<DB: Database> std::cmp::Eq for Dependency<DB> {}
+
+impl<DB: Database> std::fmt::Debug for Dependency<DB> {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.slot.fmt(fmt)
+    }
+}
--- a/src/dependency.rs
+++ b/src/dependency.rs
@@ -21,10 +21,9 @@ pub(crate) struct Dependency<DB: Database> {
 
 impl<DB: Database> Dependency<DB> {
     pub(crate) fn new(slot: Arc<dyn DatabaseSlot<DB> + '_>) -> Self {
-        // Forgive me for I am a horrible monster and pray for death:
-        //
-        // Hiding these bounds behind a trait object is a total hack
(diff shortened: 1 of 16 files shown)


src/dependency.rs
```diff
@@ -0,0 +1,65 @@
...
+    slot: Arc<dyn DatabaseSlot<DB> + Send + Sync>,
+    phantom: std::marker::PhantomData<Arc<DB::DatabaseData>>,
+}
+
+impl<DB: Database> Dependency<DB> {
+    pub(crate) fn new(slot: Arc<dyn DatabaseSlot<DB> + '_>) -> Self {
+        // Forgive me for I am a horrible monster and pray for death:
+        //
+        // Hiding these bounds behind a trait object is a total hack
+        // but I just want to see how well this works. -nikomatsakis
```
matklad: Comment should be adjusted

src/input.rs
```diff
@@ -232,3 +200,69 @@ where
...
+        debug!("maybe_changed_since: changed_at = {:?}", changed_at);
+
+        changed_at.changed_since(revision)
+    }
+}
+
+/// Check that `Slot<DB, Q, MP>: Send + Sync` as long as
+/// `DB::DatabaseData: Send + Sync`, which in turn implies that
+/// `Q::Key: Send + Sync`, `Q::Value: Send + Sync`.
+#[allow(dead_code)]
```
matklad: I tend to use `#[test]` for such checks, seems neat, but not necessary better

src/lru.rs
```diff
@@ -0,0 +1,343 @@
...
+use std::fmt::Debug;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+
+mod test;
+
+/// A very simple concurrent lru list, built using a doubly linked
+/// list of Arcs. Since a doubly linked list forms a cycle, these Arcs
+/// will never be freed.
```
matklad: `impl Drop` is missing

src/lru/test.rs
```diff
@@ -0,0 +1,89 @@
...
+const LRU_SEED: &str = "Hello, Rustaceans";
+const PICK_SEED: &str = "Wippity WIP";
+
+/// Randomly requests nodes and compares the performance of a
+/// *perfect* LRU vs our more approximate version. Since all the
+/// random number generators use fixed seeds, these results are
+/// reproducible. Returns (oracle_hits, lru_hits) -- i.e., the number
+/// of times that the oracle had something in cache vs the number of
+/// times that our LRU did.
+fn compare(num_nodes: usize, capacity: usize, requests: usize) -> (usize, usize) {
```
matklad: Hm, frankly speaking, this does not make sense to me :-) Either I miss something crucial,  or this indeed is not a useful comparison. 

I think if you pick requests **at random**, than eviction strategy is completely irrelevant, because the probability of each hit is independent of previous requests and equals to `lru_size / total_size`. In other words, if requests are uniformly and independently distributed, the contents of lru is completely irrelevant for hit/miss probability, only the size matters.

It only makes sense to measure LRU performance with respect to a particular request distribution which exhibits temporal locality.

src/lru.rs
```diff
@@ -0,0 +1,343 @@
...
+        self.data.lock().record_use(node)
+    }
+}
+
+impl<Node> LruData<Node>
+where
+    Node: LruNode,
+{
+    fn new() -> Self {
+        Self::with_rng(SmallRng::from_entropy())
```
matklad: Let's just use a fixed seed here: reproducibility is nice!

Also (and this is a bit of rant), I feel like `rand` is a bit of an overkill. It's a big crate with many dependencies. I often feel that just hard-coding xor-shift, as stdlib does, would be better:

https://github.com/rust-lang/rust/blob/dbec74ffa7982de8e066a93969ce70a891d7908b/src/libcore/slice/sort.rs#L486-L493

src/lru/test.rs
nikomatsakis: Yeah, you're not wrong. I wanted to introduce a normal distribution or something like that but the rand crate for doing so looked a bit complicated and I never got around to writing the code. The RLS tests however seem to suggest it works pretty decently. But it'd be good indeed to connect this to a proper distribution and see what it looks like. (In particular, I think there is some more tuning we could do to achieve a better hit-rate.)

src/lru.rs
nikomatsakis: the comment is out of date, fixed

src/lru/test.rs
nikomatsakis: Create tests using normal distributions and various standard deviations. Approx. LRU performs very well.

src/lru.rs
nikomatsakis: Switched to fixed seed. I agree rand is probably overkill but I'm not inclined to "roll our own" -- though I'd probably merge a PR if you feel that strongly. =) 

(A "max-min-rand" might be a nice crate to publish)

src/input.rs
nikomatsakis: I wanted it to be checked on every build, but yeah...

matklad:
@nikomatsakis wanna rebase?

nikomatsakis:
@matklad been doing that =)

src/input.rs
matklad: `fn _check_static` is another option

PR #178, opened 2019-06-25
Change authors to Salsa developers

matklad:
Salsa is a team effort, and builds on a mountain of pre-existing
knowledge!

@nikomatsakis wdyt? I personally don't have any strong feelings here :)


nikomatsakis:
Definitely!

PR #180, opened 2019-07-02
Durability

nikomatsakis:
See also https://github.com/salsa-rs/salsa-rfcs/pull/6


src/debug.rs
```diff
@@ -33,7 +35,7 @@ pub trait DebugQueryTable {
 pub struct TableEntry<K, V> {
     /// key of the query
     pub key: K,
-    /// value of the query, if it is stored
+    /// value of the query, if it =is stored
```
theotherphil: (I have a few trivial nits. Sorry if this PR isn’t ready for review yet)

Typo - errant “=“

src/derived/slot.rs
```diff
@@ -72,22 +72,26 @@ where
     /// Last revision when the memoized value was observed to change.
     changed_at: Revision,
 
+    /// If true, then this value was considered a constant when last
```
theotherphil: Out of date comment

src/revision.rs
```diff
@@ -0,0 +1,70 @@
+use std::num::NonZeroU64;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+/// Value if the initial revision, as a u64. We don't use 0
```
theotherphil: Typo: “if” for “of”

src/revision.rs
```diff
@@ -0,0 +1,70 @@
...
+/// because we want to use a `NonZeroU64`.
+const START_U64: u64 = 1;
+
+/// A unique identifier for the current version of the database; each
+/// time an input is changed, the revision number is incremented.
+/// `Revision` is used internally to track which values may need to be
+/// recomputed, but not something you should have to interact with
```
theotherphil: Typo: “but not” for “but is not”

src/runtime.rs
```diff
@@ -479,11 +571,12 @@ struct ActiveQuery<DB: Database> {
     /// What query is executing
     database_key: DB::DatabaseKey,
 
-    /// Maximum revision of all inputs thus far;
-    /// we also track if all inputs have been constant.
-    ///
-    /// If we see an untracked input, this is not terribly relevant.
-    changed_at: ChangedAt,
+    /// True if all inputs were constant (and no untracked inputs).
+    durability: Durability,
```
theotherphil: Out of date comment

src/runtime.rs
```diff
@@ -494,52 +587,45 @@ pub(crate) struct ComputedQueryResult<DB: Database, V> {
...
     pub(crate) value: V,
 
-    /// Maximum revision of all inputs observed; `is_constant` is true
-    /// if all inputs were constants.
-    ///
-    /// If we observe an untracked read, this will be set to a
-    /// non-constant value that changed in the most recent revision.
-    pub(crate) changed_at: ChangedAt,
+    /// True if all inputs were constant (and no untracked inputs).
+    pub(crate) durability: Durability,
```
theotherphil: Out of date comment

src/debug.rs
nikomatsakis:
```suggestion
    /// value of the query, if it is stored
```

matklad:
observation: `changed_at` and `durability` are stored together, perhaps we need to group them in a struct

nikomatsakis:
@matklad they used to be together, in the `ChangedAt` struct, but I removed that in [this commit](https://github.com/salsa-rs/salsa/pull/180/commits/30714022fd65e1eb461180c0c25f5068e3f96474) -- I found overall that struct was kind of annoying.

src/input.rs
```diff
@@ -86,7 +84,7 @@ where
         // need to read from this input. Therefore, we wait to acquire
         // the lock on `map` until we also hold the global query write
         // lock.
-        db.salsa_runtime().with_incremented_revision(|next_revision| {
+        db.salsa_runtime().with_incremented_revision(|guard| {
```
matklad: This comment seems slightly outdated. Additionally, looks like a subtle thing is going on here, which probably warrants a comment as well.

I would expect that we `mark_durability_as_changed` using the maximum of the durabilities of the old and the new slot. However, we advance only the old durability. I think this is ok:

* if there is no previous entry, no one could have observed it, so it's ok to not advance revisions
* if the previous entry has higher durability, that's equivalent to the "obviously correct" max of durabilites
* if it has lower durability, it is correct to not advance the higher durability, because no query of this durability could have observed the input.

This also makes me think that it would even be sound to not bump the base revision when we set previously non-existent key.

src/input.rs
```diff
@@ -100,26 +98,16 @@ where
             // racing with somebody else to modify this same cell.
             // (Otherwise, someone else might write a *newer* revision
             // into the same cell while we block on the lock.)
-            let changed_at = ChangedAt {
-                is_constant: is_constant.0,
-                revision: next_revision,
+            let stamped_value = StampedValue {
+                value,
+                durability: durability,
```
matklad: field init shorthand

src/input.rs
```diff
@@ -145,17 +133,22 @@ where
...
         Ok(value)
     }
 
-    fn is_constant(&self, _db: &DB, key: &Q::Key) -> bool {
+    fn durability(&self, _db: &DB, key: &Q::Key) -> Durability {
         self.slot(key)
-            .map(|slot| slot.stamped_value.read().changed_at.is_constant)
-            .unwrap_or(false)
+            .map(|slot| slot.stamped_value.read().durability)
+            .unwrap_or(Durability::LOW)
```
matklad: why would we ask a durability of a non-existing key?

src/input.rs
```diff
@@ -191,13 +184,20 @@ where
     fn set(&self, db: &DB, key: &Q::Key, database_key: &DB::DatabaseKey, value: Q::Value) {
```
matklad: It seems like we can remove `set` method from this plumbing trait?

src/input.rs
```diff
@@ -191,13 +184,20 @@ where
...
-    fn set_constant(&self, db: &DB, key: &Q::Key, database_key: &DB::DatabaseKey, value: Q::Value) {
+    fn set_with_durability(
+        &self,
+        db: &DB,
+        key: &Q::Key,
+        database_key: &DB::DatabaseKey,
+        value: Q::Value,
+        durability: Durability,
+    ) {
         log::debug!("{:?}({:?}) = {:?}", Q::default(), key, value);
```
matklad: let's log durability as well?

src/runtime.rs
```diff
@@ -139,6 +159,16 @@ where
         db.for_each_query(|query_storage| query_storage.sweep(db, strategy));
     }
 
+    /// True if `d` represents the max durability level.
+    pub(crate) fn is_constant(&self, d: Durability) -> bool {
+        d == self.max_durability()
+    }
```
matklad: unused function. 

Should we remove `allow(unused)` from lib.rs?

src/derived/slot.rs
```diff
@@ -671,19 +690,24 @@ where
...
                 }
-
-                false
             }
         };
 
+        Some(self.verify_value(revision_now))
+    }
+
+    fn verify_value(&mut self, revision_now: Revision) -> StampedValue<Q::Value> {
```
matklad: `fn makr_verified`? I'd expect an `(&self) -> bool` signature from a `verify` method

src/input.rs
nikomatsakis: That's really just the method for the debug trait that lets you access the durability of a key. I suppose it could panic if the key is not defined.

src/runtime.rs
nikomatsakis: I presume you mean `#[allow(dead_code)]` -- but yeah, we should.

src/input.rs
nikomatsakis: > This also makes me think that it would even be sound to not bump the base revision when we set previously non-existent key.

I think this is correct. It wasn't always so, as we used to have `Default` key values -- but now we panic on access to a non-existent key, so that's ok.

nikomatsakis: Also, I want (but haven't) add some way to get the "set of valid keys" on an input, in which case we would need a new revision.

src/input.rs
nikomatsakis: made it panic

src/input.rs
matklad: >Also, I want (but haven't) add some way to get the "set of valid keys" on an input, in which case we would need a new revision.

If we do that, we'll need to modify the logic to always use max durability, without shortcutrs

src/derived/slot.rs
```diff
@@ -696,16 +720,9 @@ where
         );
 
         if self.verified_at == revision_now {
-            let is_constant = match self.inputs {
-                MemoInputs::Constant => true,
-                _ => false,
-            };
-
             return Some(StampedValue {
```
matklad: Let's inline `probe_memoized_value` into the call site? It doesn't do much, and we can put the condition into a match arm guard.

matklad:
observation: we call `validate_memoized_value` in `read_upgrade`, after replacing QueryState with InProgress. It seems to me that, after slots, we actually can do that in `read`, without aquiring a write lock at all.

matklad:
Also, we can switch back to upgradable reads here, the issue was fixed in parking lot a while ago

src/derived/slot.rs
```diff
@@ -215,20 +218,27 @@ where
         // old value.
```
matklad: Several lines above, `let database_key = self.database_key(db);` has only a single usage. Let's move that several lines up still, to reuse this local variable instead of two `self.database_key(db)` calls?

src/derived/slot.rs
```diff
@@ -671,19 +690,24 @@ where
...
         };
 
+        Some(self.mark_value_as_verified(revision_now))
+    }
+
+    fn mark_value_as_verified(&mut self, revision_now: Revision) -> StampedValue<Q::Value> {
+        let value = match &self.value {
+            Some(v) => v.clone(),
+            None => panic!("invoked `verify_value` without a value!"),
+        };
```
matklad: Alternative

```rust
let value = self.value.clone().unwrap_or_else(|| panic!("invoked `verify_value` without a value!"))
```

matklad:
>observation: we call validate_memoized_value in read_upgrade, after replacing QueryState with InProgress. It seems to me that, after slots, we actually can do that in read, without aquiring a write lock at all.

Also `maybe_check_since` looks suspiciously similar to `probe` + `validate_memoized_value`. Could it be the case that, after slots, we can share more code between the two?

matklad:
Reviewed! 

Looks good, https://github.com/salsa-rs/salsa/pull/180#issuecomment-508762722 is the single non-nit comment I have

src/derived/slot.rs
```diff
@@ -671,48 +701,27 @@ where
...
+        StampedValue {
+            durability: self.durability,
+            changed_at: self.changed_at,
+            value,
         }
-
-        None
     }
 
+    /// Returns the memoized value *if* it is known to be update in the given revision.
```
matklad: This comment should have need removed when inlinijg that function :)

src/derived/slot.rs
nikomatsakis: I sort of find the explicit match more readable, but it's also (very mildly) more efficient, since we only check if the value is `Some` once. I suppose post optimization that's perhaps not even true. =)

nikomatsakis: In any case I think I'll leave as is...

PR #181, opened 2019-07-03
use ptr::eq/hash for readability

matklad:
no strong opinion here, but seems just a hair slightly more direct than exposing `*const`


nikomatsakis:
yeah seems good

PR #182, opened 2019-07-03
use AtomicU64 for RuntimeId

matklad:
We don't use `RuntimeId` to index memory, so using usize is
not the best fit semantically


cormacrelf:
Wasn’t it because we don’t want to pay for double width on sub-64bit platforms? I think I remember you guys choosing it on purpose and being okay with the headroom on 32bit.

matklad:
Originally we used AtomicUsize because AtomicU64 wasn’t stable.

PR #186, opened 2019-07-28
s/query struct/database struct/

rvolgers:
I found this very confusing. I think this is the intended meaning. Unfortunately I don't know anything about Salsa so I can't be sure.


rvolgers:
On second look, this probably isn't it. Instead I think is trying to say something about the actual storage being split out by query group, and the database struct just puts them all together?

PR #195, opened 2019-09-24
document the selection pattern

nikomatsakis:
- extend some of the empty sections, add a new common pattern
- also, show how to use anchors and include so we can test the sources for common patterns

r? @matklad


.travis.yml
```diff
@@ -10,10 +10,9 @@ cache:
 matrix:
   allow_failures:
       - rust: nightly
-before_script:
-  - (test -x $HOME/.cargo/bin/cargo-install-update || cargo install cargo-update)
-  - (test -x $HOME/.cargo/bin/mdbook || cargo install --vers "^0.1" mdbook)
-  - cargo install-update -a
+before_install:
+  - cargo install mdbook --vers '0.3.2' --debug --force
+  - cargo install mdbook-linkcheck --vers '0.4.0' --debug --force
```
nikomatsakis: I copied this from the async-book setup; I actually don't know what the difference is.

book/src/common_patterns/on_demand_inputs.md
```diff
@@ -0,0 +1,49 @@
...
+That is, when someone requests the text of a file for the first time:
+
+1. Read the file from disk and cache it.
+2. Setup a file-system watcher for this path.
+3. Invalidate the cached file once the watcher sends a change notification.
+
+This is possible to achieve in salsa, using a derived query and `report_synthetic_read` and `invalidate` queries.
+The setup looks roughtly like this:
+
+```rust,ignore
```
nikomatsakis: We should move this to the `examples` directory too.

nikomatsakis:
One question: I've also called the *selection* pattern the *firewall* pattern, and I like that name too (very evocative). But somehow selection felt more "obvious". Thoughts?

nikomatsakis:
(Also, sneak preview: I want to propose some specialized features to make the selection pattern more efficient.)

.travis.yml
```diff
@@ -10,10 +10,9 @@ cache:
 matrix:
   allow_failures:
       - rust: nightly
-before_script:
-  - (test -x $HOME/.cargo/bin/cargo-install-update || cargo install cargo-update)
-  - (test -x $HOME/.cargo/bin/mdbook || cargo install --vers "^0.1" mdbook)
-  - cargo install-update -a
+before_install:
+  - cargo install mdbook --vers '0.3.1' --debug --force
```
matklad:
```suggestion
  - cargo install mdbook --version '0.3.1' --debug --force
```

(`--vers` seems to be some old back-compat thing, `-h` tells to use `--version`)

.travis.yml
```diff
@@ -10,10 +10,9 @@ cache:
...
+  - cargo install mdbook-linkcheck --vers '0.4.0' --debug --force
```
matklad:
```suggestion
  - cargo install mdbook-linkcheck --version '0.4.0' --debug --force
```

book/src/common_patterns/selection.md
```diff
@@ -0,0 +1,79 @@
...
+In particular, Qsel does not combine values from other queries. In some sense,
+then, Qsel is redundant -- you could have just extracted the information
+the information from Qbase yourself, and done without the salsa machinery. But
+Qsel serves a role in that it limits the amount of re-execution that is required
+when Qbase changes.
+
+## Example: the base query
+
+For example, imagine that you have a query `parse` that parses the input text of a request to return a
+only consume small parts of itheader and a body:
```
matklad:
```suggestion
only consume small parts of it (a header and a body):
```

book/src/common_patterns/selection.md
```diff
@@ -0,0 +1,79 @@
...
+
+## A note on cloning and efficiency
+
+In this example, we used common Rust types like `Vec` and `String`,
+and we cloned them quite frequently. This will work just fine in Salsa,
+but it may not be the most efficient choice. This is because each clone
+is going to produce a deep copy of the result. As a simple fix, you
+might convert your data structures to use `Arc` (e.g., `Arc<Vec<ParsedHeader>>`),
+which makes cloning cheap. Or you might prefer to adopt persistent data
+structures such as those found in [the `im` crate](https://crates.io/crates/im).
```
matklad: I would maybe drop the advice about persistent data structure: they are useful to preserve old copies when you do mutation (ie, O(log n) update is the core property). Here, we don't do mutation, we only project, so `Arc<Vec>` seems to be a better choice.

book/src/common_patterns/selection.md
nikomatsakis: man emacs was doing weird stuff like this whenever I tried to "reflow" paragraphs :( I thought I caught them all...

book/src/common_patterns/selection.md
nikomatsakis: Seems fine. If only `Arc<Vec>` were less annoying. In Lark we introduced some specialized types for this. Someday maybe we'll have some companion util libraries for Salsa along those lines... or not. =)

matklad:
>One question: I've also called the selection pattern the firewall pattern, and I like that name too (very evocative). But somehow selection felt more "obvious". Thoughts?

Heh, I personally tend to prefer "projection". I think "firewall" is a good metaphor, but I would prefer `selection` / `projection` as the official name: that way, you can use select and project verbs to name the queries.

book/src/common_patterns/selection.md
matklad: Replying in zulip: https://salsa.zulipchat.com/#narrow/stream/145099-general/topic/selection.20pattern

nikomatsakis:
@matklad I started with "projection" but switched to "selection" since projection seemed kind of like "PL jargon" to me.

matklad:
Hm, yeah, agree that "selection" is better. I usually love jargon, because it creates unique names which are hard to confuse with anything else. However, because salsa is used in the domain where PL jargon is common, a "layman" term would actually stand out better!

nikomatsakis:
It also reminded me of SQL, what with our "db" terminology and all. =)

PR #196, opened 2019-09-26
permit gc to collect slot keys

nikomatsakis:
In the older codebase, GC could never collect a slot key. We now enable this by transforming slots to a *removed* state. There is a slight race condition in that someone *reading* the slot might now race with the GC which would remove it in the meantime; in that case we simply try again.

### DIFF
--- a/src/derived.rs
+++ b/src/derived.rs
@@ -136,14 +136,18 @@ where
     MP: MemoizationPolicy<DB, Q>,
 {
     fn try_fetch(&self, db: &DB, key: &Q::Key) -> Result<Q::Value, CycleError<DB::DatabaseKey>> {
-        let slot = self.get_or_create_slot(key);
+        let mut slot;
         let StampedValue {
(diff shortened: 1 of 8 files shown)


nikomatsakis:
Question I am wondering: Should I try to create a test case that deliberately engineers this race condition? Not sure how to do it, though.

src/derived.rs
```diff
@@ -150,7 +161,10 @@ where
     }
 
     fn durability(&self, db: &DB, key: &Q::Key) -> Durability {
-        self.slot(key).durability(db)
+        match self.get_slot(key) {
+            Some(slot) => slot.durability(db),
+            None => Durability::LOW,
```
matklad: It's non-obvious why this `::LOW` is correct. I think the reason why it is correct is that we actually don't care, as it is only used for debug. Perhaps we should rename this method to `debug_durability`?

PR #197, opened 2019-09-27
require `&mut self` for moving to a new revision (breaking)

nikomatsakis:
We always *intended* that to create a new revision you needed `&mut self`, but this was enforced not by the `SalsaRuntime` itself but by the DB. The reason for this was primarily that we needed to have (a) a reference to the DB that overlapped with (b) a reference to query tables owned by the database, and that didn't permit `&mut`. But if we move those query tables into an `Arc`, we can resolve this problem, and then we can have `with_incremented_revision` take `&mut self` (which is what this PR does).

This would be largely unobservable to users, except that it it fixes an existing hole in the API and hence is a breaking change (it also requires a new `salsa_runtime_mut` method). The hole is that before you could invoke `synthetic_write` with only an `&DB`, which means you could do it from a `Frozen<DB>`. This was not intended.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,6 +48,9 @@ pub trait Database: plumbing::DatabaseStorageTypes + plumbing::DatabaseOps {
     /// Gives access to the underlying salsa runtime.
     fn salsa_runtime(&self) -> &Runtime<Self>;
 
+    /// Gives access to the underlying salsa runtime.
+    fn salsa_runtime_mut(&mut self) -> &mut Runtime<Self>;
+
     /// Iterates through all query storage and removes any values that
     /// have not been used since the last revision was created. The
     /// intended use-cycle is that you first execute all of your
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -33,6 +33,7 @@ use crate::revision::Revision;
 use derive_new::new;
 use std::fmt::{self, Debug};
 use std::hash::Hash;
+use std::sync::Arc;
 
 pub use crate::durability::Durability;
 pub use crate::intern_id::InternId;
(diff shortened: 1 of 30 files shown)


src/lib.rs
```diff
@@ -497,8 +501,8 @@ where
     DB: plumbing::GetQueryTable<Q>,
     Q: Query<DB> + 'me,
 {
-    db: &'me DB,
-    storage: &'me Q::Storage,
+    db: &'me mut DB,
+    storage: Arc<Q::Storage>,
```
matklad: Hm, this encoding seems somewhat unoptimal now. Like we already have `&mut DB`, and that's strictly more than `Arc<Q::Storage>`. I wonder if we should change `InputQueryStorageOps::set` to not take `DB` as an argument. 

However, this is internal impl detail, so I am ok with landing this as is.

nikomatsakis: Hmm, interesting thought. If we didn't have a `DB`, but only a runtime, it might work. However, it would prevent us from doing `salsa_event` calls -- at least from within the `InputStorage`. I'll take a look later. If we could do away with the `Arc` and `db` argument, it would mean that the storage could take `&mut self`, which could allow us to eliminate the mutex -- that seems kind of interesting.

nikomatsakis: OK so it's going to be a bit tricky. The problem is this: *somebody* needs to be able to give us a "simultaneously live" `&mut Q::Storage` and `&mut SalsaRuntime`. Unfortunately, right now, those two bits of data are supplied by two distinct traits.

One thing that might help us here would be if we gave up the idea that the user defines the Salsa database type. Instead, salsa database would be a struct that **we** define -- but it has a "userdata" field that users can specify.

I've thought about this for other reasons. In particular, for the purposes of guaranteeing safety, it seems suboptimal that the DB -> runtime link is defined by end-users, who could do all manor of crazy things in there.

nikomatsakis: (However that also seems out of scope for this PR)

nikomatsakis: My inclination is to spin out my previous comment into an issue to be explored shortly, but land this PR in the meantime... agree, @matklad?

matklad: :+1:

PR #199, opened 2019-09-30
Fix deploy book

funkill:
`mdbook` generate two directories: `html` with rendered book and `linkcheck` with output of `mdbook-linkcheck`. Set deploy only `html` dir.


matklad:
:man_facepalming: 

Thanks @funkill !

PR #200, opened 2019-09-30
Fix documentation links + on demand example

declanvk:
I can squash these commits if that works better.


matklad:
Changes look good to me! @declanvk, could you rebase the PR?

matklad:
Thanks!

PR #201, opened 2019-10-03
Implement QueryTable constructor manually instead of using derive-new

lnicola:
While `derive-new` isn't on the critical path, it still takes some 6 seconds to compile on a reasonably fast laptop, and the freed up CPU can be used for something else:

image

image

Perhaps it's worth writing a couple of lines of code by hand.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,6 @@ readme = "README.md"
 
 [dependencies]
 crossbeam = "0.7.1"
-derive-new = "0.5.5"
 indexmap = "1.0.1"
 log = "0.4.5"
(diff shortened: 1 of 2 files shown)


matklad:
:+1: for me! In general, I have a pretty high tolerance for trivial boilerplate and pretty low tolerance for proc-macros :) But let's see what @nikomatsakis thinks here. I would also want to remove `derive(new)` from chalk.

lnicola:
That would be https://github.com/rust-lang/chalk/pull/254 :-).

nikomatsakis:
Seems ok.

PR #206, opened 2019-11-26
Speedup mdbook install on CI

matklad:


matklad:
Hm, since we are using mdbook-linkcheck, we still have all networking stuff =/

With this PR we build it once instead of twice, but this doesn't help that much in relative terms. It's better just to setup proper caching

Issue #188, opened 2019-08-15
optimize locks and things post slots


nikomatsakis:
@matklad made a number of comments in https://github.com/salsa-rs/salsa/pull/180 that seem worth following up on, but which are not particularly related to that PR, so I'm pulling them into a separate comment to be pursued separately:

---

> observation: we call `validate_memoized_value` in `read_upgrade`, after replacing QueryState with InProgress. It seems to me that, after slots, we actually can do that in `read`, without aquiring a write lock at all.

_Originally posted by @matklad in https://github.com/salsa-rs/salsa/pull/180#issuecomment-508762722_

---

> Also, we can switch back to upgradable reads here, the issue was fixed in parking lot a while ago

_Originally posted by @matklad in https://github.com/salsa-rs/salsa/pull/180#issuecomment-508764489_

---

> Also `maybe_check_since` looks suspiciously similar to `probe` + `validate_memoized_value`. Could it be the case that, after slots, we can share more code between the two?

_Originally posted by @matklad in https://github.com/salsa-rs/salsa/pull/180#issuecomment-508773298_

michalt:
Since I already started reading this code (when profiling rust-analyzer), I can look into fixing these.

Veykril:
Seems no longer relevant given the rewrite

PR #208, opened 2020-01-04
Use parking_lot's `upgradable_read` in `read_upgrade`

michalt:
The bug in parking_lot that made it necessary to use a write-lock has
been fixed and we can go back to using an upgradable read-lock.

This cleanup is part of issue #188.

Signed-off-by: Michal Terepeta <michal.terepeta@gmail.com>


nikomatsakis:
*And* it has taken over a year to look at this -- but @michalt if you want to rebase, I would r+  :)

michalt:
I think this PR is obsolete (#230 already fixed this)

PR #209, opened 2020-01-12
Update some comments in slot.rs

michalt:
I'm trying to better understand the code and noticed that there are some outdated comments. Let me know if something is inaccurate! (or more places need updating)


matklad:
I believe the changes are accurate! I am not sure if anything else needs to be updated: it's been a while since I've looked at salsa closely.

michalt:
Thanks for the quick review! :)

PR #212, opened 2020-02-09
A few simplifications of `maybe_changed_since`

michalt:
As observed in #188, the method looks very similar to the combination of `probe` and `validate_memoized_value`, but is different enough that I didn't see to completely replace it. The core difference is what the name suggest ;) -- it really is focused on answering the question of whether something changed, whereas the other methods are focused on providing/validating an actual value.  Do let me know if I'm missing something though!

For reviewing, consider looking at the commits one by one -- I tried to separate the change to formatting/location from the ones affecting the behavior.

### DIFF
--- a/src/derived/slot.rs
+++ b/src/derived/slot.rs
@@ -522,53 +522,51 @@ where
         }
 
         // Either way, we have to update our entry.
-        //
-        // Keep in mind, though, that we released the lock before checking the ipnuts and a lot
-        // could have happened in the interim. =) Therefore, we have to probe the current
-        // `self.state`  again and in some cases we ought to do nothing.
-        {
-            let mut state = self.state.write();
-            match &mut *state {
-                QueryState::Memoized(memo) => {
-                    if memo.verified_at == revision_now {
-                        // Since we started verifying inputs, somebody
-                        // else has come along and updated this value
-                        // (they may even have recomputed
-                        // it). Therefore, we should not touch this
-                        // memo.
-                        //
-                        // FIXME: Should we still return whatever
-                        // `maybe_changed` value we computed,
-                        // however..? It seems .. harmless to indicate
-                        // that the value has changed, but possibly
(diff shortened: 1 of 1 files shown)


src/derived/slot.rs
```diff
@@ -429,55 +429,29 @@ where
             self, revision, revision_now,
         );
 
-        // Acquire read lock to start. In some of the arms below, we
-        // drop this explicitly.
-        let state = self.state.read();
+        // Check the current state and potentially block if there's a computation in progress.
+        let state = match self.probe(db, self.state.read(), runtime, revision_now) {
```
nikomatsakis, line 433: So the problem here is that we are going to clone the value in this case, which isn't necessary

src/derived/slot.rs
```diff
@@ -522,53 +522,51 @@ where
...
+        // the interim. =) Therefore, we have to probe the current `self.state`  again and in some
+        // cases we ought to do nothing.
+        let mut state = self.state.write();
+        match &mut *state {
+            QueryState::Memoized(memo) => {
+                if memo.verified_at == revision_now {
+                    // Since we started verifying inputs, somebody else has come along and updated
+                    // this value (they may even have recomputed it). Therefore, we should not touch
+                    // this memo.
+                    return memo.changed_at > revision_since;
```
nikomatsakis, line 545: This change seems correct though.

nikomatsakis:
Sorry for taking forever to review this PR -- overall I like the idea of simplifying that function, but I don't think introducing a clone of the values is the right way to go. That said, I'm wanting to kind of refactor and simplify this code-path anyway, so maybe it's a moot point...

src/derived/slot.rs
michalt: Indeed! Thanks for catching this.

michalt:
Thanks for taking a look! Since you're planning to simplify this code anyway, I'll just close the PR.

PR #221, opened 2020-05-16
Fix doc typos

RadicalZephyr:
Found a few more typos in the book.

### DIFF
--- a/book/src/common_patterns/selection.md
+++ b/book/src/common_patterns/selection.md
@@ -62,8 +62,8 @@ This will result in a dependency graph like so:
 request_text  -->  parse  -->  header -->  content_type  -->  (other queries)
 ``` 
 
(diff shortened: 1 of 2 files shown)


book/src/common_patterns/selection.md
```diff
@@ -62,8 +62,8 @@ This will result in a dependency graph like so:
 request_text  -->  parse  -->  header -->  content_type  -->  (other queries)
 ``` 
 
-The advantage of this is that changes that only effect the "body" of
-only consume small parts of itthe request will
+The advantage of this is that changes that only effect the "body" or
+only consume small parts of the request will
```
RadicalZephyr, line 66: Is this the correct meaning of this sentence? It seems to make sense to me but I'm not 100% certain.

PR #222, opened 2020-06-24
Document query group plumbing

nikomatsakis:
I'm trying to write-up some proposals to overhaul how salsa work and I found I needed some base documentation for what we do today. This PR extends the book to cover that.


nikomatsakis:
Well, it at least covers the first part.

book/src/plumbing.md
```diff
@@ -0,0 +1,279 @@
...
+* a "group struct" named `HelloWorldStorage` that represents the group; this struct implements `plumbing::QueryGroup`
+  * somewhat confusingly, this struct doesn't actually contain the storage itself, but rather has an associated type that leads to the "true" storage struct
+* an impl of the `HelloWorld` trait, for any database type
+* for each query, a "query struct" named after the query; these structs implement `plumbing::Query` and sometimes other plumbing traits
+* a group key, an enum that can identify any query within the group and store its key
+* the associated storage struct, which contains the actual hashmaps that store the data for all queries in the group
+
+Note that there are a number of structs and types (e.g., the query structs, or
+the group descriptor) that represent things which don't have "public" names. We
+currently generate mangled names but those names are not meant to be exposed to
```
matklad, line 35: Not exactly true, QueryStructs are a part of public API. We use them, eg, to measure memory: https://github.com/rust-analyzer/rust-analyzer/blob/dff62def2ea31e7eb0b010c35d60cd53318eab80/crates/ra_ide_db/src/change.rs#L167-L269

book/src/plumbing.md
```diff
@@ -0,0 +1,279 @@
+# Plumbing
+
+This chapter documents the code that salsa generates and its "inner workings".
+We refer to this as the "plumbing".
```
matklad, line 4: Such internals docs tend to go obsolete, so it might be a good idea to add "as of 2020-06-24"

nikomatsakis: +1, although I think it also falls on us to ensure that any major edits to the plumbing edit this book in the same PR.

book/src/plumbing.md
nikomatsakis: Well, I think it's true that this is in violation of our semver rules. It may also be true that we should expose some way to do this that we *do* want to commit to.

nikomatsakis: (Or just commit to the current way.)

nikomatsakis: I rewrote the text to be a bit ambiguous here. I'd be ok with "committing" to these structs existing, at least, and maybe (I can't remember...) we already did, though in that case we should probably document it more clearly.

nikomatsakis: Well, I guess we basically did commit to that, what with the `db.query(Q)` methods.

nikomatsakis:
Huh, ok, the travis results for the LRU oracle tests and what I see locally don't agree, and I have no idea why that is.

matklad:
>Huh, ok, the travis results for the LRU oracle tests and what I see locally don't agree, and I have no idea why that is.

Different version of `rand` perhaps?

nikomatsakis:
@matklad sigh, yeah, I guess that could be it.

nikomatsakis:
I can either disable them, or else that that the results are within 10% or something

nikomatsakis:
Decided to disable them.

PR #224, opened 2020-06-24
Setup GitHub actions

### LINKED ISSUE #223: port to github actions
I'd like to move our CI from Travis to Github Actions. It needs to do two things:

* run tests
* build and publish the mdbook

Anybody want to help?

mbrobbel:
This is an attempt to close #223.


nikomatsakis:
Woohoo, thanks!

.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: ${{ matrix.rust }}
+        default: true
+    - uses: actions-rs/cargo@v1
+      with:
+        command: test
+        args: --all-features --all-targets
```
Marwes, line 36: `--all-targets` ends up not testing `src/doctest.rs`. Either this should be removed or multiple commands needs to be run.

nikomatsakis:
Regarding tokens, can we not make use of the [`GITHUB_TOKEN` that is always available](https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token)? I would rather not put a personal access token in there, as it gives access to way more than is needed.

mbrobbel:
Yes, let's try with `GITHUB_TOKEN`. I've had issues before where using the provided `GITHUB_TOKEN` wouldn't trigger a GitHub Pages build, but maybe that is fixed now.

mbrobbel:
I switched to a different action for the deployment to GitHub Pages, which supports using `GITHUB_TOKEN`.

nikomatsakis:
Should I merge, or is there more to be done?

mbrobbel:
It should be good to go. If there are issues I can follow up with another PR.

PR #228, opened 2020-06-27
move the accepted RFCs to the book and describe a new process

nikomatsakis:


matklad:
lgtm! 

Maybe merge the durability rfcs as well?

nikomatsakis:
Yeah, I figured we would move https://github.com/salsa-rs/salsa-rfcs/pull/5/ (LRU) and https://github.com/salsa-rs/salsa-rfcs/pull/6 (Durability), both of which are pretty solidly "deployed".

PR #229, opened 2020-06-27
Bump deps

lnicola:


src/derived/slot.rs
```diff
@@ -165,13 +166,13 @@ where
...
-        // can sometimes encounter deadlocks.
-        let old_memo = match self.probe(db, self.state.write(), runtime, revision_now) {
+        let old_memo = match self.probe(db, self.state.upgradable_read(), runtime, revision_now) {
             ProbeState::UpToDate(v) => return v,
-            ProbeState::StaleOrAbsent(mut state) => {
+            ProbeState::StaleOrAbsent(state) => {
+                type RwLockUpgradableReadGuard<'a, T> =
+                    lock_api::RwLockUpgradableReadGuard<'a, RawRwLock, T>;
+
+                let mut state = RwLockUpgradableReadGuard::upgrade(state);
```
lnicola, line 175: This looks weird, but I couldn't figure out how to call this otherwise.

matklad: Let's maybe just keep `.write` here, to not mix up library update with changes?

lnicola: Done.

matklad:
released as 0.14.2

PR #231, opened 2020-06-29
[RFC] Dynamic databases

nikomatsakis:
This PR introduces an RFC describing a shift to `dyn`-capable databases. The primary goal is to make it so that query-group code can be compiled in the query-group crate, rather than waiting to be monomorphized in the final database crate.

**The main user-visible effect of this proposal is that it requires all salsa query group traits to be dyn-safe.**

[Rendered form of the RFC](https://github.com/nikomatsakis/salsa/blob/dynamic-databases-rfc/book/src/rfcs/RFC0006-Dynamic-Databases.md)

### Current status

Largely implemented. 

Pending work:

* [x] Refactor slots and dependencies and try to minimize code generic over Q in the derived impl -- the current branch doesn't match the RFC in that it still creates slots that store the values.
* [x] Update the salsa book
* [x] Test effectiveness perhaps?

Pending updates the RFC text itself:

* [x] Describe the `'static` limitation on databases and how it might be overcome
* [x] I removed the `db.query(Q)` methods entirely in favor of `Q.in_db(&db)`. This does not require an extension trait and also works better with `dyn Db` coercions.
* [x] Describe how we are adding `salsa::Database` and `HasQueryGroup<G>` as an automatic supertrait to each query-group trait.


book/src/rfcs/RFC0006-Dynamic-Databases.md
```diff
@@ -0,0 +1,635 @@
...
+### Our goal
+
+The primary goal of this RFC is to make it so that the glue code we generate for
+query groups is not dependent on the database type, thus enabling better
+incremental rebuilds.
+
+## User's guide
+
+Most of the changes in this RFC are "under the hood". But there are various user
+visibile changes proposed here.
```
matklad, line 97: What happens to query definitions? 

Presumably, users now should always write them as `fn my_query(db: &dyn MyDatabase)`? Or, on the contrary, because we generate all the code in the context of query defining crate, it might makes sense to use `fn my_query(db: &iml MyDatabase)` as that would be monomorphised only once?

book/src/rfcs/RFC0006-Dynamic-Databases.md
```diff
@@ -0,0 +1,635 @@
...
+Today, the `Query`, `QueryFunction`, and `QueryGroup` traits are generic over
+the database `DB`, which allows them to name the final database type and
+associated types derived from it. In the new scheme, we never want to do that,
+and so instead they will now have an associated type, `DynDb`, that maps to the
+`dyn` version of the query group trait that the query is associated with.
+
+Therefore `QueryFunction` for example can become:
+
+```rust
+pub trait QueryFunction: Query<DB> {
```
matklad, line 283:
```suggestion
pub trait QueryFunction: Query {
```

?

book/src/rfcs/RFC0001-Query-Group-Traits.md
```diff
@@ -1,12 +1,20 @@
-# Motivation
+# Query group traits
+
+## Metadata
+
+* Author: nikomatsakis
+* Date: 2019.01.15
```
bjorn3, line 6:
```suggestion
* Date: 2019-01-15
```

book/src/rfcs/RFC0006-Dynamic-Databases.md
```diff
@@ -0,0 +1,635 @@
+# Dynamic databases
+
+## Metadata
+
+* Author: nikomatsakis
+* Date: 2020.06.29
```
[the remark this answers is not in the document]
nikomatsakis: over my dead body

book/src/rfcs/RFC0006-Dynamic-Databases.md
nikomatsakis: Yes, they must write `dyn MyDatabase`. You could write `impl` but you'd have to write `impl MyDatabase + ?Sized`.

nikomatsakis:
Regarding following the setup, it is tricky, I was thinking about how to diagram it but I couldn't quite come up with the right diagram to explain everything. I didn't put a lot of effort into it yet, but I think it'd be worth it.

nikomatsakis:
I can probably put in some effort into prototyping this but I'd also be happy to work with someone else on that. Also, I forgot that I had planned to add a few notes in the alternatives section about ways we could collect `DatabaseKeyIndex` values and/or why I chose to make `dyn` mandatory (the latter is basically "simpler to pick one and not have to maintain multiple options").

book/src/rfcs/RFC0006-Dynamic-Databases.md
```diff
@@ -0,0 +1,635 @@
...
+### The query method on `salsa::Database` moves to a helper method
+
+As a consequence of the previous point, the existing `query` and `query_mut`
+methods on the `salsa::Database` trait will be moved to an extension trait,
+automatically implemented for all salsa databases:
+
+```rust,ignore
+pub trait DatabaseExt {
+    #[allow(unused_variables)]
+    fn query<Q>(&self, query: Q) -> QueryTable<'_, Self, Q>
```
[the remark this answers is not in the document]
nikomatsakis: Ah, interesting, that's true, we could do that, though it seems strictly less good. This is an interesting case in the language, it'd be nice if we had a way to express "final" functions in a trait that could not be overridden by impls (which would then be exempted from object safety requirements).

book/src/rfcs/RFC0006-Dynamic-Databases.md
```diff
@@ -0,0 +1,639 @@
...
+once and always remains up-to-date, even when the query is re-executed (because
+its inputs have changed). Thus all of the slot's data is effectively under a
+lock.
+
+This RFC proposes to create a fresh `Arc<DependencySlot>` each time the query is
+re-executed. This makes *almost all* the fields in the `DependencySlot`
+immutable, including the list of dependencies. This in turn means that those
+fields can be traversed during revalidation without acquiring any read-locks.
+
+There is one complication, though. It sometimes happens that we a query Q has a
```
matklad, line 339:
```suggestion
There is one complication, though. It sometimes happens that a query Q has a
```

book/src/rfcs/RFC0006-Dynamic-Databases.md
```diff
@@ -0,0 +1,639 @@
...
+
+```rust
+trait Database {
+    fn debug_fmt_database_key(
+        &self,
+        key: DatabaseKey,
+        fmt: &mut std::fmt::Formatter<'_>,
+    ) -> std::fmt::Result<()>;
+}
+```
```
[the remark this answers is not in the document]
nikomatsakis: Yes. Lark had a `debug_with` trait for this purpose, so you would write `foo!("{:?}", bar.debug_with(context))`. It worked ok but I wish there was a standard solution.

matklad:
>Regarding following the setup, it is tricky, I was thinking about how to diagram it but I couldn't quite come up with the right diagram to explain everything. I didn't put a lot of effort into it yet, but I think it'd be worth it.

What I usually do is just ask salsa to dump all the code generated for hello world example. It might make sense to write this as a test  (piping the output through `rustftmt`), so that it's easy to see what exactly is going on.

nikomatsakis:
I started playing around with implementation (you can see the commits) and I realized I may want to change one aspect of the current proposal. Instead of the `DatabaseKeyIndex` being an index into a central vector, it may be better to have it be a set of three integers:

```rust
struct DatabaseKeyIndex {
    group_index: u16,
    query_index: u16,
    key_index: u32,
}
```

This was actually my original idea, but I changed it to a centralized index at some point.  This "3 integer" approach means that we can create fresh keys without any central locking or coordination. It works just as well for the main job of dispatching revalidation requests -- perhaps better. It also works particularly well with the "interning"-style queries, as the `key_index` is simply the intern request (trying to fix interning-style queries into the code was what drove me to this approach).

I think that for the the other kinds of queries, we would basically use a `FxIndexMap<Q::Key, ..>`. Already with slots we never remove keys and instead just replace them with a "not computed" placeholder, so the indices won't get invalidated.

book/src/rfcs/RFC0006-Dynamic-Databases.md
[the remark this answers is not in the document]
nikomatsakis: e.g., it'd be great if you could do `debug!(context, "{:?}", foo)` and have the context be supplied to all the things within. Hmm. That's not the worst idea...

book/src/rfcs/RFC0006-Dynamic-Databases.md
```diff
@@ -0,0 +1,639 @@
...
+
+    fn query_mut<Q>(&mut self, query: Q) -> QueryTableMut<'_, Self, Q> ...  { }
+}
+
+impl<DB: ?Sized + salsa::Database> DatabaseExt for DB {
+    ...
+}
+```
+
+### The salsa-event mechanism will move to dynamic dispatch
```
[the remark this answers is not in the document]
nikomatsakis: We'll just have to measure whether using a dyn interface here makes a perf difference, but I don't really see much of an alternative.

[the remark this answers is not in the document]
matklad: For cancellation specifically, I think we should just build it into salsa.

nikomatsakis:
> What I usually do is just ask salsa to dump all the code generated for hello world example. 

Yeah this is why I wrote up those "plumbing" chapters in the book, so I didn't have to keep doing that. I think there's a way to get a good diagram, but I didn't want to invest too much energy into it until we had worked out the changes due to this RFC. Note that the latest pushes implement the "triple index" scheme I described and also handle the tracking of dependencies via indices instead of direct references to slots. I think it worked out quite cleanly.

nikomatsakis:
Update: I think this RFC implies that the salsa::requires feature must be removed, as described in the latest commit.

book/src/rfcs/RFC0006-Dynamic-Databases.md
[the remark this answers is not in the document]
nikomatsakis: This turned out to be not that relevant as the only place that was using debug that I found was code generated by the procedural macro.

nikomatsakis:
A few notes:

I did some more implementation work and updated the RFC text a little. Everything seems to be going fairly smoothly. One thing I've noticed though:

* We used to take advantage of doing `db.salsa_runtime().xxx()` in order to call various plumbing methods (the `xxx()`) which were defined as `pub(crate)` inherent methods on `salsa::Runtime<DB>`. Since those methods have to be called from the storage code, and that code can no longer name the `Runtime` (as that type involves `DB`), those methods have to move to either methods on the `salsa::Database` trait (or some super trait) or methods on a `dyn Runtime`. The downside of the latter is that we now need to do `db.salsa_runtime().xxx()`, which is two virtual calls, and feels a bit silly. On the other hand, they are more "encapsulated" that way.
* I'm currently leaning towards moving the methods into the `DatabaseOps` trait, still, so that we are doing `db.xxx()` instead -- and perhaps giving them names like `salsa_xxx()` or `_xxx()` to make them a bit more obviously "plumbing methods". Good use case for a more flexible trait privacy system I guess!
* However, even if they are defined on the `salsa::DatabaseOps` trait, that raises the question of how to define their implementation. The methods can all be implemented via default methods on `salsa::Database` that invoke `db.salsa_runtime()`, but that would still compile to virtual dispatch. An alternative would be to have the procedural macro assume that the runtime is in a field named `runtime` (perhaps with the ability to pick a different one via a`#[salsa::runtime_field(my_runtime)]` attribute), so that I can then generate the implementations for those methods by having them do `self.runtime.xxx()`.
* Or maybe I should go with defining the `dyn Runtime` trait, combined perhaps with moving those methods that runtime methods that users are potentially encouraged to call (e.g., `synthetic_write`) into the `salsa::Database` (they can still forward through the `salsa_runtime().xxx()`). This would mean that we can define all methods in `dyn Runtime` as "plumbing and liable to change" and draw a kind of clear boundary, at the cost of potentially some extra virtual dispatch... hmm. Seems not terrible. I might go that way to start at least.
* (An alternative to all this would be to somehow have the `Runtime` type not include the `DB`, but given that the `Runtime` presently owns the `DB::DatabaseStorage` values (and puts them in shared state etc) I'm not sure how we would do that.)

nikomatsakis:
OK, I realized the solution to my dilemma and pushed an implementation of it. You now define databases via `storage: Storage<Self>` instead of directly embedding the runtime, but the runtime just has the type `Runtime` and doesn't require the DB type parameter. This allows us to continue using `pub(crate)` for plumbing which is very convenient.

nikomatsakis:
(I suppose it would have been a less *invasive* change to split out `Runtime` into something else, so that users still wrote `runtime::Runtime<Self>`... but I sort of like the name "storage"...)

nikomatsakis:
OK, I've basically implemented this RFC now, but I hit one final obstacle I hadn't anticipated: In the absence of GATs, the structure that I've setup here, lacking GATs, basically makes it impossible for databases to carry non-static borrows.

The problem stems from one of the nicest simplifications in this RFC. Instead of having everything generic over `DB`, things are instead just generic over a `Q: Query`, and for the database they get a `Q::DynDb`, which is short for something like `dyn MyQueryGroup`. This makes a lot of gnarly generic signatures a lot simpler. However, the *problem* is that `type DynDb = dyn MyQueryGroup` is really short for `type DynDb = dyn MyQueryGroup + 'static`. 

This is annoying because the parameter is `db: &Q::DynDb`, and if we were writing things out without generics we'd have `&'a dyn MyQueryGroup` which would default to `&'a dyn (MyQueryGroup + 'a)`. Unfortunately, we are not writing things out by hand, we're working with generics, and it doesn't work like that.

If we had GATs, we could make it `db: &Q::DynDb<'_>`, and things would work out fine. Even without GATs, there's probably a way to get support with a bit of "fancy footwork". We'd have another trait that carries the `DynDb` associated type which also has a lifetime parameter, so that we wind up with `db: <Q as DynDbRef<'_>>::DynDb` or something like that. 

That said, right *now*, database traits don't really support generic parameters anyway, so we're not actually losing anything afaik by requiring `'static`.

nikomatsakis:
OK, fully rebased, tests pass, but I do have a few pending updates to the RFC text itself:

* Describe the `'static` limitation
* I removed the `db.query(Q)` methods entirely in favor of `Q.in_db(&db)`. This does not require an extension trait and also works better with `dyn Db` coercions.

book/src/rfcs/RFC0006-Dynamic-Databases.md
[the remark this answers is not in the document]
nikomatsakis: In the latest version, I would up with `TheQuery.in_db(&db)` instead. This works better because we need to coerce the `&db` type to a `&dyn DB` anyway. It's also nicely ergonomic and avoids turbofish. Not as discoverable. We could, side note, add a method like `db.xxx_query()` that returns the `QueryTable`.

book/src/rfcs/RFC0006-Dynamic-Databases.md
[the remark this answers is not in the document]
nikomatsakis: Yeah, that makes sense

nikomatsakis:
BTW, I adjusted the docs, so if you get a chance, do a checkout and run `mdbook serve`. I think the diagram + links is reasonably comprehensible, or at least to help in "refreshing one's memory"

components/salsa-macros/src/query_group.rs
```diff
@@ -213,7 +213,15 @@ pub(crate) fn query_group(args: TokenStream, input: TokenStream) -> TokenStream
...
         query_fn_definitions.extend(quote! {
             fn #fn_name(&self, #(#key_names: #keys),*) -> #value {
-                salsa::plumbing::get_query_table::<#qt>(self).get((#(#key_names),*))
+                // Create a shim to force the code to be monomorphized in the
+                // query crate. Our experiments revealed that this makes a big
+                // difference in total compilation time in rust-analyzer, though
+                // it's not totally obvious why that should be.
+                fn __shim(db: &dyn #trait_name,  #(#key_names: #keys),*) -> #value {
+                    salsa::plumbing::get_query_table::<#qt>(db).get((#(#key_names),*))
+                }
```
matklad, line 222: Will this work before polymorphisation work lands? I hope that rustc is not dumb enough to monomorphise non-generic functions, but still, one can not be sure... I guess, I'll just benchmak this.

matklad: We also want the same for `input` queries I think (for both get & set), I saw then in `cargo llvm-lines` output as well (with not as high impact obviously).

nikomatsakis: This should not require polymorphization, as this function is not generic (unlike a closure, a nested function item does not inherit generics from its context).

nikomatsakis: But by all means we should test

nikomatsakis: (Pushed a commit extending this to input queries)

matklad:
Some doctests fail

matklad:
See https://github.com/rust-analyzer/rust-analyzer/issues/1987#issuecomment-654696185 for visualization of impact on compile times.

nikomatsakis:
OK, doc-tests are fixed. I think we're going to merge this once CI is happy.

PR #232, opened 2020-06-30
feat: Prune unnecessary dependencies

Marwes:
Only AtomicCell is used from crossbeam and SmallRng from rand so we can
easily disable the default features


nikomatsakis:
The LRU tests fail, though in my branch for moving over to a purely "dyn" setup, I removed those tests ... I wouldn't be opposed to removing them in this PR necessarily.

PR #234, opened 2020-07-03
Make salsa buildable on power

matklad:
rust-analyzer fails in rust-lang/rust repo on the power target,
because it doesn't have `AtomicU64`. So, lets just use `AtomicUsize`
as a revision number.

Semantically this is wrong, as we are not using revision to address
arrays, but, practically, it's a nice compromise to build on targets
without AtomicU64 and to have large revision numbers on most relevant
targets.


matklad:
going to merge this and do a minor release, want  to ship ra with rustup already :-)

PR #239, opened 2020-07-14
RFC: Idempotent-synchronized

nikomatsakis:
This is an RFC to make queries idempotent by default. It also describes various resulting changes to the implementation strategy. The end result is that executing "default" (non-synchronized) queries requires zero explicit locks, unless the query key has never been seen before, though it does require a lookup in a concurrent hashmap.

[Rendered RFC text.](https://github.com/nikomatsakis/salsa/blob/idempotent-synchronized/book/src/rfcs/RFC0007-Idempotent-Synchronized.md)

### DIFF
--- a/book/src/rfcs/RFC0007-Idempotent-Synchronized.md
+++ b/book/src/rfcs/RFC0007-Idempotent-Synchronized.md
@@ -55,7 +55,20 @@ The proposal is to have four categories of cached queries:
 | Transparent | ❌ | ❌ | ❌ |
 | Dependencies | ✔️ | ❌ | ❌ |
 | **Cached** (default) | ✔️ | ✔️ | ❌ |
-| Syncronized | ✔️ | ✔️ | ✔️ |
+| Synchronized | ✔️ | ✔️ | ✔️ |
+
+### When do you want to synchronize?
+
+There are two reasons to synchronize queries:
+
+* Not deterministic: Queries that are not fully deterministic need to be
+  synchronized. If executing the query twice might yield truly different
+  results, the system will behave in unpredictable ways. An example of a
+  non-deterministic query might be a query that executes a user supplied
+  procedural macro. Marking such a query as synchronized will ensure that we
+  execute the macro at most once and use the result.
+* Performance: If a query is expensive, it might be better to block waiting
+  for the result than to execute it twice.
 
 ## Reference guide
 
@@ -69,36 +82,35 @@ and structures:
 * `key_indices: DashMap<Q::Key, u32>` -- this maps from a key to an index in the
   `key_data` slab.
 * `key_data: ShardedSlab<KeyData<Q>>` -- this permits indexed access to a
-  particular key with no locks. 
+  particular key with no locks.
 * the `KeyData<Q>` struct stores the following information for each key:
   * the `Q::Key` value
-  * an `ArcSwapOption<Q::Value>` storing the value
-  * an `ArcSwapOption<Dependencies>` storing the dependencies
-* the `Dependencies` structure is similar to today's `MemoRevisions`, storing:
-  * a `verified_at: AtomicCell<Revision>`
-  * a `changed_at: Revision`
-  * a `durability: Durability`
-  * an `inputs: Option<Arc<[DependencyKeyIndex]>>`:
-    * where `None` means that there is an untracked input
+  * an `ArcSwapOption<Memo<Q::Value>>` storing the memoized value/dependencies
+* a `Memo<V>` combines a (optional) memoized value and its dependencies, and stores:
+  * the value `Option<V>`
+  * a `MemoRevisions` roughly equivalent to today, but with an `AtomicCell<Revision>`
+    for the verified-at field
+  * a lru-index field for use with the LRU code (only used if the value is provided)
 * a `synchronization: Mutex<FxHashMap<usize, InProgress>>` map used for `#[synchronized]` queries,
-  but otherwise unused
-* the `InProgress` struct stores the following data, taken from today's `QueryState`:
-  * `id: RuntimeId`, which stores id of the leader runtime so that we can detect cycles
-  * `waiting: Mutex<SmallVec<[Promise<...>; 2]>>`, which stores a promise for each blocked runtime,
-    so that they can be sent the value once it is computed
+  but otherwise unused, where `InProgress` stores the same data as today's `QueryState::InProgress` variant
 
-Within the runtime, we still store the active query stack in a
+### Runtime structures
+
+Within the runtime, we store the active query stack in a
 `FxIndexMap<DatabaseKeyIndex, ActiveQuery>`. This permits fast access to the
 current query (it's the "last" item in the map) and also quick detection of
-"intra-thread cycles".
+"intra-thread cycles" (if there is already a value for a given key, that's a
+problem). Since this stack is thread-local, use of an index map is not a
+problem.
 
 ### Executing a query
 
 Executing a query `db.query(key)` now takes the following steps,
 both of which are described in more detail below:
 
-* Create an index for `key` if one does not already exist
-* Compute the value given the inex
+* Create an index for `key` if one does not already exist.
+* If there is a memo, check if the dependencies are still valid and return the enclosed value, if so.
+* Compute the value given the index.
 
 ### Create an index for a key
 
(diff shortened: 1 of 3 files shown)


book/src/rfcs/RFC0007-Idempotent-Synchronized.md
```diff
@@ -0,0 +1,226 @@
...
+`FxIndexMap<DatabaseKeyIndex, ActiveQuery>`. This permits fast access to the
+current query (it's the "last" item in the map) and also quick detection of
+"intra-thread cycles" (if there is already a value for a given key, that's a
+problem). Since this stack is thread-local, use of an index map is not a
+problem.
+
+### Executing a query
+
+Executing a query `db.query(key)` now takes the following steps,
+both of which are described in more detail below:
```
matklad, line 109: off by one error, there are three steps

book/src/rfcs/RFC0007-Idempotent-Synchronized.md
```diff
@@ -0,0 +1,226 @@
...
+This RFC does not address the problem that we cannot accommodate queries whose
+values do not implement `Eq` but which *are* cached. We used to permit this
+with volatile queries, but we no longer support those. Note that if we permitted
+values to be cached but only within one revision, this would address the same
+use cases as volatile (particularly combined with volatile reads).
+
+### Backdating with hashes
+
+Another way to handle "backdating" besides using `Eq` is to permit storing a
+cryptographic hash of the value instead of the value itself, which would permit
```
matklad, line 216: Unimportant, but is there a better term than "cryptographic" here? We don't need hashing to be one-way, we just need the probability of collision to be negligible.

matklad:
Hm, I've slept on it, and I think I have concern about **cascading duplicate work**. Usually, when similar racy writers synchronization is involved, the amount of work is small. In salsa, queries call other queries, and that can balloon the CPU time. 

Let's say we have `a -> b -> c` query structure. And let's say threads 1 & 2 try to compute `a` simultaneously. They see that the cache is empty, so they both start executing it. Then, they both notice uncached `b`, and try to compute it in turn. And, because computing `b` needs `c`, it seems pretty likely that they'll actually conflict on `b`. 

Also, I am still not convinced that synchronising all queries necessary adds a mutexes... I haven't tried writing the code, but I think something like this should work (storing waiters in memo):

* load `Memo`
  * if it has value, `return`
  * if it is None
    * `ArcSwapOption::compare_and_swap` memo for a memo with waiters
    * compute value, store `Memo`, wake waiters (`event_listener` might be useful here)
  * if it is non-None, but the value is absent
    * add oneself to the list of waiters

I *think* we could make waiter list operations wait free at least for case where there are no waiters?

It seems the main trick in the RFC is `ShardedSlab`, and not necessary waitlist avoidance?

nikomatsakis:
@matklad 

Yeah, you're right that it doesn't have to involve a *mutex*, when I wrote that I was thinking more that it has to involve some kind of atomic write and blocking (which I kind of think of as a 'mutex' in my head).

In any case, the setup that you described feels basically the same as what we had today, except that instead of storing a *lock* inside the `Slot`, we use compare-and-swap to modify it. 

I suppose though that, even if we adopted the idempotent distinction, the "synchronized" operation might be better implemented this way. Have to think about that one.

I just remembered another of my motivations: I do not think we can accommodate chalk-style monotonic queries that span across threads. It's just too complicated, since in the event of cycles they have to re-execute continuously. But that might just be a distinct kind of query, that is built separately from the "derived" codebase (i.e.,`#[monotonic]` or something). I've already been leaning towards the idea that a `#[monotonic]` query is only well-defined in the event that it calls itself (i.e., the cycles must all consist of the same monotonic query constituent). That's all chalk needs, anyway.

I think the core question here though is this point you raised:

> Hm, I've slept on it, and I think I have concern about cascading duplicate work. Usually, when similar racy writers synchronization is involved, the amount of work is small. In salsa, queries call other queries, and that can balloon the CPU time.

This is definitely a *possibility* with this design, though I'm not sure if it's a problem. One thing I was thinking yesterday is that I'd like to try and do a "case study" of rust-analyzer's (and perhaps rustc's) existing queries, to see if we can loosely characterize them into categories and also look at what kinds of keys they use. In terms of rust-analyzer, it'd also be nice to instrument and take a look at their dependencies. This might then inform the question of whether this is a problem.

Other random thoughts:

* It's true that we'll be acquiring the mutex during setup. I didn't think too hard about this phase as it didn't seem that important, but a mutex isn't really required. It seems likely that `dashmap` has some kind of method to "insert if not already present", which is what we really need, but maybe it doesn't (I can't tell what sort of guarantees `entry` gives you). From peering at the implementation of dashmap, it uses some form of sharded mutex, and presumably we could so something similar as well.
* It's also true that there exist designs for concurrent hashmaps that don't acquire any locks *at all* on reads, so we could in principle use those, but I don't know if they've been implemented in Rust or how they perform in an absolute sense =)
* I didn't mention it here, but it seems like we would want to share the "hash-key-and-get-usize" step between derived and interned queries.
* Regarding `#[dependencies]`, I'm not surprised you're not using it much, I don't think it seems very useful myself. I'd be ok removing it, but I am interested to know if you still would like the "drop the value after each revision" query you once mentioned.

nikomatsakis:
Some separate musing:

I've mentioned "no locks at all" a few times. I'm thinking specifically of data structures like [masstree](https://github.com/rmind/masstree) or maybe this [concurrent trie hash-map](https://github.com/rmind/thmap), both implemented in C AFAIK. Those are also fairly old and there's probably newer stuff. In any case, the summary is that you have a tree-like structure such that you can walk down it to find a key without acquiring any locks. Writing involves locking the nodes on the path down, or something like that. At least that's what I dimly recall from a presentation I once watched. I've not read the source code.

One thing to note is that the API for [thmap](https://github.com/rmind/thmap), for example, says that insertion operations, if the key is already present, return the existing value *without changing it*. This means that, for our purposes of "create an index", they already do what we want without the need for any mutex.

In any case, I was thinking more about this, and also about the "explicit entity/component" we've talked about a few times. Certainly for the case where we knew that the key was an entity-id, we could kind of skip the first step of converting the key to an index. We still have to contend with the fact that our indices will be sparse, but we could conceivably use a tree like one of those, and store the component data as the value of the tree. It'd presumably just be a kind of `ArcSwap<Memo>` as described here -- we could then implement the 'synchronization', as you suggested, using atomic operations on that `ArcSwap`.

If we were to go so far as to say that *all keys have to be entity ids*, one could use interning to get entity-ids for other combinations, which is *basically* what we're doing here. Or we could just have salsa generate that step separately.

Anyway, that's appealing.

nikomatsakis:
(Side note, binary trees are also what specs uses to store component data, so that suggests that it's a decent choice.)

matklad:
> I do not think we can accommodate chalk-style monotonic queries that span across threads

Idle thought -- can we *require* that all cyclic queries has to be local? Ie, that the cycle across threads is always a bug.

nikomatsakis:
@matklad I don't think that is feasible. The cycles arise naturally from user input and aren't really in control of the compiler author.

nikomatsakis:
@matklad I think that roughly *is* my argument, just elaborated. In other words, I think most realistic applications will not be able to fully prevent cycles, so saying "cycles if your problem if you use threads" is effectively saying "don't use threads".

In any case, I'm still thinking about these ideas. One thing I realized is that it would be "non-trivial" to fully eliminate a mutex in order to handle the cross-thread registration case. There are just a couple of steps that need to be taken (e.g., adding the future into the vector, checking for cycles, etc) that need to be taken. I'm sure we could work out a protocol.

I'm also curious to read a bit more about masstree etc and bring that stuff back into cache. The idea of having some kind of specialized case around "ids as keys" is sort of appealing.

In any case, at this point I'm wondering if it would be a better approach to do some experimentation where we refactor the implementation in various ways (e.g., introducing concurrent maps etc) without changing the user model, and/or adding the monotonic queries I want for chalk integration as a separate thing first so that we have a clearer picture of that story.

matklad:
>but I am interested to know if you still would like the "drop the value after each revision" query you once mentioned.

Hm, I've realised that we accidentally hacked something like this already:

https://github.com/rust-analyzer/rust-analyzer/blob/d9d00988f2d03c61088332bd6546694b43209942/crates/ra_hir/src/semantics.rs#L89-L94

This `Semantics` structure *wraps* a database, and adds some per-request caches. This is sound, because caches are outside of the database, so we can't get to them from inside the queries. This is useful, because we cache data (syntax trees pimarily) which we want to evict aggressively. 

One difference from "drop the value after each revision" is that we drop value after "request", so that, if the user browses code-base without typing, we don't accumulate unbounded number of syntax trees. 

In terms of salsa, we set per-snapshot caches, which are efficiently thread-local. I think it makes sense to add such `#[snapshot]`  queries to salsa (with the assetion that they can only be called with an empty query stack).

PR #240, opened 2020-07-22
Add purge method

matklad:
This is mostly useful for debugging, rust-analyzer uses it to measure
memory usage of tables

### DIFF
--- a/src/derived.rs
+++ b/src/derived.rs
@@ -205,6 +205,10 @@ where
             slot.sweep(revision_now, strategy);
         }
     }
+    fn purge(&self) {
+        self.lru_list.purge();
+        *self.slot_map.write() = Default::default();
+    }
 }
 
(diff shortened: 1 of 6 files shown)


matklad:
rust-analyzer's side of things: https://github.com/rust-analyzer/rust-analyzer/pull/5494

matklad:
Specifically, this allows us to measure:

* the overhead of tables themselves 
* size of input queries.

matklad:
`#[doc(hidden)]` ?

nikomatsakis:
I feel like doc(hidden) is too big of a hammer.

matklad:
I guess, I wouldn't worry about discouraging use -- seems like even if you missue this, you'll notice the problem pretty quickly :)

nikomatsakis:
Yeah, seems fine.

PR #242, opened 2020-07-23
feat: Allow the dynamic db to be non-static 

Marwes:
Thought this would let me implement a non-refcell based query stack but that unfortunately did not work out. Still, this lifts the `'static` restriction around `DynDb`.

This doesn't actually allow databases with lifetimes to be declared as that would (at least) require changes in the proc macro to accept and propagate lifetimes but this is a step on the way towards that at least.

Getting this to work with the interned storage was tricky as I think there may be a rustc bug around higher ranked lifetimes but I did come up with a workaround at least.

### DIFF
--- a/book/src/rfcs/RFC0006-Dynamic-Databases.md
+++ b/book/src/rfcs/RFC0006-Dynamic-Databases.md
@@ -310,8 +310,8 @@ Therefore `QueryFunction` for example can become:
 
 ```rust,ignore
 pub trait QueryFunction: Query {
-    fn execute(db: &Self::DynDb, key: Self::Key) -> Self::Value;
-    fn recover(db: &Self::DynDb, cycle: &[DB::DatabaseKey], key: &Self::Key) -> Option<Self::Value> {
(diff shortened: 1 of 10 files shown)


Marwes:
Not directly, however the lifetime parameter on `QueryDb` ended up useful for the `Db` type added for async queries #244 . There might be some slightly simpler solution perhaps but not overly so.

Kixiron:
Anything that allows lifetimes to be used with queries is a big deal for me, I'd love to help if there's issues or tasks that need doing

nikomatsakis:
OK, well I'm inclined to merge, curious @matklad if you have any thoughts. It seems like it makes the code a bit messier but it's an interesting enabler, and I don't think it affects end users.

matklad:
LGTM!

I guess, I have some vague concerns about trying to make salsa fit for each and every use-case. Philosophically I tend to prefer "many specialized libraries" to "one general library". So, in that sense I think it's fine for salsa to make some specific trade offs about allocations, API, etc. 

But I don't have super strong feelings here, and this definitely doesn't affect end users, so :+1:

nikomatsakis:
OK, sorry for the delay, things (as always) got crazy, but let's merge this PR at least.

PR #243, opened 2020-07-25
Made proc-macros panic less

Kixiron:
Replaced the panics in the `query_group` macro with `syn::Error`s for better user feedback and experience
Panicking inside of macros doesn't give any locational feedback and can make very annoying-to-debug errors, so this replaces those with returning a `syn` compile error which lets us tag the specific area that's going wrong


nikomatsakis:
This seems great. I wonder if there is some way we can test it?

Kixiron:
I know that Dtolnay created [trybuild](https://github.com/dtolnay/trybuild) for this, but I'm not familiar with using it

PR #244, opened 2020-07-25
feat: Allow queries to be defined as async

Marwes:
This is a complete rewrite of my previous attempt of async queries and solves basically all the issues of that PR!

* The database only needs to be `Send` if an `async` query are used, normal synchronous queries can still have non-sync values, keys etc,

* Async queries require a mutable reference to the database but sync queries still works with shared references. To prevent async queries from calling mutating methods the database gets wrapped with a newtype which only provides public access to the async query methods or a shared reference which gives access to the sync queries.

* While the main derived query path is now async, sync queries still uses blocking and going from async to sync is a very light wrapper so there should be basically now overhead (not even any boxed futures!).

There are still some issues I need to cleanup and I need to add some more tests but basically everything is there already. You may note that there are a few hacks to get the types to work out (which I point out with comments in the source). Due to the use of higher ranked lifetimes rustc is unable to see that `Send` exists or that `DynDb` is the same in interned queries but with a pair of helper traits I appear to get around these issues (may need some deeper explanation).

### DIFF
--- a/book/src/rfcs/RFC0007-Async-Queries.md
+++ b/book/src/rfcs/RFC0007-Async-Queries.md
@@ -1,0 +1,102 @@
+
+## Metadata
+
+* Author: marwes
+* Date: 2020-08-04
+* Introduced in: [salsa-rs/salsa#1](https://github.com/salsa-rs/salsa/pull/1) (please update once you open your PR)
+
+## Summary
+
+Allow salsa databases to define derived queries that execute asynchronous code.
+
+## Motivation
+
+Asynchronous code is typically not needed for compilers as they typically do not need to wait on IO. However there is still utility in writing asynchronous queries
+as asynchronous tasks model queries waiting for eachother really well. There may also be non-compiler uses for salsa where the async nature may be useful (think caching a resource fetch over the network).
+
+### Our goal
+
+Allow `async fn` to be written for any query that needs to be `async`.
+
+## User's guide
+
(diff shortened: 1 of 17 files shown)


nikomatsakis:
Oh, wow, I should have looked at this PR before the parallel one. Seems like I'll need to re-evaluate the question of async vs sync. One thing that would be really nice, @Marwes, is if there were a new RFC added as part of this PR to help give the "big picture" of how it works.

Marwes:
Rebased and pushed an RFC. I think I covered everything important (though there is mostly just internal changes, for users all that is new is that `async fn` works).

book/src/rfcs/RFC0007-Async-Queries.md
```diff
@@ -0,0 +1,102 @@
+
+## Metadata
+
+* Author: marwes
+* Date: 2020-08-04
+* Introduced in: [salsa-rs/salsa#1](https://github.com/salsa-rs/salsa/pull/1) (please update once you open your PR)
```
akraines, line 6: Shouldnt this say pull/244

Marwes:
Any chance of this (and #245) getting merged? Would be nice to remove my fork of salsa.

Kixiron:
This would be really nice to have for working with things like [`lspower`](https://github.com/silvanshade/lspower)

nikomatsakis:
I'm going to read this over.

nikomatsakis:
I'm mass-closing PRs that aren't related to #305 -- @Marwes thank you for the tremendous effort here. At the moment, I don't think it's the direction we want to go in salsa. When async trait in fn lands, and other progress is made, it may be worth reconsidering, but at the moment I feel like the threaded model is treating us decently well.

PR #245, opened 2020-07-29
feat: Allow sub-queries to run concurrently

Marwes:
This is an update of [gluon's](https://github.com/gluon-lang/gluon) fork of salsa on top of #244. So it has been in use for a bit, but the implementation isn't great atm. However I would like to discuss if this API is good enough, or if something else is needed.

This adds a `forker` method to each `ParallelDatabase`. Through the returned `Forker` object it is possible to create new database instances on which queries can be executed concurrently as long as they finish before the `Forker` object is dropped. This is only checked dynamically since a scoped api like crossbeams scoped threads doesn't work well with futures. This is also the reason for cloning instead of borrowing the query stack.

In the synchronous case both of these could be fixed if necessary at the cost of some duplication.

cc #80

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ readme = "README.md"
 async-trait = "0.1"
 crossbeam-utils = { version = "0.7.1", default-features = false }
 indexmap = "1.0.1"
-futures-util = { version = "0.3", default-features = false }
 futures-channel = { version = "0.3", features = ["alloc"], default-features = false, optional = true }
+futures-util = { version = "0.3", default-features = false, optional = true }
 lock_api = "0.4"
 log = "0.4.5"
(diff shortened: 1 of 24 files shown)


nikomatsakis:
We've been discussing how we might manage parallel subqueries a bit on the Zulip, too. One thing that @matklad and I were discussing was adding the ability to have a kind of "prefetch", which then got me to thinking about the question of where the supply the thread-pool that would execute these prefetches concurrently -- I'm still debating about it, although the idea of "just use rayon" was growing on me. 

I guess a big question mark here is what to do about #244 -- I personally am still not too keen on shifting to async, and it seemed like that fork had some fairly serious downsides, but I guess I'd have to revisit it, since it's been a while since I've looked at it.

PR #247, opened 2020-09-13
update on_demand_inputs invalidation

ChristopherBiscardi:
The query in the on demand input example in the book seems out of date since at least https://github.com/salsa-rs/salsa/commit/9b9dbcca8c8e12b24b7c25470af9e070684821e1#diff-c86c61ef2a1841ef63294b8494f6a783

I also have a more complete example of integrating with notify [in this git repo](https://github.com/ChristopherBiscardi/salsa-file-watch-example/blob/f968dc8ea13a90373f91d962f173de3fe6ae24cd/main.rs) with a [writeup](https://www.christopherbiscardi.com/on-demand-lazy-inputs-for-incremental-computation-in-salsa-with-file-watching-powered-by-notify-in-rust) to go with it. I'm happy to put that code and such in this file if you think that would be better.

### DIFF
--- a/book/src/common_patterns/on_demand_inputs.md
+++ b/book/src/common_patterns/on_demand_inputs.md
@@ -43,7 +43,9 @@ impl FileWatcher for MyDatabase {
 }
 ```
 
-* We declare the query as a derived query (which is the default).
-* In the query implementation, we don't call any other query and just directly read file from disk.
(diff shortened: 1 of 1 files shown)


matklad:
LGTM! I think it's also fine to link to that writeup from the book. Could you add a link in this PR?

ChristopherBiscardi:
yup, will do

PR #248, opened 2020-09-17
Allow using a borrowed key in `DerivedQueryStorageOps::invalidate`

andrewhickman:


matklad:
What is the motivation for this change? `plumbing` is a semi-internal module, users shouldn't generally interact with its types. And, internally, I think we'll never be able to gain from the advanced flexibility here, as always use type parameters here.

andrewhickman:
This was motivated by the (lazy inputs example)[https://salsa-rs.github.io/salsa/common_patterns/on_demand_inputs.html] in the book. On the latest version of salsa, it becomes:

```
impl FileWatcher for MyDatabase {
    fn watch(&self, path: &Path) { ... }
    fn did_change_file(&mut self, path: &PathBuf) {
        ReadQuery.in_db_mut(self).invalidate(path);
    }
}
```

The key is an `&PathBuf`, which is slightly inefficient. 

Possibly it would be more idiomatic to use a `Arc<Path>` here, in which case this PR could be closed?

matklad:
Ah, right, this **is** a public API. I think we should have this then, thanks!

PR #250, opened 2020-10-13
Update crossbeam utils

matklad:


cynecx:
https://github.com/salsa-rs/salsa/pull/242 introduced some changes around the generated derive output and must be used with salsa-macro v0.15.3. So wouldn't that be a breaking change? (salsa-macros `0.15.2` is incompatible with salsa `0.15.3`).

matklad:
Ah indeed, my bad. I am yanking 0.15.3 and will take a close look at the set of changes since 0.15.2

PR #257, opened 2021-03-20
Add failing test for cycle revalidation

flodiebold:
The following test hits the `unwrap` in [runtime.rs:415](https://github.com/salsa-rs/salsa/blob/789cf5e99b9ec40715f30e0aac5bd53e6191d052/src/runtime.rs#L415). I encountered this panic while working on rust-analyzer.

As far as I understand, `report_unexpected_cycle` assumes the query we hit the cycle on has to be executing right now, but that's not the case here: it's just being revalidated (through `validate_memoized_value`). I'm not sure what the correct way to handle this is?

### DIFF
--- a/tests/cycles.rs
+++ b/tests/cycles.rs
@@ -123,6 +123,14 @@ fn inner_cycle() {
 }
 
 #[test]
+fn cycle_revalidate() {
+    let mut db = DatabaseImpl::default();
+    assert!(db.cycle_a().is_err());
+    CycleLeafQuery.in_db_mut(&mut db).invalidate(&());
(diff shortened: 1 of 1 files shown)


nikomatsakis:
Hmm. I dont' have time to think hard about this right now (working through notifications) but I'll try to ping on salsa zulip

matklad:
cc #147, the PR that added support for cycles, and also cc @Marwes (I wonder if you might see this panic in gluon's impors as well?)

nikomatsakis:
We've fixed the bugs here and I think we have good test coverage.

PR #258, opened 2021-03-23
Add tests for "dynamic" cycles

matklad:


nikomatsakis:
I think we have these cycle tests now.

PR #262, opened 2021-05-15
RFC: Opinionated cancelation

nikomatsakis:
@matklad and @jonas-schievink have been discussing a fresh take on salsa that builds on the experience we've gained with it, and one of the first simple steps we wanted to take was to make a more opinionated cancelation model. This is a write-up.

### DIFF
--- a/book/src/rfcs/RFC0007-Opinionated-Cancelation.md
+++ b/book/src/rfcs/RFC0007-Opinionated-Cancelation.md
@@ -1,0 +1,63 @@
+# Description/title
+
+## Metadata
+
+* Author: nikomatsakis
+* Date: 2021-05-15
+* Introduced in: N/A
+
+## Summary
+
+* Define panic as the one true way to handle cancellation in salsa queries
+* Modify salsa queries to automatically panic when they are canceled
+* Use a distinguished value for this panic so that people can test if the panic was a result of cancellation
+
+## Motivation
+
+Salsa's database model is fundamentally like a read-write lock. There is always a single *master copy* of the database which supports writes, and any number of concurrent *snapshots* that support reads. Whenever a write to the database occurs, any queries executing in those snapshots are considered *canceled*, because their results are based on stale data. The write blocks until they complete before it actually takes effect. It is therefore advantageous for those reads to complete as quickly as possible.
+
+Cancellation in salsa is currently quite minimal. Effectively, a flag becomes true, and queries can manually check for this flag. This is easy to forget to do. Moreover, we support two modes of cancellation: you can either use `Result` values or use `panic!`. In practice, though, there isn't much point to using `Result`: you can't really "recover" from cancellation.
+
+The largest user of salsa, rust-analyzer, uses a fairly opinionated and aggressive form of cancellation:
+
+* Every query is instrumented, using salsa's various hooks, to check for cancellation before it begins.
+* If a query is canceled, then it immediately panics, using a special sentinel value.
+* Any worker threads holding a snapshot of the DB recognize this value and go back to waiting for work.
+
+We propose to make this model of cancellation the *only* model of cancellation.
+
+## User's guide
+
+When you do a write to the salsa database, that write will block until any queries running in background threads have completed. You really want those queries to complete quickly, though, because they are now operating on stale data and their results are therefore not meaningful. To expedite the process, salsa will *cancel* those queries. That means that the queries will panic as soon as they try to execute another salsa query. Those panics occur using a sentinel value that you can check for if you wish. If you have a query that contains a long loop which does not execute any intermediate queries, salsa won't be able to cancel it automatically. You may wish to check for cancelation yourself by invoking the `panic_if_canceled` method.
+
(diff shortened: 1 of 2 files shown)


book/src/rfcs/RFC0007-Opinionated-Cancelation.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Reference guide
+
+The changes required to implement this RFC are as follows:
+
+* Adjust comments on `is_current_revision_canceled`.
+* Introduce a sentinel value `static SALSA_CANCELATION_TOKEN: SalsaCancelationToken = SalsaCancelationToken` that can be used with `panic_any!`
+* Introduce a `panic_if_canceled` method into the runtime which checks whether cancelation has occured and panics if so.
+    * This should probably be inline for the `if` with an outlined function to do the actual panic.
+* Modify the code for the various queries to invoke `panic_if_canceled` when they are invoked.
```
matklad, line 41: Not only invoked, but also when validated. If you have a complex query graph which is fresh, but we don’t know it yet, validating it will might take significant time, and it’s advantageous to cancel validation midway.

We did hit “slow validation causes lag” in rust-analyzer, although *maaaybe* that was before we introduced durability to speed up validation?

book/src/rfcs/RFC0007-Opinionated-Cancelation.md
```diff
@@ -0,0 +1,63 @@
...
+## User's guide
+
+When you do a write to the salsa database, that write will block until any queries running in background threads have completed. You really want those queries to complete quickly, though, because they are now operating on stale data and their results are therefore not meaningful. To expedite the process, salsa will *cancel* those queries. That means that the queries will panic as soon as they try to execute another salsa query. Those panics occur using a sentinel value that you can check for if you wish. If you have a query that contains a long loop which does not execute any intermediate queries, salsa won't be able to cancel it automatically. You may wish to check for cancelation yourself by invoking the `panic_if_canceled` method.
+
+## Reference guide
+
+The changes required to implement this RFC are as follows:
+
+* Adjust comments on `is_current_revision_canceled`.
+* Introduce a sentinel value `static SALSA_CANCELATION_TOKEN: SalsaCancelationToken = SalsaCancelationToken` that can be used with `panic_any!`
```
matklad, line 38: Does this need to be a value, as opposed to a ZST type?

matklad (approved):

book/src/rfcs/RFC0007-Opinionated-Cancelation.md
nikomatsakis: I was assuming it would be a ZST, yes. Admittedly these parts I was just sketching, I don't really know how the unwinding APIs work.

nikomatsakis:
@matklad thanks. tweaked.

matklad:
LGTM! I think we’ll want to tweak the specific API naming and such, not sure if it makes sense to nail them in the RFC. But, anyway:

* the sentinel value/type should probably named just `Cancelled`. `CancellationToken` is usually a name for something with `isCancelled` method, it’s not a result of cancellation.
* We should have some form of utility function that turns cancellation into Result. Basically, catch_unwind+downcast.
* `unwind_if_cancelled` is a good name. Alternatively, `unwind_if_cancellation_requestive`.


In principle, I quite like `request_cancellation/is_cancellation_requested` over `cancel/is_cancelled`. The *requested* part emphasizes the cooperative nature of cancellation, the fact that you need to wait for the cancellation to actually be acknowledged. In practice, as in salsa most of cancellation is automatic, we don’t really need the much more verbose version.

nikomatsakis:
I'm ok to remove the specific names, we'll have to bikeshed them sooner or later, of course.

nikomatsakis:
The implementation here seems pretty straightforward.

book/src/rfcs/RFC0007-Opinionated-Cancelation.md
```diff
@@ -0,0 +1,62 @@
...
+
+## User's guide
+
+When you do a write to the salsa database, that write will block until any queries running in background threads have completed. You really want those queries to complete quickly, though, because they are now operating on stale data and their results are therefore not meaningful. To expedite the process, salsa will *cancel* those queries. That means that the queries will panic as soon as they try to execute another salsa query. Those panics occur using a sentinel value that you can check for if you wish. If you have a query that contains a long loop which does not execute any intermediate queries, salsa won't be able to cancel it automatically. You may wish to check for cancelation yourself by invoking the `unwind_if_canceled` method.
+
+## Reference guide
+
+The changes required to implement this RFC are as follows:
+
+* Adjust comments on `is_current_revision_canceled`.
```
jonas-schievink, line 37: Will we even need this function with the new approach? It seems to me that there isn't much value is having this publicly available if all you can do is `resume_unwind(Canceled)`

nikomatsakis: it's true, I can't come up with a concrete scenario where you would want it.

PR #264, opened 2021-05-16
Fix book workflow

mbrobbel:
Removes the [deprecated](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) `add-path` command and bumps the versions.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -7,29 +7,25 @@ on:
   pull_request:
     paths:
     - 'book/**'
+    - '.github/workflows/book.yml'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/book.yml
```diff
@@ -7,29 +7,25 @@ on:
   pull_request:
     paths:
     - 'book/**'
+    - '.github/workflows/book.yml'
```
matklad, line 10: :+1:

matklad (approved):
bors r+

Thanks!

PR #268, opened 2021-06-16
Expose proc macro docs via the main crate

1tgr:
Improve visibility by placing the `database` and `query_group` proc macros at the bottom of the main docs page, instead of manually browsing to the docs for the `salsa-macros` crate.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -605,5 +605,4 @@ impl std::error::Error for Cancelled {}
 #[macro_use]
 extern crate salsa_macros;
 use plumbing::HasQueryGroup;
-#[doc(hidden)]
 pub use salsa_macros::*;


nikomatsakis (approved):

nikomatsakis:
bors r+

Thanks!

cormacrelf:
One question, does this make them inline or just show them as re-exports that link to the salsa-macros page? Should it have `#[doc(inline)]` instead?

1tgr:
They appear inline, ie as if the macros were originally defined in the `salsa` crate.

Which is nice, because (in my local docs at least), the page heading states:

> **Attribute Macro `salsa::query_group`**

(instead of `salsa_macros::query_group`, which is not how the macro would be used)

PR #269, opened 2021-06-16
Add `#[salsa::update]`

### LINKED ISSUE #46: allow derived queries to re-use the old result?
I was thinking that it would be useful if queries could get access to the old, invalidated value and be allowed to "mine it for spare parts" -- at minimum this would permit vectors to be truncated and hence re-use the memory, but one can also imagine scenarios where we want to do "delicate", hand-coded incremental updates.

Some examples:

- Incremental re-parse after a diff is applied
- Trait matching, chalk-style, where we can generally keep the tables in between queries but if the set of impls etc changes we may want to throw them away

I suppose incremental re-parse could be done by having the **inputs** to the system not be raw text but rather the parse trees.

The trait matching integration is a bit trickier. I have to think about how that would best work. The most naive thing would be to use a "volatile" query and just recreate the cached state on each revision; a trickier question would be whether we ever know which bits can be re-used.

1tgr:
Specifies a second function on a query that can see the previously cached value for the same query.

Closes #46.

[Rendered RFC](https://github.com/1tgr/salsa/blob/update/book/src/rfcs/RFC0009-Updatable-Derived-Queries.md)

### DIFF
--- a/book/src/rfcs/RFC0009-Updatable-Derived-Queries.md
+++ b/book/src/rfcs/RFC0009-Updatable-Derived-Queries.md
@@ -63,7 +63,7 @@ fn formatted_value(db: &dyn MyDatabase, key: u32) -> String {
     db.some_input(key).to_string()
 }
 
-fn update_formatted_value(db: &dyn MyDatabase, value: &mut String, key: u32) -> salsa::ValueChanged {
+fn update_formatted_value(db: &dyn MyDatabase, key: u32, value: &mut String) -> salsa::ValueChanged {
     // Write a new value into the existing String.
   
     // 1. Clear the existing String while keeping the memory allocation intact.
--- a/book/src/rfcs/RFC0009-Updatable-Derived-Queries.md
+++ b/book/src/rfcs/RFC0009-Updatable-Derived-Queries.md
@@ -47,8 +47,8 @@ fn formatted_value(db: &dyn MyDatabase, key: u32) -> String {
 When you apply the `#[salsa::update]` attribute to the function, Salsa looks for a second function,
 which it calls to update an existing value in response to changes in the query's dependencies. By
 default the update function has the word `update_` prepended to its name. Salsa passes an
-`&mut Value` reference to the update function, and it expects the update function to return `true`
-unless the value is unchanged.
+`&mut Value` reference to the update function, and it expects the update function to return
+`ValueChanged::True` unless the value is unchanged.
 
 ```rust
 #[salsa::query_group]
@@ -63,7 +63,7 @@ fn formatted_value(db: &dyn MyDatabase, key: u32) -> String {
     db.some_input(key).to_string()
 }
 
-fn update_formatted_value(db: &dyn MyDatabase, value: &mut String, key: u32) -> bool {
+fn update_formatted_value(db: &dyn MyDatabase, value: &mut String, key: u32) -> salsa::ValueChanged {
     // Write a new value into the existing String.
   
     // 1. Clear the existing String while keeping the memory allocation intact.
@@ -73,8 +73,8 @@ fn update_formatted_value(db: &dyn MyDatabase, value: &mut String, key: u32) -> 
     use std::fmt::Write;
     let _ = write!(value, "{}", db.some_input(key));
 
-    // 3. Return true, as we assume the value in the String has changed.
-    true
+    // 3. We assume the value in the String has changed.
+    salsa::ValueChanged::True
 }
 ```
 
@@ -84,7 +84,7 @@ fn update_formatted_value(db: &dyn MyDatabase, value: &mut String, key: u32) -> 
 The `QueryFunction` trait has functions `init` and `update`, corresponding to the two functions
 implemented by the user. The `update` function is responsible for making a best-effort determination
 of whether it changed the value: it has access to the `MP: MemoizationPolicy` generic parameter and
-returns `value_changed: bool`.
+returns `Value_changed`.
 
 ### Query execution
 In the case where no cached value exists (either the query is being called for the first time, or
@@ -94,8 +94,8 @@ Where a cached value does exist, `Slot::read_upgrade` does the following:
 1. `Option::take` on the cached value, to move the cached value to a local variable within
    `read_upgrade` and set the cached value to `None` while the query is being executed.
 2. Call `QueryFunction::update` with a mutable reference to the value. `update` modifies the value
-   in place through the reference and returns `value_changed: bool`.
-3. When `!value_changed`, resets `memo.revisions.changed_at` to the revision it had before the
+   in place through the reference and returns `ValueChanged`.
+3. When `ValueChanged == True`, resets `memo.revisions.changed_at` to the revision it had before the
    update
 4. Moves the value back to the cache
 
@@ -109,8 +109,7 @@ fn init(db, key) {
 
 fn update<MP>(db, key, value) {
     let old_value = mem::replace(value, execute(db, key));
-    let value_is_eq = MP::memoized_value_eq(&old_value, value);
-    !value_is_eq
+    MP::memoized_value_changed(&old_value, value)
 }
 ```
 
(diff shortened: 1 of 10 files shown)


nikomatsakis:
I'm interesting in this, but I think it merits an [RFC](https://salsa-rs.github.io/salsa/rfcs.html) -- can you push a commit with a writeup using [this template](https://github.com/salsa-rs/salsa/blob/master/book/src/rfcs/template.md)?

cormacrelf:
Agreed; in particular I would like to see if it's possible to expose information about which dependencies or maybe inputs had actually changed, to maximise the computation that can be salvaged.

(My personal use case involves an `O(n*m)` computation (a bit like the python list comprehension `[f(x, y) for x in m_list for y in n_list]`) for which a change to one of the "m"s should result in an `O(n)` scan, and a change to one of the "n" inputs should result in an `O(m)` scan. If only you could see which ns and ms caused the recomputation!)

1tgr:
> in particular I would like to see if it's possible to expose information about which dependencies or maybe inputs had actually changed, to maximise the computation that can be salvaged

I don't address that here, but elsewhere I've seen a feature where a query function can ask the framework to decorate individual input values with their revision ids. If the query function stashes `max(revision_id)` in its own value then it can inspect  incoming revision ids which ones are more recent.

cormacrelf:
<details>
<summary>off-topic</summary>

That's a pretty cool idea! My first impression is that you could use it for another problem, which is where the scan for changed dependencies (using Eq) gets too expensive. Because the return value includes a revision number, that would encode the semantics "assume this query's output is different for every input change". Imagine a factorial query producing bignums, where the factorial is slow to compute but you get no advantage from comparing its output equality (bignum, slower) as opposed to just assuming it changes when its input does (usize or whatever salsa uses these days).

Unfortunately not perfect out of the box if you still need to prune downstream query execution based on the Eq implementation as salsa::memoize does normally, because now no two query outputs are ever Eq. You would want two wrapper types, one that ignores the value in the PartialEq impl and defers to the revision number (`AssumeChanged`), and one that does the opposite (`WithRevision`). The former optimises the case where there are no changes for faster verification, the latter just stores the revision for use by downstream queries without defeating said Eq-pruning of those same queries.

I'll have a think about this, thank you.

</details>

nikomatsakis:
Thanks for the RFC! Reading now

nikomatsakis:
This is a great write-up! I quite like the design. I am thinking about how it interacts with the direction I had intended to go for parallel computation. At first, I thought there was a conflict, now but I think it should work ok. The basic idea that I am working towards is that the *global caching* is really just a simple hashmap for completed queries that stores the value + dependencies -- unlike now where it serves to both "lock" the query and encode the overall computation state.

I think that we could adapt that design to the `update` function by saying that, if we find that the cached value is out of date, we will remove it from the hashmap (and take ownership) and thus be able to invoke the `update` function with an `&mut` reference.

components/salsa-macros/src/query_group.rs
```diff
@@ -499,10 +509,27 @@ pub(crate) fn query_group(args: TokenStream, input: TokenStream) -> TokenStream
...
             let invoke = query.invoke_tt();
 
+            let update = if let Some(update) = &query.update {
+                quote! {
+                    fn update<MP: salsa::plumbing::MemoizationPolicy<Self>>(
+                        db: &<Self as salsa::QueryDb<'_>>::DynDb,
+                        #key_pattern: <Self as salsa::Query>::Key,
+                        value: &mut <Self as salsa::Query>::Value,
+                    ) -> bool {
+                        #update(db, value, #(#key_names),*)
```
nikomatsakis, line 519: I'm curious why the keys come *after* the value -- that somewhat surprised me. It doesn't really matter, of course.

src/derived/slot.rs
```diff
@@ -206,14 +206,25 @@ where
 
                 return Ok(value);
             }
-        }
+
+            memo.value.take()
```
nikomatsakis, line 210: I see. This change in behavior with respect to panics applies regardless of whether update is used or not. Nonetheless, I think it's ok. I think dropping the cached value is a reasonable recovery strategy for panics.

book/src/rfcs/RFC0009-Updatable-Derived-Queries.md
```diff
@@ -0,0 +1,224 @@
...
+    fn formatted_value(&self, key: u32) -> String;
+}
+
+fn formatted_value(db: &dyn MyDatabase, key: u32) -> String {
+    // Our query is being called for the first time and it has no value already cached.
+    // Return a new String.
+    db.some_input(key).to_string()
+}
+
+fn update_formatted_value(db: &dyn MyDatabase, value: &mut String, key: u32) -> bool {
```
nikomatsakis, line 66: Nit: As noted elsewhere, I think the key(s) should come *first*. 

Other nit: I think I would prefer to return an enum like `salsa::ValueChanged::True` vs `salsa::ValueChanged::False`. We can also add a comment to that enum explaining that this result is **not a hint** and must be accurate, etc.

nikomatsakis (changes requested):
Did a real review. A few nits and minor thoughts.

components/salsa-macros/src/query_group.rs
1tgr: I think it's my habit to put `&mut` parameters at the front, `strcpy`-style. I agree there is a tension vs `key`-`value`, and the framework internals (eg `QueryFunction`) have `key` then `value`.

book/src/rfcs/RFC0009-Updatable-Derived-Queries.md
1tgr: I pushed this, although what about `ValueChanged::No` and `ValueChanged::Maybe`?

matklad:
This looks reasonable to me! I must confess I don't fully understand why we need two functions though -- it seems the following would work:

```rust
fn formatted_value(db: &dyn MyDatabase, prev_value: Option<String>, key: u32) -> (String, ValueChanged)
```

But that's a minor detail -- it's OK to have two functions in the API and do code-sharing in the user's code.

I have some doubts on how useful this would be in practice. It seems in the general case you need both old and new values to determine `ValueChange` exactly. If it so happens that every usage of `update` does return `ValueChange::True` unconditionally, than we can simplify API. But it's also the case that returning `ValueChange::False` after update is what makes salsa truly efficient. There seems to be an unfortunate tradeoff between storage re-use and early stopping of the change propagation. 

The second point is that I think this doesn't help with incremental updates for situations where diff exist (that is, segment trees). To give an example, let's say we have a database whith an input query which returns a mulitiset, and a derived query which returns a union of all inputs:

```rust
trait Db {
  fn input(&self, input_index: u32) -> Map<String, u32>
  fn union_all(&self) -> Map<String, u32>
}
```

In this setup, the ideal update algorithm for when an input changes is to compute the diff of this specific input, and apply the diff to the end result. 

This all sounds pretty abstract, but is how, eg, IntelliJ maintains its indexes. 

Obviously, this RFC shouldn't solve this problem -- it'd just be nice if it could :-)

cormacrelf:
> In this setup, the ideal update algorithm for when an input changes is to compute the diff of this specific input, and apply the diff to the end result.

You can just use Arcs and store the previous individual dependency outputs in the union_all output. @1tgr's idea from above https://github.com/salsa-rs/salsa/pull/269#issuecomment-863388522 strikes again. This basically gets you over the line, as the inputs map is available (and reusable) in the update function:

```rust
trait Db {
    fn input(&self, key: InputKey) -> Arc<Value>;
    #[update]
    fn union_all(&self) -> (Value, Map<InputKey, Arc<Value>);
}
```

Obviously it affects the API a bit, but the potential of the update function on its own is not to be underestimated. You could of course do all sorts of macro things for specific requirements like this (and my one above) to inject dependency info into the update function. But use a struct and implement Deref and it's good enough I think.

The other benefit is that each query that needs this behaviour knows which inputs it cares about, but salsa can't know. So we don't have to change the way we store inputs on a slot just to make all of them visible in the update function always, when doing so is possibly wasteful.

nikomatsakis:
I'm "mass closing" open PRs that aren't related to #305 -- this one is pretty old, but I would want to reframe the approach in terms of #305 before we did anything here (although definitely still relevant). Apologies @1tgr for dropping the ball on follow-up here.

PR #271, opened 2021-06-16
CI runs Clippy

1tgr:


src/derived/slot.rs
```diff
@@ -663,6 +664,7 @@ where
     /// that work completes. This helper does that; it returns a port
     /// where you can wait for the final value that wound up being
     /// computed (but first drop the lock on the map).
+    #[allow(clippy::type_complexity)]
```
nikomatsakis, line 667: Can we just put this allow at the crate level? I'm not into this lint.

tests/parallel/independent.rs
```diff
@@ -8,11 +8,11 @@ fn in_par_two_independent_queries() {
     let mut db = ParDatabaseImpl::default();
 
     db.set_input('a', 100);
-    db.set_input('b', 010);
-    db.set_input('c', 001);
+    db.set_input('b', 10);
+    db.set_input('c', 1);
     db.set_input('d', 200);
-    db.set_input('e', 020);
-    db.set_input('f', 002);
```
nikomatsakis, line 15: the formatting here was intentional, but...it's fine

nikomatsakis (changes requested):
I've never really used clippy before, but for the most part these changes seem good.

tests/parallel/independent.rs
1tgr: I agree it's not a big deal, but maybe 200, 220, 202?

nikomatsakis:
cargo fmt errors

nikomatsakis (approved):

nikomatsakis:
bors r

PR #273, opened 2021-06-17
Expose the ability to update the value of an input query in place

1tgr:
Related to, but separate from, #46 and #269.

### DIFF
--- a/book/src/rfcs/RFC0009-Updatable-Inputs.md
+++ b/book/src/rfcs/RFC0009-Updatable-Inputs.md
@@ -1,0 +1,63 @@
+# Updatable inputs
+
+## Metadata
+
+* Author: 1tgr
+* Date: 2021-06-17
+* Introduced in: https://github.com/salsa-rs/salsa/pull/1 (please update once you open your PR)
+
+## Summary
+
+- Update the value of an input query in place
+
+## Motivation
+
+- Adds an `update_some_input` function to input queries, which allows amendment of an input query's
+  value without cloning it.
+
+## User's guide
+
+Input queries in Salsa are defined in terms of a pair of functions:
+
+- `fn some_input(&self, key: u32) -> String`
+- `fn set_some_input(&self, key: u32, value: String)`
+
+Initial values are typically set when the database is instantiated, and are updated using a pattern
+like this:
+
+```rust
+let mut value = db.some_input(123);
+value.push_str("hello");
+db.set_some_input(123, value);
+```
+
+The `some_input` function works by fetching the value from the input storage and returning a clone.
+Cloning this value can be expensive.
+
+This proposal exposes a third function, which can be used to update the value on an input query in
+place, through a mutable reference to the underlying storage:
+
+- `fn update_some_input<F>(&self, key: u32, value_fn: F) where F: FnOnce(&mut String)`
+
+Under this proposal, the update pattern becomes:
+
+```rust
+db.update_some_input(|value: &mut String| {
+    value.push_str("hello");
+});
+```
+
+## Reference guide
+
+We expose a `fn update` on the `InputQueryStorageOps` trait. The implementation of this function on
+`InputStorage` requests a new revision and acquires a write lock (like `InputStorage::write`), but
+it panics if the input query does not already have a value set (like `InputStorage::try_fetch`).
+
+## Frequently asked questions
+
+### What if the closure doesn't change the value?
+
+Salsa is designed to invalidate downstream queries when an input has its value set, regardless of
+whether the value changed. This proposal does not change this assumption, and a call like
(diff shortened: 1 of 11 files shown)


book/src/rfcs/RFC0009-Updatable-Inputs.md
```diff
@@ -0,0 +1,63 @@
...
+db.set_some_input(123, value);
+```
+
+The `some_input` function works by fetching the value from the input storage and returning a clone.
+Cloning this value can be expensive.
+
+This proposal exposes a third function, which can be used to update the value on an input query in
+place, through a mutable reference to the underlying storage:
+
+- `fn update_some_input<F>(&self, key: u32, value_fn: F) where F: FnOnce(&mut String)`
```
nikomatsakis, line 40: I think it should return `-> salsa::ValueChanged`, just like the other update function you proposed. It sometimes happens that the input value was the same as the new value you need.

nikomatsakis: Ah, I see you addressed that point. We will have to make a new revision whatever happens, but I'm not sure why we should force the value to be assumed to have changed.

src/input.rs
```diff
@@ -242,6 +242,30 @@ where
...
+            .with_incremented_revision(&mut |next_revision| {
+                let slots = self.slots.write();
+
+                let slot = slots
+                    .get(key)
+                    .unwrap_or_else(|| panic!("no value set for {:?}({:?})", Q::default(), key));
+
+                let mut slot_stamped_value = slot.stamped_value.write();
+                let old_durability = slot_stamped_value.durability;
+                value_fn(&mut slot_stamped_value.value);
```
nikomatsakis, line 263: What happens in the event of a panic? That seems like a problem (and the question should be added to the RFC).

nikomatsakis: I guess the answer is probably that the panic propagates out and the input is in an invalid state. In other words, unlike for derived queries, we can't recover from panics in the mutator thread. This seems reasonable, but we should document it. (At least, I don't see how we could recover.)

nikomatsakis (changes requested):
Looks good. Left some thoughts.

book/src/rfcs/RFC0009-Updatable-Inputs.md
1tgr: Right, I think it's helpful for the closure to be able to change its mind by returning early and requesting that the value not be invalidated. Let's make these two new `update` functions consistent and update this one to return `salsa::ValueChanged`?

src/input.rs
1tgr: Agreed, I'll add to the RFC. For consistency with the other RFC we would unmemoize the value, by calling `take` then `value_fn`. This works on the other RFC because it just means the derived query re-calculates the value. But a missing input query is different from an unmemoized derived value, as a missing input query will panic the next time.

matklad:
(on mobile, haven’t read through carefully)

A plausible alternative is to add remove function to input queries. Then,
update is modeled as remove/set pair.

One benefit is that we don’t need to think about semantics of closure based
API (what if update recursively calls update, etc).

Another benefit is that remove function is useful in itself: today rust
analyzer works around absence of remove by setting text of deleted files to
an empty string.

I don’t remember though why we don’t already have remove?

1tgr:
With the remove/set pair, would there be a race after the remove and before the set?

Under this proposal, the user's closure is called with the slot's write lock held.

1tgr:
Here's how it looks in terms of a `remove_input` function: #275

book/src/rfcs/RFC0009-Updatable-Inputs.md
1tgr: > We will have to make a new revision whatever happens

So we would have the roll back the update in the `ValueChanged::False` case? In any case, this means decrementing the current revision number. But we may have also cancelled some queries via `SharedState::pending_revision`, which can't be un-cancelled.

nikomatsakis: I think we would just always increment the revision -- but we can still avoid altering the "last changed revision" for this input, which means that revalidating derived queries should be cheap.

matklad:
>With the remove/set pair, would there be a race after the remove and before the set?

I don't think so -- we require unique (runtime checked) access to the database to mutate it's state. Ie, if you do `let x = db.remove(key); db.set(key, modify(x))`, than nothing can access the database from some other thread. What might happen is that `modify` itself can look at the db, but that seems unlikely.

godcodehunter:
Is it still valid?

nikomatsakis:
closing in favor of #275

PR #274, opened 2021-06-25
RFC: Parallel/Chalk friendly caching

nikomatsakis:
# Motivation

Two-fold:

* To integrate chalk caching with salsa more deeply, we need to be able to handle cycles more gracefully.
    * In chalk, cycles are handled by iterating until a fixed-point is reached. This could be useful in other salsa applications as well.
    * For this to work, we need caching of *provisional results* that are produced during those iterations. These results should not be exposed outside the cycle.
    * In general, it would be simpler if we could move cycle handling to *per-thread*, although there are limits to how much we can do this.
* We are moving towards a 'parallel by default' future for Salsa.   
    * Eventual goal: Cache hit requires only reads (no locks).
    * Creates the option of queries where a cache hit requires no atomic writes.
    * Validation and execution require no locks and few atomic writes.
        * Although synchronization remains an option.
    * This RFC brings that within sight, although it is not achieved.

[Rendered form.](https://github.com/nikomatsakis/salsa/blob/global-vs-local-storage/book/src/rfcs/RFC0009-Parallel-Friendly-Caching.md)

# Status

Still a WIP! Feedback requested as I evolve the RFC and implementation.

# Implementation status

* [x] Split global and local state
* [ ] Rework global state to separate out the 'slot map', 'cached values', and 'synchronization state', as described in the notes


matklad:
cc @flodiebold

flodiebold:
I did read this, but I don't really know enough about Salsa internals to have much to say about it :sweat_smile:

mheiber:
The link to the rendered RFC is broken, how about this uri instead? https://github.com/nikomatsakis/salsa/blob/global-vs-local-storage/book/src/rfcs/RFC0010-Parallel-Friendly-Caching.md

Also, the diff seems to contain docs changes to more than just the RFC, is that intentional?

mheiber:
> Salsa is free software and I'm no position to make demands, please take with a grain of salt

Since this is an RFC, my understanding is you're looking for freedback from Salsa users.
As a user, it's hard to evaluate, since the "Users' Guide" section of the RFC says "No user visible change".

It would be easier to evaluate something that *does* change Salsa's capabilities. Since it sounds like there is only one use case for local state so far (fixed-point queries), seeing what this local state API looks like would help.

I'm a bit nervous about changing architecture with the idea that it will be helpful later, when sometimes "later" doesn't come. For example, when first trying to understand Salsa internals [I immediately hit a wall with this higher-ranked trait bound](https://stackoverflow.com/questions/70557666/what-does-this-higher-ranked-trait-bound-mean). The HRTB was introduced in [a PR from Sept. 2020](https://github.com/salsa-rs/salsa/pull/242) with the intention of using it later–iuc the HRTB still doesn't do anything currently.

nikomatsakis:
I'm going to close this, actually. I've taken a different approach in the meantime. I may get back to this split at some point.

PR #275, opened 2021-06-25
Expose the ability to remove the value from an input query, taking ownership of the value

1tgr:


nikomatsakis (approved):
I know it's been super delayed, but I still think this is a good idea. I'm inclined to merge it. I'll try to adapt the other RFC.

nikomatsakis:
Hmm, I just realized that the impl of this is not going to work. By removing the item from the map, we invalidate it's `DatabaseKeyIndex`.

PR #276, opened 2021-06-28
Keep parameter names of queries in `query_group` attribute

Veykril:
Now that RA gets to a point where it can expand attributes it would be helpful not to discard parameter names.
Before:
Code_1XD5LiRfJw
After:
Code_kwRhvPySuL


nikomatsakis:
Test falure!

Veykril:
@nikomatsakis should be fixed now

PR #285, opened 2021-10-30
Cleanup and fix cycle handling

nikomatsakis:
This PR is...kind of long. It reshapes the core logic of salsa to fix various bugs in cycle handling and generally simplify how we handle cross-thread coordination.

Best read commit by commit: every commit passes all tests, afaik.

The core bug I was taking aim at was the fact that, when you invoke `maybe_changed_since`, you can sometimes wind up detecting a cycle without having pushed some of the relevant queries onto the stack. This is now fixed.

From a user's POV, ~~nothing changes from this PR~~, there are only minimal changes to the public interface. The biggest one is that recover functions now get a `&salsa::Cycle` which has methods for accessing the participants; the other is that queries that are participating in cycle fallback will use unwinding to avoid executing past the point where the cycle is discovered. Otherwise, things work the same as before:

* If you encounter a cycle and all participant queries are marked with `#[salsa::recover]`, then they will take on the recovery value. (At the moment, they continue executing after the cycle is observed, but their final result is ignored; I plan to change this in a follow-up PR, or maybe some future commit to this PR.)
* If you encounter a cycle and some or all participants are NOT marked with `#[salsa::recover]`, then the code panics. This is treated like any other panic, cancelling all other work.

Along the way, I made... a few... other changes:

* Cross-thread handling is simplified. When we block on another thread, it no longer sends us a final result. Instead, it just gets re-awoken and then it retries the original request. This is helpful when you encounter cycles in `maybe_changed_since` vs `read`, but it's also more compatible with some of the directions I have in mind.
* Cycle detection is simplified and more centrally coordinated. Previously, when a cycle was detected, we would mark all the participants on the current thread, but then we would mark other threads 'lazilly'. Now, threads move ownership of their stack into the shared dep graph when they block, so that we can mark all the stack frames at once. This also means less cloning on blocking, so it should be mildly more efficient.
* The code is DRY-er, since `maybe_changed_since` has been re-implemented in terms of the same core building blocks as `read` (`probe` and friends). I originally tried to unify them, but I realized that they behave somewhat differently from one another and both of them make sense. (In particular, we want to be able to free values with the LRU cache while still checking if they are up to date.)

Ah, I realize now that I had planned to write a bunch of docs in the salsa book before I landed this. Well, I'm going to open the PR anyway, as I've let this branch go far too long.

r? @matklad


tests/incremental/log.rs
```diff
@@ -11,6 +11,6 @@ impl Log {
     }
 
     pub(crate) fn take(&self) -> Vec<String> {
-        std::mem::replace(&mut *self.data.borrow_mut(), vec![])
+        std::mem::take(&mut *self.data.borrow_mut())
```
matklad, line 14: `self.data.take()` would work I think

tests/cycles.rs
```diff
@@ -40,10 +77,11 @@ trait Database: salsa::Database {
     fn volatile_b(&self) -> ();
 
     #[salsa::input]
-    fn cycle_leaf(&self) -> bool;
+    fn should_create_cycle(&self) -> bool;
```
matklad, line 81: :+1:

matklad (commented):
Going to come back to this tomorrow, but this is exciting!

components/salsa-macros/src/query_group.rs
```diff
@@ -539,6 +539,17 @@ pub(crate) fn query_group(args: TokenStream, input: TokenStream) -> TokenStream
         });
     }
 
+    let mut cycle_recovery_strategy_ops = proc_macro2::TokenStream::new();
+    for (Query { fn_name, .. }, query_index) in non_transparent_queries().zip(0_u16..) {
+        cycle_recovery_strategy_ops.extend(quote! {
+            #query_index => {
+                salsa::plumbing::QueryStorageOps::cycle_recovery_strategy(
+                    &*self.#fn_name
+                )
```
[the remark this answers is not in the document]
nikomatsakis: I don't remember, to be honest, if it got removed or not. I'm trying to remember why I did it this particular way... I guess you mean as opposed to something like `<&*self.#fn_name as Query>::CYCLE_STRATEGY` or something?

nikomatsakis:
Oh, man, reviewing my history, I see that one of the major changes I described is still in a stash! 

> Cycle detection is simplified and more centrally coordinated. Previously, when a cycle was detected, we would mark all the participants on the current thread, but then we would mark other threads 'lazilly'. Now, threads move ownership of their stack into the shared dep graph when they block, so that we can mark all the stack frames at once. This also means less cloning on blocking, so it should be mildly more efficient.

I'm going to try to unstash that and see what happens! Also, add some more tests, because I'm surprised the code works now. =)

src/plumbing.rs
```diff
@@ -237,3 +236,49 @@ where
...
         Q::Key: Borrow<S>;
 }
+
+/// The error returned when a query could not be resolved due to a cycle
+#[derive(Eq, PartialEq, Clone, Debug)]
+pub struct CycleError<K> {
+    /// The queries that were part of the cycle
+    pub(crate) cycle: Vec<K>,
+    pub(crate) changed_at: Revision,
+    pub(crate) durability: Durability,
```
matklad, line 245: Might make sense to add comments explaining the semantics of these two fields.

src/runtime/dependency_graph.rs
```diff
@@ -56,15 +62,32 @@ impl DependencyGraph {
...
         from_id: RuntimeId,
         database_key: DatabaseKeyIndex,
         to_id: RuntimeId,
         path: impl IntoIterator<Item = DatabaseKeyIndex>,
-    ) {
+        query_mutex_guard: QueryMutexGuard,
+    ) -> Option<WaitResult> {
         me.add_edge(from_id, database_key, to_id, path);
+
+        // Release the mut&mut meex that prevents `database_key`
```
matklad, line 144: Particularly confusing typo :)

Cargo.toml
```diff
@@ -26,5 +26,6 @@ linked-hash-map = "0.5.2"
 rand = "0.7"
 rand_distr = "0.2.1"
 test-env-log = "0.2.7"
+insta = "1.8.0"
```
matklad, line 29: FWIW, in rust-analyzer we switched from `insta` to https://github.com/rust-analyzer/expect-test quite a while ago. expect-test is  a bit more minimal. Rather than needing a `cargo-insta` command to upgrade the thest, the test runtime modifies code in-place. 

I kinda like it more, but no strong opinion!

matklad (commented):
Finish reading by commit, looks reasonable. Going to grab some more tea and try to understand the whole thing in detail now!

matklad:
One thing which seems odd to me is the usage of a single `condvar` for all cycles. It seems that we already do all the work to track who depends on what and who needs to be woken up (as we communicate results via `wait_results`), and that we just don't use that info. I wonder if that would be a good idea to say `type RuntimeId = std::thread::Thread` and use std::thread::park/unpark to wake specific threads. But, as `park` allows for spurious wakeups, it doesn't seem like that'll actually change the overall logic.

src/lib.rs
```diff
@@ -438,6 +509,11 @@ where
     {
         self.storage.purge();
     }
+
+    #[allow(dead_code)] // FIXME
```
matklad, line 513: unaddressed fixme

src/lib.rs
```diff
@@ -595,12 +642,87 @@ impl Cancelled {
...
+    pub(crate) fn new(participants: plumbing::CycleParticipants) -> Self {
+        Self { participants }
+    }
+
+    /// Iterate over the [`DatabaseKeyIndex`] for each query participating
+    /// in the cycle. The start point of this iteration within the cycle
+    /// is arbitrary but deterministic, but the ordering is otherwise determined
+    /// by the execution.
+    pub fn participant_keys(&self) -> impl Iterator<Item = DatabaseKeyIndex> + '_ {
+        let min = self.participants.iter().min().unwrap();
```
matklad, line 674: Can we rotate `participants` in `Cycle::new` instead, so that they are always just in the right order?

src/lib.rs
```diff
@@ -595,12 +642,87 @@ impl Cancelled {
...
+            .iter()
+            .chain(self.participants[..index].iter())
+            .copied()
+    }
+
+    /// Returns a vector with the debug information for
+    /// all the participants in the cycle.
+    pub fn all_participants<DB: ?Sized + Database>(&self, db: &DB) -> Vec<String> {
+        self.participant_keys()
+            .map(|d| format!("{:?}", d.debug(db)))
```
matklad, line 686: I think we have an invariant here that db's current revision matches the revision when the Cycle was created. I wonder if it makes sense to add a revision field to CycleError and a debug assert here? Or, rathre, I am fairly certain it makes sense, but I don't know whether it is worth it -- it's quitly unlikely that you'd use CycleErrors arcros revisions accidentally I would think?

src/lib.rs
```diff
@@ -595,12 +642,87 @@ impl Cancelled {
...
+    }
+
+    /// Returns a "debug" view onto this strict that can be used to print out information.
+    pub fn debug<'me, DB: ?Sized + Database>(&'me self, db: &'me DB) -> impl std::fmt::Debug + 'me {
+        struct UnexpectedCycleDebug<'me> {
+            c: &'me Cycle,
+            db: &'me dyn Database,
+        }
+
+        impl<'me> std::fmt::Debug for UnexpectedCycleDebug<'me> {
```
matklad, line 707: I'd

fmt

here to not repeat std::fmt

src/runtime.rs
```diff
@@ -280,113 +247,151 @@ impl Runtime {
...
+        debug!("create_cycle_error(database_key={:?})", database_key_index);
+
+        let mut from_stack = self.local_state.take_query_stack();
+        let from_id = self.id();
+
+        // Extract the changed_at and durability values from the top of the stack;
+        // these reflect the queries which were executed thus far and which
+        // led to the cycle.
+        let changed_at = from_stack.last().unwrap().changed_at;
+        let durability = from_stack.last().unwrap().durability;
```
matklad, line 271: Not sure this logic is correct -- here we look at the top of our stack, but what if:

* we have different values on different stacks
* we have a different value on *our* stack, which is before the last element, and still participates in a cycle

I think we should compute that as minimum across all queries in `for_each_cycle_participant`.

src/runtime.rs
```diff
@@ -280,113 +247,151 @@ impl Runtime {
...
-            .take_while(|active_query| {
-                err.cycle
-                    .iter()
-                    .any(|e| *e == active_query.database_key_index)
-            })
+    fn mutual_cycle_recovery_strategy(
+        &self,
+        db: &dyn Database,
+        cycle: &Cycle,
+    ) -> CycleRecoveryStrategy {
```
matklad, line 324: Do we envision other stategies beyond Panic/Fallback? If not, this can be written in a simpler way by just chceking if any element is a Panic.

src/runtime.rs
```diff
@@ -280,113 +247,151 @@ impl Runtime {
...
+        } else {
+            db.salsa_event(Event {
+                runtime_id: self.id(),
+                kind: EventKind::WillBlockOn {
+                    other_runtime_id: other_id,
+                    database_key,
+                },
+            });
+
+            let stack = self.local_state.take_query_stack();
```
matklad, line 364: Love this! Very neat trick to just move the stack to the shared state!

src/runtime.rs
```diff
@@ -479,36 +485,17 @@ struct ActiveQuery {
...
-    /// untracked read, this will be set to the most recent revision.
-    pub(crate) changed_at: Revision,
-
-    /// Complete set of subqueries that were accessed, or `None` if
-    /// there was an untracked read.
-    pub(crate) dependencies: Option<FxIndexSet<DatabaseKeyIndex>>,
-
-    /// The cycle if one occured while computing this value
-    pub(crate) cycle: Vec<DatabaseKeyIndex>,
+    cycle: Option<Cycle>,
```
matklad, line 488: I wonder it it'd be better to pull `Arc` out and say: `cycle: Option<Arc<Cycle>>`, as that more obviously reflects the setup where all active queries share the same cycle.

src/runtime.rs
```diff
@@ -280,113 +247,151 @@ impl Runtime {
...
+            CycleRecoveryStrategy::Panic => {}
+            CycleRecoveryStrategy::Fallback => {
+                // Mark the cycle participants, so they know to recover:
+                dg.for_each_cycle_participant(
+                    from_id,
+                    &mut from_stack,
+                    database_key_index,
+                    to_id,
+                    |aq| {
+                        aq.cycle = Some(cycle.clone());
```
matklad, line 302: Here, I'd expect us to also set `durability` and `changed_at`.

matklad: Let me see if I can write a test for this...

src/derived/slot.rs
```diff
@@ -489,210 +492,137 @@ where
...
     /// When we encounter an `InProgress` indicator, we need to either
     /// report a cycle or else register ourselves to be notified when
-    /// that work completes. This helper does that; it returns a port
-    /// where you can wait for the final value that wound up being
-    /// computed (but first drop the lock on the map).
-    fn register_with_in_progress_thread(
+    /// that work completes. This helper does that. If a cycle is detected,
+    /// it returns immediately, but otherwise it blocks `self.database_key_index`
+    /// has completed and returns the corresponding result.
+    fn block_on_in_progress_thread<MutexGuard>(
```
matklad, line 613: I wonder if `block_on_in_progress_query` would be a better name? It's a bit confusing that the `thread` might be this thread.

tests/cycles.rs
```diff
@@ -21,6 +61,36 @@ impl ParallelDatabase for DatabaseImpl {
...
+        // Default configuration:
+        //
+        //     A --> B <-- C
+        //     ^     |
+        //     +-----+
+
+        res.set_a_invokes(CycleQuery::B);
+        res.set_b_invokes(CycleQuery::A);
+        res.set_c_invokes(CycleQuery::B);
+        res
```
matklad, line 79: I'd maybe duplicate this setup in every test? When writing a new test, it took me a while to realize why I observe revisons higher than I expect.

src/runtime.rs
matklad: Yup, here's the test: https://github.com/nikomatsakis/salsa/pull/3 :) Really like that the new logic is straightforward -- I am able to just read and understand the code now, without spending quite some time to just get back into salsa.

nikomatsakis:
@matklad I'll take a look at your detailed comments in a bit, but I just pushed a few more commits. These have started to introduce some user-visible changes, so I'm working on an RFC documenting the explanations. The main gist of it is that when we encounter a recoverable cycle, we panic as soon as we have detected it. I realized that, if we don't do this, you can get in some weird situations-- for example, if you have code that creates a cycle C1 and then continues to execute and creates a second cycle C2, we might e.g. panic as a result of C2 instead of recovering from C1. That seems ungreat. (Actually, the current setup has some similarly questionable possibilities if the recover function goes and executes queries that result in a cycle, but I'm trying to decide if I care about *that*.)

I think the code is cleaner as a result anyway.

src/plumbing.rs
nikomatsakis: Actually, now that you mention it, I don't think we need this type in the final version, so I've removed it.

Cargo.toml
nikomatsakis: That does look reasonably nice! I also don't have a super strong opinion.

src/lib.rs
nikomatsakis: Bah. Thanks.

nikomatsakis: And indeed the function appears to be dead. lol.

src/lib.rs
nikomatsakis: Hmm, Database key ids don't go stale-- at least, not presently. That said, if we ever did try to garbage collect keys then they *could*... really they are basically exactly like interned values (they are, indeed, interned values).

I don't think we want to assert that the cycle is only being used in the most recent revision. I feel like people may embed cycles in the recovery value and it could be possible for those to be judged as "nothing changed".

src/lib.rs
nikomatsakis: Hmm, I tend to feel the opposite: `std` is short enough that I get annoyed when people shorten it to `fmt` instead of `std::fmt`. That said, I think this aversion dates from the days when I didn't have rust-analyzer to manage the `use` statements for me =) 

No strong opinion, really.

nikomatsakis: This particular code seems to have been removed in the final version anyway.

src/runtime.rs
nikomatsakis: Yes, I do. =) I'm working towards "fixed point" to better integrate chalk and salsa.

nikomatsakis: (that said, I do often also try to write my code "as if" more match variants will be added unless I am quite certain they won't)

src/runtime.rs
nikomatsakis: This code was removed in the latest version. That said, there is similar logic in the places where we catch cycle panics, and when you asked earlier I got to having similar doubts! We could certainly detect that information. I'd like to write a test that exposes the problem first.

src/runtime.rs
nikomatsakis: Hmm. The `Cycle` type is exposed to end-users -- the intent is that they might store it in their recovery information -- so I kind of prefer to present a simple API to them and not expose the underlying `Arc`. Thoughts?

src/runtime.rs
nikomatsakis: Nice. Let me check that out!

nikomatsakis: Is this the test I was wanting earlier? Quite possible =)

tests/cycles.rs
nikomatsakis: Yeah, good idea.

src/lib.rs
nikomatsakis: We could. Cute idea.

Cargo.toml
nikomatsakis: One nit: you should crate a `cargo expect` command or something instead of using an environment variable! :)

src/runtime.rs
nikomatsakis: OK, I've pushed a fix for this test, and we now compute the changed-at and durability across all cycle participants more cleanly. I am reasonably happy with this state, though I am not *thrilled* that we invoke the recover function for C0 twice and from within probe. I want to create some tests for what happens when the recover function itself induces a cycle. =)

nikomatsakis:
@matklad 

> One thing which seems odd to me is the usage of a single condvar for all cycles. It seems that we already do all the work to track who depends on what and who needs to be woken up (as we communicate results via wait_results), and that we just don't use that info. I wonder if that would be a good idea to say type RuntimeId = std::thread::Thread and use std:thread:park/unpark to wake specific threads. But, as park allows for spurious wakeups, it doesn't seem like that'll actually change the overall logic.

Indeed, it wouldn't be hard to use multiple condvars. My assumption was that cross-thread blocking is typically not a hot-path and so it made sense to keep the communication as simple as possible (although, it's debatable whether having spurious wakeups is simpler). I hadn't considered invoking `std::thread::park` directly. I've never actually looked closely at those functions. That might indeed be a good idea, we have a very specific idea of who is blocking. An alternative would be to allocate a `Condvar` for each blocking thread. I'm inclined to push those changes to follow-up PRs, though.

nikomatsakis:
Also, I think you are correct that we should be using thread-ids and not runtime-ids. I think I only introduced runtime-ids because I didn't want to think about thread-ids (and maybe that API wasn't available at the time).

nikomatsakis:
Pushed the first RFC draft.

book/src/rfcs/RFC0009-Cycle-recovery.md
```diff
@@ -0,0 +1,126 @@
...
+    db.do_query()
+}) {
+    Ok(v) => v,
+    Err(Cancelled::UnexpectedCycle(cycle)) => {
+        eprintln("cycle results: {:#?}", cycle.debug(db))
+    }
+    Err(_) => {
+        /* ... */        
+    }
+}
```
matklad, line 37: Need to read the proposal in more detail later, but this approach of treating unexpected cycle as a variant of cancelled seems "obviously wrong" to me, so I wonder if I might be missing something :) 

The old `Cancelled` panic wasn't an error -- it was serendipitous success. The new `Cancelled::UnexpectedCycle` seems like it signals an error, specifically, a programmers error, a bug in a program. That is, it seems that we are now mixing two very distinct cases in the same `Cancelled::catch` API:

* recoverable (even successful) "normal operation"
* erroneous "bug in the program", which is unrecoverable 

I would expect that, for unexpected cycles, we'd just `panic!` with a message printing out the cycle, or, if we care a lot about giving access to `Cycle` for debugging purposes, with throwing the `Cycle` itself.

nikomatsakis: Yeah, I might have chosen wrong here. I debated about precisely this question and I was thinking that, in the future, I expect to move salsa so that it manages a pool of helper threads, and that (for those threads) irrecoverable cycles would want to be caught and forwarded, so it was kind of "akin" to cancellation, but really I agree with you that they are distinct things. I can change the code to just throw the `Cycle`, as suggested.

That said, this is not the "interesting" part of the change -- we used to panic with a string, we now panic with a useful value. The more interesting part is how we treat *recoverable* cycles. There we unwind with `CycleParticipant` and catch it internally.

nikomatsakis: The latest commit throws `Cycle` values directly.

nikomatsakis:
Update: wrote a bunch more docs.

src/runtime/cycle_participant.rs
```diff
@@ -0,0 +1,26 @@
...
+    pub(crate) fn new(cycle: Cycle) -> Self {
+        Self { cycle }
+    }
+
+    pub(crate) fn throw(self) -> ! {
+        std::panic::resume_unwind(Box::new(self));
+    }
+
+    pub(crate) fn recover<T>(execute: impl FnOnce() -> T, recover: impl FnOnce(Cycle) -> T) -> T {
+        std::panic::catch_unwind(AssertUnwindSafe(execute)).unwrap_or_else(|err| {
```
matklad, line 19: Can we, instead of asserting unwind-safeness, just require it by, eg, adding some trait bounds to db?

matklad:
Ok, I think I know understand the big bit I was missing before: why we even need panicking here. Basically the situation is like this. We have the following user-derfined query function: 

```rust
fn q(db: &Db) {
    db.q1();
    db.q2();
    db.q3();
}
```

and in the process of executing the `db.q2()` line we found a cycle. Now, the problem is that `q` is just usual rust code, there's q activation frame on the stack, and we are in the middle of q's body. At the moment we detect the cycle, the code *after* `db.q2();` becomes irrelevant -- we want to abort the execution of `q2` entirely. The only way we can do this is via a panic, because the usual Rust code in q is written without anticipation of cancellation.

matklad:
:thinking: one thing I am not sure about is that we require *all* participants to specify fallback. Intuitively, it seems that in a lot of cases you have only some queries which are fundamental to the cycle, and which should specify recovery.

For example, let's say we have the following cycle: `a -> b -> a` where both a and b specify refactor. Intuitively, I think it should be possible to insert an identity query in-between them: `a -> id1 -> b -> id2 -> a` such that `id1&2` just forward to a or b, and are not themselves annotated with `#[fallback]`. 

A specific example I have in mind is resolving type aliases. Let's consider a subset of Rust which has only unit types and type aliases: 

```rust
type A = ();
type B = A;
type C1 = C2;
type C2 = C3;
type C3 = C1;
```

I would like to model it like this in salsa (pseudo code):

```rust
enum AstType {
    Unit,
    Alias(String),
}

enum IrType {
    Unit,
    Err(&'static str),
}

#[input]
fn definition(name: String) -> Option<AstType> {}

#[derived]
fn type_of(name: String) -> IrType {
    match definition(name) {
        Some(ast_type) => match ast_type {
            AstType::Unit => IrType::Unit,
            AstType::Alias(name) => resolve_alias(name),
        },
        None => return IrType::Err("undefined name"),
    }
}

#[derived]
#[cycle_fallback(resolve_alias_fallback)]
fn resolve_alias(name: String) -> IrType {
    type_of(name)
}
fn resolve_alias_fallback() -> IrType {
    IrType::Err("cyclic aliases")
}
```

That is, I don't want the "ast case analysis" to care about cycles, but only the specific "resolve type alias" bit. This is a minimal example which I think shows the issue, but it is useful to extend it to the language which supports tuples as well: 

```rust
type A = ();
type B = A;
type C = ((A, C), A);
```

This I'd like to model as

```rust
enum AstType {
    Unit,
    Alias(String),
    Tuple(Vec<AstType>),
}

enum IrType {
    Unit,
    Tuple(Vec<IrType>),
    Err(&'static str),
}

#[input]
fn definition(name: String) -> Option<AstType> {}

#[query]
fn type_of(name: String) -> IrType {
    match definition(name) {
        Some(ast_type) => lower(ast_type),
        None => return IrType::Err("undefined name"),
    }
}
// NB: not a query
fn lower(ast_type: AstType) -> IrType {
    match ast_type {
        AstType::Unit => IrType::Unit,
        AstType::Alias(name) => resolve_alias(name),
        AstType::Tuple(ts) => IrType::Tuple(ts.into_iter().map(lower).collect()),
    }
}

#[fallback(resolve_alias_fallback)]
fn resolve_alias(name: String) -> IrType {
    type_of(name)
}
fn resolve_alias_fallback() -> IrType {
    IrType::Err("cyclic aliases")
}
```

That is, I want `type_of("C")` to be `(((), !), ())`. So, I don't actually want to unwind from a recursive call of `lower` when I notice there's a `Cycle`, I *do* want to continue comutation with recovery value as usual.

src/runtime/cycle_participant.rs
nikomatsakis: Yeah, thanks for calling that out, I wasn't happy with that part of the PR. I think we probably *do* want `salsa::Database` to just require `UnwindSafe`, and that would suffice. It seems like we're pretty "all in" on unwinding so...

nikomatsakis:
@matklad I just pushed a refactoring and a fix for a bug that I found while working on a follow-up branch; will read your comments in more detail and reply this morning I think. :)

nikomatsakis:
@matklad Nice examples. Let me talk this through a bit in comment form.

To start, why did I want all participants to specify fallback... I am concerned about non-deterministic results that depend on which query you invoke first. 

For example, suppose we have `A -> B -> C -> A` and only `A` specifies fallback. Well, it's easy enough for me to handle that case: on the edge `C -> A`, we give `A` the fallback value and let `B` and `C` continue. Once they complete, then `A` can (internally) unwind and supply the fallback value.

But what happens if the user calls `C` first? Then we have `C -> A -> B -> C`. Now we have a bit of a challenge, because we have the `B -> C` edge to manage, and we don't have a fallback value to supply for `C`.

I think this problem only arises when you have *two* non-fallback queries in a row. It wouldn't arise for your example. 

I guess what we could do, conceivably, is to unwind both B and C and just have A recover. A would then supply the fallback and C would complete as normal. 

I am now wondering, what happens if a query D is blocked on B? But that doesn't seem like a problem -- the new "retry" handling means it would just retry and it would attempt to execute B. B would invoke C, which has a cached result.

OK, I'm convinced. I think I can implemented this this morning, even.

Interestingly, this whole thing is kind of a "special case" of the "fixed point" cycle handling that I want for chalk. The way that it would work *there* is that you would have an initial value to use for cycles. Once execution completes, you get some value V for each participant, and you run again. This time if a cycle results, you use the value V. You now get a value V1. If V1 == V, you can stop. (Actually, if V == the dummy value, you could stop). This way, you wind up with a set of values that are "consistent". 

We are basically doing the same thing here, where fallback is supply that initial value (and that value never changes), except that we don't re-run B, we instead wait for some other query to come along and re-run it for us. Interesting!

matklad:
>Nice examples.

Looking at them, I am wondering if we perhaps want to have an extended "minirust" example somewhere? Or rather, it definitely would be nice to have one, but it's unclear if that's the best time investment. I wonder if this an E-good-first-issue even?

nikomatsakis:
Ok, I don't think this causes a problem, but I wanted to record another interesting case so I can add it as a test. What if there are multiple cycles?

Example:

A will invoke B and then C. Both B and C invoke A and return the identity value. But both A *and* C have recovery.

```rust
#[derived] #[fallback = Err(())]
fn a() -> Resuilt<()> { b(); c(); Ok(()) }

#[derived]
fn b() -> Resuilt<()>  { a(); Ok(()) }

#[derived] #[fallback = Err()]
fn c() -> Resuilt<()>  { a(); Ok(()) }
```

So what happens here?

* A invokes B
    * B invokes A: cycle! A is flagged as a participant. B is unwound.
    * A is unwound and stores the fallback value of `Err`.
* Later, C is invoked.
    * C invokes A, it reads `Err` (fallback value) but it does not consider itself a participant in the cycle, so it returns `Ok`.

This last part was worrying me, but I think it's fine. If you invoked in another order:

* C invokes A
* A invokes B
* B invokes A: cycle! A is flagged as a participant, B is unwound.
* C continues as normal and returns `Ok`

The key bit is (again) that `A` stops executing immediately and returns. Otherwise you would get into a weird situation where `A` invokes `C` while it is still on the stack, but that value is later discarded.

nikomatsakis:
> > Nice examples.
> 
> Looking at them, I am wondering if we perhaps want to have an extended "minirust" example somewhere? Or rather, it definitely would be nice to have one, but it's unclear if that's the best time investment. I wonder if this an E-good-first-issue even?

Sounds reasonable, but it may make sense to wait before investing a lot of time in it. At least I have some reasonably big changes to propose for Salsa 2022 so I'd like to talk those over on Friday :)

nikomatsakis:
OK, I see a complication, though not (I think) an unsolvable one. The fixed point query example provides the clue.

You can have a situation where you have two threads:

* Thread X has some queries C0 -> C1
* Thread Y has some queries C2 -> C3
* C1 from thread X invokes C2, and so X blocks on Y
* C3 from thread Y invokes C0, and so we detect a cycle

What happens if C0 and C1 have recovery but C2 and C3 don't?

Under the current code I've implemented, C2 will panic and thread Y would have to catch that panic and retry. But that retry will be successful, because (this time) C0 has a recovery value installed. Hmm, that's not great.

Probably the *best* result is that C2 retries automatically. I think we only want that in the particular where the first query in some particular thread has does not have recovery set, but we could do it for other queries too I suppose. I have to think this one through, but it's time for my morning run!

This is making me wonder if this wouldn't be a nice follow-up to this particular PR, I could amend the RFC to indicate that it is a "WIP".

nikomatsakis:
Thought of a nice approach on my run :)

nikomatsakis:
@matklad ok, the code now supports cycle fallback so long as any one query has it. I also added tests for all the wicked situations I could come up with. The main thing that's not done is updating the RFC and Salsa book to describe how the algorithm works.

nikomatsakis:
Although in the process of writing those comments I already found one bug (the fix for which is already part of these commits), so I wouldn't be surprised if doing more writing helped to flush out more bugs.

nikomatsakis:
OK, the RFC is more up to date now.

Cargo.toml
matklad: Heh, not *having* `cargo expect` is one of the design goals for `expect-test`. I don't like cargo subcommands because:

* they require extra installation steps, you need to teach a human to run `cargo install foo` instead of things just working
* they are not hermetic -- the version of command you use isn't pinned by the Cargo.lockfile, so it can drift over time. 

So that's why expect-test is strictly only a library, but that necessitates somewhat unusual env-var workflow.

book/src/plumbing/cycles.md
```diff
@@ -0,0 +1,65 @@
...
+* **Handle the current thread:** Finally, we have to choose how to have the current thread proceed. If the current thread includes any cycle with recovery information, then we can begin unwinding. Otherwise, the current thread simply continues as if there had been no cycle, and so the cyclic edge is added to the graph and the current thread blocks. This is possible because some other thread had recovery information and therefore has been awoken.
+
+Let's walk through the process with a few examples.
+
+### Example 1: Recovery on the detecting thread
+
+Consider the case where only the query QA2 has recovery set. It and QA3 will be marked as cycle participants. No threads will be unblocked, as they do not have any cycle recovery nodes. The current thread will unwind with the cycle. This will unwind through QA3 and be caught by QA2. QA2 will substitute the recovery value and return normally. QA1 and QC3 will then complete normally and so forth, on up until all queries have completed.
```
matklad, line 53: I think you are overloading "cycle participant" terminology. It seems that some paragraphs use it to mean "any query participating in the cycle", and some paragraphs (including this one) use it to mean "subset of all nodes in the cycle for which we set `aq.cycle = Some(cycle.clone());`".

src/runtime.rs
```diff
@@ -280,113 +263,205 @@ impl Runtime {
...
+        // dependencies to things outside the SCC that control whether it will
+        // form again.
+        cycle_query.remove_cycle_participants(&cycle);
+
+        // Mark the cycle participants so they know to recover.
+        // This only matters for queries that have a fallback value specified;
+        // the others will just unwind without storing any recovery information.
+        dg.for_each_cycle_participant(from_id, &mut from_stack, database_key_index, to_id, |aqs| {
+            aqs.iter_mut()
+                .skip_while(
```
matklad, line 345: The comment looks a bit misleading -- it implies that we should `.filter` the queries, and not `skp_while`.

src/derived/slot.rs
```diff
@@ -489,210 +468,131 @@ where
...
+            // Note that we don't need the actual value to be available.
+            ProbeState::NoValue(_, changed_at)
+            | ProbeState::UpToDate(StampedValue {
+                value: _,
+                durability: _,
+                changed_at,
+            }) => MaybeChangedSinceProbeState::ChangedAt(changed_at),
+
+            // If we have nothing cached, then value may have changed.
+            ProbeState::NotComputed(_) => MaybeChangedSinceProbeState::ChangedAt(revision_now),
```
matklad, line 513: Hm, this seems questionable to me. If we return `revision_now` here, than in `maybe_changed_since` we actually return `false` from `changed_at > revision` condition. That is, we treat the input as if it didn't change. 

It seems that this can actually lead to the bugs, when the query state flips from some state to non-computed, but triggering that should be tricky.

src/runtime/dependency_graph.rs
```diff
@@ -0,0 +1,280 @@
...
+                .take_while(|p| p.database_key_index != key)
+                .count();
+            let next_id = edge.blocked_on_id;
+            let next_key = edge.blocked_on_key;
+
+            if let Some(cycle) = edge.stack[prefix..]
+                .iter()
+                .rev()
+                .filter_map(|aq| aq.cycle.clone())
+                .next()
```
matklad, line 158:
```suggestion
                .find_map(|aq| aq.cycle.clone())
```

src/runtime/dependency_graph.rs
```diff
@@ -0,0 +1,280 @@
...
+        // Finally, we copy in the results from `from_stack`.
+        let prefix = from_stack
+            .iter_mut()
+            .take_while(|p| p.database_key_index != key)
+            .count();
+        closure(&mut from_stack[prefix..]);
+    }
+
+    /// For each runtime that is blocked as part of this cycle excluding the current one,
+    /// execute `should_unblock` with its portion of the stack. If that function returns true,
```
matklad, line 125: I guess there's no more `should_unblock`?

src/derived/slot.rs
nikomatsakis: Well, it's not really valid to call `maybe_change_since` with revision now. The meaning of "maybe changed since R" is "did this change in revision (R+1)..now", but it'd be good to document that. The reason for this is that we invoke `maybe_changed_since(Rv)` where `Rv` is the "revision in which the data was last verified". What we really want to know, then, is did something change *after* that revision (if it changed at the *start* of Rv, we've already taken that into account).

src/runtime/dependency_graph.rs
nikomatsakis: Oh, yes, I removed that.

book/src/plumbing/cycles.md
nikomatsakis: I'll revisit, probably true.

src/derived/slot.rs
nikomatsakis: Renamed to `maybe_changed_after` and documented this more clearly. Also added various debug assertions.

nikomatsakis:
@matklad I believe I have addressed all of your review comments.

nikomatsakis:
I discussed this with @matklad and we agreed to merge this PR.

bors r+

PR #286, opened 2021-12-05
Make `query_group` attribute helpers actual attribute items

Veykril:
This is a small exploration in proc-macro space to see if we can make attribute helper macros(not the derive kind) more visible for users as well as more IDE friendly. As things are currently, these are basically just syntax the attribute consumes and as such they do not really exist anywhere.

This PR tries to change that by introducing these helpers as actualy attribute items. They error our if actually invoked as their main use is to be consumed by the actual `query_group` macro instead.
Making these proper items gives us a few nice properties, for one we can document them and have rustdoc show them as well, OTOH this could also be considered problematic as this pollutes the attribute listing of the crate, or rather the attribute(macro) namespace in general.

Note the documentation is rather spotty, partially just copied from the main doc of the `query_group` attribute, as I do not have too much knowledge about the purpose of these helper attributes myself.

IDEs can also benefit from this, as prior to this, the attributes actually never resolved to anything that existed, IDEs couldn't do anything with them and instead treated them as mere tokens sometimes resolving them to potentially incorrect things. Now we can actually resolve them properly, giving us hover messages
image
as well as autocompletion
image
and basically anything else an IDE might offer.

### DIFF
--- a/components/salsa-macros/src/lib.rs
+++ b/components/salsa-macros/src/lib.rs
@@ -146,3 +146,81 @@ pub fn query_group(args: TokenStream, input: TokenStream) -> TokenStream {
 pub fn database(args: TokenStream, input: TokenStream) -> TokenStream {
     database_storage::database(args, input)
 }
+
+/// Helper attribute for [`macro@query_group`] setting the storage
+/// to memoized for this query. This is the default storage.
+///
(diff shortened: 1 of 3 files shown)


Veykril:
Forgot about this PR actually, will close it as I never really intended to land any of this properly as it was more of an experiment.

PR #287, opened 2021-12-26
Remove unused crossbeam_utils dependency

DSPOM2:
crossbeam_utils is currently not used anywhere withing this crate so I propose to remove the dependency to (slightly) reduce compile times


nikomatsakis:
Thanks

PR #301, opened 2022-05-12
Remove RefCell and make database Sync

### LINKED ISSUE #300: RefCell makes !Sync
Thank you for a great library. I want to use salsa in a `Sync` context, but I cannot do so because of `RefCell`.

I could to send a PR to replace it with `Rwlock` or something.

ryo33:
Closes #300 

I just replaced `RefCell` with `RwLock`. Do we need a `sync` feature that uses `RwLock` in order to keep using `RefCell` in default?

### DIFF
--- a/src/runtime/local_state.rs
+++ b/src/runtime/local_state.rs
@@ -5,7 +5,7 @@ use crate::runtime::ActiveQuery;
 use crate::runtime::Revision;
 use crate::Cycle;
 use crate::DatabaseKeyIndex;
-use std::cell::RefCell;
+use parking_lot::RwLock;
 use std::sync::Arc;
 
 use super::StampedValue;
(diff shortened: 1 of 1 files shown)


matklad:
@ryo33 local state intentionally uses RefCell, because it is, by design, restricted to a single thread. This does not preclude using salsa in multi-threaded scenarious though! There's snapshot method which allows you to create sendable handles to the database: 

https://docs.rs/salsa/latest/salsa/trait.ParallelDatabase.html#tymethod.snapshot

Here's an example of parallel processing from rust analyzer: 

https://github.com/rust-lang/rust-analyzer/blob/1b120216de987f2e7700b7707ef6b5d5b5545d94/crates/ide/src/prime_caches.rs#L80-L84

ryo33:
My use case is a little bit weird, I use salsa with Bevy! I want to store many salsa databases as components or resources of Bevy ECS and want to use them from Bevy systems but components and resources require `Sync` (because systems can run on any thread).

Is it restricted to a single thread even if `RwLock` is used?

PR #304, opened 2022-08-01
entity-based salsa preview

nikomatsakis:
This PR is a squashed and updated version of my entity branch. This API is, I believe, what will become the long awaited (by me, anyway) Salsa 1.0 release. It's a complete overhaul of the Salsa API and implementation that feels a lot cleaner to me.

The PR includes a tutorial and sample application (a compiler) that covers the major concepts. There is still more work needed to complete the tutorial and in particular to write up a bunch of tests (I had more tests but they were hacky so I deleted them). 

Right now, all the new stuff is in a pair of new crates, salsa-entity-mock and salsa-entity-macros, leaving the existing salsa intact. I would like to remove the old salsa but we have to reconcile the two and port over the tests.

There is still a fair bit of work to do before we can release this new version of Salsa, but I expect a lot of it can happen after this PR is merged. The list is on #305.


nikomatsakis:
bors r+

I'm going to go ahead and land this. We can iterate on master.

Issue #305, opened 2022-08-01
Salsa 2022


nikomatsakis:
I'm shooting to release a new take on Salsa this year. I'm calling it Salsa 2022 to give a deadline, but I think it will be the first 1.0 release once it's ready. There's a preview in #304. This is the meta tracking issue for the Salsa 2022 effort.

ppaulweber:
@nikomatsakis is there any update on the status of the proposed `salsa v1` release?

PR #308, opened 2022-08-02
Tracked proposal

nikomatsakis:
I sketched out a variation of the entity API that I am calling tracked (this is a hat-tip to Ember's terminology, cc @wycats). It is the mostly the same ideas, but repackaged in a way that I think is more intuitive. This was inspired by writing the previous tutorial and thinking how things could be a bit cleaner.

To read the proposal, check out the "overview" page on the netlify preview. The tutorial and code have not been fully updated to match yet.

cc #305


8573:
> To read the proposal, check out the "overview" page on the netlify preview.

...which is [here](https://deploy-preview-308--salsa-rs.netlify.app/overview.html), though it disappeared from the Netlify bot's comment.

nikomatsakis:
OK!  This is looking good.

@bors r+

matklad:
bors=nikomatsakis

I think we use the other one here

matklad:
bors r=nikomatsakis

I think we use the other one here

Issue #328, opened 2022-08-06
Test expected reuse and the like


nikomatsakis:
We have one very test that checks that we are getting the re-use we expect:

https://github.com/salsa-rs/salsa/blob/03a27a7054ee234a566231243dd5025edfbcdde6/salsa-2022-tests/tests/hello_world.rs#L69-L77

We need more tests! For example, we should check that 

* [ ] if field X of an entity changes but not field Y, functions that depend on X re-execute, but those depending only on Y do not

nikomatsakis:
I'm tagging this as good-first-issue, since it's an easy way to get involved with salsa. Here are some instructions for what to do:

* Create a test file in `salsa-2022-tests/tests` with a suitable name like `expect-reuse-field-X-changes-but-fn-depends-on-field-Y.rs`.
* You can copy the basic setup from the [hello world](https://github.com/salsa-rs/salsa/blob/master/salsa-2022-tests/tests/hello_world.rs) test -- in particular, you will want to use the `push_log` and `assert_logs` mechanism to test which tracked functions actually executed.
* Open a PR with the new test!

Another great way to contribute is to come up with alternate testing scenarios (and, hopefully, implementing them).

karpa4o4:
Hi, @nikomatsakis. Perhaps more tests are needed? I could take care of them.

nikomatsakis:
@karpa4o4 Yes, I think we could probably use more tests, though tbh I don't know exactly *what* needs testing.

nikomatsakis:
No longer relevant.

PR #334, opened 2022-08-07
test: expect reuse field X changes but fn depends on field Y

XFFXFF:
ref: #328


nikomatsakis (approved):

MihailMihov:
There seems to be a typo in the filenames of both tests. deponds should be depends.

XFFXFF:
> There seems to be a typo in the filenames of both tests. deponds should be depends. Edit: The word depends is misspelled in the files too.

@MihailMihov oh my fault, thanks! I fixed them in https://github.com/salsa-rs/salsa/pull/363

PR #336, opened 2022-08-09
Add options to tracked functions for cycle recovery

### LINKED ISSUE #331: Add options to tracked functions for cycle recovery
Salsa allows tracked functions to declare a "cycle recovery" function, [as described here](https://salsa-rs.github.io/salsa/cycles/fallback.html) (actually, that page is somewhat out of date, we no longer require all cycle participants to define a recovery function, for example, should fix that --ed.). This capability is not exposed yet for tracked functions. 

We should support users writing something like

```rust
#[salsa::tracked(recovery_fn = my_recover_fn)]
fn my_tracked_fn(db: &dyn crate::Db, ...) { }

fn my_recover_fn(
    db: &dyn MyDatabase,
    cycle: &salsa::Cycle,
) -> MyResultValue
```

XFFXFF:
closes #331 

This pr ports the old salsa tests for cycle in a single thread, except for [cycle_disappears_durability](https://github.com/salsa-rs/salsa/blob/03a27a7054ee234a566231243dd5025edfbcdde6/tests/cycles.rs#L326), since we don't have the api that permits setting durability.  

~I haven't ported parallel related tests, which would be some work, wondering if we can merge this in first~

### DIFF
--- a/salsa-2022-tests/tests/cycles.rs
+++ b/salsa-2022-tests/tests/cycles.rs
@@ -1,8 +1,68 @@
-
 #![allow(warnings)]
 
+use std::panic::{RefUnwindSafe, UnwindSafe};
+
+use expect_test::expect;
+use salsa::storage::HasJarsDyn;
+
+// Axes:
+//
+// Threading
+// * Intra-thread
+// * Cross-thread -- part of cycle is on one thread, part on another
+//
+// Recovery strategies:
+// * Panic
+// * Fallback
+// * Mixed -- multiple strategies within cycle participants
+//
+// Across revisions:
+// * N/A -- only one revision
+// * Present in new revision, not old
+// * Present in old revision, not new
+// * Present in both revisions
+//
+// Dependencies
+// * Tracked
+// * Untracked -- cycle participant(s) contain untracked reads
+//
+// Layers
+// * Direct -- cycle participant is directly invoked from test
+// * Indirect -- invoked a query that invokes the cycle
+//
+//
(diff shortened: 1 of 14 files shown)


salsa-2022-tests/tests/cycles.rs
```diff
@@ -0,0 +1,425 @@
...
+    panic!("unexpected value: {:?}", v)
+}
+
+#[test]
+fn cycle_memoized() {
+    let mut db = Database::default();
+    let input = MyInput::new(&mut db);
+    let cycle = extract_cycle(|| memoized_a(&db, input));
+    let expected = expect![[r#"
+        [
```
XFFXFF, line 190: We haven't implemented `debug` for [`DependencyKey`](https://github.com/salsa-rs/salsa/blob/03a27a7054ee234a566231243dd5025edfbcdde6/components/salsa-2022/src/key.rs#L37)

nikomatsakis:
> I haven't ported parallel related tests, which would be some work, wondering if we can merge this in first

yes, let's do that, but let's also create issues for the remaining to-do items...

XFFXFF:
@nikomatsakis  I ported all old salsa tests for cycle recovery, except for [cycle_disappears_durability](https://github.com/salsa-rs/salsa/blob/03a27a7054ee234a566231243dd5025edfbcdde6/tests/cycles.rs#L326), I added a [issue](#339)  for this.  
I also added #340 for more human readble output from  `DependencyIndex.debug`

nikomatsakis (approved):
This is awesome, thanks!

nikomatsakis:
Egads.

bors r+

PR #342, opened 2022-08-11
Make input setters return old value

MihailMihov:
Added an `InputFieldIngredient` which replaces the `FunctionIngredient` for inputs. The end goal is for the inputs to use an ingredient simpler than `FunctionIngredient` and that would allow us to return the old values when updating an input.


components/salsa-2022/src/input_field.rs
```diff
@@ -0,0 +1,67 @@
+use std::hash::Hash;
+use rustc_hash::FxHashMap;
+use crate::{Durability, IngredientIndex, Revision, Runtime};
+use crate::cycle::CycleRecoveryStrategy;
+use crate::ingredient::Ingredient;
+use crate::key::DependencyIndex;
+use crate::runtime::local_state::QueryInputs;
+use crate::runtime::StampedValue;
+
+pub struct InputFieldIngredient<K, F> {
```
nikomatsakis, line 10: Let's add a comment here, e.g.

```suggestion
/// Ingredient used to represent the fields of a `#[salsa::input]`.
/// These fields can only be mutated by an explicit call to a setter
/// with an `&mut` reference to the database,
/// and therefore cannot be mutated during a tracked function or in parallel.
/// This makes the implementation considerably simpler.
pub struct InputFieldIngredient<K, F> {
```

components/salsa-2022/src/input_field.rs
```diff
@@ -0,0 +1,67 @@
...
+        } else {
+            None
+        }
+    }
+
+    pub fn fetch(
+        &self,
+        key: K,
+    ) -> &F {
+        &self.map.get(&key).unwrap().value
```
nikomatsakis, line 50: You also need to record the read here!

You need a call like this one:

https://github.com/salsa-rs/salsa/blob/8b7324dca886001df7e7bf7c45567520954fec6b/components/salsa-2022/src/function/fetch.rs#L26-L30

This will tell the active query (if any) that the function read this input.

The `database_key_index` function will also have to be added (it will read the `index` field), and should look something like this:

https://github.com/salsa-rs/salsa/blob/8b7324dca886001df7e7bf7c45567520954fec6b/components/salsa-2022/src/function.rs#L145-L150

components/salsa-2022/src/input_field.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+impl<DB: ?Sized, K, F> Ingredient<DB> for InputFieldIngredient<K, F>
+{
+    fn cycle_recovery_strategy(&self) -> CycleRecoveryStrategy {
+        CycleRecoveryStrategy::Panic
+    }
+
+    fn maybe_changed_after(&self, _db: &DB, _input: DependencyIndex, _revision: Revision) -> bool {
```
nikomatsakis, line 60: This needs to 

First, convert `input` into a `K` by doing something like this:

https://github.com/salsa-rs/salsa/blob/8b7324dca886001df7e7bf7c45567520954fec6b/components/salsa-2022/src/function.rs#L193

Then, lookup the `StampedValue` in `self.map` and compare `_revision` against the `changed_at` field, something like this:

```rust
self.map[&input].changed_at > revision
```

similar to this:

https://github.com/salsa-rs/salsa/blob/master/components/salsa-2022/src/function/maybe_changed_after.rs#L40

except that the `memo` is just the `StampedValue` we read from `self.map`

nikomatsakis (changes requested):
Nice! The suggestions here should help fix the tests that I presume are broken :)

components/salsa-2022/src/input_field.rs
MihailMihov: https://github.com/salsa-rs/salsa/pull/342/commits/9024a7bb880e1b469e07f39bfc06574ee3f99cd6 should fix this. I only passed the runtime instead of the db, that shouldn't be a problem right?

components/salsa-2022/src/input_field.rs
MihailMihov: In https://github.com/salsa-rs/salsa/pull/342/commits/740fe86fe98c4ab8cf4bdfce752d7b70b0ea191e I implemented something similar to what `FunctionIngredient` has, but without adding the `key_from_id` helper function.

MihailMihov:
Addressing these suggestions fixes the tests, except the `specify_tracked_fn_in_rev_1_but_not_2` fails. The second `WillCheckCancellation` event is missing after the call to `create_tracked`

nikomatsakis:
@MihailMihov I think that test will be removed by #343 =)

components/salsa-2022/src/input_field.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn maybe_changed_after(&self, _db: &DB, input: DependencyIndex, revision: Revision) -> bool {
+        let key: K = AsId::from_id(input.key_index.unwrap());
+        self.map.get(&key).unwrap().changed_at > revision
+    }
+
+    fn inputs(&self, _key_index: Id) -> Option<QueryEdges> {
+        None
+    }
+
+    fn remove_stale_output(&self, _executor: DatabaseKeyIndex, _stale_output_key: Option<Id>) {
```
nikomatsakis, line 93: Once #343 lands, you'll be able to remove this method

nikomatsakis (changes requested):
Can we add a test where we do something like

```rust
let mut my_string = some_input.set_field(&mut db, String::new());
my_string.push_str("Hi");
some_input.set_field(&mut db, my_string);
```

just to test that it works?

components/salsa-2022/src/input_field.rs
MihailMihov: Added a test called `compare_and_swap`, which a name that you used in the initial issue, but I'm not sure if it applies here. I also added `assert_eq`'s checking if it actually returns the expected values. Is there anything else that would be good to have in the test?

components/salsa-2022/src/input_field.rs
```diff
@@ -0,0 +1,96 @@
...
+impl<DB: ?Sized, K, F> Ingredient<DB> for InputFieldIngredient<K, F>
+where
+    K: AsId,
+{
+    fn cycle_recovery_strategy(&self) -> CycleRecoveryStrategy {
+        CycleRecoveryStrategy::Panic
+    }
+
+    fn maybe_changed_after(&self, _db: &DB, input: DependencyIndex, revision: Revision) -> bool {
+        let key: K = AsId::from_id(input.key_index.unwrap());
```
nikomatsakis, line 85: Total bikeshed nit, but I generally prefer this way of writing it:

```suggestion
        let key = K::from_id(input.key_index.unwrap());
```

salsa-2022-tests/tests/compare_and_swap.rs
```diff
@@ -0,0 +1,57 @@
+//! Test that a setting a field on a `#[salsa::input]`
+//! overwrites and returns the old value.
+
+use salsa_2022_tests::{HasLogger, Logger};
```
nikomatsakis, line 4: Yeah, let's rename this, you're right that the name isn't right. Perhaps "swap-and-replace" or "mutate-in-place"?

nikomatsakis (approved):
This looks great! I'd r+ right now but I'd prefer to rename the test.

nikomatsakis (approved):

Issue #315, opened 2022-08-04
Implement garbage collection: when a tracked struct is not created in new round, delete


nikomatsakis:
When a tracked function executes, we also track what tracked structs it creates. When it re-executes, we can delete any tracked structs that were not created the second time, and we can also go and delete the values of their fields -- as well as tracked functions defined against them (but that is a bit more complex and should be moved into a new section). This will help keep memory use under control.

8573:
In terms of balancing saving memory vs saving recomputation, how does such garbage collection compare to keeping the structs in a cache of fixed capacity?

> When it re-executes, we can delete any tracked structs that were not created the second time

I don't know enough to say, but is this implying that structs not computed the second time should not be needed again? What if Salsa is powering an LSP server the user of which deletes a large chunk of code, rewrites that code in a different way, and then changes their mind and "Undo"es back to the original code?

I apologize in advance if my uninformed questions are too uninformed or irrelevant.

jonas-schievink:
Undoing edits to arrive at an earlier state already requires computing everything that changed again with current-day salsa and is unlikely to change (the amount of recomputation is typically very small).

nikomatsakis:
I'm working on this -- as part of fixing #338 and #313 I'm realizing that I likely have to add all the infra we need here.

Issue #313, opened 2022-08-04
Clear accumulators when re-executing functions


nikomatsakis:
When you *push* onto an accumulated for the first time, it creates an empty vector:

https://github.com/salsa-rs/salsa/blob/6bd6f6ee0b0243b8ab9f87e02bd172e3bc5b1103/components/salsa-entity-mock/src/accumulator.rs#L39-L43

But that vector never gets cleared out!  Later it gets read here:

https://github.com/salsa-rs/salsa/blob/6bd6f6ee0b0243b8ab9f87e02bd172e3bc5b1103/components/salsa-entity-mock/src/accumulator.rs#L51-L67

We need to clear the vector when we are in a new revision and we see a `push`, and we need to delete the vector when we see a read but the function has re-executed without pushing.

nikomatsakis:
I am working on this

PR #347, opened 2022-08-13
Track outputs and specified fields

### LINKED ISSUE #338: Panics when functions read fields of tracked structs they created
If a function creates a tracked struct, reads its fields, and then the inputs change but the function doesn't have to re-execute, you get a panic. We don't have the dependencies for tracked struct fields quite right.

nikomatsakis:
This PR implements a new strategy for tracking output edges and specified fields correctly. Core ideas:

* Track the *outputs* of each query Q and not only its inputs. An output is a specified or assigned field, a created entity, or a pushed accumulator.
* When marking a query Q as validated (no need to execute) in a new revision, you mark ALL of its outputs as validated.

Fixes #338.

The PR also includes some commits that lay the groundwork for #313 and #315 (but I left those for a future PR). In particular, it includes the logic to diff outputs to find values that are no longer output by Q. This will be used to delete obsolete entities. (Originally, I was going to use this to clear outdated outputs of all kinds, but I realized that marking them as validated was a better approach.)


nikomatsakis:
I do not understand why this branch is marked as having merge conflicts. grrrrr.

PR #352, opened 2022-08-18
Add options to tracked funcitons for lru capacity 

### LINKED ISSUE #344: Add options to tracked funcitons for lru capacity
The [set_lru_capacity](https://docs.rs/salsa/0.16.1/salsa/struct.QueryTableMut.html#method.set_lru_capacity) method, in old salsa, can be used to fix the maximum capacity for a query. In salsa-2022, we already have a method [`set_capacity`](https://github.com/salsa-rs/salsa/blob/367c0bfe1154d8848600bf4029343882652a358c/components/salsa-2022/src/function.rs#L151) to do the same thing, but without exposing the api to users.  

We could support users writing somthing like

```rust
#[salsa::tracked(lru_capacity=32)]
fn my_tracked_fn(db: &dyn crate::Db, ...) { }
```  

cc @nikomatsakis

XFFXFF:
fixes #344 

Now we can write something like the following to set the lru capacity of tracked functions  
```rust
#[salsa::tracked(lru=32)]
fn my_tracked_fn(db: &dyn crate::Db, ...) { }
```

some details:  
* lru should not be combined with specify. We will report an error if people do #[salsa::tracked(lru = 32, specify)]
* set 0 as default capacity to disable LRU (Because I think doing this would make the code simpler when implementing `create_ingredients` of tracked functions).
* old salsa support to change lru capacity at runtime, [as noted here](https://salsa-rs.github.io/salsa/rfcs/RFC0004-LRU.html?highlight=change#reference-guide), but we do not support this now

### DIFF
--- a/components/salsa-2022-macros/src/accumulator.rs
+++ b/components/salsa-2022-macros/src/accumulator.rs
@@ -32,6 +32,8 @@ impl crate::options::AllowedOptions for Accumulator {
     const DB: bool = false;
 
     const RECOVERY_FN: bool = false;
+
+    const LRU: bool = false;
 }
 
 fn accumulator_contents(
--- a/components/salsa-2022-macros/src/jar.rs
+++ b/components/salsa-2022-macros/src/jar.rs
@@ -41,6 +41,8 @@ impl crate::options::AllowedOptions for Jar {
     const DB: bool = true;
 
     const RECOVERY_FN: bool = false;
+
+    const LRU: bool = false;
 }
 
 pub(crate) fn jar_struct_and_friends(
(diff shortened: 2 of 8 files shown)


nikomatsakis (approved):

nikomatsakis:
This looks great. We probably do need to keep those methods for defining lru at runtime, I guess, I forgot about that! I imagine it should be something like: if `lru` is specified to any value but 0, then there is a method provided to alter the capacity? We can file a follow-up issue for it.

nikomatsakis:
I don't know how to test the fact that you get an error when lru/specify are listed together -- I guess maybe a compile-fail doctest?

PR #353, opened 2022-08-18
start documenting plumbing

nikomatsakis:
Feedback desired! I am trying to document an overview of the new salsa 2022 plumbing. I'd love for folks to [read these docs and tell me if they make sense](https://deploy-preview-353--salsa-rs.netlify.app/plumbing.html).


nikomatsakis:
cc @XFFXFF, @MihailMihov as recent contributors -- please take a look at this [in-progress plumbing docs](https://deploy-preview-353--salsa-rs.netlify.app/plumbing.html) and let me know if they are understandable and/or helpful to get an overview of how the pieces fit together.

XFFXFF (commented):
I can easily understand what is the relationship between `Ingredient`, `Jar` and `Database` from the docs.  
I would also like to know the relationship between `Database`, `Storage` and `Runtime`

nikomatsakis:
@XFFXFF ok, I added some material trying to explain those things better. Maybe it helps.

book/src/plumbing/jars_and_ingredients.md
```diff
@@ -0,0 +1,231 @@
+# Jars and ingredients
+
+{{#include caveat.md}}
```
MihailMihov, line 3: Should the caveat be included here, as it is already in `plumbing.md`? If it should, the path isn't relative and it isn't showing up correctly.
```suggestion
{{#include ../caveat.md}}
```

book/src/plumbing/jars_and_ingredients.md
```diff
@@ -0,0 +1,231 @@
...
+
+Tbis page covers how data is organized in salsa and how links between salsa items (e.g., dependency tracking) works.
```
MihailMihov, line 5:
```suggestion
This page covers how data is organized in salsa and how links between salsa items (e.g., dependency tracking) works.
```

MihailMihov:
The new docs are a nice overview, the only thing that I think would be good is if we take the snippets that are currently written by hand and replace them with anchors to a file before and after the macros are expanded, so that we don't have to worry about them becoming outdated.

nikomatsakis:
yeah, that's a good suggestion. it'd also be good if I could replace the links into github with some other form of link that is more persistent, but I don't know the best choice there.

PR #356, opened 2022-08-18
make Id::from_u32 public

nikomatsakis:
I made it private 'just because', but it turns out that dada uses it, and it seems reasonable.


nikomatsakis:
I'm going to quickly merge this because I want to be able to rebase dada atop the latest changes and use it as a testing bed.

PR #359, opened 2022-08-18
fix: typos in tutorial ir chapter

dawnofmidnight:
Looking at the tutorial, I believe `ProgramSource` is a typo for `SourceProgram`. I'm also pretty sure the `define` -> `derive` is correct, unless `define` is some attribute I'm not aware of.


nikomatsakis (approved):

nikomatsakis:
bors r+

Thanks!

PR #360, opened 2022-08-18
Permit renaming constructors, getters and setters

### LINKED ISSUE #332: Permit renaming getters, setters, and constructors
It's sometimes annoying that salsa defines the `new` function along with all the getters and setters. I've sometimes wanted to override them, e.g. to accent any `impl Into<Foo>` instead of just `Foo`. We should permit users to override them in some way. I am imagining a `#[method_names(foo=bar, ...)]` attribute that can be placed on the struct or on fields, e.g.

```rust
#[salsa::interned]
#[method_names(new = from_string)]
struct Id {
    #[method_names(get = raw_text)] #[return_ref]
    text: String
}

impl Id {
    pub fn new(db: &dyn crate::Db, s: impl Display) {
        Id::from_string(db, s.to_string())
    }

    pub fn text(self, db: &dyn Crate::Db) -> &str {
        self.raw_text(db)
    }
}
```

MihailMihov:
The goal is to add an option `constructor_name` to `#[salsa::input]`, `#[salsa::interned]` and `#[salsa::tracked]` that allows changing the name of the generated constructor. After that add attributes `get` and `set` to the fields which allow overriding the names of the getters and setters. In the end the following snippet should compile and work:
```rust
#[salsa::interned(constructor = from_string)]
struct MyInterned {
    #[get(text)] #[set(set_text)] #[return_ref]
    field: String,
}

impl MyInterned {
    pub fn new(db: &dyn Db, s: impl Display) -> MyInterned {
        MyInterned::from_string(db, s.to_string())
    }

    pub fn field(self, db: &dyn Db) -> &str {
        self.text(db)
    }

    pub fn set_field(self, db: &mut dyn Db, id: String) {
        self.set_text(&mut db, id)
    }
}
```

resolves #332

### DIFF
--- a/components/salsa-2022-macros/src/accumulator.rs
+++ b/components/salsa-2022-macros/src/accumulator.rs
@@ -34,6 +34,8 @@ impl crate::options::AllowedOptions for Accumulator {
     const RECOVERY_FN: bool = false;
 
     const LRU: bool = false;
+
+    const CONSTRUCTOR_NAME: bool = false;
 }
 
 fn accumulator_contents(
--- a/components/salsa-2022-macros/src/input.rs
+++ b/components/salsa-2022-macros/src/input.rs
@@ -64,13 +64,14 @@ impl InputStruct {
         let input_index = self.input_index();
 
         let field_indices = self.all_field_indices();
-        let field_names: Vec<_> = self.all_field_names();
+        let field_names = self.all_field_names();
         let field_tys: Vec<_> = self.all_field_tys();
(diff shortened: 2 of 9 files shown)


MihailMihov:
Currently the `constructor` option is implemented and working as expected, but I have yet to finish the `get` and `set` attributes.

nikomatsakis (approved):
Nice! The main thing missing is docs, but given that we don't have an obvious "reference" place, I'm not sure where to suggest you put this.

Issue #339, opened 2022-08-09
Porting durability related tests for cycle recovery


XFFXFF:
https://github.com/salsa-rs/salsa/pull/336 adds options to tracked functions for cycle recovery and ports over old salsa tests, except for [cycle_disappears_durability](https://github.com/salsa-rs/salsa/blob/03a27a7054ee234a566231243dd5025edfbcdde6/tests/cycles.rs#L326), since we don't have the api that permits setting durability.  This test can be ported when #329 is complete.

PR #362, opened 2022-08-19
Implement Setter API

### LINKED ISSUE #329: Add "builder style" setter APIs that permit setting durability
The `salsa::input` structs generate a setter for each field. We should make an alternate version of the setter that takes a durability. The obvious thing is something like `set_foo_with_durability`, but now that I think about it, maybe a builder API would be more appropriate, in order to be more future proof.

I'm inclined to modify the setters to be builder style. You would write `input.set_foo(&mut db)` and that would return a `salsa::Setter` struct that captures the input, a reference to the field (a `FunctionIngredient`), and the database. The setter would have a `to_value(self, value)` that stores the new value, so that the "normal" call looks like

```rust
input.set_foo(&mut db).to_value(value)
```

but then we can add additional builder style methods, e.g.

```rust
input.set_foo(&mut db).with_durability(Durability::HIGH).to_value(value)
```

This makes those method highly discoverable and easily documented. It also means we can add new "final actions", e.g. maybe for "taking", deleting, or something else in the future.

crlf0710:
Fixes #329

### DIFF
--- a/components/salsa-2022/src/setter.rs
+++ b/components/salsa-2022/src/setter.rs
@@ -1,0 +1,39 @@
+use crate::input_field::InputFieldIngredient;
+use crate::{AsId, Durability, Runtime};
+use std::hash::Hash;
+
+#[must_use]
+pub struct Setter<'setter, K, F> {
+    runtime: &'setter mut Runtime,
+    key: K,
+    ingredient: &'setter mut InputFieldIngredient<K, F>,
+    durability: Durability,
+}
+
+impl<'setter, K, F> Setter<'setter, K, F>
+where
+    K: Eq + Hash + AsId,
+{
+    pub fn new(
+        runtime: &'setter mut Runtime,
+        key: K,
+        ingredient: &'setter mut InputFieldIngredient<K, F>,
+    ) -> Self {
+        Setter {
+            runtime,
+            key,
+            ingredient,
+            durability: Durability::LOW,
+        }
+    }
+
+    pub fn with_durability(self, durability: Durability) -> Self {
+        Setter { durability, ..self }
+    }
+
+    pub fn to(self, value: F) -> F {
+        self.ingredient
+            .store(self.runtime, self.key, value, self.durability)
+            .unwrap()
+    }
+}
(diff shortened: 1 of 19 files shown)


components/salsa-2022/src/setter.rs
```diff
@@ -0,0 +1,39 @@
...
+            ingredient,
+            durability: Durability::LOW,
+        }
+    }
+
+    pub fn with_durability(self, durability: Durability) -> Self {
+        Setter { durability, ..self }
+    }
+
+    pub fn to_value(self, value: F) -> F {
```
nikomatsakis, line 34: Nit: maybe we should just call it `to`, just to keep the simplest case short? 

e.g.

```rust
func.set_name(db).to(value)
```

components/salsa-2022/src/setter.rs
```diff
@@ -0,0 +1,39 @@
...
+    ) -> Self {
+        Setter {
+            runtime,
+            key,
+            ingredient,
+            durability: Durability::LOW,
+        }
+    }
+
+    pub fn with_durability(self, durability: Durability) -> Self {
```
nikomatsakis, line 30: Nit: we don't have any tests of durability! Maybe you want to add one? That said, I'm ok to land this PR and port the older tests over as follow-up.

nikomatsakis (changes requested):
This looks great! I left one :bike:-shed question (`to` instead of `to_value`), but the code itself looks ready to land. Do you think you'd be up to port over some of the durability tests from old salsa?

components/salsa-2022/src/setter.rs
crlf0710: Done~

crlf0710:
Rebased and renamed `to_value` to `to` following review.

components/salsa-2022/src/setter.rs
crlf0710: That's #339 so I'd like to leave it to a follow-up PR. I'll see if i can get it right.

nikomatsakis:
Neat.

bors r+

nikomatsakis (approved):

crlf0710:
Oops, there's one left-over in the calc example. Fixed now.

PR #363, opened 2022-08-21
Add method to change LRU capacity at runtime

### LINKED ISSUE #355: Add method to change LRU capacity at runtime

XFFXFF:
closes #355

### DIFF
--- a/salsa-2022-tests/tests/lru.rs
+++ b/salsa-2022-tests/tests/lru.rs
@@ -103,3 +103,47 @@ fn lru_doesnt_break_volatile_queries() {
         }
     }
 }
+
+#[test]
+fn lru_can_be_changed_at_runtime() {
+    let mut db = Database::default();
+    assert_eq!(load_n_potatoes(), 0);
+
+    for i in 0..128u32 {
+        let input = MyInput::new(&mut db, i);
+        let p = get_hot_potato(&db, input);
+        assert_eq!(p.0, i)
+    }
+
+    // Create a new input to change the revision, and trigger the GC
+    MyInput::new(&mut db, 0);
+    assert_eq!(load_n_potatoes(), 32);
+
+    get_hot_potato::set_lru_capacity(&db, 64);
+    assert_eq!(load_n_potatoes(), 32);
+    for i in 0..128u32 {
+        let input = MyInput::new(&mut db, i);
+        let p = get_hot_potato(&db, input);
+        assert_eq!(p.0, i)
+    }
+
+    // Create a new input to change the revision, and trigger the GC
+    MyInput::new(&mut db, 0);
+    assert_eq!(load_n_potatoes(), 64);
+
+    // Special case: setting capacity to zero disables LRU
+    get_hot_potato::set_lru_capacity(&db, 0);
+    assert_eq!(load_n_potatoes(), 64);
+    for i in 0..128u32 {
+        let input = MyInput::new(&mut db, i);
+        let p = get_hot_potato(&db, input);
+        assert_eq!(p.0, i)
+    }
+
+    // Create a new input to change the revision, and trigger the GC
+    MyInput::new(&mut db, 0);
+    assert_eq!(load_n_potatoes(), 192);
+
+    drop(db);
+    assert_eq!(load_n_potatoes(), 0);
+}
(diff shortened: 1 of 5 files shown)


salsa-2022-tests/tests/lru.rs
```diff
@@ -103,3 +118,74 @@ fn lru_doesnt_break_volatile_queries() {
...
+    assert_eq!(load_n_potatoes(), 64);
+    for i in 0..128u32 {
+        let input = MyInput::new(&mut db, i);
+        let p = get_hot_potato(&db, input);
+        assert_eq!(p.0, i)
+    }
+
+    // Create a new input to change the revision, and trigger the GC
+    MyInput::new(&mut db, 0);
+    assert_eq!(load_n_potatoes(), 192);
```
XFFXFF, line 160: Old salsa thinks it should be 128. https://github.com/salsa-rs/salsa/blob/80d0d141945051ec86ecd979ac06cca1e364b27e/tests/lru.rs#L119

salsa-2022-tests/tests/lru.rs
```diff
@@ -103,3 +118,74 @@ fn lru_doesnt_break_volatile_queries() {
...
+
+    drop(db);
+    assert_eq!(load_n_potatoes(), 0);
+}
+
+#[test]
+fn lru_keeps_dependency_info() {
```
XFFXFF, line 167: I forgot to port this test in https://github.com/salsa-rs/salsa/pull/352 and ran into some issues with it.  

The old test used `synthetic_write`  to mock some input of durability has changed. we don't have this API right now
https://github.com/salsa-rs/salsa/blob/80d0d141945051ec86ecd979ac06cca1e364b27e/tests/lru.rs#L151  
https://github.com/salsa-rs/salsa/blob/80d0d141945051ec86ecd979ac06cca1e364b27e/src/runtime.rs#L100-L110

salsa-2022-tests/tests/lru.rs
```diff
@@ -103,3 +118,74 @@ fn lru_doesnt_break_volatile_queries() {
...
+
+    // We want to test that calls to `get_hot_patato2` are still considered
+    // clean. Check that no new executions occur as we go here.
+    db.assert_logs_len((capacity + 1) * 2);
+
+    // calling `get_hot_potato2(0)` has to check that `get_hot_potato(0)` is still valid;
+    // even though we've evicted it (LRU), we find that it is still good
+    let p = get_hot_potato2(&db, *inputs.first().unwrap());
+    assert_eq!(p, 0);
+    db.assert_logs_len(2); // FIXME: the length of logs should be 0
```
XFFXFF, line 190: LRU doesn't seem to keep dependencies, the length of the logs should be 0.

XFFXFF:
fixed filename typos I introduced in https://github.com/salsa-rs/salsa/pull/334#issuecomment-1209346974 (@MihailMihov found those.)

salsa-2022-tests/tests/lru.rs
nikomatsakis: Hmmm, I'm not sure why I didn't add this API, I think it's still on the runtime, giving warnings? I'd be ok to add it, though I'd keep it on the runtime.

salsa-2022-tests/tests/lru.rs
nikomatsakis: We did land some PRs fixing things about the LRU, most notably https://github.com/salsa-rs/salsa/pull/298, maybe that's related? Not obvious how. I have to file an issue to forward port those changes in particular.

MihailMihov:
> fixed filename typos I introduced in [#334 (comment)](https://github.com/salsa-rs/salsa/pull/334#issuecomment-1209346974)

The same typo is also in the files, not just the filenames

nikomatsakis (changes requested):
Can you fix the "deponds/depends" type in the content of the test cases?

XFFXFF:
> The same typo is also in the files, not just the filenames

@MihailMihov I will fix them. thanks again!

XFFXFF:
> Can you fix the "deponds/depends" type in the content of the test cases?
Yes!

salsa-2022-tests/tests/lru.rs
XFFXFF: I just found out that 192 is because we created new inputs for `get_hot_potato`. Although the input value has not changed, the revision has changed and needs to be re-executed. We need to create all the inputs first, then here will be 128, logically equal to old salsa.

nikomatsakis (approved):

PR #369, opened 2022-08-21
Try to fix issue#340

### LINKED ISSUE #340: `debug` method of `DependencyIndex` should be more human readable.
https://github.com/salsa-rs/salsa/blob/03a27a7054ee234a566231243dd5025edfbcdde6/components/salsa-2022/src/key.rs#L36-L38  
Currently, we just return `self` in `debug` method of `DependecyIndex`, so when call `debug`, will get something like  
```rust
DependencyIndex { ingredient_index: IngredientIndex(8), key_index: Some(Id { value: 1 }) }
```
which is not human friendly. Instead of the index, we can output information about the raw values of the ingradient and key. An expected output could be `func_name(arg_value)` for a tracked function.

zjp-CN:
(WIP) 

Follow the instructions from https://github.com/salsa-rs/salsa/issues/340#issuecomment-1210153085.

Still not sure what to do with `fmt_index`...

Fix #340


components/salsa-2022/src/function.rs
```diff
@@ -30,15 +30,17 @@ mod store;
 mod sync;
 
 /// Function ingredients are the "workhorse" of salsa.
-/// They are used for tracked functions, for the "value" fields of tracked structs, and for the fields of input structs.
-/// The function ingredient is fairly complex and so its code is spread across multiple modules, typically one per method.
-/// The main entry points are:
+/// They are used for tracked functions, for the "value" fields of tracked structs, and for the
+/// fields of input structs. The function ingredient is fairly complex and so its code is
+/// spread across multiple modules, typically one per method. The main entry points are:
```
nikomatsakis, line 35: I don't have a strong opinion about this, but the newlines in this comment were intentionally spaced this way to permit easier diffs when text is edited in the future.

nikomatsakis: I'm guessing some kind of formatting tool adjusted these?

components/salsa-2022/src/input.rs
```diff
@@ -95,6 +97,14 @@ where
             "unexpected call: input ingredients do not register for salsa struct deletion events"
         );
     }
+
+    fn fmt_index(
+        &self,
+        _index: Option<crate::Id>,
+        _fmt: &mut std::fmt::Formatter<'_>,
+    ) -> std::fmt::Result {
```
nikomatsakis, line 105: This should be something like:

```rust
if let Some(i) = index {
    write!(fmt, "{}({})", self.debug_name, u32::from(i))
} else {
    write!(fmt, "{}()", self.debug_name)
}
```

components/salsa-2022/src/accumulator.rs
```diff
@@ -148,6 +148,14 @@ where
     fn salsa_struct_deleted(&self, _db: &DB, _id: crate::Id) {
         panic!("unexpected call: accumulator is not registered as a dependent fn");
     }
+
+    fn fmt_index(
+        &self,
+        _index: Option<crate::Id>,
+        _fmt: &mut std::fmt::Formatter<'_>,
+    ) -> std::fmt::Result {
+        todo!()
```
nikomatsakis, line 157: Ideally, we'd give a debug name to accumulators, and then do something like

```rust
if let Some(i) = index {
    write!(fmt, "{}({})", self.debug_name, u32::from(i))
} else {
    write!(fmt, "{}()", self.debug_name)
}
```

(Actually, given how often this little snippet will occur across various ingredients, maybe we should add a helper function to do it.)

components/salsa-2022/src/interned.rs
```diff
@@ -234,6 +234,14 @@ where
     fn salsa_struct_deleted(&self, _db: &DB, _id: crate::Id) {
         panic!("unexpected call: interned ingredients do not register for salsa struct deletion events");
     }
+
+    fn fmt_index(
+        &self,
+        _index: Option<crate::Id>,
+        _fmt: &mut std::fmt::Formatter<'_>,
+    ) -> std::fmt::Result {
+        todo!()
```
nikomatsakis, line 243: As above

nikomatsakis (changes requested):
I left a preliminary suggestion for `fmt_index`. I *think* that's probably good enough for a start, though I think we can do better.

components/salsa-2022/src/function.rs
zjp-CN: Yes, rustfmt did this on autosave. And fmt is reverted now.

XFFXFF:
@nikomatsakis Looks like we are not running salsa-2022-tests in CI

XFFXFF (changes requested):
Could you run `UPDATE_EXPECT=1 cargo test --package salsa-2022-tests` under the salsa folder? You will see some changes in the tests files, which may help us check that `fmt_index` is working as expect

nikomatsakis:
Looks we still need another `cargo fmt` run, @zjp-CN

nikomatsakis:
never mind, I pushed it myself

nikomatsakis (approved):

zjp-CN:
> run `UPDATE_EXPECT=1 cargo test --package salsa-2022-tests` under the salsa folder

I did it, and pushed it. But CI still failed.

XFFXFF:
> I did it, and pushed it. But CI still failed.  

try `UPDATE_EXPECT=1 cargo test --all`, there are some tests in examples.

zjp-CN:
This time it will be fine (hopefully).

XFFXFF (approved):
Thanks!  
bors r+

Issue #366, opened 2022-08-21
Compare the "main salsa" runtime with the "salsa 2022" runtime and identify differences


nikomatsakis:
The function ingredient in salsa 2022 was based off of the code from "old salsa", but there were also some improvements made after the fork. This issue exists to track the process of doing a detailed diff between the two and making sure that we've ported over any changes of note. The actual porting of changes will be covered by distinct issues:

* [x] https://github.com/salsa-rs/salsa/issues/365

Issue #365, opened 2022-08-21
Port LRU improvements from 'old salsa' to salsa 2022 crate


nikomatsakis:
The function ingredient in salsa 2022 was based off of the code from "old salsa", but there were also some improvements made after the fork. We had to do a more detailed comparison (#366), but one improvement in particular was https://github.com/salsa-rs/salsa/pull/298, which modified the behavior when LRU eviction occurs -- instead of losing all the dependencies, we recreate the memo with the same dependencies but a different value.

nikomatsakis:
Here are some mentoring instructions.

### Code in old salsa

Let's first trace the code in old salsa. [When an LRU overflow occurs](https://github.com/salsa-rs/salsa/blob/8dfc578edc2c62cef84c49da247dc199b50740ae/src/derived/fetch.rs#L30-L32), the `evict` method is called:

https://github.com/salsa-rs/salsa/blob/8dfc578edc2c62cef84c49da247dc199b50740ae/src/derived/fetch.rs#L106-L108

which invokes the [`evict` method on `memo_map`](https://github.com/salsa-rs/salsa/blob/8dfc578edc2c62cef84c49da247dc199b50740ae/src/derived/memo.rs#L33-L61), which (crucially) replaces the memo with a version that no value:

https://github.com/salsa-rs/salsa/blob/8dfc578edc2c62cef84c49da247dc199b50740ae/src/derived/memo.rs#L52-L58

### Eviction in salsa 2022

The salsa 2022 code triggers eviction [here](https://github.com/salsa-rs/salsa/blob/master/components/salsa-2022/src/function/fetch.rs#L26-L28), which leads to calling [`evict`](https://github.com/salsa-rs/salsa/blob/8dfc578edc2c62cef84c49da247dc199b50740ae/components/salsa-2022/src/function/fetch.rs#L96-L115), which invokes `delete_memo`:

https://github.com/salsa-rs/salsa/blob/8dfc578edc2c62cef84c49da247dc199b50740ae/components/salsa-2022/src/function/fetch.rs#L110-L112

`delete_memo` is also used elsewhere, and it has a very simple job of removing the memo from the map:

https://github.com/salsa-rs/salsa/blob/8dfc578edc2c62cef84c49da247dc199b50740ae/components/salsa-2022/src/function/delete.rs#L9-L19

### What we need to do

[...]

PR #370, opened 2022-08-22
Port LRU improvements from 'old salsa' to salsa 2022 crate #365

macovedj:
Solve for [this](https://github.com/salsa-rs/salsa/issues/366).

### DIFF
--- a/components/salsa-2022/src/function/fetch.rs
+++ b/components/salsa-2022/src/function/fetch.rs
@@ -108,7 +108,7 @@ where
                 }
 
                 QueryOrigin::Derived(_) => {
-                    self.delete_memo(key);
+                    self.memo_map.evict(key);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


macovedj:
Looked like the 2022 salsa didn't have a concept of `QueryInput`.  Considered porting it, but it looked like the same thing might be able to be handled correctly through the 2022 concept of `QueryOrigin`.  Gave it a shot, but wondering if maybe with more context I'd realize that the `QueryInput` port is actually necessary.

nikomatsakis:
`QueryOrigin` is the "evolved" version of `QueryInput`, I believe.

nikomatsakis (changes requested):

nikomatsakis:
See comment [here](https://github.com/salsa-rs/salsa/pull/371#issuecomment-1222141346) -- there's overlap with another PR, so let's land one but not both :) -- but also there is a need for a test. 

Which PR we land, thanks and I hope you'll pick up another issue!

Issue #364, opened 2022-08-21
Add an API to mock some input of durability has changed.


XFFXFF:
`synthetic_write` in old salsa causes the system to act *as though* some  input of durability `durability` has changed. Maybe we should support it in salsa-2022 too. 
https://github.com/salsa-rs/salsa/blob/80d0d141945051ec86ecd979ac06cca1e364b27e/src/runtime.rs#L100-L110

nikomatsakis:
Yeah, it's useful for testing. Let's do it. Not sure why I left it out.

PR #371, opened 2022-08-22
Ported LRU improvements from old salsa to salsa 2022

OLUWAMUYIWA:
## What?
In handling evictions from the LRU, I replaced the `delete_memo`  with `evict` in the while keeping the `delete_memo` intact. 

## How?
This mechanism was already implemented in old `salsa `. I only had to port it.

Fixes: #365


nikomatsakis:
Ack! We had a race condition with @macovedj, who opened #370. Sorry for the overlap.

This PR looks like it already has the suggestion I made in #370, but it needs a run of `cargo fmt`! I'm happy to land either one, really -- but both of them could use a test. 

I'm a bit confused though because the test I wrote at the time:

https://github.com/salsa-rs/salsa/blob/0f3c887cb186c9a3977b5e2e7e9d5514d33230c2/tests/lru.rs#L139-L163

*appears* to have been ported to salsa-2022

https://github.com/salsa-rs/salsa/blob/0f3c887cb186c9a3977b5e2e7e9d5514d33230c2/salsa-2022-tests/tests/lru.rs#L166-L190

Oh, wait, I guess the problem is that we don't have the "synthetic write" in the new version. I believe @XFFXFF pointed out that this API wasn't added to salsa-2022 yet in https://github.com/salsa-rs/salsa/issues/364, though it would be straight-forward to add it I think.

https://github.com/salsa-rs/salsa/blob/0f3c887cb186c9a3977b5e2e7e9d5514d33230c2/tests/lru.rs#L151

OLUWAMUYIWA:
Sure. I'll look into it now.

XFFXFF:
I created a pr https://github.com/salsa-rs/salsa/pull/376 to add `synthetic_write`, the test `lru_keeps_dependency_info` in that pr was broken. I fetch this pr locally, and it makes `lru_keeps_dependency_info` with `synthetic_write` pass.

XFFXFF:
@OLUWAMUYIWA looks like we need to run a `cargo fmt`

macovedj:
Looks like this one's a bit further along than mine at this point.  I'll go ahead and close mine.

PR #372, opened 2022-08-22
run `cargo test -p salsa-2022-tests` in CI

XFFXFF:
* run `cargo test -p salsa-2022-tests` in CI
* check format first


nikomatsakis:
Does `--all` not run the `salsa-2022-tests` tests?

nikomatsakis:
Oh! We were..not using `--all`. Maybe we should do that instead?

XFFXFF:
> Oh! We were..not using `--all`. Maybe we should do that instead?

Yes! that's better

PR #375, opened 2022-08-22
Editing pass over the Overview, Tutorial, and Reference sections of the book

seanchen1991:
This PR includes a number of proposed changes, mostly fixing typos, a few additions/rearrangements where I thought it made sense, as well as a couple of questions delineated in brackets, to the book in an effort to polish it up a bit. 

There's also a question around whether the book should standardize around 'Salsa' or 'salsa'.


book/src/reference/algorithm.md
```diff
@@ -65,7 +65,11 @@ If the module text is changed, we saw that we have to re-execute `parse_module`,
 
 ## Durability: an optimization
 
-As an optimization, salsa includes the concept of **durability**. When you set the value of a tracked function, you can also set it with a given _durability_:
+As an optimization, salsa includes the concept of **durability**, which is the notion of how often some piece of tracked data changes. 
+
+For example, when compiling a Rust program, you might mark the inputs from crates.io as _high durability_ inputs, since they are unlikely to change. The current workspace could be marked as _low durability_, since changes to it are happening all the time.
```
XFFXFF, line 70: The example is great and makes it more clear

book/src/tutorial/parser.md
```diff
@@ -63,7 +63,7 @@ It's generally better to structure tracked functions as functions of a single sa
 
 ### The `return_ref` annotation
 
-You may have noticed that `parse_statements` is tagged with `#[salsa::tracked(return_ref)]`.
+You may have noticed that `parse_statements` is tagged with `#[salsa::tracked(return_ref)]`. [The function isn't actually tagged with `return_ref`; is the text incorrect or the code example?]
```
nikomatsakis, line 66: I'm planning to do some major changes to this section

nikomatsakis (changes requested):
OK, after reading these changes, I think I came around to the POV that we should have Salsa be capitalized when using it as a proper noun (but `salsa` when referring to the *crate* specifically). It just ... looks a bit odd when I see all the diffs. Sorry @seanchean1991, do you mind reworking it with that in mind?

book/src/tutorial/parser.md
nikomatsakis: By this I mean the tutorial

nikomatsakis: I have a branch that extends the tutorial to include the type checker etc, and I have to go back and rewrite some of the sections from the parser as a result.

book/src/plumbing/jars_and_ingredients.md
```diff
@@ -167,15 +159,16 @@ We can then do things like ask, "did this input change since revision R?" by
 * using the ingredient index to find the route and get a `&dyn Ingredient<DB>`
 * and then invoking the `maybe_changed_since` method on that trait object.
 
-### HasJarsDyn
+### `HasJarsDyn`
 
 There is one catch in the above setup.
-We need the database to be dyn-safe, and we also need to be able to define the database trait and so forth without knowing the final database type to enable separate compilation.
+We need the database to be dyn-safe [what does this mean?], and we also need to be able to define the database trait without knowing the final database type to enable separate compilation.
```
seanchen1991, line 165: It's not entirely clear to me what "dyn-safe" means here. An example or a definition would help.

book/src/plumbing/jars_and_ingredients.md
```diff
@@ -205,13 +198,13 @@ The implementation of this method is defined by the `#[salsa::db]` macro; it sim
...
 
 ```rust,ignore
 {{#include ../../../components/salsa-2022-macros/src/jar.rs:create_jar}}
 ```
 
 The code to create the ingredients for any particular item is generated by their associated macros (e.g., `#[salsa::tracked]`, `#[salsa::input]`), but it always follows a particular structure.
-To create an ingredient, we first invoke `Routes::push` which creates the routes to that ingredient and assigns it an `IngredientIndex`.
-We can then invoke (e.g.) `FunctionIngredient::new` to create the structure.
+To create an ingredient, we first invoke `Routes::push`, which creates the routes to that ingredient and assigns it an `IngredientIndex`.
+We can then invoke (e.g.) [Missing an example here] `FunctionIngredient::new` to create the structure.
```
seanchen1991, line 209: Looks like we're missing an example here.

book/src/tutorial/db.md
```diff
@@ -14,11 +14,11 @@ In `calc`, the database struct is in the [`db`] module, and it looks like this:
 ```
 
 The `#[salsa::db(...)]` attribute takes a list of all the jars to include.
-The struct must have a field named `storage` whose types is `salsa::Storage<Self>`, but it can also contain whatever other fields you want.
+The struct must have a field named `storage` whose type is `salsa::Storage<Self>`, but it can also contain whatever other fields you want.
 The `storage` struct owns all the data for the jars listed in the `db` attribute.
 
 The `salsa::db` attribute autogenerates a bunch of impls for things like the `salsa::HasJar<crate::Jar>` trait that we saw earlier.
-This means that
+This means that ... [what goes here?]
```
seanchen1991, line 21: This sentence just abruptly ends; what should go here?

book/src/tutorial/parser.md
seanchen1991: Ok, I'll leave this text as-is for now then.

book/src/plumbing/jars_and_ingredients.md
nikomatsakis: Ah, yeah, that's "Rust jargon". It means "usable with dyn". I'm trying to find something to link to and failing! That...seems like a gap in the Rust reference.

book/src/plumbing/jars_and_ingredients.md
nikomatsakis: Yes, that is what it is referring to. The idea is that it's the `new` method of whichever ingredient is being created, I guess.

book/src/tutorial/db.md
nikomatsakis: I have no idea :) let's just remove it?

book/src/plumbing/jars_and_ingredients.md
nikomatsakis: Notably, a trait with generic methods cannot be used `dyn`. Let me see if I can rewrite this.

```suggestion
The user's code always interacts with a `dyn crate::Db` value, where `crate::Db` is the trait defined by the jar; the `crate::Db` trait extends `salsa::HasJar` which in turn extends `salsa::Database`.
Ideally, we would have `salsa::Database` extend `salsa::HasJars`, which is the main trait that gives access to the jars data.
But we don't want to do that because `HasJars` defines an associated type `Jars`, and that would mean that every reference to `dyn crate::Db` would have to specify the jars type using something like `dyn crate::Db<Jars = J>`.
This would be unergonomic, but what's worse, it would actually be impossible: the final Jars type combines the jars from multiple crates, and so it is not known to any individual jar crate.
To workaround this, `salsa::Database` in fact extends *another* trait, `HasJarsDyn`, that doesn't reveal the `Jars` or ingredient types directly, but just has various method that can be performed on an ingredient, given its `IngredientIndex`.
```

nikomatsakis: More verbose, but hopefully clearer!

seanchen1991:
I'm going to consider this PR as 'ready to merge' once all of the feedback has been adequately addressed. I'll open a separate PR for edits to the rest of the book.

nikomatsakis (approved):

book/src/plumbing/jars_and_ingredients.md
8573: > It's not entirely clear to me what "dyn-safe" means here. An example or a definition would help.

> Ah, yeah, that's "Rust jargon". It means "usable with dyn". I'm trying to find something to link to and failing! That...seems like a gap in the Rust reference.

Isn't it https://doc.rust-lang.org/reference/items/traits.html#object-safety ?

PR #376, opened 2022-08-23
add `synthetic_write`

### LINKED ISSUE #364: Add an API to mock some input of durability has changed.

XFFXFF:
fixes #364 

add `synthetic_write` and use it in test `lru_keeps_dependency_info`, the test will now be broken. We can use `lru_keeps_dependency_info` as a test for https://github.com/salsa-rs/salsa/issues/365, which already has a pr https://github.com/salsa-rs/salsa/pull/371.

### DIFF
--- a/components/salsa-2022-macros/src/db.rs
+++ b/components/salsa-2022-macros/src/db.rs
@@ -120,6 +120,10 @@ fn has_jars_dyn_impl(input: &syn::ItemStruct, storage: &syn::Ident) -> syn::Item
                 self.#storage.runtime()
             }
 
+            fn runtime_mut(&mut self) ->&mut salsa::Runtime {
+                self.#storage.runtime_mut()
+            }
+
             fn maybe_changed_after(
                 &self,
                 input: salsa::key::DependencyIndex,
@@ -159,9 +163,14 @@ fn has_jars_dyn_impl(input: &syn::ItemStruct, storage: &syn::Ident) -> syn::Item
                 let ingredient = self.#storage.ingredient(ingredient);
                 ingredient.salsa_struct_deleted(self, id);
             }
+
             fn fmt_index(&self, index: salsa::key::DependencyIndex, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                 let ingredient = self.#storage.ingredient(index.ingredient_index());
                 ingredient.fmt_index(index.key_index(), fmt)
+            }
+
+            fn synthetic_write(&mut self, durability: salsa::Durability) {
+                self.runtime_mut().synthetic_write(durability);
             }
         }
     }
--- a/components/salsa-2022-macros/src/db.rs
+++ b/components/salsa-2022-macros/src/db.rs
@@ -168,10 +168,6 @@ fn has_jars_dyn_impl(input: &syn::ItemStruct, storage: &syn::Ident) -> syn::Item
                 let ingredient = self.#storage.ingredient(index.ingredient_index());
                 ingredient.fmt_index(index.key_index(), fmt)
             }
-
-            fn synthetic_write(&mut self, durability: salsa::Durability) {
-                self.runtime_mut().synthetic_write(durability);
-            }
         }
     }
 }
--- a/components/salsa-2022/src/database.rs
+++ b/components/salsa-2022/src/database.rs
@@ -22,6 +22,14 @@ pub trait Database: HasJarsDyn + AsSalsaDatabase {
     fn synthetic_write(&mut self, durability: Durability) {
         self.runtime_mut().synthetic_write(durability);
     }
+
+    /// Reports that the query depends on some state unknown to salsa.
+    ///
+    /// Queries which report untracked reads will be re-executed in the next
+    /// revision.
+    fn report_untracked_read(&self) {
+        self.runtime().report_untracked_read();
+    }
 }
 
 /// Indicates a database that also supports parallel query
--- a/components/salsa-2022/src/database.rs
+++ b/components/salsa-2022/src/database.rs
@@ -1,4 +1,4 @@
-use crate::{storage::HasJarsDyn, DebugWithDb, Event};
+use crate::{storage::HasJarsDyn, DebugWithDb, Durability, Event};
 
 pub trait Database: HasJarsDyn + AsSalsaDatabase {
     /// This function is invoked at key points in the salsa
@@ -9,6 +9,18 @@ pub trait Database: HasJarsDyn + AsSalsaDatabase {
     /// the standard `log` facade.
     fn salsa_event(&self, event: Event) {
         log::debug!("salsa_event: {:?}", event.debug(self));
+    }
+
+    /// A "synthetic write" causes the system to act *as though* some
+    /// input of durability `durability` has changed. This is mostly
+    /// useful for profiling scenarios.
+    ///
+    /// **WARNING:** Just like an ordinary write, this method triggers
+    /// cancellation. If you invoke it while a snapshot exists, it
+    /// will block until that snapshot is dropped -- if that snapshot
+    /// is owned by the current thread, this could trigger deadlock.
+    fn synthetic_write(&mut self, durability: Durability) {
+        self.runtime_mut().synthetic_write(durability);
     }
 }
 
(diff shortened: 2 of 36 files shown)


OLUWAMUYIWA:
Hi @XFFXFF how can we get this merged? Since both this and #371 won't pass tests individually? I had some trouble porting `synthetic_write`. Still trying to get familiar with the code base. I've read your commit.

nikomatsakis:
@XFFXFF 

```
---- lru_keeps_dependency_info stdout ----
thread 'lru_keeps_dependency_info' panicked at 'assertion failed: `(left == right)`
  left: `2`,
 right: `0`', /home/runner/work/salsa/salsa/salsa-2022-tests/src/lib.rs:30:9
```

Any idea what's up with that? :)

XFFXFF:
> Any idea what's up with that? :)

 @nikomatsakis  This test will pass after #371 lands

XFFXFF:
> Hi @XFFXFF how can we get this merged? Since both this and #371 won't pass tests individually? I had some trouble porting `synthetic_write`. Still trying to get familiar with the code base. I've read your commit.

Hmmm, I am thinking of rebasing this branch to #371

XFFXFF:
@OLUWAMUYIWA I leave you some instructions to rebase this pr to your branch if you want.  
```bash
# add upstream remote if you have not
git remote add upstream https://github.com/salsa-rs/salsa.git

# fetch my branch
git fetch upstream pull/376/head:synthetic_write

# rebase it to your branch
git rebase synthetic_write

# push with --force
git push origin --force lru_port
```

salsa-2022-tests/tests/lru.rs
```diff
@@ -178,6 +183,9 @@ fn lru_keeps_dependency_info() {
         assert_eq!(x as usize, i);
     }
 
+    db.salsa_runtime_mut()
+        .synthetic_write(salsa::Durability::HIGH);
```
XFFXFF, line 187: > Any idea what's up with that? :)

@nikomatsakis (add some details, hope it's clear) `get_hot_potato(0)` was evicted due to lru. `synthetic_write` will add `current_revision`,  `get_hot_potato2(0)` and `get_hot_potato(0)` will be re-executed, since we don't keep lru dependency info in this branch. #371 ported lru improments that will make this test pass

nikomatsakis: I see

nikomatsakis:
It seems to me that this PR plus #371 should be combined into one, so that the tests can pass. I don't care which one gets the commits of the other in particular -- 

You could either `git rebase`, as @XFFXFF described [here](https://github.com/salsa-rs/salsa/pull/376#issuecomment-1225002360), or `git cherry-pick` the individual commits (you still have to add and fetch the remote, though).

nikomatsakis:
Oh, I see, #371 only needs cargo fmt. OK, I pushed the commit.

(@XFFXFF -- you *should* have permission to instruct bors, but I guess you don't technically have write permission. I'm not sure the best way to set that up, I'd prefer to have it so that only bors writes to master without some "extra" setting...)

nikomatsakis:
OK, @XFFXFF, #371 is merged, so do you want to rebase this PR?

XFFXFF:
> OK, @XFFXFF, #371 is merged, so do you want to rebase this PR?

Yes, I'm going to do it

salsa-2022-tests/tests/hello_world.rs
```diff
@@ -44,6 +44,10 @@ impl salsa::Database for Database {
     fn salsa_runtime(&self) -> &salsa::Runtime {
         self.storage.runtime()
     }
+
+    fn salsa_runtime_mut(&mut self) -> &mut salsa::Runtime {
+        self.storage.runtime_mut()
+    }
```
nikomatsakis, line 50: Hmm, I am not thrilled about requiring everyone to add this method just so that they can make synthetic writes!

I think I would rather add the "synthetic write" method to `HasJarsDyn` or some plumbing trait, or else add the `salsa_runtime_mut` method into that trait.

(For that matter, why do we make everyone add the `salsa_runtime` method...? I don't think there's a good reason, the code in `salsa::db` already assumes it can find the `storage` field...)

nikomatsakis (changes requested):

nikomatsakis:
@XFFXFF OK -- I read the code a bit more carefully and left some comments. I don't think we're quite ready to land.

XFFXFF:
> @XFFXFF OK -- I read the code a bit more carefully and left some comments. I don't think we're quite ready to land.

I'll hack them after dinner.

components/salsa-2022-macros/src/db.rs
```diff
@@ -159,10 +163,15 @@ fn has_jars_dyn_impl(input: &syn::ItemStruct, storage: &syn::Ident) -> syn::Item
...
             }
+
             fn fmt_index(&self, index: salsa::key::DependencyIndex, fmt: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
                 let ingredient = self.#storage.ingredient(index.ingredient_index());
                 ingredient.fmt_index(index.key_index(), fmt)
             }
+
+            fn synthetic_write(&mut self, durability: salsa::Durability) {
+                self.runtime_mut().synthetic_write(durability);
+            }
```
XFFXFF, line 174: Add `runtime_mut` and `synthetic_write` to `HasJarsDyn`

components/salsa-2022/src/database.rs
```diff
@@ -11,7 +11,9 @@ pub trait Database: HasJarsDyn + AsSalsaDatabase {
         log::debug!("salsa_event: {:?}", event.debug(self));
     }
 
-    fn salsa_runtime(&self) -> &Runtime;
+    fn salsa_runtime(&self) -> &Runtime {
+        self.runtime()
+    }
```
XFFXFF, line 16: Give `salsa_runtime` a default implementation.  
I wanted to delete it directly, because we already have `db.runtime()`, but found that it is used in many places in the salsa code

salsa-2022-tests/tests/hello_world.rs
XFFXFF: @nikomatsakis  I added `synthetic_write` to `HasJarsDyn`, but using `HasJarsDyn` exposes many APIs that users don't care about, such as `maybe_changed_after`, `remote_stale_output`.

nikomatsakis: > Would it be a good idea to add synthetic_write to the salsa::Database trait and provide a default implementation for it?

I think that makes sense, yes.

components/salsa-2022/src/database.rs
nikomatsakis: Those places can probably be rewritten to use `db.runtime()`?

nikomatsakis: I think a default implementation is fine, though

components/salsa-2022/src/database.rs
```diff
@@ -11,7 +11,25 @@ pub trait Database: HasJarsDyn + AsSalsaDatabase {
...
+        self.runtime_mut().synthetic_write(durability);
+    }
+
+    /// Reports that the query depends on some state unknown to salsa.
+    ///
+    /// Queries which report untracked reads will be re-executed in the next
+    /// revision.
+    fn report_untracked_read(&self) {
+        self.runtime().report_untracked_read();
+    }
```
XFFXFF, line 32: I also added `report_untracked_read` to `salsa::Database`, so users can use `db.report_untracked_read` instead of `db.runtime().report_untracked_read`, I don't think users need to know the existence of `Runtime`

components/salsa-2022/src/database.rs
XFFXFF: I removed `salsa_runtime` and replaced it with `db.runtime()` wherever it appears

nikomatsakis (approved):

PR #379, opened 2022-08-23
Update test.yml

agluszak:
- Add names to steps
- No need to run clippy AND check (https://stackoverflow.com/questions/57449356/is-cargo-clippy-a-superset-of-cargo-check)


nikomatsakis:
This seems good, but clippy is giving various errors for some reason.

nikomatsakis (approved):

Issue #337, opened 2022-08-09
Support singleton inputs


nikomatsakis:
[Zuo Baoquan points out that singleton inputs would be convenient](https://salsa.zulipchat.com/#narrow/stream/145099-general/topic/Access.20input.20without.20handle.3F). A proposed API looks something like this:

```rust
#[salsa::input(singleton)]
struct Configuration {
    verbose: bool
}
```

With the `singleton` option, there is only ever a single input in existence. You create it with `Configuration::new` (perhaps we should rename it, not sure, but let's separate that question):

```rust
let c1 = Configuration::new(&mut db, false); // verbose is false by default
```

You can then get back the global value by `Configuration::get`...

```rust
let c2: Option<Configuration> = Configuration::get(&db);
assert_eq!(Some(c1), c2);
```

...this will return `None` if the configuration is not yet created.

Like any input you can modify its fields, triggering a new revision:

```rust
c1.set_verbose(&mut db, true);
assert_eq!(c1.verbose(&db), true);
```

You can also invoke `new` again. Since the value already exists, that just updates its fields:

```rust
let c3 = Configuration::new(&mut db, false);
assert_eq!(c1, c3); // same instance as before is returned
assert_eq!(c3.verbose(&db), false); // field is updated
```

nikomatsakis:
A bit of bike-shedding is needed.

Some questions

### Should `get` return `Option<Configuration>` or just `Configuration`?

Reading the verbose field is kind of painful...

```rust
Configuration::get(&db).unwrap().verbose(&db)
```

It might be nice if `get` returned `Configuration`, so you don't have to write `::get(&db).unwrap()`...

```rust
Configuration::get(&db).verbose(&db)
```

If we did that, it'd probably be nice to have *some* way to test if the result is initialized, maybe add a method with a different name, like `try_get`?

### Convenience getters

Even this is kind of long...

```rust
Configuration::get(&db).verbose(&db)
```

...might be nice to write `Configuration::verbose(&db)`, or something similar.

### new vs init

Maybe instead of `Configuration::new` we should write `Configuration::init`, and maybe it should panic if already initialized?

nikomatsakis:
Personally, I'm somewhat inclined towards consistency -- making it easier to convert code that was non-singleton into singleton and vice-versa. I have a vague hunch that there will be times when you don't know yet if the singleton has been initialized or not, so making `new` simply reinitialize fields would be more convenient. 

This leads me to:

* `new`, not `init`
[...]

PR #383, opened 2022-08-25
Have users of `SalsaStruct` specify allowed options

DropDemBits:
This allows adding new options exclusively to `tracked`, `input`, `interned`, or any combination of them, without having to add support code to `SalsaStruct`.

The motivating example is #337, where `input` needs to accept a new `singleton` option, but `tracked` and `interned` should reject it.

### DIFF
--- a/components/salsa-2022-macros/src/input.rs
+++ b/components/salsa-2022-macros/src/input.rs
@@ -17,14 +17,34 @@ pub(crate) fn input(
     }
 }
 
-struct InputStruct(SalsaStruct);
+struct InputStruct(SalsaStruct<Self>);
 
 impl std::ops::Deref for InputStruct {
-    type Target = SalsaStruct;
+    type Target = SalsaStruct<Self>;
 
     fn deref(&self) -> &Self::Target {
         &self.0
     }
+}
+
+impl crate::options::AllowedOptions for InputStruct {
+    const RETURN_REF: bool = false;
+
+    const SPECIFY: bool = false;
+
+    const NO_EQ: bool = false;
+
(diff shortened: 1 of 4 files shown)


DropDemBits:
(I honestly wasn't expecting this refactor to be such a simple change, I was imagining it'd be a much larger one 😅)

XFFXFF (commented):
@DropDemBits  Looks like we need to run a `cargo fmt`

nikomatsakis (approved):
Looks good! But yes, we need a cargo fmt run

components/salsa-2022-macros/src/input.rs
```diff
@@ -17,16 +17,36 @@ pub(crate) fn input(
     }
 }
 
-struct InputStruct(SalsaStruct);
+struct InputStruct(SalsaStruct<Self>);
 
 impl std::ops::Deref for InputStruct {
-    type Target = SalsaStruct;
+    type Target = SalsaStruct<Self>;
```
seanchen1991, line 23: Could someone explain to me why adding the `Self` parameter here is necessary? What does it enable that couldn't be done without the parameter present?

DropDemBits: Adding the `Self` parameter there is necessary as it needs to match the type of the newtype's tuple field, which is also a `SalsaStruct<Self>`. Aside from that, the functionality of the impl is the same (we're still providing access to the same fields and functions of `SalsaStruct`).

XFFXFF (approved):
Thanks!
bors r+

PR #390, opened 2022-08-27
Add some doc on `specify`

XFFXFF:
Add some documentation to explain why `specify` only works if the key is a tracked struct created in the current query, as per [this](https://salsa.zulipchat.com/#narrow/stream/146365-good-first-issue/topic/questions.20about.20.60specify.60/near/295416521) zulip thread.  
I also added some tests to test that "specify" shouldn't work in some cases.


salsa-2022-tests/tests/compile-fail/specify-does-not-work-if-the-key-is-a-salsa-input.stderr
```diff
@@ -0,0 +1,14 @@
+error[E0277]: the trait bound `MyInput: TrackedStructInDb<dyn Db>` is not satisfied
+   --> tests/compile-fail/specify-does-not-work-if-the-key-is-a-salsa-input.rs:21:28
+    |
+20  | #[salsa::tracked(jar = Jar, specify)]
+    | ------------------------------------- required by a bound introduced by this call
+21  | fn tracked_fn(db: &dyn Db, input: MyInput) -> MyTracked {
+    |                            ^^^^^ the trait `TrackedStructInDb<dyn Db>` is not implemented for `MyInput`
```
XFFXFF, line 7: I wonder if there is a way to make the error message more friendly. Maybe something like "specify only works if the key is the tracked struct", but get a `salsa::input` here.

XFFXFF:
Oops, the error message is a bit different between `nintly` and `stable`

salsa-2022-tests/tests/compile-fail/specify-does-not-work-if-the-key-is-a-salsa-input.stderr
```diff
@@ -0,0 +1,14 @@
+error[E0277]: the trait bound `MyInput: TrackedStructInDb<dyn Db>` is not satisfied
+   --> tests/compile-fail/specify-does-not-work-if-the-key-is-a-salsa-input.rs:21:28
```
XFFXFF, line 2: It says "test/compile-fail/specify-does-not-work-if-the-key-is-a-salsa-input.rs:20:1" on the nightly version, the difference is in the span

salsa-2022-tests/tests/compile_fail.rs
```diff
@@ -1,3 +1,4 @@
+#[rustversion::not(nightly)]
```
XFFXFF, line 1: Add this to make `compile_fail' only run when the rust version is not nightly.
Is it ok to not test on the nightly version?

salsa-2022-tests/tests/compile_fail.rs
```diff
@@ -1,3 +1,4 @@
+#[rustversion::stable]
```
XFFXFF, line 1: Add this to make `compile_fail' only run on stable.
Is it ok?

nikomatsakis (approved):

nikomatsakis:
Nice :)

bors r+

salsa-2022-tests/tests/compile_fail.rs
nikomatsakis: seems ok for now

PR #392, opened 2022-08-28
Allow creation of tracked methods

### LINKED ISSUE #319: Support tracked methods
In the Salsa 2022 system, tracked functions are almost always defined on some kind of salsa struct, and I'm contemplating making that mandatory, even:

```rust
#[salsa::tracked]
struct MyType { ... }

#[salsa::tracked]
fn some_function(db: &dyn crate::Db, arg: MyType) {
    ...
}
```

Often, this method is "wrapped" in an inherent impl:

```rust
impl MyType {
    fn some_method(db: &dyn crate::Db) {
        some_function(db, arg)
    }
}
```

It'd be nice if you could have Salsa do the wrapping for you:

```rust
#[salsa::tracked]
struct MyType { ... }

#[salsa::tracked]
impl MyType {
    fn some_method(db: &dyn crate::Db) { ... }
}
```

Skepfyr:
Fixes #319.

This allows users to annotate impl blocks with `#[salsa::tracked]` and
then create tracked methods by marking individual functions with
`#[salsa::tracked]`.

Note this requires your code that looks like:
```rust
#[salsa::tracked(jar = Jar)]
impl MyInput {
    #[salsa::tracked]
    fn tracked_fn(self, db: &dyn Db) -> u32 {
        self.field(db) * 2
    }
}
```
You get an error if you annotate a method with `#[salsa::tracked]` but forget to mark the impl block.

It got messier than I was hoping but I think it turned out alright, this would look really pretty if we had [inherent associated types](https://github.com/rust-lang/rust/issues/8995), but we don't. Annoyingly even if that landed I think we'd still need the attribute on the impl block just so that it was possible to create the associated struct somewhere as you can't put types inside an impl block (and they aren't accessible if placed inside a function).

### DIFF
--- a/components/salsa-2022-macros/src/tracked_fn.rs
+++ b/components/salsa-2022-macros/src/tracked_fn.rs
@@ -18,6 +18,13 @@ pub(crate) fn tracked_fn(
         ));
     }
 
+    if let syn::FnArg::Receiver(receiver) = &item_fn.sig.inputs[0] {
+        return Err(syn::Error::new(
+            receiver.span(),
+            "#[salsa::tracked] must also be applied to the impl block for tracked methods",
+        ));
+    }
+
     if let Some(s) = &args.specify {
         if requires_interning(&item_fn) {
             return Err(syn::Error::new(
@@ -96,10 +103,12 @@ pub(crate) fn tracked_impl(
                 _ => return None,
             };
             let salsa_tracked_attr = item_method.attrs.iter().position(|attr| {
-                attr.path
-                    .get_ident()
-                    .map(|ident| ident == "tracked")
-                    .unwrap_or(false)
+                let path = &attr.path.segments;
+                path.len() == 2
+                    && path[0].arguments == syn::PathArguments::None
+                    && path[0].ident == "salsa"
+                    && path[1].arguments == syn::PathArguments::None
+                    && path[1].ident == "tracked"
             })?;
             let salsa_tracked_attr = item_method.attrs.remove(salsa_tracked_attr);
             let inner_args = if !salsa_tracked_attr.tokens.is_empty() {
--- a/components/salsa-2022-macros/src/tracked_fn.rs
+++ b/components/salsa-2022-macros/src/tracked_fn.rs
@@ -94,6 +94,15 @@ pub(crate) fn tracked_impl(
             ))
         }
     };
+    let self_type_name = &self_type.path.segments.last().unwrap().ident;
+    let name_prefix = match &item_impl.trait_ {
+        Some((_, trait_name, _)) => format!(
+            "{}_{}",
+            self_type_name,
+            trait_name.segments.last().unwrap().ident
+        ),
+        None => format!("{}", self_type_name),
+    };
     let extra_impls = item_impl
         .items
         .iter_mut()
@@ -116,10 +125,18 @@ pub(crate) fn tracked_impl(
             } else {
                 Ok(FnArgs::default())
             };
-            Some(match inner_args {
-                Ok(inner_args) => tracked_method(&args, inner_args, item_method, self_type),
-                Err(e) => Err(e),
-            })
+            let inner_args = match inner_args {
+                Ok(inner_args) => inner_args,
+                Err(err) => return Some(Err(err)),
+            };
+            let name = format!("{}_{}", name_prefix, item_method.sig.ident);
+            Some(tracked_method(
+                &args,
+                inner_args,
+                item_method,
+                self_type,
+                &name,
+            ))
         })
         // Collate all the errors so we can display them all at once
         .fold(Ok(Vec::new()), |mut acc, res| {
@@ -166,6 +183,7 @@ fn tracked_method(
     mut args: FnArgs,
     item_method: &mut syn::ImplItemMethod,
     self_type: &syn::TypePath,
+    name: &str,
 ) -> syn::Result<TokenStream> {
     args.jar_ty = args.jar_ty.or_else(|| outer_args.jar_ty.clone());
 
@@ -182,14 +200,7 @@ fn tracked_method(
         sig: item_method.sig.clone(),
         block: Box::new(rename_self_in_block(item_method.block.clone())?),
     };
-    item_fn.sig.ident = syn::Ident::new(
-        &format!(
-            "{}_{}",
-            self_type.path.segments.last().unwrap().ident,
-            item_fn.sig.ident
-        ),
-        item_fn.sig.ident.span(),
-    );
+    item_fn.sig.ident = syn::Ident::new(name, item_fn.sig.ident.span());
     // Flip the first and second arguments as the rest of the code expects the
     // database to come first and the struct to come second. We also need to
     // change the self argument to a normal typed argument called __salsa_self.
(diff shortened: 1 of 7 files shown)


nikomatsakis:
Ah, interesting, I hadn't thought of the idea of a `#[tracked]` attribute on the methods *in* the block. Does the code support impl blocks with both tracked *and* untracked methods?

Skepfyr:
Yes, it does. Although I couldn't actually come up with a use for why you'd want that. The biggest reason I went with having `#[tracked]` on the methods in the block is that it lets you annotate them as return_ref etc separately.

nikomatsakis:
@Skepfyr I've found that most methods are not tracked, but there are a few "main" methods that are tracked. Usually I've managed this by just making tracked functions for those particular method and writing "forwarding" methods, but this change would be more convenient.

components/salsa-2022-macros/src/tracked_fn.rs
```diff
@@ -85,6 +71,206 @@ impl crate::options::AllowedOptions for TrackedFn {
...
 }
 
+type ImplArgs = Options<TrackedImpl>;
+
+pub(crate) fn tracked_impl(
+    args: proc_macro::TokenStream,
+    mut item_impl: syn::ItemImpl,
+) -> syn::Result<TokenStream> {
+    let args: ImplArgs = syn::parse(args)?;
+    let self_type = match &*item_impl.self_ty {
```
nikomatsakis, line 81: Do we have to check for trait impls?

e.g.,

```rust
#[salsa::tracked]
impl Foo for MyType {
}
```

components/salsa-2022-macros/src/tracked_fn.rs
```diff
@@ -85,6 +71,206 @@ impl crate::options::AllowedOptions for TrackedFn {
...
+            let item_method = match item {
+                syn::ImplItem::Method(item_method) => item_method,
+                _ => return None,
+            };
+            let salsa_tracked_attr = item_method.attrs.iter().position(|attr| {
+                attr.path
+                    .get_ident()
+                    .map(|ident| ident == "tracked")
+                    .unwrap_or(false)
+            })?;
```
nikomatsakis, line 103: Can we detect `salsa::tracked` here instead? 

Maybe we just use that for consistency?

Alternatively, we could give a better error that way.

salsa-2022-tests/tests/tracked_method.rs
```diff
@@ -0,0 +1,44 @@
...
+struct Jar(MyInput, MyInput_tracked_fn, MyInput_tracked_fn_ref);
+
+trait Db: salsa::DbWithJar<Jar> {}
+
+#[salsa::input(jar = Jar)]
+struct MyInput {
+    field: u32,
+}
+
+#[salsa::tracked(jar = Jar)]
```
nikomatsakis, line 15:
```suggestion
#[salsa::tracked]
```

This is the default, presumably?

nikomatsakis (approved):
Yeah, this looks good! The main question is whether to call the attribute on the methods `#[tracked]` or `#[salsa::tracked]`. 

And of course we're going to have to rework the tutorial, but that can be done after the PR lands.

components/salsa-2022-macros/src/tracked_fn.rs
Skepfyr: I probably should do but I'm not sure what your suggesting, with this change trait impls kinda just work, the methods must take (self, &dyn Db, ...) and it doesn't include the trait name in the generated struct so you can still only have one method with that name. The two options I see:
1. Accept trait impls but include the trait name in the generaated struct to reduce conflicts
2. Ban trait impls to keep things simple until we find a reason that they're needed

salsa-2022-tests/tests/tracked_method.rs
Skepfyr: Yes I was just copying the other tests which appear to all specify the jar type for some reason.

Skepfyr: I've removed it anyway.

components/salsa-2022-macros/src/tracked_fn.rs
Skepfyr: Changed to #[salsa::tracked] and added an error message for if you don't put #[salsa::tracked] on the impl block.

components/salsa-2022-macros/src/tracked_fn.rs
nikomatsakis: Hmm! Interesting. It hadn't occurred to me that they might just work but... I guess ... they probably do, so long as they take a `&dyn crate::Db` as argument? 

I don't see any reason to make them an error, in that case, but can we add a test that shows they work?

nikomatsakis: I don't have a strong opinion about the name of the generated type, but including the trait name seems like a good idea.

nikomatsakis: Maybe `Type_Trait_method_name` ?

components/salsa-2022-macros/src/tracked_fn.rs
nikomatsakis: Great! Can you add a test in the `salsa-2022-tests/tests/compile-fail` directory to show the error cases? You can use the [trybuild workflow](https://github.com/dtolnay/trybuild#workflow) to generate the stderr files.

components/salsa-2022-macros/src/tracked_fn.rs
nikomatsakis: (request: add a test showing that trait impls work, adjust name of generated type)

components/salsa-2022-macros/src/tracked_fn.rs
nikomatsakis: (request: add an error test into salsa-2022-tests/tests/compile-fail)

nikomatsakis (changes requested):
Left two requests! (I'm just leaving this review to help me stay on top of the status of each PR).

This is coming along really nicely.

components/salsa-2022-macros/src/tracked_fn.rs
Skepfyr: Done. Although I did notice this also works:
```rust
#[salsa::jar(db = Db)]
struct Jar(
    MyInput,
    MyInput_TrackedTrait_tracked_trait_fn,
);

trait TrackedTrait<T> {
    fn tracked_trait_fn(self, db: &dyn Db, val: T) -> T;
}

#[salsa::tracked]
impl TrackedTrait<u32> for MyInput {
    #[salsa::tracked]
    fn tracked_trait_fn(self, db: &dyn Db, val: u32) -> u32 {
        self.field(db) * val
    }
}
```
I think it's fine but could require breaking changes in the future if we want to support multiple generic impls in the same module in the future.

components/salsa-2022-macros/src/tracked_fn.rs
Skepfyr: I've added a simple test for the error message I added.

nikomatsakis:
neat :) 

bors r+

Skepfyr:
Note I edited the PR description so the merge commit was more accurate and the PR is now linked to the issue it fixes.

XFFXFF:
bors single on

XFFXFF:
bors retry

XFFXFF:
Oh, I see, there is a change request, please approve it first and then `bors r+` @nikomatsakis

nikomatsakis (approved):

PR #393, opened 2022-08-28
Allow "constant" tracked functions

### LINKED ISSUE #323: Consider some kind of "constants"
In porting dada over to the new version, I found that we were using memoized functions as a kind of simple constant:

```rust
#[salsa::memoized]
fn keywords(db: &dyn crate::Db) -> Map<...> { ... }
```

the idea was that we needed to create some constants that made use of interning etc, and this was convenient.

In the new version, I made every tracked function require some kind of salsa struct. The idea for this requirement was that we would be able to tell when to delete the memoized values (when that salsa struct goes away). But as a side-effect that removed the idea of 0-argument functions.

You can still emulate them like by creating a salsa interned unit struct:

```rust
fn keywords(db: &dyn crate::Db) -> Map<...> {
    _keywords(db, Dummy::new(db))
}

#[salsa::interned]
struct Dummy { }
#[salsa::memoized]
fn keywords(db: &dyn crate::Db) -> Map<...> { ... }
```

but that's annoying. 

Maybe we should add them back? In that case, functions could either (a) have just a `db` parameter or (b) have a salsa struct as their next parameter (plus maybe others).

Skepfyr:
Fixes #323.

This adds support for tracked functions with only a database as input,
that is, it does not take a salsa struct.

I'm not entirely convinced by this, it feels like it works somewhat accidentally and may be fragile to changes. I'm happy if this just get closed as I was mostly playing around to see how this worked.

This change has the odd side-effect of making this code work:
```rust
#[salsa::tracked]
fn tracked_fn(db: &dyn Db, unit: ()) -> u32 {
    44
}

```

### DIFF
--- a/components/salsa-2022-macros/src/tracked_fn.rs
+++ b/components/salsa-2022-macros/src/tracked_fn.rs
@@ -26,7 +26,7 @@ pub(crate) fn tracked_fn(
     }
 
     if let Some(s) = &args.specify {
-        if requires_interning(&item_fn) {
+        if function_type(&item_fn) == FunctionType::RequiresInterning {
             return Err(syn::Error::new(
                 s.span(),
                 "tracked function takes too many arguments to have its value set with `specify`",
@@ -309,11 +309,17 @@ fn configuration_struct(item_fn: &syn::ItemFn) -> syn::ItemStruct {
     let visibility = &item_fn.vis;
 
     let salsa_struct_ty = salsa_struct_ty(item_fn);
-    let intern_map: syn::Type = if requires_interning(item_fn) {
-        let key_ty = key_tuple_ty(item_fn);
-        parse_quote! { salsa::interned::InternedIngredient<salsa::Id, #key_ty> }
-    } else {
-        parse_quote! { salsa::interned::IdentityInterner<#salsa_struct_ty> }
+    let intern_map: syn::Type = match function_type(item_fn) {
+        FunctionType::Constant => {
+            parse_quote! { salsa::interned::IdentityInterner<()> }
+        }
+        FunctionType::SalsaStruct => {
+            parse_quote! { salsa::interned::IdentityInterner<#salsa_struct_ty> }
+        }
+        FunctionType::RequiresInterning => {
+            let key_ty = key_tuple_ty(item_fn);
+            parse_quote! { salsa::interned::InternedIngredient<salsa::Id, #key_ty> }
+        }
     };
 
     parse_quote! {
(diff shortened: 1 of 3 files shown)


vemoo:
> This change has the odd side-effect of making this code work:
> 
> ```rust
> #[salsa::tracked]
> fn tracked_fn(db: &dyn Db, unit: ()) -> u32 {
>     44
> }
> ```

This seems to work now
```rust
#[salsa::tracked(jar = Jar)]
fn intermediate_result(db: &dyn Db, input: MyInput, unit: ()) -> u32 {
  ...
}
```
so it doesn't seem to be related to this PR.

nikomatsakis (commented):
I think you're correct that it works "by accident" to some extent -- what I *think* winds up happening is that we create an interning table and intern '()' to create the input key. It's inefficient but it ought to *work*. It's probably ok for now, it might be nice to do a deeper edit at some point, I'm not sure how hard that would be.

nikomatsakis:
Oh, hmm, I might be wrong about what I said. Actually, I think this is not using interning and is probably reasonably efficient. The trick is this `impl` that I added at some point (I think anticipating exactly this branch):

https://github.com/salsa-rs/salsa/blob/657b85682b5371ce7b5ef5697ef0fcd96351406d/components/salsa-2022/src/id.rs#L82-L92

That, combined with the fact that `requires_interning` is false:

https://github.com/salsa-rs/salsa/blob/657b85682b5371ce7b5ef5697ef0fcd96351406d/components/salsa-2022-macros/src/tracked_fn.rs#L122-L125

means that we will expect the input (`()`) to be an actual `salsa::Id`, and in this case it will always be zero.

There's *still* room to make things more efficient -- e.g., we're using a whole hashmap that will only ever have a single key -- but I think it's "good enough".

nikomatsakis:
Ah, I guess that this...

```rust
#[salsa::tracked]
fn tracked_fn(db: &dyn Db, unit: ()) -> u32 {
    44
}
```

...works because we are using `()` as the salsa struct? 

Hmm. That doesn't feel great. We could probably fix this by introducing a `Singleton` unit type (e.g., `salsa::id::Singleton`) and using that as the `SalsaKey` instead of `()`.

What do you think, @Skepfyr ?

Skepfyr:
> Hmm. That doesn't feel great. We could probably fix this by introducing a `Singleton` unit type (e.g., `salsa::id::Singleton`) and using that as the `SalsaKey` instead of `()`

This doesn't trivially work, `()` has the benefit that it is the empty tuple which means the existing code to intern a tuple of all the arguments naturally produces it. Doing better would require some deeper changes and at the point it's probably worth doing it more properly.

I briefly thought that I could introduce a `ConstantIngredient` but `report_untracked_read` means that you need (most of the machinery of) a `FunctionIgredient`.

nikomatsakis (approved):

nikomatsakis:
Thanks!

bors r+

nikomatsakis:
Oh, looks like there are merge conflicts.

PR #395, opened 2022-08-30
reduce amount of generated code for `DebugWithDb`

vemoo:
Extract common code that was being repeated for each field of each salsa handled struct to `components/salsa-2022/src/debug.rs` module `helper`.

### DIFF
--- a/components/salsa-2022-macros/src/salsa_struct.rs
+++ b/components/salsa-2022-macros/src/salsa_struct.rs
@@ -284,41 +284,33 @@ impl<A: AllowedOptions> SalsaStruct<A> {
         let db_type = self.db_dyn_ty();
         let ident_string = ident.to_string();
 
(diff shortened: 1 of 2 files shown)


huang12zheng (commented):
This may also be an option(just about clone):
Fallback remove info about Debug,
and add Debug to method
```rs
trait Fallback<Db>
where
    Db: ?Sized,
{
    fn match_data_type<'a, 'b: 'a, T: Clone>(a: &T, db: &'b Db) -> T {
        a.clone()
    }
}

impl<T, Db> Fallback<Db> for T
where
    T: ?Sized,
    Db: ?Sized,
{
}
```

nikomatsakis (approved):
Nice.

nikomatsakis:
@huang12zheng I didn't quite understand your suggestion-- seems like it has to do with `clone`, not with `debug` generation

PR #396, opened 2022-08-31
Extend salsa 2022 tutorial to cover type-checking, evaluation

### LINKED ISSUE #367: Extend calc example, tutorial with type checker

### LINKED ISSUE #368: Extend calc example, tutorial with interpreter

nikomatsakis:
This (wip) PR completes the tutorial:

* [x] Extend code to cover and test type-checking and evaluation, demonstrating incremental re-use
* [x] Update tutorial to cover changes to the jar/database
* [ ] Update tutorial to explain anchored spans during parsing and how they enable re-use even when whitespace changes
* [ ] Update tutorial to explain when and where to use tracked structs vs interned structs vs regular Rust structs
* [ ] Update tutorial to explain type-checking
* [ ] Update tutorial t0 explain evaluation

You can view the tutorial in the [netlify deploy preview](https://630f31397f2c8800085c06e3--salsa-rs.netlify.app/tutorial).

Fixes #367 
Fixes #368 

cc @seanchen1991 and other @salsa-rs/contributors -- I'd like to get some feedback, either on the writing or the code. 

( Also, there is some odd behavior I uncovered in writing the evaluation tests, where some things seem to execute more than once even though they ought to be memoized. I don't know what's up with that, could be a bug or an artifact of the logs. )


XFFXFF (commented):
Can we start with `IR`, `parser`, `type checker` and then move on to `Jar` and `Database`? Although I know we need `Db` and `jar` when implementing `IR`, `parser`, etc. My concern is that the reader may be confused by the previous chapters, but there is actually little progress in implementing the `calc language` itself.
I'm not sure, just an idea

calc-example/calc/src/ir.rs
```diff
@@ -90,19 +99,80 @@ pub struct Function {
 // ANCHOR_END: functions
 
 #[salsa::tracked]
+pub(crate) fn find_function(
```
XFFXFF, line 102: There is alread a tracked `find_function` in [`type_check.rs`](https://github.com/salsa-rs/salsa/blob/657b85682b5371ce7b5ef5697ef0fcd96351406d/calc-example/calc/src/type_check.rs#L26-L35), which will make the reuse of test `evaluate_example` look not working as expected ?

nikomatsakis: Ah, *that* explains why I was seeing it in the log twice. Thanks!

book/src/tutorial/structure.md
```diff
@@ -77,7 +77,8 @@ which is a bit less intuitive but allows for more incremental re-use.
 The idea is to compute, for a given expression, which variables it references.
 Then there is a function `check` which ensures that those variables are a subset of those that are already defined.
 
-## Interpreter
+## Evaluator
```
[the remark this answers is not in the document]
nikomatsakis: No, I guess not. I just like the word evaluator. I can change it back.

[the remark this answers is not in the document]
nikomatsakis: Ah, I guess the reason I change it was because I had named the modules in a "verb-like" fashion and `interpret` felt strange to me.

book/src/tutorial/jar.md
```diff
@@ -42,10 +42,19 @@ The database trait for our `calc` crate is very simple:
 When you define a database trait like `Db`, the one thing that is required is that it must have a supertrait `salsa::DbWithJar<Jar>`,
 where `Jar` is the jar struct. If your jar depends on other jars, you can have multiple such supertraits (e.g., `salsa::DbWithJar<other_crate::Jar>`).
```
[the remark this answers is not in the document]
nikomatsakis: Improved, I think, per @XFFXFF's suggestion. Take a look?

calc-example/calc/src/type_check.rs
```diff
@@ -72,14 +61,12 @@ impl CheckExpression<'_> {
...
-        find_function(self.db, self.program, f)
+        self.program.find_function(self.db, f)
     }
 
     fn report_error(&self, span: Span, message: String) {
-        Diagnostics::push(
-            self.db,
-            Diagnostic::new(span.start(self.db), span.end(self.db), message),
-        );
+        let (start, end) = span.absolute_locations(self.db, self.anchor);
```
Skepfyr, line 68: Doesn't resolving the anchors here break the reuse? You're going to have to re-execute this function if the source code changes and you re-collect the Diagnostics to get the updated line numbers.I'd expect the Diagnostics to contain spans, and them to be resolved after the call to `accumulated`.

calc-example/calc/src/ir.rs
```diff
@@ -90,19 +104,169 @@ pub struct Function {
...
+/// Spans are not stored as absolute values but rather relative to some enclosing anchor
+/// (some struct that implements the `Anchor` trait).
+/// This way, although the location of the anchor may change, the spans themselves rarely do.
+/// So long as a function doesn't convert the span into its absolute form,
+/// and thus read the anchor's precise location, it won't need to re-execute, even if the anchor
+/// has moved about in the file.
+///
+/// **NB:** It is your job, when converting the span into relative positions,
+/// to supply the correct anchor! For example, the anchor for the expressions
+/// within a function body is the function itself.
```
Skepfyr, line 219: This feels really awkward and I'd hope we could support something better. I wonder if something like this might work:
```rust
#[tracked]
struct Anchor {
    #[id]
    name: String,
    parent: Option<Anchor>,
    offset: Offset,
}

struct Span {
    anchor: Anchor,
    start: Offset,
    end: Offset,
}
```

Skepfyr (commented):
Looks mostly sensible to me. The biggest thing I'm worried about is whether the anchor system works properly. I've left a couple of comments detailing the worry and a suggestion.

calc-example/calc/src/ir.rs
nikomatsakis: I considered that setup. 

There are a few problems that would lead me to avoid it in a real compiler:

* The `Span` is a very common data structure and it now occupies an additional 32 bits, at least. Ideally you'd have spans be super tiny so you can have them all over the place (even the current setup, with two offsets, is kinda large).
* The anchor winds up duplicating parts of the things it anchors (e.g., the name) which makes the code a bit more awkward.

The latter one is not so bad, but the former pushed me over the edge and made me happier with this setup. 

Thoughts?

calc-example/calc/src/type_check.rs
nikomatsakis: It will break the re-use, but given that the function is producing errors, that seems fine to me. Usually only a small minority of the code is producing errors at any point in time, and it's likely being actively edited.


The reason that I prefer this route is that the diagnostics now represent data that can be consumed wiithout reading into the database. My general preference is that the inputs/outputs of the program don't contain any tracked structs. If they do, you run the risk of modifying an input and then reading a tracked struct from the previous revision. Although I guess that's not necessarily worse than reading any output from the past revision.

book/src/tutorial/ir.md
```diff
@@ -111,13 +137,52 @@ Apart from the fields being immutable, the API for working with a tracked struct
...
 To get better reuse across revisions, particularly when things are reordered, you can mark some entity fields with `#[id]`.
 Normally, you would do this on fields that represent the "name" of an entity.
 This indicates that, across two revisions R1 and R2, if two functions are created with the same name, they refer to the same entity, so we can compare their other fields for equality to determine what needs to be re-executed.
 Adding `#[id]` attributes is an optimization and never affects correctness.
 For more details, see the [algorithm](../reference/algorithm.md) page of the reference.
+(If you're familiar with React, the `#[id]` fields serve the same role as ["key" fields on the React dom](https://reactjs.org/docs/reconciliation.html#recursing-on-children).)
+
+### `Anchor`: Significant points in the program text
+
+The final tracked struct is called `Anchor`:
```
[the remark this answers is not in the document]
nikomatsakis: Yeah, the text is out of date from the code.

components/salsa-2022/src/debug.rs
```diff
@@ -160,6 +173,34 @@ where
     }
 }
 
+// These impls are here for convenience, it's not really a scalable sol'n.
```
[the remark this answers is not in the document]
nikomatsakis: I'm actually not sure why I added those impls now -- the "deriving" code made use of a good trick in this respect to get around the problem. I should try removing the impls and see what breaks.

calc-example/calc/src/type_check.rs
Skepfyr: I'm mostly convinced by the errors are probably being actively edited argument. It's a trade-off between more reuse in the face of lots of errors and everything being slightly slower because the spans are more complicated.

I wasn't imagining exposing the tracked structs in the program io, just that the thing accumulating the diags does the resolution rather than pushing the diags. I agree that exposing tracked structs is bad, I think that using an old tracked struct could cause panics or completely invalid data if it was removed or replaced in the intern table.

calc-example/calc/src/ir.rs
Skepfyr: You could always intern them, but I can see that being unacceptable slow for a real compiler. I've been unable to come up with a nicer solution so I guess I'm happiest with this setup.

gavrilikhin-d:
What's the status of this PR? This seems to be a huge improvement to documentation. I can help you to finalise it, if you want

nikomatsakis:
I'm just going to rewrite this from scratch.

PR #399, opened 2022-08-31
Make getters and setters have the visibility of their field

### LINKED ISSUE #398: Visibility of accessors is tied to struct, not field
I noticed something weird:

```rust
#[salsa::input]
pub struct Foo {
    bar: u32
}
```

generates a `bar(db)` getter with `pub` access! I think the access should be the same as field (so priv here). I would expect to have to write

```rust
#[salsa::input]
pub struct Foo {
    pub bar: u32
}
```

if I want a `pub` setter/getter.

Same for tracked structs and interned structs.

MihailMihov:
resolves #398 

Currently `#[salsa::input]` and `#[salsa::tracked]` structs made all getters and setters public, regardless of the visibility of the field and the getters for `#[salsa::interned]` had the visibility of the struct itself. This pull request changes the generated impl's to use the visibility of each field for it's getters and setters.

### DIFF
--- a/calc-example/calc/src/ir.rs
+++ b/calc-example/calc/src/ir.rs
@@ -7,7 +7,7 @@ use ordered_float::OrderedFloat;
 #[salsa::input]
 pub struct SourceProgram {
     #[return_ref]
-    text: String,
+    pub text: String,
 }
 // ANCHOR_END: input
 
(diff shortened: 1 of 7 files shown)


nikomatsakis (changes requested):
Can you add some compile-fail tests that demonstrate this is working?

nikomatsakis:
To do that, you add a test in this directory https://github.com/salsa-rs/salsa/tree/master/salsa-2022-tests/tests/compile-fail and then follow the [trybuild workflow](https://github.com/dtolnay/trybuild#workflow) to bless the stderr output.

nikomatsakis (approved):

Issue #354, opened 2022-08-18
Exhaustive tests for macros and the options you can and cannot use at each place (the calc example uses only the "main" route, e.g. when jar is not specified explicitly)


nikomatsakis:

nikomatsakis:
@XFFXFF added the `lru` option to tracked functions, but we should test that it cannot be used with `specify` somehow.

XFFXFF:
https://github.com/dtolnay/trybuild.git  
I found this crate useful and easy to use to test for compiling failures. I can add a pr as an example to show how it works so you can see if this is the right way to go

PR #401, opened 2022-09-02
Support singleton inputs #337

### LINKED ISSUE #337: Support singleton inputs

OLUWAMUYIWA:
## What?
In #337, it  was requested `Singleton` input option be made available for Input structs with a single field. This PR addresses that

## How?
I basically followed the mentoring instructions made available.

Fixes #337


components/salsa-2022-macros/src/salsa_struct.rs
```diff
@@ -25,41 +25,46 @@
...
     configuration,
+    modes::AllowedModes,
     options::{AllowedOptions, Options},
 };
+use heck::ToUpperCamelCase;
+use proc_macro2::{Ident, Literal, Span, TokenStream};
+use syn::spanned::Spanned;
 
-pub(crate) struct SalsaStruct<A: AllowedOptions> {
+pub(crate) struct SalsaStruct<A: AllowedOptions, M: AllowedModes> {
```
DropDemBits, line 38: You can get away with not adding an extra bound if you do something like this:
```suggestion
pub(crate) struct SalsaStruct<A: AllowedOptions + AllowedModes> {
```

Though, I don't think AllowedModes is necessary right now, since it's already rejected by option parsing before we even get to `check_singleton`

components/salsa-2022-macros/src/input.rs
```diff
@@ -73,8 +87,16 @@ impl InputStruct {
     }
 
     fn validate_input(&self) -> syn::Result<()> {
+        // check for dissalowed fields
         self.disallow_id_fields("input")?;
 
+        // check if an input struct labeled singleton truly has one field
+        if self.0.is_isingleton() && self.0.num_fields() != 1 {
```
XFFXFF, line 94: Why can't we support multiple fields here?

components/salsa-2022-macros/src/salsa_struct.rs
OLUWAMUYIWA: Oh yeah. True. Thanks!

About the second part, It's not exactly rejected by option parsing. Option parsing still allows any salsa struct to be `singleton`. I tried doing the validation there, but apparently the `syn::parse::ParseStream` supplied to the `stn::parse` trait only contains the args to the macro.

XFFXFF (commented):
We also need some compile-fail tests to test `singleton` is not allowed on `interned` and `tracked struct`

UPDATE: you can follow [this](https://github.com/salsa-rs/salsa/issues/354#issue-1342908047) to add some compile-fail tests

components/salsa-2022-macros/src/input.rs
OLUWAMUYIWA: I'll work on the tests. I assumed wrongly that Singleton structs should have only one field. I see that it's wrong now. Will remove it.
Thanks!

components/salsa-2022-macros/src/input.rs
```diff
@@ -73,8 +87,16 @@ impl InputStruct {
     }
 
     fn validate_input(&self) -> syn::Result<()> {
+        // check for dissalowed fields
```
nikomatsakis, line 90:
```suggestion
        // check for disallowed fields
```

components/salsa-2022-macros/src/input.rs
nikomatsakis: I agree the limitation is not needed. Sounds good.

components/salsa-2022-macros/src/salsa_struct.rs
nikomatsakis: I'm a bit confused by this. The option parsing for the struct itself would be e.g. to parse things like `#[salsa::input(jar = Foo)]`, so it seems like it ought to also be usable to parse the singleton flag?

nikomatsakis (changes requested):
Cool! Some suggestions.

components/salsa-2022-macros/src/salsa_struct.rs
OLUWAMUYIWA: I'm the confused one. Will work on it. Thanks

components/salsa-2022-macros/src/input.rs
OLUWAMUYIWA: Thanks for the comment and recommendation

components/salsa-2022-macros/src/salsa_struct.rs
OLUWAMUYIWA: @DropDemBits  I was wrong. I could just use `AllowedOptions` and do away with `AllowedModes` altogether. 
Thanks!

nikomatsakis (approved):
Nice! This looks good.

nikomatsakis:
```
error[E0046]: not all trait items implemented, missing: `SINGLETON`
Error:    --> components/salsa-2022-macros/src/tracked_fn.rs:163:1
    |
163 | impl crate::options::AllowedOptions for TrackedImpl {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing `SINGLETON` in implementation
    |
   ::: components/salsa-2022-macros/src/options.rs:91:5
    |
91  |     const SINGLETON: bool;
    |     ---------------------- `SINGLETON` from trait

For more information about this error, try `rustc --explain E0046`.
error: could not compile `salsa-2022-macros` due to previous error
Error: warning: build failed, waiting for other jobs to finish...
Error: The process '/home/runner/.cargo/bin/cargo' failed with exit code 101
```

OLUWAMUYIWA:
Fixed

PR #404, opened 2022-09-03
On-demand input

### LINKED ISSUE #394: Support on-demand inputs
This is kinda possible to implement in salsa-2022, but it would be nice to have better support. I'm not sure what that would look like though.

Janky implementation: https://gist.github.com/Skepfyr/03648830d4171e604898442b6fa0f105

Skepfyr:
Fixes #394 

This is not yet done, but I'd like to get some feedback on the approach.
I think these things are remaining:
- [x] Clean up some now incorrect comments in `input_field.rs`
- [x] ~Update calc-example to use on-demand inputs, and remove the lazy-input example (this is quite fun as you end up with an interactive calculator)~
- [x] Update the book


lazy-input/src/main.rs
```diff
@@ -0,0 +1,143 @@
...
+        if let salsa::EventKind::WillExecute { .. } = event.kind {
+            self.logs
+                .lock()
+                .unwrap()
+                .push(format!("Event: {:?}", event.debug(self)));
+        }
+    }
+}
+
+impl Db for Database {
```
nikomatsakis, line 91: I feel like we should create a `salsa_fs_watch` create that people can use to manage this for them. But that can come later.

nikomatsakis (changes requested):
Yeah, I'm convinced. This is very cool.

I want to land the PR, but I'd like to move the `lazy-input` example into a test in `salsa-2022-tests`, or perhaps it can be named `lazy-input-example` and live alongside `calc-example`?

nikomatsakis:
BTW, I am not convinced we should update the calc-example to use lazy-inputs. I feel like it will make that example more complicated than it has to be, and that it might be better to separate out the two.

lazy-input/src/main.rs
Skepfyr: Maybe, I suspect that the driving code will end up pretty domain specific though. For example with LSP servers, the LSP client can take control of a file and so you need to swap from watching the filesystem to watching the LSP message stream for file change notifications.

nikomatsakis (approved):

nikomatsakis:
This is really cool, @Skepfyr. I feel like it's a nice story now.


bors r+

Skepfyr:
That rebase got slightly out-of-hand, what I ended up doing was requiring that singleton inputs take a `&mut db` to their new, because it might modify an existing value. This means singleton's can't be on-demand but that seems fine.

nikomatsakis (approved):

nikomatsakis:
Argh, more rebase conflicts:

> calc-example/calc/src/parser.rs

nikomatsakis:
Filed https://github.com/salsa-rs/salsa/issues/415 for the question of singleton behavior

Skepfyr:
That rebase went through fine, so I'll be slightly cheeky and see if I can

bors r=nikomatsakis

Issue #397, opened 2022-08-31
Limit DebugWithDb impls to avoid generating false dependencies


nikomatsakis:
Hmm, so @zjp-CN added DebugWithDb impls in #369, but I'm beginning to think it might be good to make them a bit more limited.

The current impls for tracked structs, for example, dump the values of all fields. But this means that salsa will consider the function that uses them to read all the tracked fields, which means the function will re-execute more than you might like. That seems bad, and also likely to cause "heisen-bugs", where adding a `{x:?}` print-out to a function changes the behavior. It also seems like it might cause problems for #386 or other attempts to debug salsa's behavior.

An alternative is to limit the fields that get printed -- for example, we could only print out the `#[id]` fields for a tracked struct, since those are part of its identity. Following this model, for interned structs, we can print all the fields, but for inputs we would print only the salsa-id.

Another option would be to add some form of "opt-in", e.g. by adding a method like `debug_all` in the `DebugWithDb` trait that dumps out *all* the fields (and a corresponding `fmt_all` method that gets invoked; it might, by default, simply call `fmt`).

Finally, I think it would be nice to let users disable or control the automatic `DebugWithDb` impl.

vemoo:
When I tried to update https://github.com/dada-lang/dada to latest salsa-2022 I run against this issue, so I'd like to implement it.

To disable the `DebugInDb` the idea I had is to add an `no_debug_with_db` option to use like `#[salsa::interned(no_debug_with_db)]`.

nikomatsakis:
@vemoo yeah, that was part of why I noticed it too :) 

In terms of deciding whether or not to derive `DebugWithDb` *at all*, the other way to do it would be to disable the *automatic* `DebugWithDb` impls and instead search for something like

```rust
#[salsa::input]
#[derive(DebugWithDb)]
```

...that seems more consistent with Rust, but also kind of annoying.

vemoo:
For opt-in another option could be `#[salsa::input(derive(DebugWithDb))]`

PoignardAzur:
Thinking out loud: another way to achieve this could be to have some sort of "defer" method, eg:

```rust
db.defer(|db| {
    eprintln!("my_value = {:?}", my_value.debug(db));
    eprintln!("my_foo = {:?}", my_value.foo(db));
});
```

Because the code would be guaranteed to run after the query, salsa would know it doesn't influence the query's results.

PoignardAzur:
[...]

PR #405, opened 2022-09-04
Include only identity fields by default in `DebugWithDb::debug` and add `DebugWithDb::debug_all`

vemoo:
This addresses a couple of items of #397 

I initially added a separate `DebugWithDb::fmt_all` method but then changed it to an extra parameter to make it easier to propagate to inner fields. Also tried introducing something like:
```rust
pub struct FormatterWithDb<'me, 'f, Db: ?Sized>{
  fmt: &'me mut std::fmt::Formatter<'f>,
  db: &'me Db,
  include_all_fields: bool
}
```
to pass to `fmt`, but I gave up because the lifetimes got a bit unwieldy.

### DIFF
--- a/components/salsa-2022-macros/src/salsa_struct.rs
+++ b/components/salsa-2022-macros/src/salsa_struct.rs
@@ -97,14 +97,11 @@ impl<A: AllowedOptions> SalsaStruct<A> {
         self.fields.iter()
     }
 
-    /// Iterator over fields that are part of structs identity.
-    ///
-    /// If this is an input, empty iterator.
-    pub(crate) fn identity_fields(&self) -> Vec<&SalsaField> {
+    pub(crate) fn is_identity_field(&self, field: &SalsaField) -> bool {
         match self.kind {
-            SalsaStructKind::Input => Vec::new(),
-            SalsaStructKind::Tracked => self.all_fields().filter(|f| f.has_id_attr).collect(),
-            SalsaStructKind::Interned => self.all_fields().collect(),
+            SalsaStructKind::Input => false,
+            SalsaStructKind::Tracked => field.has_id_attr,
+            SalsaStructKind::Interned => true,
         }
     }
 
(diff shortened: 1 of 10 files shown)


components/salsa-2022-macros/src/salsa_struct.rs
```diff
@@ -89,6 +99,14 @@ impl<A: AllowedOptions> SalsaStruct<A> {
         self.fields.iter()
     }
 
+    pub(crate) fn is_identity_field(&self, field: &SalsaField) -> bool {
+        match self.kind {
+            SalsaStructKind::Input => false,
```
nikomatsakis, line 104: This is correct, but it's going to be annoying in practice and I'm not sure what to do about it. :/

I'm imagining that e.g. you have `InputFile { path: String, contents: String }`, and you would like to include `path` when you print out the input-file, so you don't just see `InputFile(3)`. 

I wonder if it makes sense to have the ability to mark input fields as `#[id]` and then include them in the printouts (and maybe even make them immutable?). It seems like it might be confusing to do that since, in fact, they are not part of the "identity" of the input.

nikomatsakis (commented):
The code looks good. I'm just debating on what exact behavior we really want.

components/salsa-2022-macros/src/salsa_struct.rs
vemoo: Maybe the method can be renamed to `is_stable_field`, `is_immutable_field` or `is_non_changing_field` and another attribute can be added like `#[stable]`, `#[immutable]` or `#[meta]`.

nikomatsakis: Why don't we file a bug about this problem. I think we need some way to tag fields in a `salsa::input` and have them show up, and I think those fields should be immutable. The `#[id]` attribute feels nice to me in that we already use it on `salsa::tracked` for a "similar in spirit" sort of meaning -- if this field were to change, it'd be a different struct, so to speak. So e.g. I can imagine an `InputFile` might mark the path as `#[id]` to indicate that it just doesn't change (and if it did, we'd make a new input).

But it's also confusing in that, unlike with tracked structs, there is no "id" in the same way.

components/salsa-2022/src/debug.rs
```diff
@@ -12,6 +13,32 @@ pub trait DebugWithDb<Db: ?Sized> {
...
+            value: BoxRef::Ref(self),
+            db,
+            include_all_fields,
+        }
+    }
+
+    /// Be careful when using this method inside a tracked function,
+    /// because the default macro generated implementation will read all fields,
+    /// maybe introducing undesired dependencies.
+    fn debug_all<'me, 'db>(&'me self, db: &'me Db) -> DebugWith<'me, Db>
```
nikomatsakis, line 34: Question -- do we want *want* debug-all? I'll file an issue to discuss.

nikomatsakis (approved):

components/salsa-2022-macros/src/salsa_struct.rs
nikomatsakis: Filed https://github.com/salsa-rs/salsa/issues/410

components/salsa-2022/src/debug.rs
nikomatsakis: https://github.com/salsa-rs/salsa/issues/411

nikomatsakis:
@vemoo looks like this needs a rebase!

vemoo:
rebased

nikomatsakis (approved):

nikomatsakis:
Merge conflict again :(

@vemoo you should be able to do `bors r=nikomatsakis` afterwards, btw

nikomatsakis:
bors delegate

vemoo:
bors r=nikomatsakis

PR #406, opened 2022-09-05
no need to generate a config struct for each field of `salsa::input`

XFFXFF:
https://github.com/salsa-rs/salsa/pull/342 added `InputFieldIngredient` which replaces the `FunctionIngredient` for `salsa::input`, so we don't need to generate a config struct and `salsa::function::Configuration` impl for each field, beacuse they are only needed when we use `FunctionIngredient`


nikomatsakis:
Oh, nice catch.

bors r+

Issue #386, opened 2022-08-26
`debug` method for `DependencyKeyIndex` should leverage `DebugWithDb` 


nikomatsakis:
In #369, @zjp-CN added a debug method for `DependencyKeyIndex` that is (more) human-readable. In #378, @agluszak improved the `DebugWithDb` impl for salsa structs. We should put these together! Right now if you have a struct like

https://github.com/salsa-rs/salsa/blob/1363d78fe5fa85a7f92ee03ad366559b983454b2/salsa-2022-tests/tests/deletion-cascade.rs#L38-L41

and a function like

https://github.com/salsa-rs/salsa/blob/1363d78fe5fa85a7f92ee03ad366559b983454b2/salsa-2022-tests/tests/deletion-cascade.rs#L51-L55

you wind up with output like

https://github.com/salsa-rs/salsa/blob/1363d78fe5fa85a7f92ee03ad366559b983454b2/salsa-2022-tests/tests/deletion-cascade.rs#L124

but it would be more readable if we got `contribution_from_struct(MyTracked { salsa::id: 2 })`.

Another case to consider are functions with multiple inputs. They'll be touch tricker because they first intern those inputs.

nikomatsakis:
Brief mentoring instructions:

The code that generates the output in question is here:

https://github.com/salsa-rs/salsa/blob/1363d78fe5fa85a7f92ee03ad366559b983454b2/components/salsa-2022/src/function.rs#L276-L278

The simplest fix would be to modify the `fmt_index` helper function:

https://github.com/salsa-rs/salsa/blob/1363d78fe5fa85a7f92ee03ad366559b983454b2/components/salsa-2022/src/ingredient.rs#L68-L79

so that instead of taking a `salsa::Id` it took a `Option<impl DebugWIthDb>`, and then (in the function ingredient) to convert the `salsa::Id` into a `C::Key` using this method:

https://github.com/salsa-rs/salsa/blob/1363d78fe5fa85a7f92ee03ad366559b983454b2/components/salsa-2022/src/function.rs#L118-L122

This will work for the example above, but it won't work for a function with multiple inputs. To handle *that* case... hmm.... that will be harder, because the interning actually takes place outside of the function ingredient. The macro creates an interner:

https://github.com/salsa-rs/salsa/blob/1363d78fe5fa85a7f92ee03ad366559b983454b2/components/salsa-2022-macros/src/tracked_fn.rs#L376

[...]

PR #408, opened 2022-09-07
Exhaustive macros test

OLUWAMUYIWA:
### What
Used `trybuild` to verify that incompatible options with salsa structs will not compile

Ref: #354 

### How:
I created a `.rs` file for each salsa struct. I decided not to create one for each incompatible option, as this would only greatly increase the number of files.


OLUWAMUYIWA:
@nikomatsakis  please let me know if i left any check out. 
Also, I asked a question under #386 , please check.
Finally, I made changes to PR #337, as requested. Please check.
Thanks!

nikomatsakis (approved):
Thank you! This looks good to me. I didn't *see* anything missing, we can always add things later.

PR #412, opened 2022-09-12
Fix race condition in interned data

Skepfyr:
In some scenarios the `InternedIngredient` could return two different ids for the same data if called in parallel.

### DIFF
--- a/components/salsa-2022/src/interned.rs
+++ b/components/salsa-2022/src/interned.rs
@@ -83,23 +83,28 @@ where
             self.reset_at,
         );
 
+        // Optimisation to only get read lock on the map if the data has already
+        // been interned.
         if let Some(id) = self.key_map.get(&data) {
             return *id;
         }
 
-        loop {
-            let next_id = self.counter.fetch_add(1);
-            let next_id = Id::from_id(crate::id::Id::from_u32(next_id));
-            match self.value_map.entry(next_id) {
-                // If we already have an entry with this id...
-                dashmap::mapref::entry::Entry::Occupied(_) => continue,
-
-                // Otherwise...
-                dashmap::mapref::entry::Entry::Vacant(entry) => {
-                    self.key_map.insert(data.clone(), next_id);
-                    entry.insert(Box::new(data));
-                    return next_id;
-                }
+        match self.key_map.entry(data.clone()) {
+            // Data has already been interned, just use the existing ID
+            dashmap::mapref::entry::Entry::Occupied(entry) => *entry.get(),
+            // Insert the new value at the next available id
+            // This holds a lock on the entry so other can't attempt to insert
+            // at the same time.
+            dashmap::mapref::entry::Entry::Vacant(entry) => {
+                let next_id = self.counter.fetch_add(1);
+                let next_id = Id::from_id(crate::id::Id::from_u32(next_id));
+                let old_value = self.value_map.insert(next_id, Box::new(data));
+                assert!(
+                    old_value.is_none(),
+                    "next_id is guaranteed to be unique, bar overflow"
+                );
+                entry.insert(next_id);
+                next_id
             }
         }
     }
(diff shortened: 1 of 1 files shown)


components/salsa-2022/src/interned.rs
```diff
@@ -83,23 +83,28 @@ where
...
+        match self.key_map.entry(data.clone()) {
+            // Data has already been interned, just use the existing ID
+            dashmap::mapref::entry::Entry::Occupied(entry) => *entry.get(),
+            // Insert the new value at the next available id
+            // This holds a lock on the entry so other can't attempt to insert
+            // at the same time.
+            dashmap::mapref::entry::Entry::Vacant(entry) => {
+                let next_id = self.counter.fetch_add(1);
+                let next_id = Id::from_id(crate::id::Id::from_u32(next_id));
+                let old_value = self.value_map.insert(next_id, Box::new(data));
```
nikomatsakis, line 101: My concern is that in `delete_index` we hold the locks in the reverse order:

https://github.com/salsa-rs/salsa/blob/61e3bdfc81463e7bf1ea9fa6d3896e3e29b2a7a4/components/salsa-2022/src/interned.rs#L161-L166

As the comment says:

https://github.com/salsa-rs/salsa/blob/61e3bdfc81463e7bf1ea9fa6d3896e3e29b2a7a4/components/salsa-2022/src/interned.rs#L35

If there were a call to `delete_index` occurring in parallel, it could deadlock.

That said, I think `delete_index` might be dead code? I wanted at some point to experiment with removing interned values, or having an LRU-like system, but I don't think we're using it at all now. We could delete it and address the question later.

nikomatsakis (changes requested):

components/salsa-2022/src/interned.rs
Skepfyr: Well spotted, I read that comment but didn't quite internalise what it meant. We do use delete_index to clean up tracked structs, so I've implemented what I hope is a proper fix.

components/salsa-2022/src/interned.rs
```diff
@@ -83,23 +83,39 @@ where
...
-                // If we already have an entry with this id...
-                dashmap::mapref::entry::Entry::Occupied(_) => continue,
-
-                // Otherwise...
-                dashmap::mapref::entry::Entry::Vacant(entry) => {
-                    self.key_map.insert(data.clone(), next_id);
-                    entry.insert(Box::new(data));
-                    return next_id;
-                }
+        let next_id = self.counter.fetch_add(1);
```
nikomatsakis, line 92: So actually I was a bit nervous about the `fetch_add` to begin with. It seems like this could cause "gaps" in the interned values. I guess that might be ok. 

Overall this code feels too complex. I'm pondering a few options.

One of them is to just add a "modification lock" that we have to acquire first and foremost. i.e., something like

```rust
let modification_lock = self.modification_lock.lock();
```

that would mean you can only intern (or delete) one thing at a time, but it would remove the concern about deadlock.

nikomatsakis: Actually, wait. Does `delete_index` really have to be *atomic*? Can't we do something like this?

```rust
let key = self.value_map.remove(id).unwrap();
self.key_map.remove(&key);
self.deleted_entries.push(key);
```

After all, the function invariant specifies that there can be no "This should only be used when you are certain that the given `id` has not (and will not) be used in the current revision." OK, well, that's not strong enough, but I think we could strengthen the invariant to also say this should only be used when...

* you are certain that the given `id` has not been and will not be used in the current revision
* you are certain that the interned key corresponding to `id` will not be interned in this revision

The one place we call this is when deleting a tracked struct, and we know that the second condition holds then.

nikomatsakis: That would then allow us to go back to your original formulation where we hold the lock on key-map whilst inserting into value-map.

nikomatsakis (changes requested):

components/salsa-2022/src/interned.rs
Skepfyr: Fair enough, I've done exactly that

nikomatsakis (approved):

PR #413, opened 2022-09-13
panic when reading fields of tracked structs from older revisions

### LINKED ISSUE #407: Panic when reading fields of tracked structs from older revisions
```rust
use test_log::test;

#[salsa::jar(db = Db)]
struct Jar(MyInput, MyTracked, final_result, intermediate_result);

trait Db: salsa::DbWithJar<Jar> {}

#[salsa::input(jar = Jar)]
struct MyInput {
    field: u32,
}

#[salsa::tracked(jar = Jar)]
struct MyTracked {
    field: u32,
}

#[salsa::tracked(jar = Jar)]
fn final_result(db: &dyn Db, tracked: MyTracked) -> u32 {
    tracked.field(db) * 2
}

#[salsa::tracked(jar = Jar)]
fn intermediate_result(db: &dyn Db, input: MyInput) -> MyTracked {
    MyTracked::new(db, input.field(db) / 2)
}

#[salsa::db(Jar)]
#[derive(Default)]
struct Database {
    storage: salsa::Storage<Self>,
}

impl salsa::Database for Database {}

impl Db for Database {}


#[test]
fn execute() {
    let mut db = Database::default();

    let input = MyInput::new(&mut db, 22);
    let tracked = intermediate_result(&db, input);

    dbg!(final_result(&db, tracked));

    input.set_field(&mut db).to(24);
    dbg!(final_result(&db, tracked));
}
```
it will output
```sh
[salsa-2022-tests/tests/hello_world.rs:46] final_result(&db, tracked) = 22
[salsa-2022-tests/tests/hello_world.rs:49] final_result(&db, tracked) = 22
```
not sure if this is expected behavior, or if we should get
```sh
[salsa-2022-tests/tests/hello_world.rs:46] final_result(&db, tracked) = 22
[salsa-2022-tests/tests/hello_world.rs:49] final_result(&db, tracked) = 24
```

XFFXFF:
fix #407

### DIFF
--- a/components/salsa-2022/src/function/accumulated.rs
+++ b/components/salsa-2022/src/function/accumulated.rs
@@ -65,9 +65,7 @@ impl Stack {
     /// Extend the stack of queries with the dependencies from `origin`.
     fn extend(&mut self, origin: Option<QueryOrigin>) {
         match origin {
-            None
-            | Some(QueryOrigin::Assigned(_))
-            | Some(QueryOrigin::BaseInput) => {}
+            None | Some(QueryOrigin::Assigned(_)) | Some(QueryOrigin::BaseInput) => {}
             Some(QueryOrigin::Derived(edges)) | Some(QueryOrigin::DerivedUntracked(edges)) => {
                 for DependencyIndex {
                     ingredient_index,
--- a/components/salsa-2022/src/function/accumulated.rs
+++ b/components/salsa-2022/src/function/accumulated.rs
@@ -67,8 +67,7 @@ impl Stack {
         match origin {
             None
             | Some(QueryOrigin::Assigned(_))
-            | Some(QueryOrigin::BaseInput)
-            | Some(QueryOrigin::Field) => {}
+            | Some(QueryOrigin::BaseInput) => {}
             Some(QueryOrigin::Derived(edges)) | Some(QueryOrigin::DerivedUntracked(edges)) => {
                 for DependencyIndex {
                     ingredient_index,
--- a/components/salsa-2022/src/function/accumulated.rs
+++ b/components/salsa-2022/src/function/accumulated.rs
@@ -1,7 +1,6 @@
 use crate::{
     hash::FxHashSet,
-    key::DependencyIndex,
-    runtime::local_state::QueryOrigin,
+    runtime::local_state::{EdgeKind, QueryOrigin},
     storage::{HasJar, HasJarsDyn},
     DatabaseKeyIndex,
 };
@@ -67,18 +66,15 @@ impl Stack {
         match origin {
             None | Some(QueryOrigin::Assigned(_)) | Some(QueryOrigin::BaseInput) => {}
             Some(QueryOrigin::Derived(edges)) | Some(QueryOrigin::DerivedUntracked(edges)) => {
-                for DependencyIndex {
-                    ingredient_index,
-                    key_index,
-                } in edges.inputs().iter().copied()
+                for (_, dependency_index) in edges
+                    .input_outputs
+                    .iter()
+                    .filter(|edge| edge.0 == EdgeKind::Input)
+                    .copied()
                 {
-                    if let Some(key_index) = key_index {
-                        let i = DatabaseKeyIndex {
-                            ingredient_index,
-                            key_index,
-                        };
+                    if let Ok(i) = DatabaseKeyIndex::try_from(dependency_index) {
                         if self.s.insert(i) {
-                            self.v.push(i);
+                            self.v.push(i)
                         }
                     }
                 }
--- a/components/salsa-2022/src/function/accumulated.rs
+++ b/components/salsa-2022/src/function/accumulated.rs
@@ -1,6 +1,6 @@
 use crate::{
     hash::FxHashSet,
-    runtime::local_state::{EdgeKind, QueryOrigin},
+    runtime::local_state::QueryOrigin,
     storage::{HasJar, HasJarsDyn},
     DatabaseKeyIndex,
 };
@@ -66,12 +66,7 @@ impl Stack {
         match origin {
             None | Some(QueryOrigin::Assigned(_)) | Some(QueryOrigin::BaseInput) => {}
             Some(QueryOrigin::Derived(edges)) | Some(QueryOrigin::DerivedUntracked(edges)) => {
-                for (_, dependency_index) in edges
-                    .input_outputs
-                    .iter()
-                    .filter(|edge| edge.0 == EdgeKind::Input)
-                    .copied()
-                {
+                for dependency_index in edges.inputs() {
                     if let Ok(i) = DatabaseKeyIndex::try_from(dependency_index) {
                         if self.s.insert(i) {
                             self.v.push(i)
--- a/components/salsa-2022/src/function/maybe_changed_after.rs
+++ b/components/salsa-2022/src/function/maybe_changed_after.rs
@@ -174,14 +174,6 @@ where
                 // This value was `set` by the mutator thread -- ie, it's a base input and it cannot be out of date.
                 return true;
             }
-            QueryOrigin::Field => {
-                // This value is the value of a field of some tracked struct S.
-                // The fact that we are here means that we are accessing fields from old revisions, which is not allowed.
-                panic!(
-                    "accessing fields of tracked struct {:?} from older revisions",
-                    std::any::type_name::<<C as Configuration>::SalsaStruct>()
-                )
-            }
             QueryOrigin::DerivedUntracked(_) => {
                 // Untracked inputs? Have to assume that it changed.
                 return false;
--- a/components/salsa-2022/src/function/maybe_changed_after.rs
+++ b/components/salsa-2022/src/function/maybe_changed_after.rs
@@ -200,8 +200,8 @@ where
                             // since all prior inputs were green. It's important to do this during
                             // this loop, because it's possible that one of our input queries will
                             // re-execute and may read one of our earlier outputs
-                            // (e.g., in a scenario where we do something like 
-                            // `e = Entity::new(..); query(e);` and `query` reads a field of `e`). 
+                            // (e.g., in a scenario where we do something like
+                            // `e = Entity::new(..); query(e);` and `query` reads a field of `e`).
                             //
                             // NB. Accumulators are also outputs, but the above logic doesn't
                             // quite apply to them. Since multiple values are pushed, the first value
--- a/components/salsa-2022/src/function/maybe_changed_after.rs
+++ b/components/salsa-2022/src/function/maybe_changed_after.rs
@@ -194,6 +194,22 @@ where
                             }
                         }
                         EdgeKind::Output => {
+                            // Subtle: Mark outputs as validated now, even though we may
+                            // later find an input that requires us to re-execute the function.
+                            // Even if it re-execute, the function will wind up writing the same value,
+                            // since all prior inputs were green. It's important to do this during
+                            // this loop, because it's possible that one of our input queries will
+                            // re-execute and may read one of our earlier outputs
+                            // (e.g., in a scenario where we do something like 
+                            // `e = Entity::new(..); query(e);` and `query` reads a field of `e`). 
+                            //
+                            // NB. Accumulators are also outputs, but the above logic doesn't
+                            // quite apply to them. Since multiple values are pushed, the first value
+                            // may be unchanged, but later values could be different.
+                            // In that case, however, the data accumulated
+                            // by this function cannot be read until this function is marked green,
+                            // so even if we mark them as valid here, the function will re-execute
+                            // and overwrite the contents.
                             db.mark_validated_output(database_key_index, dependency_index);
                         }
                     }
--- a/components/salsa-2022/src/function/maybe_changed_after.rs
+++ b/components/salsa-2022/src/function/maybe_changed_after.rs
@@ -5,7 +5,7 @@ use crate::{
     debug::DebugWithDb,
     key::DatabaseKeyIndex,
     runtime::{
-        local_state::{ActiveQueryGuard, QueryOrigin, EdgeKind},
+        local_state::{ActiveQueryGuard, EdgeKind, QueryOrigin},
         StampedValue,
     },
     storage::HasJarsDyn,
@@ -192,10 +192,10 @@ where
                             if db.maybe_changed_after(dependency_index, last_verified_at) {
                                 return false;
                             }
-                        },
+                        }
                         EdgeKind::Output => {
                             db.mark_validated_output(database_key_index, dependency_index);
-                        },
+                        }
                     }
                 }
             }
--- a/components/salsa-2022/src/function/maybe_changed_after.rs
+++ b/components/salsa-2022/src/function/maybe_changed_after.rs
@@ -170,11 +170,17 @@ where
                 // during this revision or is otherwise stale.
                 return false;
             }
-            QueryOrigin::BaseInput | QueryOrigin::Field => {
-                // BaseInput: This value was `set` by the mutator thread -- ie, it's a base input and it cannot be out of date.
-                // Field: This value is the value of a field of some tracked struct S. It is always updated whenever S is created.
-                // So if a query has access to S, then they will have an up-to-date value.
+            QueryOrigin::BaseInput => {
+                // This value was `set` by the mutator thread -- ie, it's a base input and it cannot be out of date.
                 return true;
+            }
+            QueryOrigin::Field => {
+                // This value is the value of a field of some tracked struct S.
+                // The fact that we are here means that we are accessing fields from old revisions, which is not allowed.
+                panic!(
+                    "accessing fields of tracked struct {:?} from older revisions",
+                    std::any::type_name::<<C as Configuration>::SalsaStruct>()
+                )
             }
             QueryOrigin::DerivedUntracked(_) => {
                 // Untracked inputs? Have to assume that it changed.
--- a/components/salsa-2022/src/function/maybe_changed_after.rs
+++ b/components/salsa-2022/src/function/maybe_changed_after.rs
@@ -5,7 +5,7 @@ use crate::{
     debug::DebugWithDb,
     key::DatabaseKeyIndex,
     runtime::{
-        local_state::{ActiveQueryGuard, QueryOrigin},
+        local_state::{ActiveQueryGuard, QueryOrigin, EdgeKind},
         StampedValue,
     },
     storage::HasJarsDyn,
@@ -186,9 +186,16 @@ where
                 // valid, then some later input I1 might never have executed at all, so verifying
                 // it is still up to date is meaningless.
                 let last_verified_at = old_memo.verified_at.load();
-                for &input in edges.inputs().iter() {
-                    if db.maybe_changed_after(input, last_verified_at) {
-                        return false;
+                for &(edge_kind, dependency_index) in edges.input_outputs.iter() {
+                    match edge_kind {
+                        EdgeKind::Input => {
+                            if db.maybe_changed_after(dependency_index, last_verified_at) {
+                                return false;
+                            }
+                        },
+                        EdgeKind::Output => {
+                            db.mark_validated_output(database_key_index, dependency_index);
+                        },
                     }
                 }
             }
--- a/components/salsa-2022/src/function/specify.rs
+++ b/components/salsa-2022/src/function/specify.rs
@@ -91,19 +91,6 @@ where
         self.insert_memo(db, key, memo);
     }
 
-    /// Specify the value for `key` but do not record it is an output.
-    /// This is used for the value fields declared on a tracked struct.
-    /// They are different from other calls to specify because we KNOW they will be given a value by construction,
-    /// so recording them as an explicit output (and checking them for validity, etc) is pure overhead.
-    pub fn specify_field<'db>(&self, db: &'db DynDb<'db, C>, key: C::Key, value: C::Value)
-    where
-        C::Key: TrackedStructInDb<DynDb<'db, C>>,
-    {
-        self.specify(db, key, value, |_| QueryOrigin::Field);
-        let database_key_index = self.database_key_index(key);
-        db.runtime().add_output(database_key_index.into());
-    }
-
     /// Specify the value for `key` *and* record that we did so.
     /// Used for explicit calls to `specify`, but not needed for pre-declared tracked struct fields.
     pub fn specify_and_record<'db>(&self, db: &'db DynDb<'db, C>, key: C::Key, value: C::Value)
@@ -140,7 +127,6 @@ where
         // assigneed by `executor`.
         match memo.revisions.origin {
             QueryOrigin::Assigned(by_query) => assert_eq!(by_query, executor),
-            QueryOrigin::Field => {}
             _ => panic!(
                 "expected a query assigned by `{:?}`, not `{:?}`",
                 executor.debug(db),
--- a/components/salsa-2022/src/function/specify.rs
+++ b/components/salsa-2022/src/function/specify.rs
@@ -100,6 +100,8 @@ where
         C::Key: TrackedStructInDb<DynDb<'db, C>>,
     {
         self.specify(db, key, value, |_| QueryOrigin::Field);
+        let database_key_index = self.database_key_index(key);
+        db.runtime().add_output(database_key_index.into());
     }
 
     /// Specify the value for `key` *and* record that we did so.
@@ -138,6 +140,7 @@ where
         // assigneed by `executor`.
         match memo.revisions.origin {
             QueryOrigin::Assigned(by_query) => assert_eq!(by_query, executor),
+            QueryOrigin::Field => {}
             _ => panic!(
                 "expected a query assigned by `{:?}`, not `{:?}`",
                 executor.debug(db),
--- a/components/salsa-2022/src/function/specify.rs
+++ b/components/salsa-2022/src/function/specify.rs
@@ -46,7 +46,8 @@ where
         //
         // Now, if We invoke Q3 first, We get one result for Q2, but if We invoke Q4 first, We get a different value. That's no good.
         let database_key_index = key.database_key_index(db);
-        if !runtime.is_output_of_active_query(database_key_index) {
+        let dependency_index = database_key_index.into();
+        if !runtime.is_output_of_active_query(dependency_index) {
             panic!("can only use `specfiy` on entities created during current query");
         }
 
--- a/components/salsa-2022/src/runtime/active_query.rs
+++ b/components/salsa-2022/src/runtime/active_query.rs
@@ -3,7 +3,7 @@ use crate::{
     hash::{FxIndexMap, FxIndexSet},
     key::{DatabaseKeyIndex, DependencyIndex},
     tracked_struct::Disambiguator,
-    Cycle, Revision,
+    Cycle, Revision, Runtime,
 };
 
 use super::local_state::{EdgeKind, QueryEdges, QueryOrigin, QueryRevisions};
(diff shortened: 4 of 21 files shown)


components/salsa-2022/src/function/specify.rs
```diff
@@ -100,6 +100,8 @@ where
         C::Key: TrackedStructInDb<DynDb<'db, C>>,
     {
         self.specify(db, key, value, |_| QueryOrigin::Field);
+        let database_key_index = self.database_key_index(key);
+        db.runtime().add_output(database_key_index.into());
```
XFFXFF, line 104: It may be more complicated than we previous thought.  What I'm doing here is adding the fields (`FunctionIngredient`) of tracked structs to the query output, so that we can mark all fields as valid (update `validate_at`) when we `deep_verify_memo` and find there are no changes.  

But there is still a problem when we try to read directly or indirectly the fields of the tracked struct created in the current query.
like 
https://github.com/salsa-rs/salsa/blob/f0fe691d3f902fab4ea3f419ea4f7493b46a5030/salsa-2022-tests/tests/tracked_fn_read_own_entity.rs#L29-L35
https://github.com/salsa-rs/salsa/blob/f0fe691d3f902fab4ea3f419ea4f7493b46a5030/salsa-2022-tests/tests/deletion-cascade.rs#L51-L60

XFFXFF: I was thinking if we can store the query (`DatabaseKeyIndex`) that created the field, when we find that a field may be outdated, we can check if the query stack contains the query that created the field, if yes, the old value is valid, otherwise, panic .

nikomatsakis: Oh, right, I remembered that I spent a while tinkering with this setup. =) I think I had excluded them as outputs for efficiency reasons, but logically they *are* outputs. Let me take a look, thanks for giving this a try.

XFFXFF: Yes, it panics if we try to reuse `intermidiate_result` after creating a distinct input to trigger a new revision.
https://github.com/salsa-rs/salsa/blob/f0fe691d3f902fab4ea3f419ea4f7493b46a5030/salsa-2022-tests/tests/tracked_fn_read_own_entity.rs#L85-L107
```
thread 'red_herring' panicked at 'accessing fields of tracked struct "tracked_fn_read_own_entity::MyTracked" from older revisions'
```

XFFXFF: > I would expect that we would update the revision to the current revision in the call to `specify`?

We would update the revision to the current revision in the current query when we *execute* it, but not when we *resue (deep verify)* it.

nikomatsakis: OK, sorry, what I meant @XFFXFF was -- now that you've added these as outputs -- are those tests failing.

nikomatsakis: What we should probably do is just remove the `specify_field` method...

https://github.com/salsa-rs/salsa/blob/f8f6dbd3490a439d970fde4020e06bbe7432b60f/components/salsa-2022/src/function/specify.rs#L98-L103

...and only have `specify_and_record`...

https://github.com/salsa-rs/salsa/blob/f8f6dbd3490a439d970fde4020e06bbe7432b60f/components/salsa-2022/src/function/specify.rs#L105-L109

...`specify_field` was a micro-optimization and I think it was wrong. I think we can make tracked structs more efficient, but for now this makes sense until we get everything right.

nikomatsakis: This probably means we can remove `QueryOrigin::Field` altogether.

nikomatsakis: I see one test still fails, I don't 100% know why, but I can dig into it later or tomorrow if you get stuck, @XFFXFF .

XFFXFF: > now that you've added these as outputs -- are those tests failing

now only [tracked_fn_read_own_entity::red_herring](https://github.com/salsa-rs/salsa/blob/f0fe691d3f902fab4ea3f419ea4f7493b46a5030/salsa-2022-tests/tests/tracked_fn_read_own_entity.rs#L85-L107) and [deletion-cascade](https://github.com/salsa-rs/salsa/blob/f0fe691d3f902fab4ea3f419ea4f7493b46a5030/salsa-2022-tests/tests/deletion-cascade.rs#L90-L130) panic. They all panic with something like   
```
thread 'red_herring' panicked at 'accessing fields of tracked struct "tracked_fn_read_own_entity::MyTracked" from older revisions'
```
The panic is what we add in this pr. Let me try to explain why it panics.  


Take the example as 
```rust
 #[salsa::tracked(jar = Jar)] 
 fn intermediate_result(db: &dyn Db, input: MyInput) -> MyTracked { 
     db.push_log(format!("intermediate_result({:?})", input)); 
     let tracked = MyTracked::new(db, input.field(db) / 2); 
     let _ = tracked.field(db); // read the field of an entity we created 
     tracked 
 } 

fn main() {
    ...
    let input = MyInput::new(&mut db, 22)
    intermediate_result(&db, input)  // OK

    let input2 = MyInput::new(&mut db, 33)
    intermediate_result(&db, input) // panic
}
```

We create an input and call `intermediate_result`, it's ok. Now we create a different input to trigger a new revision, and then call `intermediate_result` with original input (which hasn't changed), it panics. 

The second time we call `intermediate_reuslt(&db, input)`, we need to deep verify the result because of the new revision introduced by `input2`, and need to check the dependencies of the query. Since we call `tracked.field(db)` in `intermediate_result`, so `tracked.field(db)` is one of the dependencies of the query, we should verify `tracked.field(db)` first, then we will find the `verified_at` of `tracked.field(db)`  smaller than the current revision, so panic. 

We will mark `tracked.field(&db)` as valid (update `verified_at` to the current revision) after we deep verify `intermediate_result`, and then it is OK to use `tracked.field(&db)`, but we use it before the deep verification is done

```
intermediate_result -> deep verify intermediate_reuslt -> deep verify tracked.field -> panic
```

components/salsa-2022/src/function/maybe_changed_after.rs
```diff
@@ -170,12 +170,18 @@ where
...
             }
-            QueryOrigin::BaseInput | QueryOrigin::Field => {
-                // BaseInput: This value was `set` by the mutator thread -- ie, it's a base input and it cannot be out of date.
-                // Field: This value is the value of a field of some tracked struct S. It is always updated whenever S is created.
-                // So if a query has access to S, then they will have an up-to-date value.
+            QueryOrigin::BaseInput => {
+                // This value was `set` by the mutator thread -- ie, it's a base input and it cannot be out of date.
                 return true;
             }
+            QueryOrigin::Field => {
```
XFFXFF, line 177: I was thinking one way to get all the tests pass is that we could store the query that created the field (like `QueryOrigin::Assigned`), when we find that a field may be outdated, we can check if the query stack contains the query that created the field, if yes, the old value is valid, otherwise, panic.  

> This probably means we can remove QueryOrigin::Field altogether.

@nikomatsakis  Also, I think we need to keep the `QueryOrigin::Field`, which is handled differently here than `QueryOrigin::Assigned`

components/salsa-2022/src/function/specify.rs
nikomatsakis: OK, I see. This seems like a pre-existing bug that would affect specified, as well. Basically we validate all inputs, but we don't consider that some inputs are also outputs? 

I think first off we should add a test case like...

```rust
#[tracked] struct Test { }

#[tracked(specify)] fn specifiable(db: &dyn crate::Db, test: Test) -> u32 { 0 }

#[tracked] fn test(db: &dyn crate::Db) -> u32 {
    let t = Test::new();
    specifiable::specify(db, t, 22);
    specifiable(db, t)
}
```

If I understand the problem, this `test` function would also panic once an input is changed.

Then, I think the best fix would be to replace this loop

https://github.com/salsa-rs/salsa/blob/32f0ce0cffc0e3695a3292fb740d378cdcf19408/components/salsa-2022/src/function/maybe_changed_after.rs#L190-L195


with something like this

```rust
                let last_verified_at = old_memo.verified_at.load();
                for &input in edges.inputs().iter() {
                    if !edges.has_output(input) { // <-- NEW!
                        if db.maybe_changed_after(input, last_verified_at) {
                            return false;
                        }
                    }
                }
```

where `edges.has_output` is some new method we add to `QueryEdges`. It looks like we'll have to add back a helper method `outputs` and then we can use [`binary_search`](https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search) to search through it (the outputs array is sorted).

nikomatsakis: Alternatively, you're right that we could remember the query that specified the value of the field, and then modify `maybe_changed_after` so that it takes as input the id of the query being re-validated as well. 

What I don't like about it is that it makes the interface of that function more complex -- i.e., now it is saying "return true if the value have changed after this revision *or* it was assigned by this query". 

What I *do* like about it is that we don't have to search the outputs for *every* input.  There are sometimes quite a lot of inputs/outputs, so that could be a major change.

nikomatsakis: OK, I did a quick search to look at who calls `maybe_changed_after` and it seems like the only real caller is this loop right here. I think @XFFXFF it's probably better to follow your suggestion, and modify `QueryOrigin::Specified` to include the query that specified the value. Then we can add a parameter to `maybe_changed_after` with a name like `consumer` or something, with the expected value being something like "has it changed after this revision OR was it created by this consumer". 

All things considered, I suspect that my suggestion of doing an extra binary search is going to be too expensive.

components/salsa-2022/src/function/maybe_changed_after.rs
nikomatsakis: Ah, hmm, I see. Yes. But actually it doesn't  *have* to be handled differently here. I think returning false is actually better (panicking in this code makes me a bit nervous, to be honest, although I think it should be fine).

The end-result will be that we try to execute the function associated with the field, which will panic. 

Basically by returning false we're saying: we can't verify that the value in this function is up-to-date.

XFFXFF (commented):
@nikomatsakis  Now all tests pass except [deletion-cascade](https://github.com/salsa-rs/salsa/blob/f0fe691d3f902fab4ea3f419ea4f7493b46a5030/salsa-2022-tests/tests/deletion-cascade.rs#L90-L130) 

```rust
thread 'basic' panicked at '`execute` method for field `MyTracked::field` invoked
```

https://github.com/salsa-rs/salsa/blob/f0fe691d3f902fab4ea3f419ea4f7493b46a5030/salsa-2022-tests/tests/deletion-cascade.rs#L51-L60

In this case, instead of using `m.field(db)` directly in `contribution_from_struct` that creates `m`, we use it in `copy_field`, a dependency of `contribution_from_struct`

```
contribution_from_struct -> deep verify contribution_from_struct -> copy_field -> deep verify copy_field -> tracked.field -> deep verify tracked.field -> panic
```

I think one way to fix this is instead fo  checking  the `specfied` was created by the consumer query, but by any query in the query stack.

the code might like this
```rust
match &old_memo.revisions.origin {
    QueryOrigin::Assigned(by_query) => {
        if db.runtime().query_stack().contain(by_query) {
            return true
        }
        return false;
    }
    ...
}
```
Also, we can use binary_search to search through it

nikomatsakis:
Argh, of course. Searching the query stack doesn't seem quite right to me, unfortunately. That includes queries that are *actively* executing, and I think it would be possible to have a scenario like this

* in revision R1:
    * query X specifies value for query Y and invokes query Z
    * query Z reads query Y
* in revision R2
    * query X is actively executing but does NOT specify value for query Y and invokes query Z
    * query Z reads query Y -- this should be marked as "not valid", but because we see X in the stack, we thing it is valid

This scenario is actually a bit thorny, I have to think about it for a second. 

I *think* the right answer might be to merge the input/output lists into *one* list which (annoyingly) would require an extra bit per input or something like that (maybe we can come up with a more compact representation). We would then mark the outputs as verified as we encounter them in the deep-verify list.

It might seem like this is not valid, because we could still encounter an input later that causes our query to re-execute. But actually I think it *is* valid, because we know that all the inputs seen thus far to the query were green, and therefore even if the query re-executes, it will store the same value (or else the program is not deterministic).

The most annoying part is figuring out how to represent the input-output list in a compact fashion. We could probably shave off one bit from one of the sections, such that we support only 2^15 bits for jars or something like that (I have to check how a `DatabaseKey` is structred, I've forgotten). That would let us have a "input vs output" bit.

I think there's a universe where we can be clever and figure out if the read/write *must* have been an output, but we'd have to rejuggle a few things, and we might not be able to do it unless we remove `specify` functions (or make it so that they *must* be specified and cannot have a "fallback" value).

If @XFFXFF you're still game to keep pushing on this PR, I think what I would want to do is to try and validate my original idea. This means...

(1) Changing the input-output list so that it's not a separated list of `DatabaseKeyIndex` but instead a list of enums, something like `(EdgeKind, DatabaseKeyIndex)` pairs:

```rust
enum EdgeKind {
    Input, Output
}
```

I'm not sure how much of a pain that will be. 

(2) Modifying `deep_verify` so that when it encounters an *output* it marks it as validated.

nikomatsakis:
Something I can't remember --- 

Is it important that we have a sorted list of outputs? 

I think initially I thought it would be, because we were going to hash the outputs as part of the final value, and so they needed to be in a consistent order etc across revisions. But we don't do that.

Alternatively, maybe we search through the outputs using binary search? But I'm pretty sure we don't do that either, as there isn't even an `outputs` method to access the list of outputs.

In fact, I think it doesn't matter at all.

salsa-2022-tests/tests/accumulate-reuse-workaround.rs
```diff
@@ -89,5 +89,6 @@ fn test1() {
         [
             "accumulated(List(Id { value: 1 }))",
             "compute(List(Id { value: 1 }))",
+            "compute(List(Id { value: 2 }))",
```
XFFXFF, line 92: This is a bug.

> It might seem like this is not valid, because we could still encounter an input later that causes our query to re-execute. But actually I think it is valid, because we know that all the inputs seen thus far to the query were green, and therefore even if the query re-executes, it will store the same value (or else the program is not deterministic).

There are three kinds of output  
* tracked structs created
* invocations of `specify`
* accumulators pushed to

I think you are right about first two cases, but for *accumulators*, things might be different.

take the example as 

```rust
#[salsa::accumulator]
struct Integers(u32);

#[salsa::tracked]
fn compute(db: &dyn Db, input: Input) -> u32 {
    Integers::push(db, 0);
    input.value(db)
}

#[test]
fn test1() {
    let mut db = Database::default();

    let input = Input::new(&mut db, 1);
    dbg!(compute::accumulated::<Integers>(&db, input));

    input.set_value(&mut db).to(2);
    dbg!(compute::accumulated::<Integers>(&db, input));
}
```
will output

```
[salsa-2022-tests/tests/accumulate-reuse-workaround.rs:53] compute::accumulated::<Integers>(&db, input) = [
    0,
]
[salsa-2022-tests/tests/accumulate-reuse-workaround.rs:56] compute::accumulated::<Integers>(&db, input) = [
    0,
    0,
]
```
If it was the first push in a new version, we would reset the accumulated values by checking the `produced_at` of the old accumulated values

https://github.com/salsa-rs/salsa/blob/07848c2818077635658ba94b245d6f5713909452/components/salsa-2022/src/accumulator.rs#L64-L68

But now we will first mark the old accumulated values as valid in `deep_verify_memo`, which will update the `produced_at' of the old accumulated values to the newest version so that when we re-execute the query and push the values, we will not reset them.

XFFXFF (commented):
I merged the input/output lists into one list, please help to see if it's going in the right direction.

XFFXFF:
> Is it important that we have a sorted list of outputs?

https://github.com/salsa-rs/salsa/blob/07848c2818077635658ba94b245d6f5713909452/components/salsa-2022/src/function/diff_outputs.rs#L21-L25

@nikomatsakis  We are assuming here that the output is in order, but I think it is also okay if the output is in order of execution?

nikomatsakis:
@XFFXFF Oh, right, *that* was why I cared about them being in sorted order. Hmm! (They are in sorted order now, I'm pretty sure, although I can't remember how that comes about -- I think we accumulate them into a `BTreeSet` or something like that.)

Well, I see a few options:

* Track the outputs twice, once in a combined list with the inputs, and once separately.

or

* Compute the diff another way. For example, iterate over the outputs of the "old node" and put them into a hashset, then iterate over the outputs of the new node and remove elements from that set when we find them. Whatever is left we can delete.

(Also, sorry for the late reply, I missed the GH notification for this somehow...)

XFFXFF:
@nikomatsakis No need to say sorry, I know you must have a lot of GH notifications from rustc.

You may have missed https://github.com/salsa-rs/salsa/pull/413#discussion_r973654052, which I think is important.

> They are in sorted order now, I'm pretty sure, although I can't remember how that comes about -- I think we accumulate them into a BTreeSet or something like that

I know you implemented it with `BTreeSet`, sorry, what I meant was that after I changed the code, which is the current code of this pr, it is in the execution order

components/salsa-2022/src/runtime/local_state.rs
```diff
@@ -114,28 +112,35 @@ pub struct QueryEdges {
...
+    pub input_outputs: Arc<[(EdgeKind, DependencyIndex)]>,
 }
 
 impl QueryEdges {
     /// Returns the (tracked) inputs that were executed in computing this memoized value.
     ///
     /// These will always be in execution order.
-    pub(crate) fn inputs(&self) -> &[DependencyIndex] {
-        &self.input_outputs[0..self.separator as usize]
+    pub(crate) fn inputs(&self) -> Vec<DependencyIndex> {
```
nikomatsakis, line 122: What code calls this now?

nikomatsakis (changes requested):

nikomatsakis:
@XFFXFF Ah, I see! I didn't realize you had updated the PR. Indeed, I think we have to modify the `diff_outputs` routine. I think I would "collect" the outputs from one side into a `FxHashSet`, then iterate over the outputs from the other and remove them.

nikomatsakis:
But this looks good otherwise!

components/salsa-2022/src/runtime/local_state.rs
XFFXFF: It is only called here.  
https://github.com/salsa-rs/salsa/blob/07848c2818077635658ba94b245d6f5713909452/components/salsa-2022/src/function/accumulated.rs#L66-L76

components/salsa-2022/src/function/accumulated.rs
```diff
@@ -65,23 +64,17 @@ impl Stack {
...
-            | Some(QueryOrigin::BaseInput)
-            | Some(QueryOrigin::Field) => {}
+            None | Some(QueryOrigin::Assigned(_)) | Some(QueryOrigin::BaseInput) => {}
             Some(QueryOrigin::Derived(edges)) | Some(QueryOrigin::DerivedUntracked(edges)) => {
-                for DependencyIndex {
-                    ingredient_index,
-                    key_index,
-                } in edges.inputs().iter().copied()
+                for (_, dependency_index) in edges
+                    .input_outputs
```
XFFXFF, line 70: I removed the `inputs` method from `QueryEdges` and iterated directly over the field `input_ouputs`. Previously we needed to iterate over `input_outputs` to get `inputs` and then over `inputs`, plus it would create a `Vec` as a container for `inputs`. I think the current implementation would be a bit more efficient

XFFXFF (commented):
A serious bug in the code now is that https://github.com/salsa-rs/salsa/pull/413/files#r973654052

components/salsa-2022/src/function/accumulated.rs
nikomatsakis: yes, but I wonder if we should add a method like

```rust
impl QueryEdges {
    pub fn inputs(&self) -> impl Iterator<Item = DependencyIndex> {
        self.iter().filter().copied()
    }
}
```

instead?

components/salsa-2022/src/runtime/local_state.rs
```diff
@@ -114,28 +111,24 @@ pub struct QueryEdges {
...
     ///
     /// These will always be in execution order.
-    pub(crate) fn inputs(&self) -> &[DependencyIndex] {
-        &self.input_outputs[0..self.separator as usize]
+    pub(crate) fn outputs(&self) -> Vec<DependencyIndex> {
+        self.input_outputs
+            .iter()
+            .filter(|(edge_kind, _)| *edge_kind == EdgeKind::Output)
+            .map(|(_, dependency_index)| *dependency_index)
+            .collect()
```
nikomatsakis, line 126:
```suggestion
    pub(crate) fn outputs(&self) -> impl Iterator<Item = DependencyIndex> {
        self.input_outputs
            .iter()
            .filter(|(edge_kind, _)| *edge_kind == EdgeKind::Output)
            .map(|(_, dependency_index)| *dependency_index)
```

maybe this?

salsa-2022-tests/tests/accumulate-reuse-workaround.rs
nikomatsakis: I see, yes. Hmm. The problem here is that it's not a single, instantaneous write, but rather a series of writes. Let me ponder this for a second.

nikomatsakis (changes requested):
Left one nit for now, let me think about the accumulator problem!

nikomatsakis:
OK, here is my suggestion. Currently, when you push to an accumulator, we check if the revision on the accumulated value is the current revision and reset it to empty if not...

https://github.com/salsa-rs/salsa/blob/2ffe4a78a824acb8c73e77497e4c2c469fcbed37/components/salsa-2022/src/accumulator.rs#L64-L68

...but why not instead check if the accumulator is an output of the current query?

```rust
        // This is the first push in a new revision. Reset.
        if !runtime.is_output_of_active_query(self.dependency_index()) {
            accumulated_values.values.truncate(0);
            accumulated_values.produced_at = current_revision;
        }
```

I think that would work.

components/salsa-2022/src/runtime/local_state.rs
XFFXFF: yes, it's better

XFFXFF:
> but why not instead check if the accumulator is an output of the current query?

Make sense to me, great solution, thanks!  

@nikomatsakis  please review again

components/salsa-2022/src/function/maybe_changed_after.rs
```diff
@@ -188,9 +186,16 @@ where
...
-                        return false;
+                for &(edge_kind, dependency_index) in edges.input_outputs.iter() {
+                    match edge_kind {
+                        EdgeKind::Input => {
+                            if db.maybe_changed_after(dependency_index, last_verified_at) {
+                                return false;
+                            }
+                        }
+                        EdgeKind::Output => {
+                            db.mark_validated_output(database_key_index, dependency_index);
```
nikomatsakis, line 197:
```suggestion
                            // Subtle: Mark outputs as validated now, even though we may
                            // later find an input that requires us to re-execute the function.
                            // Even if it re-execute, the function will wind up writing the same value,
                            // since all prior inputs were green. It's important to do this during
                            // this loop, because it's possible that one of our input queries will
                            // re-execute and may read one of our earlier outputs
                            // (e.g., in a scenario where we do something like 
                            // `e = Entity::new(..); query(e);` and `query` reads a field of `e`). 
                            //
                            // NB. Accumulators are also outputs, but the above logic doesn't
                            // quite apply to them. Since multiple values are pushed, the first value
                            // may be unchanged, but later values could be different.
                            // In that case, however, the data accumulated
                            // by this function cannot be read until this function is marked green,
                            // so even if we mark them as valid here, the function will re-execute
                            // and overwrite the contents.
                            db.mark_validated_output(database_key_index, dependency_index);
```

components/salsa-2022/src/runtime/local_state.rs
```diff
@@ -80,19 +75,23 @@ pub enum QueryOrigin {
 }
 
 impl QueryOrigin {
-    /// Indices for queries *written* by this query (or `&[]` if its value was assigned).
-    pub(crate) fn outputs(&self) -> impl Iterator<Item = DependencyIndex> + '_ {
-        let slice = match self {
+    /// Indices for queries *written* by this query (or `vec![]` if its value was assigned).
+    pub(crate) fn outputs(&self) -> Box<dyn Iterator<Item = DependencyIndex> + '_> {
```
nikomatsakis, line 79: Nothing *wrong* with returning a `Box<dyn>`, but in this kind of scenario I usually use ye olde flat-map trick...

```rust
pub(crate) fn outputs(&self) -> impl Iterator<Item = DependencyIndex> + '_ {
    let opt_edges = match self {
        QueryOrigin::Derived(edges) | QueryOrigin::DerivedUntracked(edges) => Some(edges),
        QueryOrigin::Assigned(_) | QueryOrigin::BaseInput => None,
    };
    opt_edges.flat_map(|edges| edges.outputs())
}
```

nikomatsakis: It works well when you have one "iterator" case and one "empty" case.

components/salsa-2022/src/runtime/shared_state.rs
```diff
@@ -13,9 +13,6 @@ pub(super) struct SharedState {
     /// Stores the next id to use for a snapshotted runtime (starts at 1).
     pub(super) next_id: AtomicUsize,
 
-    /// Vector we can clone
-    pub(super) empty_dependencies: Arc<[DependencyIndex]>,
-
```
nikomatsakis, line 18: Wait, why can't we clone `empty_dependencies` anymore...

nikomatsakis: ...this was, admittedly, a pure optimization. But it seemed like it might be a useful one. I guess I don't have any data to back that up. :)

nikomatsakis: It's fine, let's remove it, easy enough to add back later once we have some data.

nikomatsakis (changes requested):
Two nits:

A comment and a way to use `-> impl` and avoid a `Box` on a potentially hot path.

components/salsa-2022/src/runtime/local_state.rs
XFFXFF: This is exactly what I was looking for, thanks! @nikomatsakis

components/salsa-2022/src/runtime/shared_state.rs
XFFXFF: I thought about it for a while and I think I should add it back. I thought it was redundant before, sorry I didn't think carefully here, but now I think it makes sense, there are many queries that have no dependencies, only tracked functions have dependencies? It would be better to have them share an empty dependency. @nikomatsakis

components/salsa-2022/src/runtime/active_query.rs
```diff
@@ -86,29 +78,22 @@ impl ActiveQuery {
...
+    pub(super) fn is_output(&self, key: DependencyIndex) -> bool {
+        self.input_outputs.contains(&(EdgeKind::Output, key))
     }
 
     pub(crate) fn revisions(&self, runtime: &Runtime) -> QueryRevisions {
-        let separator = self.dependencies.len();
-
-        let input_outputs = if self.dependencies.is_empty() && self.outputs.is_empty() {
+        let input_outputs = if self.input_outputs.is_empty() {
             runtime.empty_dependencies()
```
XFFXFF, line 91: I added empty dependencies back and used it here. @nikomatsakis

components/salsa-2022/src/function/maybe_changed_after.rs
```diff
@@ -188,9 +186,32 @@ where
...
+                            // (e.g., in a scenario where we do something like
+                            // `e = Entity::new(..); query(e);` and `query` reads a field of `e`).
+                            //
+                            // NB. Accumulators are also outputs, but the above logic doesn't
+                            // quite apply to them. Since multiple values are pushed, the first value
+                            // may be unchanged, but later values could be different.
+                            // In that case, however, the data accumulated
+                            // by this function cannot be read until this function is marked green,
+                            // so even if we mark them as valid here, the function will re-execute
+                            // and overwrite the contents.
```
nikomatsakis, line 212: cc @pnkfelix -- this is an example of the kind of reasoning I would like to able to verify with a model checker.

nikomatsakis (approved):

PR #414, opened 2022-09-14
better debugging for input structs by including id fields

### LINKED ISSUE #415: Singleton inputs: best behavior when a new instance is created?
What is the best behavior when a second instance of a singleton is created?

Options:

* Panic!
* Overwrite the fields of the first one!

Right now, we do the second. Doing an extremely scientific study ([looking at the first google hit](https://www.digitalocean.com/community/tutorials/java-singleton-design-pattern-best-practices-examples#lazy-initialization)) I see that the typical Java pattern is to return the existing object, but they don't address the question of whether it has fields.

(_Originally posted by @nikomatsakis in https://github.com/salsa-rs/salsa/issues/404#issuecomment-1247902050_)

OLUWAMUYIWA:
Ref: #405 

Fixes #415


components/salsa-2022-macros/src/input.rs
```diff
@@ -112,8 +104,16 @@ impl InputStruct {
         )
         .collect();
 
+        // setters
         let set_field_names = self.all_set_field_names();
```
nikomatsakis, line 108: What's missing here is that we need to not generate a "setter" for an `#[id]` field (and we should make a negative test that shows no setter exists). The idea is that you cannot change the value of an `#[id]` field after `new` is invoked, and that's why they are safe to include in the debug impl.

It occurs to me that, on singleton inputs, with #415, their fields can be modified when a second instance is created, so we probably want to prohibit `#[id]` fields if the type is marked as singleton. This pushes me further towards the option (described in that issue) of panicking when a second instance is created.

nikomatsakis: One way to achieve this goal would be to modify

https://github.com/salsa-rs/salsa/blob/32f0ce0cffc0e3695a3292fb740d378cdcf19408/components/salsa-2022-macros/src/salsa_struct.rs#L124-L127

so that  it doesn't include fields with the `#[id]` attribute. Also, we may want to move that function out of `salsa_struct.rs` and into `input.rs`, I think only inputs have setters, so it's not really common code that applies to all kinds of salsa structs.

nikomatsakis (changes requested):

OLUWAMUYIWA:
1. So, yeah the initial commit actually took the setters into consideration, but I didn't write a  test for it. I only verified with `cargo expand`.  Now, I made it clearer, using your suggestion. and also added a `compie-fail` test. I doubt if that is the best way to check if a method is implemented for a struct. If there's a better way, please tell me. I'll update the PR.
2. I assumed that since panicking upon duplicating `singleton structs` appears to be the preferred choice, (as in #415 ). I felt it was probably better to include a patch for that here, instead of trying to prohibit `singleton structs` fields from having the `#[id]` tag.

Please check, @nikomatsakis

salsa-2022-tests/tests/compile-fail/input_struct_id_fields_no_setters.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl Db for Database {}
+
+
+fn main() {
+    let mut db = Database::default();
+    let input = MyInput::new(&mut db, 3, 4);
+    // should not compile as `id_one` should not have a setter
+    // compile error: method `set_id_one` not found in scope
+    input.set_id_one(1);
```
nikomatsakis, line 30: it's not an ideal test, indeed, but I don't know a better way :)

components/salsa-2022/src/input.rs
```diff
@@ -49,9 +49,12 @@ where
         Id::from_id(crate::Id::from_u32(next_id))
     }
 
-    pub fn new_singleton_input(&mut self, _runtime: &Runtime) -> Id {
-        // There's only one singleton so record that we've created it
-        // and return the only id.
+    pub fn new_singleton_input(&self, _runtime: &Runtime) -> Id {
+        // when one exists already, panic
+        if self.counter.load(Ordering::Relaxed) >= 1 {
+            panic!("Singleton struct may not be duplicated");
```
nikomatsakis, line 55:
```suggestion
            panic!("singleton struct may not be duplicated");
```

For some reason, starting with a lowercase letter seems to fit Rust style more to my eyes. Not sure if there's any official guidance on this point.

nikomatsakis (changes requested):

nikomatsakis (approved):

PR #416, opened 2022-09-16
Correct docs to refer to `#[salsa::cycle]`

ssbr:
Judging by the examples, `recover` is an outdated name.

E.g. https://github.com/salsa-rs/salsa/blob/master/tests/cycles.rs

(I can't really test this on version 0.16, so I'm just going by the examples.)


nikomatsakis:
Actually the docs are more outdated than this, even. At the moment, we are working to the new Salsa 2022 version (not sure if we'll have to update to Salsa 2023 though, as it's taking a bit longer than I thought :). In that version, you would declare a cycle recovery function like this:

https://github.com/salsa-rs/salsa/blob/0357e53ce96988f6c9c61619689eef67f8335e3a/salsa-2022-tests/tests/cycles.rs#L145

(I *suspect* the docs may be outdated in other ways, but I'll to read them through. We made a number of improvements to cycle recovery that made it much more robust, and I don't recall if we fixed the docs afterwards.)

nikomatsakis:
Nonetheless, I'm going to r+ this as is since it at least makes things more accurate to the current salsa version. Thanks.

bors r+

Issue #385, opened 2022-08-26
Add an API like `new_with_durability` to construct input with specific durability


xffxff:
https://github.com/salsa-rs/salsa/pull/362 added "builder style" setter APIs that permit setting durability. I think it's a little inconvenient to use it

```rust
#[salsa::input] 
struct MyInput {
   field1: u32;
   field2: u32;
}

let input = MyInput::new(&mut db, 1, 2);
input.set_field1(&mut db).with_durability(Durability::HIGH).to(1);
input.set_field2(&mut db).with_durability(Durability::HIGH).to(2);
```
* We give `field1` the value `1` when we `new` the input, but if we want to set the durability, we also need to assign it to `1` again via `to(1)`, same for `field2`
* When there are many fields, we need to set the durability one by one.

Could we have an API like `new_with_durability` to construct input with specific durability? Use it like
```rust
MyInput::new_with_durability(&mut db, field1, field2, Durability::HIGH)
```

nikomatsakis:
Maybe we can have a builder style API for creating inputs, too, to give room for future options?

I think I would keep `new` as a shorthand, and to ensure that the 'basic methods' for all salsa structs are the same:

* new to create
* getters to read

but something like `MyInput::new_builder(&mut db).with_durability(HIGH).with_fields(1, 2)` could work?

XFFXFF:
> Maybe we can have a builder style API for creating inputs, too, to give room for future options?

Yes, I think a builder stayle API is great

mdx97:
Is the durability supposed to be able to vary between the fields in a struct? If so, it might make sense to implement some sort of builder API for field values themselves:
```rust
let input = MyInput::new(
    &mut db,
    InputFieldBuilder::new(1)
        .with_durability(Durability::HIGH),
    InputFieldBuilder::new(2)
        .with_durability(Durability::MEDIUM),
);
```

PR #418, opened 2022-09-29
implemented the builder api for salsa inputs

OLUWAMUYIWA:
Ref: #385


components/salsa-2022-macros/src/input.rs
```diff
@@ -198,11 +217,86 @@ impl InputStruct {
...
+        let ident = self.id_ident();
+        let struct_name = self.builder_name();
+        let field_names = self.all_field_names();
+        let field_tys: Vec<_> = self.all_field_tys();
+        let db_dyn_ty = self.db_dyn_ty();
+        let jar_ty = self.jar_ty();
+        let input_index = self.input_index();
+        let field_indices = self.all_field_indices();
+
+        let buidler: syn::ItemStruct = parse_quote! {
```
MihailMihov, line 238: Typo
```suggestion
        let builder: syn::ItemStruct = parse_quote! {
```

components/salsa-2022-macros/src/input.rs
```diff
@@ -198,11 +217,86 @@ impl InputStruct {
...
+                        #(self.#field_names = Some(#field_names);)*
+                        self
+                    }
+
+                    #build_fn
+                }
+            }
+        };
+
+        (buidler, impls)
```
MihailMihov, line 294:
```suggestion
        (builder, impls)
```

MihailMihov (changes requested):

components/salsa-2022-macros/src/input.rs
```diff
@@ -198,11 +217,86 @@ impl InputStruct {
...
+
+        let impls: syn::ItemImpl = {
+            parse_quote! {
+                impl #struct_name {
+                    fn with_durability(&mut self, durability: salsa::durability::Durability) -> &mut Self {
+                        self.durability = Some(durability);
+                        self
+                    }
+
+                    fn with_fields(&mut self, #(#field_names: #field_tys,)*) -> &mut Self {
```
MihailMihov, line 284: `with_values` would be a better name in my opinion, but the current is fine too.

salsa-2022-tests/tests/builder_api.rs
```diff
@@ -0,0 +1,65 @@
...
+        .to(20);
+    input
+        .set_field2(&mut db)
+        .with_durability(Durability::HIGH)
+        .to(40);
+    let input_from_builder = MyInput::new_builder()
+        .with_durability(Durability::HIGH)
+        .with_fields(20, 40)
+        .build(&db)
+        .unwrap();
```
MihailMihov, line 48: Returning a `Result` from `build` isn't useless, but I also feel like almost always having to `unwrap` it will be something very repetitive and rarely useful. If you agree with my suggestion that the durability should have a default I have a different idea, which I think will make the API better, instead of having the `build` method be the terminal one, we could instead move that over to `with_fields` and we'd no longer need to return a `Result`, because we know that we have the values and the other thing we need is the db, but there a panic would be fine.
```rust    
let input_from_builder = MyInput::new_builder()
        .with_db(&db)
        .with_durability(Durability::HIGH)
        .with_fields(20, 40);
```
I think this makes the API calls shorter, with no downsides and it would be closer to the `salsa::Setter` API. The only thing that I kind of don't like is how to fields method is named similarly to the other ones, which doesn't clearly show that it's the terminal method that actually builds the type.

components/salsa-2022-macros/src/input.rs
OLUWAMUYIWA: Thanks. Will fix

components/salsa-2022-macros/src/input.rs
```diff
@@ -198,11 +217,86 @@ impl InputStruct {
...
+            struct #struct_name{
+                #(#field_names: Option<#field_tys>,)*
+                durability: Option<salsa::durability::Durability>,
+            }
+        };
+
+        let build_fn: syn::ImplItemMethod = if self.0.is_isingleton() {
+            parse_quote! {
+                fn build(&mut self, __db: &#db_dyn_ty) -> std::result::Result<#ident, std::boxed::Box<dyn std::error::Error>> {
+                    let durability = self.durability.ok_or_else(|| "durability not provided")?;
```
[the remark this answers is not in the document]
OLUWAMUYIWA: Hmmn true. Thanks. Will fix

components/salsa-2022-macros/src/input.rs
```diff
@@ -198,11 +217,86 @@ impl InputStruct {
...
 
-        // }
+                    let (__jar, __runtime) = <_ as salsa::storage::HasJar<#jar_ty>>::jar(__db);
+                    let __ingredients = <#jar_ty as salsa::storage::HasIngredientsFor< #ident >>::ingredient(__jar);
+                    let __id = __ingredients.#input_index.new_singleton_input(__runtime);
+
+                    #(
+                        __ingredients.#field_indices.store_new(__runtime, __id, self.#field_names.clone().ok_or_else(||"field not provided")?, durability);
```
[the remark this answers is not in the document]
OLUWAMUYIWA: Thought about having `with_durabilities` too, but I dropped the idea because it sorts of requires the user to keep the order of fields in mind. I don't know if that cognitive overhead is okay. @MihailMihov, @nikomatsakis , @XFFXFF 
Though personally, I think it's better.

Another option I'm considering is having `with_fields` take a tuple, i.e. its `value` and it's `durability`. That localizes concerns. But, is taking tuples neat?

components/salsa-2022-macros/src/input.rs
OLUWAMUYIWA: Hmmn. `with_fields` makes it obvious to the user what they're about to set. Wouldn't `with_values` be a little ambiguous? I really don't know.

salsa-2022-tests/tests/builder_api.rs
OLUWAMUYIWA: > The only thing that I kind of don't like is how to fields method is named similarly to the other ones, which doesn't clearly show that it's the terminal method that actually builds the type.

That's a real concern. I actually think builders should have `build`, and should naturally return `Result`s except when the value we're building can comfortably have default values all through. In this case, `field`s cannot have default values. And, panicking inside `with_fields`? hmmn maybe people don't expect that. But I wouldn't know. I don't have much experience here. If you're so sure...

Maybe others have something to say?
@XFFXFF @nikomatsakis @crlf0710

components/salsa-2022-macros/src/input.rs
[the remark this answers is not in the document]
MihailMihov: > Thought about having with_durabilities too, but I dropped the idea because it sorts of requires the user to keep the order of fields in mind. I don't know if that cognitive overhead is okay.

The cognitive overhead point is something that I didn't consider, but I did go over the idea with the tuples and decided to not suggest it, because it wouldn't look very good if we needed to add more options for the fields in the future. And then we also lose the ability to change just 1 or 2 things about a field, we'd need to set all 5 (again this only applies if we do consider having more than just a value and a durability).

Also if tuples were an option all the intermediate structs produced by the builder would have an incomplete state, and the user would need to keep track of all the fields being set.

But that's just my opinion I think we should hear what others have to say.

components/salsa-2022-macros/src/input.rs
MihailMihov: I think of a field as something which has a value, a durability and maybe other things associated with it, which is why `with_values` is better in my opinion as it communicates that you're setting just the value of the field. If I saw `with_fields` my first thought would be that it takes a struct or tuple which contains everything about a field. Maybe someone else could comment on what is the better option, but I think both are good?

salsa-2022-tests/tests/builder_api.rs
MihailMihov: > In this case, fields cannot have default values.

That and the database are the two things that we need to set for the struct to be complete. I think either one of those could be used as the `build` method and panic only if the other is missing, but it's a tradeoff where we make the API calls slightly shorter, but lose a bit of clarity of which method is the one that `build`s the type.
But panicking instead of returning `Result` and making either `with_db` or `with_fields` the `build` method are two separate things. If you're sure that the `build` should return `Result` you can still consider putting that logic in `with_db` or `with_fields`.

> But I wouldn't know. I don't have much experience here. If you're so sure...

I don't either, so take what I say with a grain of salt. I just wanted to bring it up as a talking point, so we could discuss it with the others.

nikomatsakis:
@OLUWAMUYIWA I know I disappeared a long time here! I'm still interested in this PR, are you still interested to pursue it? (Totally understand if not)

salsa-2022-tests/tests/builder_api.rs
nikomatsakis: Coming back to this with fresh eyes, a few thoughts--

1. I think that we should strive to avoid `Result` that are statically known to not apply. 
2. I don't love having `with_values` take a tuple, I think I'd rather have a method per field.
3. This seems to imply we may need to do "type tricks" to make a builder that is statically known to have a value for all fields etc.

It seems like we should back up from the PR and first we be more specific about the API that we want for an example.

OLUWAMUYIWA:
Hi @nikomatsakis I want to finish it up! I'll get on it.

nikomatsakis:
@OLUWAMUYIWA oh, awesome!

nikomatsakis:
@OLUWAMUYIWA 

What I've found really useful in the past is to write manually the code I expect to generate. Maybe you can make a gist with an example input and output and we can discuss that? Else I can do it for you if you prefer :)

nikomatsakis:
@OLUWAMUYIWA I'm still interested here, but note that we did some big renamings and things, so you'll definitely have to rebase etc.

Given its age, I'm going to go ahead and close this PR-- please reach out on Zulip or comment in the GIthub issue though if you want to discuss next steps.

PR #419, opened 2022-10-02
Remove inaccurate note about mixed cycles failing to recover

kmicklas:
This may have been true in the past, but both the Salsa 2022 and original tests demonstrate recovery from a cycle where only a subset of the cycle queries have recovery functions.

### DIFF
--- a/book/src/cycles/fallback.md
+++ b/book/src/cycles/fallback.md
@@ -13,10 +13,6 @@ fn my_recover_fn(
 ) -> MyResultValue
 ```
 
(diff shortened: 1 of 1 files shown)


nikomatsakis (approved):
Thanks! And thanks for your patience here.

nikomatsakis:
I do think we got smarter about this -- do we have a test for it, is the question...

kmicklas:
Thanks for taking a look @nikomatsakis. Any chance this could get merged?

nikomatsakis:
@kmicklas I see your OP mentioned that we have tests. Sounds good!

Issue #218, opened 2020-04-30
Volatile functions were helpful to avoid `Eq` bounds for return values


orium:
Hi,

I was looking into updating chalk to the latest version of salsa.  Given that the `salsa::volatile` attribute was removed in #174 in favor of calling `Runtime::report_untracked_read()` this means that `chalk_integration::LoweringDatabase::solver()` (https://github.com/rust-lang/chalk/blob/73a74be3bc1d0cdef3f76fa529a112a0d8367ddb/chalk-integration/src/query.rs#L54) needs to return something that implements `Eq`, which is not the case.

Obviously we can create a wrapper `struct NeverEqual<T>(T)` that will always say things are different, but that is just a weird thing to have to do.

I couldn't find the reason for volatile to be removed, but offering something like that makes sense IMO.  Was there any fundamental buggy behavior involving it or is it something that can be re-implemented in one way or another? @matklad @nikomatsakis 

Edit: CC #56

bjorn3:
> Obviously we can create a wrapper struct NeverEqual<T>(T) that will always say things are different, but that is just a weird thing to have to do.

https://doc.rust-lang.org/stable/std/cmp/trait.Eq.html

> Trait for equality comparisons which are equivalence relations.
>
> This means, that in addition to a == b and a != b being strict inverses, the equality must be (for all a, b and c):
>
> * **reflexive: a == a;**
> * symmetric: a == b implies b == a; and
> * transitive: a == b and b == c implies a == c.

Emphasis mine. For `PartialEq` it would be allowed. The fact that NaN floats compare unequal to themselves is the whole reason for the `PartialEq`/`Eq` split.

orium:
Yes, it would be a wrong implementation of `Eq`, which makes it an even worse hack than it superficially looks.

nikomatsakis:
I can't remember if we ever addressed this problem? I agree that it would be good to not require `Eq`

orium:
@nikomatsakis This is still a problem.  For example, it complicates updating salsa in chalk, which is still using salsa 0.10 where the `salsa::volatile` attribute still exists.

This attribute was removed in #174 by @matklad, but I don't know the motivation for that.  It can be the case that something was broken about `salsa::volatile`, but if there wasn't anything wrong with it we could simply revert that PR.

nikomatsakis:
[...]

PR #421, opened 2022-10-06
feat: Re-add volatile queries

Marwes:
I have some values which I would like store but which do not implement `Eq`, `volatile` queries, which used to exist, avoid this requirement by simply assuming that the value always changes and therefore do not need to compare it.

An alternative to this which is mentioned in #218 is that we might allow the equality check to be customized via, say `#[salsa::eq(my_eq)]`. It seems slightly more complicated though so I just went with this as a start (since the storage assumes it can use `#[derive(Default)]` to construct right now so some codegen changes are needed).

cc #56 #218


Marwes:
Err, so I did not understand that most of the newer changes were actually squirreled away under `components/salsa-2022` so this might not even be relevant anymore due to how much (I now see) has changed.

PR #422, opened 2022-10-13
Fix error when re-exporting salsa

fowenix:
Can't use `#[salsa::input]`, `#[salsa::interned]`, `#[salsa::tracked]` when using salsa that has been re-exported from another crate.


nikomatsakis:
Now that we've renamed everything this needs a rebase, but it's probably still a good change to do. Are you up to rebase it? Apologies for letting it sit so long. :(

nikomatsakis:
Going to close for now. Feel free to re-open, and thanks again!

PR #428, opened 2022-12-21
Fixes failed cancellation due to cvar not being notified

manapointer:
This PR fixes an issue with the `cvar` condvar field of `Shared<DB>` not being notified when `Arc<Shared<DB>>`s were getting dropped. 

Previously, the condvar was being notified here:

```rust
impl<DB> Drop for Shared<DB>
where
    DB: HasJars,
{
    fn drop(&mut self) {
        self.cvar.notify_all();
    }
}
```

However, because this is implemented on `Shared<DB>`, the `drop` code only ran after all `Arc<Shared<DB>>`s (including that of the actual database and not just its snapshots) had dropped first, even though the intention was for the `drop` code to run when each individual `Arc<Shared<DB>>` was dropped so that the condvar would be notified each time.

To fix this, I've added `WrappedShared<DB>` (a newtyped `Arc<Shared<DB>>`) which implements the desired `drop` functionality and allows cancellation to happen as intended.

## Note 
On my local branch I have a test case for this functionality, although it relied on timing using `std::thread::sleep`. I figured this was less than ideal, so I decided not to include it in this PR - I'd love to get feedback on a better way to test this!


nikomatsakis:
This seems like a good solution. I've been way too busy to keep up with Salsa but I've got a New Year's Resolution to fix that. =)

components/salsa-2022/src/storage.rs
```diff
@@ -164,6 +176,7 @@ where
     DB: HasJars,
 {
     fn drop(&mut self) {
+        drop(self.jars.take());
```
nikomatsakis, line 179: This merits a comment -- I presume the idea is that the cvar is to be notified only after our ref onto `self.jars` is removed? (which makes sense)

nikomatsakis (changes requested):
Thinking a bit about this -- why not implement Drop on the `Storage<DB>` *itself*...would that work? I'm not sure if that's cleaner or not, but I was wondering if it had been considered.

manapointer:
I didn't particularly consider this (mostly just kept it implemented on `Shared<DB>` since that how it was originally done), but thinking about it more, implementing Drop on `Storage<DB>` I think might be cleaner! We might also consider moving `cvar` out of `Shared<DB>` and into `Storage<DB>` since it's only ever waited on by `Storage<DB>`, allowing us to not have to store two `Arc`s within `Shared<DB>` and simplify its definition a bit. Basically thinking something like this:

```rust
pub struct Storage<DB: HasJars> {
    shared: Option<Arc<Shared<DB>>>,
    ...
    cvar: Arc<Condvar>,
}

struct Shared<DB: HasJars> {
    jars: DB::Jars,
}

impl<DB: HasJars> Drop for Storage<DB> {
    fn drop(&mut self) {
        drop(self.shared.take());
        self.cvar.notify_all();
    }
}
```

at which point we might actually just want to store `jars` directly on `Storage<DB>`? Not 100% sure there but overall I think this might be cleaner :D What do you think?

components/salsa-2022/src/storage.rs
manapointer: That's correct!

manapointer:
Updated the PR! Thinking about it more, I ended up switching the `Drop` implementation to be on `Storage<DB>` instead of `Shared<DB>` as discussed above (ending up being a three-line change =D) but decided to leave `cvar` as a field on `Shared<DB>` so that it remains clear that `cvar` coordinates access to `jars`

components/salsa-2022/src/storage.rs
```diff
@@ -155,16 +155,29 @@ where
...
 }
 
-impl<DB> Drop for Shared<DB>
+impl<DB> Drop for Storage<DB>
 where
     DB: HasJars,
 {
     fn drop(&mut self) {
-        self.cvar.notify_all();
+        drop(self.shared.jars.take());
```
nikomatsakis, line 179:
```suggestion
        // Drop the Arc reference before the cvar is notified,
        // since other threads are sleeping, waiting for it to reach 1.
        drop(self.shared.jars.take());
```

nikomatsakis (changes requested):

nikomatsakis (approved):

PR #429, opened 2022-12-27
Add tree construction example

ryo33:
This example is an analogy of a calculation with multiple inputs that depend on each other by a unique id. The `calc` example can handle multiple inputs, but its inputs are not referred to each other.

- `FlatNode` represents an input.
- `usize` represents an input id.
- `Tree` represents an intermediate or final result of a computation.
- `fn construct_tree` represents an internal recursive computation.
- `fn entrypoint` represents an entrypoint of a computation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,5 +37,6 @@ members = [
     "components/salsa-2022-macros",
     "examples-2022/calc",
     "examples-2022/lazy-input",
+    "examples-2022/tree-construction",
     "salsa-2022-tests",
 ]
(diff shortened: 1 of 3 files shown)


nikomatsakis:
Oooh, this is a nice idea, thanks! I'll take a look.

nikomatsakis:
It will definitely need to be rebased.

nikomatsakis:
I think though it would be useful to have a kind of "patterns" directory and perhaps include this (or a refined version of it)

nikomatsakis:
Given all the recent churn, going to close for now. Feel free to re-open, and thanks again!

PR #432, opened 2023-02-03
Recover from cyclic include in lazy-input example.

pervognsen:
A potential fix for file cycles in the lazy-input example. This one I'm much, much less sure about. Conceptually, you should be able to have mutually referential files and cycle breaking should only apply at the summation level. This seems to be an issue with the lazy input approach, so I'm curious of the intended approach here.

In a "normal" program you might handle this with a top-level worklist that is populated as you process files and discover dynamic dependencies. Tri-state recursion (unloaded -> loading -> loaded) with cycle breaking would be an option too. That second option seems more congenial to Salsa's way of doing things. When doing it manually, you might do the cycle breaking by registering a file id when transitioning unloaded -> loading (the first time a file is discovered) and then any subsequent call to load the same file would truncate the cycle by returning the file id. However, that is inherently stateful (e.g. you haven't finished loading the file contents when you register the file id initially) and I'm not sure if it's possible in Salsa.

Instead of pontificating further, I thought I'd just bring up this issue and use this mock fix (not really a fix) as a discussion starter.


pervognsen:
I think I figured out a better solution more in line with the top-level worklist approach. We create a new interned struct FileId whose field is a canonicalized PathBuf. Then a bunch of the existing uses of File and ParsedFile are replaced by FileId. For example, the links field in File changes from a Vec<ParsedFile> to a Vec<FileId>. The parse function takes a FileId as an argument and returns a ParsedFile. The sum function takes a FileId as an argument, calls parse to get the ParsedFile from the FileId and then goes from there. This means that parse is now non-recursive and only sum is recursive, so you can remove the cycle recovery to sum. I have this working and will update the PR once I've cleaned it up a bit more.

nikomatsakis:
Going to close this since there has been so much activity! Still relevant though, let's come back to it.

PR #435, opened 2023-03-15
Allow `clippy::needless_lifetimes` on tracked method getters

DropDemBits:
The tracked method generation adds an extra `__db` lifetime to the signature, but clippy complains that this lifetime can be elided in some cases, so add an allow to silence this warning.

Unfortunately clippy doesn't lint inside of the warning tests, so those tests don't do anything to check that no clippy warnings are generated.


XFFXFF:
Thanks!
bors r+

PR #440, opened 2023-04-12
Change the constructor of `salsa::Id` to const fn

Y-Nak:
This change is motivated to allow interned structs to have default const instances and prefill them in db initialization.

A specific example is managing symbols as a salsa interned struct in a compiler implementation. By declaring keywords as `const` symbols, we can handle keywords without going through their internal representation as rustc does (ref: [rustc_span::symbol](https://github.com/rust-lang/rust/blob/master/compiler/rustc_span/src/symbol.rs)). 
Conceptually, it'd look like the one below.

```rust
#[salsa::interned]
pub struct Symbol {
    data: String,
}

const SELF_SYM: Symbol = Symbol(salsa::Id::from_u32(1))
...

/// This function is called in db initialization.
fn prefill(db: &dyn HirDb) {
       Symbol::new(db, "self".to_string());
       ...
}
```

### DIFF
--- a/components/salsa-2022/src/id.rs
+++ b/components/salsa-2022/src/id.rs
@@ -26,14 +26,16 @@ impl Id {
     /// but it can be useful if you are using the type as a general
     /// purpose "identifier" internally.
     #[track_caller]
-    pub fn from_u32(x: u32) -> Self {
-        assert!(x < Self::MAX_U32);
+    pub const fn from_u32(x: u32) -> Self {
         Id {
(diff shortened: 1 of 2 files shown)


nikomatsakis:
bors r+

makes sense!

Y-Nak:
The build failure originated from the trybuild test failure. This is because of the change of the error message emitted from rustc. I fixed it in the latest commit.
Ref: https://github.com/rust-lang/rust/pull/108324

PR #442, opened 2023-05-01
Upgrade to syn 2.0

lnicola:


lnicola:
This compiles, but doesn't work because of a run-time breaking change in `syn`:

```
error: expected parentheses
 --> tests/macros.rs:3:5
  |
3 |     #[salsa::invoke(another_module::another_name)]
  |     ^
```

lnicola:
CC @fee1-dead maybe you ran into this before?

components/salsa-macros/src/query_group.rs
```diff
@@ -681,13 +681,13 @@ impl TryFrom<syn::Attribute> for SalsaAttr {
...
-        if is_not_salsa_attr_path(&attr.path) {
+        if is_not_salsa_attr_path(&attr.path()) {
             return Err(attr);
         }
 
         let span = attr.span();
-        let name = attr.path.segments[1].ident.to_string();
-        let tts = attr.tokens.into();
+        let name = attr.path().segments[1].ident.to_string();
+        let tts = attr.into_token_stream().into();
```
fee1-dead, line 690: Doesn't this return the entire attribute? Previously `tokens` were the tokens that are not the path e.g. `(foo)` in `salsa::something(foo)`.

lnicola: Yeah, that's probably it. I looked at `attr.meta`, but couldn't really make it work.

Veykril:
Would be nice if we could land https://github.com/salsa-rs/salsa/pull/443 here as well should we cut another pre-release for old salsa (I don't think syn 2 demands changes for that PR)

components/salsa-macros/src/query_group.rs
fee1-dead: The `MetaList` works differently. It doesn't have the path, but it doesn't have the parenthesis either. So you would have `foo`. You could probably `quote!` this and wrap the result around parentheses as a dirty hack.

lnicola: This is actually working now AFAICT, it fails to build because of some `Span`/`DelimSpan` changes.

fee1-dead:
I think you could just use the `DelimSpan`'s two spans in an array `[span.open(), span.close()]` and that should work

lnicola:
Thanks, that leaves [this bit](https://github.com/salsa-rs/salsa/blob/master/components/salsa-2022-macros/src/jar.rs#L161-L163) here, where `Paren::span` is a `DelimSpan`, but `Ident::span()` returns a `Span` and the only way to create a `DelimSpan` is from a `Group`.

fee1-dead:
Try using the [`Group`](https://github.com/dtolnay/syn/blob/1316ec6d1787d78d9e5722d23205f1c9395e068f/src/token.rs#L617-L623) function.

fee1-dead:
actually wait nvm. you wanted to create a paren token

fee1-dead:
Ah, I think you can should still be able to call `token::Paren(span)` with a single span because of the impl [here](https://github.com/dtolnay/syn/blob/1316ec6d1787d78d9e5722d23205f1c9395e068f/src/span.rs#L50-L56).

lnicola:
Thanks, this was such a struggle :pensive:.

nikomatsakis:
Hi @lnicola -- I apologize for letting this sit for so long. If you are still interested, would you consider rebasing it? If not, I understand.

fee1-dead:
I cloned the repository, wanting to rebase this PR, but saw that the macros already use syn 2.0. Therefore there's no need to rebase this now.

lnicola:
Yup, I guess this is no longer necessary?

PR #451, opened 2023-06-19
add a link to the older version of Salsa in the document

### LINKED ISSUE #448: book: Indicate Salsa version more clearly
At the moment the book is split between the two versions, with some some pages using `salsa-2022` _with_ a warning, some using `salsa-2022` _without_ a warning, and some using the non-`salsa-2022` version.

For many pages without warning, the version can be guessed by looking for `#[salsa::tracked]` vs `#[salsa::query_group(..)]`, but for pages like [Tuning](https://github.com/salsa-rs/salsa/blob/5aefc92d61ba205aa1749b543f0903891e650ebe/book/src/tuning.md) it can be hard to determine which version the text applies to.

While the book is in this _in-between state_, I would think it would be preferable to add the intended salsa version explicitly at the top to every document, although a bit verbose perhaps.

xffxff:
related to #448 

As Salsa-2022 is no yet complete and it is still unstable, there are some users who continue to use the old version of Salsa, or prefer using it due to its more stable API. Therefore, I think it is necessary to re-deploy the book for the older version of Salsa.

I have deployed the book for the older version from commit [754eea8b5f8a31b1100ba313d59e41260b494225](https://github.com/salsa-rs/salsa/commit/754eea8b5f8a31b1100ba313d59e41260b494225)  of Salsa on https://old-salsa-rs.netlify.app/

### DIFF
--- a/book/_redirects
+++ b/book/_redirects
@@ -1,0 +1,2 @@
+# Redirects from what the browser requests to what we serve
+/                /salsa2022
--- a/book/netlify.sh
+++ b/book/netlify.sh
@@ -13,4 +13,34 @@ curl -L https://github.com/badboy/mdbook-mermaid/releases/download/v$MDBOOK_MERM
 curl -L https://github.com/Michael-F-Bryan/mdbook-linkcheck/releases/download/v$MDBOOK_LINKCHECK_VERSION/mdbook-linkcheck.v$MDBOOK_LINKCHECK_VERSION.x86_64-unknown-linux-gnu.zip -O
 unzip mdbook-linkcheck.v$MDBOOK_LINKCHECK_VERSION.x86_64-unknown-linux-gnu.zip -d ~/.cargo/bin
 chmod +x ~/.cargo/bin/mdbook-linkcheck
-mdbook build
+
+# ======================================================================
+# The following script automates the deployment of both the latest and a
+# specified older version of the 'salsa' documentation using mdbook
+
+# Store the current branch or commit
+original_branch=$(git rev-parse --abbrev-ref HEAD)
+if [ "$original_branch" == "HEAD" ]; then
+  original_branch=$(git rev-parse HEAD)
+fi
+
+mkdir -p versions  # Create a root directory for all versions
+
+# Declare an associative array to map commits to custom version directory names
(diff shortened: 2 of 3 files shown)


oeb25:
Awesome! 👍

In regards to #448, I think this PR solves the issue for users of the older (i.e., "go read this instead"), while it unfortunately still doesn't avoid confusion for users of the new version.

xffxff:
> In regards to #448, I think this PR solves the issue for users of the older (i.e., "go read this instead"), while it unfortunately still doesn't avoid confusion for users of the new version.

@oeb25 Yes, you're correct. This PR doesn't completely address issue #448. Therefore, I updated the PR's description from "close #448" to "related to #448". This means that the issue will not be closed when the PR is merged.

nikomatsakis:
Thanks for your patience. I am slowly (*slowly*) coming back online here. This looks like a good start. It'd be nice to modify the book here to just have both versions, I suppose, rather than redirecting to an external site. How hard would that be?

xffxff:
> It'd be nice to modify the book here to just have both versions, I suppose, rather than redirecting to an external site. How hard would that be?

@nikomatsakis I've set up the book to support both versions by utilizing URL paths: `xxx/salsa2022` for the latest edition and `xxx/salsa` for the older one. You can review my deployed versions on Netlify:
- salsa2022: https://grand-meerkat-a99b5f.netlify.app  or https://grand-meerkat-a99b5f.netlify.app/salsa2022/
- old salsa: https://grand-meerkat-a99b5f.netlify.app/salsa/

To make this work on our repo, we'll need to modify the publish directory from `book/book/html` to `book/versions`

My build settings: 

image

I lack the permissions to make this change, but I believe you can assist :)

nikomatsakis:
@xffxff this looks cool, yes, thank you! Let me look into making that change.

nikomatsakis:
I updated the permissions and decided to throw the dice :game_die: and merge the PR. Let's see what happens!

nikomatsakis:
@xffxff 

image

Any thoughts ;)

nikomatsakis:
Might be I need to modify the run `netlify.sh` too...?

nikomatsakis:
Never mind, I think I see the problem.

PR #470, opened 2024-01-09
Remove `Implementing the Default trait` book section

nodir-t:
https://github.com/salsa-rs/salsa/commit/73102b1e8ed4f373fbd02f28dc7884b3d9de96b3 removed `ANCHOR: default_impl` from https://github.com/salsa-rs/salsa/blob/master/examples-2022/calc/src/db.rs, but it still in the book, which results in an confusing empty code block: https://salsa-rs.github.io/salsa/tutorial/db.html#implementing-the-default-trait

The struct is already has `#[derive(Default)]`, so at this point the section is not very useful. Maybe it is better without it?


nikomatsakis:
bors r+

Thanks.

PR #474, opened 2024-03-15
fix: wrong condvar usage

### LINKED ISSUE #473: [Bug] Deadlock may happen in cancel_other_workers even when there's only one handle
I've encountered a deadlock in salsa in `cancel_other_workers `(line 151 below) even when there are no other active snapshots.

This may have occurred when there's no other active handle.

The problem is the misusage of condvar api. When waiting for signals of changing `jars`, `jar` itself needs to be protected by a lock (which is later passed to condver), and so it is in `Drop`.

https://github.com/salsa-rs/salsa/blob/4151b09d3c37e50f1a01894fcdfa776bc2909132/components/salsa-2022/src/storage.rs#L129-L153

Chronostasys:
fix #473

### DIFF
--- a/components/salsa-2022/src/storage.rs
+++ b/components/salsa-2022/src/storage.rs
@@ -139,7 +139,14 @@ where
         loop {
             self.runtime.set_cancellation_flag();
 
-            // jars need to be protected by a lock to avoid deadlocks.
+            // Acquire lock before we check if we have unique access to the jars.
+            // If we do not yet have unique access, we will go to sleep and wait for
+            // the snapshots to be dropped, which will signal the cond var associated
+            // with this lock.
+            //
+            // NB: We have to acquire the lock first to ensure that we can check for
+            // unique access and go to sleep waiting on the condvar atomically,
+            // as described in PR #474.
             let mut guard = self.shared.noti_lock.lock();
             // If we have unique access to the jars, we are done.
             if Arc::get_mut(self.shared.jars.as_mut().unwrap()).is_some() {
--- a/components/salsa-2022/src/storage.rs
+++ b/components/salsa-2022/src/storage.rs
@@ -31,7 +31,6 @@ pub struct Storage<DB: HasJars> {
     /// The runtime for this particular salsa database handle.
     /// Each handle gets its own runtime, but the runtimes have shared state between them.
     runtime: Runtime,
-
 }
 
 /// Data shared between all threads.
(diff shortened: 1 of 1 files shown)


Chronostasys:
I've tested the code on my project, and seems like the issue has gone. However as the race condition is pretty hard to meet, I don't think it's easy to prove it with test cases.

components/salsa-2022/src/storage.rs
```diff
@@ -135,19 +139,16 @@ where
         loop {
             self.runtime.set_cancellation_flag();
 
+            // jars need to be protected by a lock to avoid deadlocks.
+            let mut guard = self.shared.noti_lock.lock();
```
nikomatsakis, line 143: Can you spell this out for me, I want to be sure I understand the deadlock...

Chronostasys: > Can you spell this out for me, I want to be sure I understand the deadlock...

Well, suppose there are two threads A and B, A is holding the original database and B is holding a snapshot, then B wants to make some changes and calls the `cancel_other_workers` function. When the code runs to line 139 the B thread is not exited yet, so `Arc::get_mut` returns `None` and continues. Just before it reaches line 151, thread B exits and the `self.shared.cvar.notify_all();` on line 182 was called, but it did nothing as there's no thread waiting for a signal right now. Then A thread reaches line 151, waiting for a signal that would never be triggered.

https://github.com/salsa-rs/salsa/blob/4151b09d3c37e50f1a01894fcdfa776bc2909132/components/salsa-2022/src/storage.rs#L129-L153

https://github.com/salsa-rs/salsa/blob/4151b09d3c37e50f1a01894fcdfa776bc2909132/components/salsa-2022/src/storage.rs#L174-L184

nikomatsakis: I see. Duh. Makes sense!

components/salsa-2022/src/storage.rs
```diff
@@ -135,19 +139,16 @@ where
         loop {
             self.runtime.set_cancellation_flag();
 
+            // jars need to be protected by a lock to avoid deadlocks.
```
nikomatsakis, line 142:
```suggestion
            // Acquire lock before we check if we have unique access to the jars.
            // If we do not yet have unique access, we will go to sleep and wait for
            // the snapshots to be dropped, which will signal the cond var associated
            // with this lock.
            //
            // NB: We have to acquire the lock first to ensure that we can check for
            // unique access and go to sleep waiting on the condvar atomically,
            // as described in PR #474.
```

nikomatsakis (approved):

PR #475, opened 2024-03-16
Added destructors to inputs

### LINKED ISSUE #345: Add `remove_xxx` API that permits remove the value from input field
Old salsa has an API `remove_xxx` for removing the value from input, [remove_input.rs](https://github.com/salsa-rs/salsa/blob/367c0bfe1154d8848600bf4029343882652a358c/tests/remove_input.rs) is a test for that API.  

Should we support it in salsa-2022? Actually, I can't think of any scenario where the input needs to be removed.

obsgolem:
This is about the simplest way to solve this problem. Note that I do not do anything regarding making input structs non-`Copy`. Instead, you can handle free-after-free by using the Option return result.

I am not familiar enough with the codebase to know if this is likely to bork things elsewhere. I also don't know enough about our thread safety guarantees to be certain that the strategy of deleting each input ingredient in turn won't cause race conditions.

Resolves #345.

### DIFF
--- a/components/salsa-2022-macros/src/input.rs
+++ b/components/salsa-2022-macros/src/input.rs
@@ -47,6 +47,8 @@ impl crate::options::AllowedOptions for InputStruct {
     const LRU: bool = false;
 
     const CONSTRUCTOR_NAME: bool = true;
+
+    const DESTRUCTOR_NAME: bool = true;
 }
 
 impl InputStruct {
@@ -192,9 +194,30 @@ impl InputStruct {
                 }
             }
         } else {
+            let destructor_name = self.destructor_name();
+            let destructor: syn::ImplItemMethod = parse_quote! {
+                pub fn #destructor_name(self, __db: &#db_dyn_ty) -> Option<(#(#field_tys,)*)>
+                {
+                    let (__jar, __runtime) = <_ as salsa::storage::HasJar<#jar_ty>>::jar(__db);
+                    let __ingredients = <#jar_ty as salsa::storage::HasIngredientsFor< #ident >>::ingredient(__jar);
+                    let __db = __db.as_salsa_database();
+                    #(
+                        let #field_names = __ingredients.#field_indices.delete_entity(__db, self)?;
+                    )*
+
+                    Some((
+                        #(
+                            #field_names,
+                        )*
+                    ))
+                }
+            };
+
             parse_quote! {
                 impl #ident {
                     #constructor
+
+                    #destructor
 
                     #(#field_getters)*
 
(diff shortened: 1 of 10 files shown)


salsa-2022-tests/tests/input_deletion.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    // Creates 3 tracked structs
+    let input = MyInput::new(&db, 3);
+    assert_eq!(final_result(&db, input), 4);
+    db.assert_logs(expect![[r#"
+        [
+            "final_result(MyInput(Id { value: 1 }))",
+        ]"#]]);
+
+    let result = input.delete(&db);
```
nikomatsakis, line 64: This seems like it ought to require `&mut db`

components/salsa-2022-macros/src/input.rs
```diff
@@ -192,10 +194,31 @@ impl InputStruct {
                 }
             }
         } else {
+            let destructor_name = self.destructor_name();
+            let destructor: syn::ImplItemMethod = parse_quote! {
+                pub fn #destructor_name(self, __db: &#db_dyn_ty) -> Option<(#(#field_tys,)*)>
+                {
+                    let (__jar, __runtime) = <_ as salsa::storage::HasJar<#jar_ty>>::jar(__db);
```
nikomatsakis, line 201: ...this could be `jar_mut`...

components/salsa-2022-macros/src/input.rs
```diff
@@ -192,10 +194,31 @@ impl InputStruct {
...
+                    let __ingredients = <#jar_ty as salsa::storage::HasIngredientsFor< #ident >>::ingredient(__jar);
+                    let __db = __db.as_salsa_database();
+                    #(
+                        let #field_names = __ingredients.#field_indices.delete_entity(__db, self)?;
```
nikomatsakis, line 205: ...and this fn would take `&mut __db`...

components/salsa-2022/src/input_field.rs
```diff
@@ -76,6 +79,23 @@ where
         }
     }
 
+    pub fn delete_entity(&self, db: &dyn crate::Database, key: K) -> Option<F> {
```
nikomatsakis, line 82: ...this has to trigger a new revision

nikomatsakis (changes requested):
Basic idea seems good, but we have to tweak some details of the approach

nikomatsakis:
Given all the recent churn, going to close, get in touch though if you want to update!

PR #481, opened 2024-04-02
Struct field ingredients

nikomatsakis:
This branch overhauls how tracked structs are implemented to make them more efficient -- it also adds tests for some edge cases along the way. In the new design, instead of treating tracked structs like a bunch of functions with specified values (omg the hashing) it creates a single shared hashmap to map the tracked struct id to its fields.

The plan is to eventually transition to a "Salsa 3.0" design in which tracked structs can be literal pointers into this shared value, but that will require more extensive changes that I have to write-up and think through (there are some edge cases in the unsafe code that are a bit subtle). This seemed like a useful stopping point regardless.

### DIFF
--- a/components/salsa-2022/src/tracked_struct.rs
+++ b/components/salsa-2022/src/tracked_struct.rs
@@ -1,14 +1,18 @@
 use std::fmt;
 
+use crossbeam::queue::SegQueue;
+
 use crate::{
     cycle::CycleRecoveryStrategy,
+    hash::FxDashMap,
     ingredient::{fmt_index, Ingredient, IngredientRequiresReset},
     ingredient_list::IngredientList,
     interned::{InternedData, InternedId, InternedIngredient},
     key::{DatabaseKeyIndex, DependencyIndex},
+    plumbing::transmute_lifetime,
     runtime::{local_state::QueryOrigin, Runtime},
     salsa_struct::SalsaStructInDb,
-    Database, Event, IngredientIndex, Revision,
+    Database, Durability, Event, IngredientIndex, Revision,
 };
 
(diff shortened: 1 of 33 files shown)


components/salsa-2022/src/tracked_struct.rs
```diff
@@ -1,21 +1,64 @@
...
+    /// The id type used to define instances of this struct.
+    /// The [`TrackedStructIngredient`][] contains the interner
+    /// that will create the id values.
+    type Id: InternedId;
+
+    /// A (possibly empty) tuple of the fields for this struct.
+    type Fields;
+
+    /// A array of [`Revision`][] values, one per each of the value fields.
+    /// When a struct is re-recreated in a new revision, the
```
solson, line 36: This comment ends mid-sentence.

nikomatsakis: Thanks! Fixed.

components/salsa-2022/src/tracked_struct/tracked_field.rs
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    fn maybe_changed_after(
+        &self,
+        _db: &DB,
+        input: crate::key::DependencyIndex,
+        revision: crate::Revision,
+    ) -> bool {
+        let id = <C::Id>::from_id(input.key_index.unwrap());
+        eprintln!("maybe_changed_after({id:?}, {revision:?})");
```
Chronostasys, line 91: Hi @nikomatsakis ,  I noticed that an eprintln!() statement was accidentally left in the PR. Could you please check and remove it if it's not needed? Thanks!

PR #482, opened 2024-04-03
[breaking-salsa-2022] Debug with db: ignore deps, support customization

nikomatsakis:
This changes how `DebugWithDb` works:

* All field values are always included (useful) and thus the `include_all_fields` flag is removed (it was annoying).
* Dependencies from those field values are ignored by salsa (useful, but potentially dangerous).
* You can use `#[customize(DebugWithDb)]` to skip generating the impl for a given type.

I think overall this is an improvement, though it is a breaking change.

### DIFF
--- a/salsa-2022-tests/tests/debug.rs
+++ b/salsa-2022-tests/tests/debug.rs
@@ -4,7 +4,13 @@ use expect_test::expect;
 use salsa::DebugWithDb;
 
 #[salsa::jar(db = Db)]
-struct Jar(MyInput, ComplexStruct, leak_debug_string);
+struct Jar(
+    MyInput,
+    ComplexStruct,
+    leak_debug_string,
+    DerivedCustom,
+    leak_derived_custom,
+);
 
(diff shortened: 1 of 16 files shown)


salsa-2022-tests/tests/debug.rs
```diff
@@ -44,15 +50,75 @@ fn input() {
...
+impl DebugWithDb<dyn Db + '_> for DerivedCustom {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>, db: &dyn Db) -> std::fmt::Result {
+        write!(
+            f,
+            "{:?} / {:?}",
+            self.my_input(db).debug(db),
+            self.value(db)
+        )
+    }
+}
```
MichaReiser, line 105: Nice: That's something I've come across too where some tracked structs have fields that are salsa ingredients and the "default" generated `DebugWithDb` implementation didn't emit them with `debug`. Thanks for making this change.

PR #490, opened 2024-05-25
salsa 3.0

nikomatsakis:
This branch implements the Salsa 3.0 plan, at least in broad strokes. It's not quite ready to merge, needs more documentation, especially around the use of unsafe. Rather than write an extensive comment here, I'm going to push some documentation commits.


nikomatsakis:
I wondered what miri would make of all this. I had forgotten we had tests for it, though. All the more reason to write up how it works, so I can try to figure out if miri is giving me a legit error or not ...

nikomatsakis:
Reproduced locally. The good news is that just running on salsa-2022-tests, lots of tests fail...

```
error: 19 targets failed:
    `-p salsa-2022-tests --test debug`
    `-p salsa-2022-tests --test deletion`
    `-p salsa-2022-tests --test deletion-cascade`
    `-p salsa-2022-tests --test expect_reuse_field_x_of_a_tracked_struct_changes_but_fn_depends_on_field_y`
    `-p salsa-2022-tests --test hello_world`
    `-p salsa-2022-tests --test is_send_sync`
    `-p salsa-2022-tests --test preverify-struct-with-leaked-data`
    `-p salsa-2022-tests --test specify-only-works-if-the-key-is-created-in-the-current-query`
    `-p salsa-2022-tests --test specify_tracked_fn_in_rev_1_but_not_2`
    `-p salsa-2022-tests --test tracked-struct-id-field-bad-eq`
    `-p salsa-2022-tests --test tracked-struct-id-field-bad-hash`
    `-p salsa-2022-tests --test tracked-struct-unchanged-in-new-rev`
    `-p salsa-2022-tests --test tracked-struct-value-field-bad-eq`
    `-p salsa-2022-tests --test tracked-struct-value-field-not-eq`
    `-p salsa-2022-tests --test tracked_fn_on_tracked`
    `-p salsa-2022-tests --test tracked_fn_on_tracked_specify`
    `-p salsa-2022-tests --test tracked_fn_read_own_entity`
    `-p salsa-2022-tests --test tracked_fn_read_own_specify`
    `-p salsa-2022-tests --test tracked_with_struct_db`
```

Good news because it indicates the error is not just some random edge case. I am pushing past the trivial stacked borrows use case, so I wouldn't be surprised if we are running into some issues in the model itself, but I don't know anything at this point. I'm going to start digging into the `hello-world` example I think as it seems .. well .. the most basic. :)

nikomatsakis:
I don't understand exactly what you are trying  to do -- can you elaborate?

components/salsa-2022/src/id.rs
```diff
@@ -65,17 +65,46 @@ impl From<Id> for usize {
...
+    /// struct, the function will panic if the tracked struct
+    /// has not yet been created in this revision). Salsa's
+    /// dependency tracking typically ensures this does not
+    /// occur, but it is possible for a user to violate this
+    /// rule.
+    fn lookup_id(id: Id, db: DB) -> Self;
+}
+
+/// Internal Salsa trait for types that are just a newtype'd [`Id`][].
+pub trait FromId: AsId + Copy + Eq + Hash + Debug {
```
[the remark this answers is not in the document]
MichaReiser: > I don't understand exactly what you are trying  to do -- can you elaborate?

I'm sorry for that. Let me follow up once I have some concrete code to share. I think that will help explaining.

book/src/reference/durability.md
```diff
@@ -0,0 +1,13 @@
+# Durability
+
+"Durability" is an optimization that can greatly improve the performance of your salsa programs.
+Durability specifies the probably that an input's value will change.
```
etaoins, line 4:
```suggestion
Durability specifies the probability that an input's value will change.
```

book/src/overview.md
```diff
@@ -147,12 +150,13 @@ Tracked functions can return any clone-able type. A clone is required since, whe
...
-Unlike inputs, they can only be created inside a tracked function, and their fields can never change once they are created.
-Getter methods are provided to read the fields, but there are no setter methods[^specify]. Example:
+Unlike inputs, they can only be created inside a tracked function, and their fields can never change once they are created (until the next revision, at least).
+Getter methods are provided to read the fields, but there are no setter methods. 
+Example:
 
 ```rust
 #[salsa::tracked]
-struct Ast {
+struct Ast<'db> {
```
MichaReiser, line 159: Does the addition of the `db` lifetime also allow queries to return data that reference the DB?

One use case that we have is that we need a mapping from `AstNode` -> `Id` where `Id` for example uniquely identifie's a scope, or a symbol in the program. 

The challenge we're facing is that our `Ast` doesn't use `Arc`s internally, thus cloning a `Node` always clones the entire sub-tree. Our "work-around" for this is to keep hold to the AST's root (wrapped in an Arc) and store a raw pointer referencing the actual node. This works pretty well, but requires heavy use of `Arc`s (a lot of cloning). I "think" your changes would allow us to directly store a `&'db Expr` instead. 

If not, then the "work-around" would just be to make the `AstNode` -> `Id` map a salsa tracked so that we get access to the db lifetime

book/src/plumbing/db_lifetime.md
```diff
@@ -0,0 +1,238 @@
...
+* A tracked function `f` executes in revision R0 and creates a tracked struct with `#[id]` fields `K` for the first time.
+    * `K` will be stored in the interning hashmap and mapped to a fresh identifier `id`.
+    * The identifier `id` will be used as the key in the `StructMap` and point to a freshly created allocation `alloc : Alloc`.
+    * A `ts: TS<'db>` is created from the raw pointer `alloc` and returned to the user.
+* The value of the field `field` is accessed on the tracked struct instance `ts` by invoking the method `ts.field(db)`
+    * *Unsafe:* This accesses the raw pointer to `alloc`.* A new revision R1 begins.
+* The tracked function `f` does not re-execute in R1.
+* The value of the field `field` is accessed on the tracked struct instance `ts` by invoking the method `ts.field(db)`
+    * *Unsafe:* This accesses the raw pointer to `alloc`.* A new revision R2 begins.
+* The tracked function `f` does reexecute in R1 and it again creates a tracked struct with key `K` and with (Some) distinct field values.
```
MichaReiser, line 30: Should this be `R2`?
```suggestion
* The tracked function `f` does reexecute in R2 and it again creates a tracked struct with key `K` and with (Some) distinct field values.
```

book/src/plumbing/db_lifetime.md
```diff
@@ -0,0 +1,238 @@
...
+When we say in this page "users cannot do X", we mean without Undefined Behavior (e.g., by transmuting integers around etc).
+
+## Proof obligations
+
+Here is a typical sequence of operations for a tracked struct along with the user operations that will require us to prove unsafe assertions:
+
+* A tracked function `f` executes in revision R0 and creates a tracked struct with `#[id]` fields `K` for the first time.
+    * `K` will be stored in the interning hashmap and mapped to a fresh identifier `id`.
+    * The identifier `id` will be used as the key in the `StructMap` and point to a freshly created allocation `alloc : Alloc`.
+    * A `ts: TS<'db>` is created from the raw pointer `alloc` and returned to the user.
```
MichaReiser, line 24: It's not entirely clear to readers what `ts` refers to. There's an explanation below in the definitions section. Maybe add it to the sentence in the first bullet point like this

> creates a tracked struct (ts) with `#[id]` fields

book/src/plumbing/db_lifetime.md
```diff
@@ -0,0 +1,238 @@
...
+* what about weird `Eq` implementations and the like? Do we have to make those unsafe?
+
+## Theorem: To access a tracked struct `ts` in revision R, the defining query `f(..)` must have either *executed* or been *validated* in the revision R.
+
+This is the core bit of reasoning underlying most of what follows.
+The idea is that users cannot "forge" a tracked struct instance `ts`.
+They must have gotten it through salsa's internal mechanisms.
+This is important because salsa will provide `&`-references to fields within that remain valid during a revision.
+But at the start of a new revision salsa may opt to modify those fields or even free the allocation.
+This is safe because users cannot have references to `ts` at the start of a new revision.
```
MichaReiser, line 89: Nit: I'm not sure if that's mentioned above. But I think that's because the operations require a `&'db mut Db`

MichaReiser (commented):
This is very exciting. I've already ran into an issue once where I hold on to a tracked strut across db versions and Salsa then panicked at runtime. Catching this at compile time would be great. It should also allow us to implement garbage collection under safe assumptions.

book/src/overview.md
```diff
@@ -114,9 +114,12 @@ Finally, you can also modify the value of an input field by using the setter met
 Since this is modifying the input, the setter takes an `&mut`-reference to the database:
 
 ```rust
-file.set_contents(&mut db, String::from("fn foo() { /* add a comment */ }"));
+file.set_contents(&mut db).to(String::from("fn foo() { /* add a comment */ }"));
 ```
 
+Note that the setter method `set_contents` returns a "builder".
+This gives the ability to set the [durability](./reference/durability.md) and other advanced concepts.
```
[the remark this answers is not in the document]
nikomatsakis: This change predates the PR, btw, the docs were just not updated.

book/src/overview.md
nikomatsakis: I did intend to support storing `&T` references like that, but it's a subtle case, and I've gone back and forth on whether it works with the stacked borrows rules etc. 

Suppose you do `let f = ast.field(db)` in R1 and it yields a `&'db Foo` (reference to some field of `ast`) and then you store that in the database as the result of a query (or part of the result). Now say we start a new revision R2 and, in R2, the value of `field` changes. This means that `f` (considered as a pointer) still points to the same memory, but the value behind `f` has changed. There are two challenges: (a) under the stacked borrow rules, it is UB to use `f` again; (b) should we consider functions that were dependent on `f` as needing to be re-executed?

I've tried to write an exploration of this question in this comment like 3 times but it keeps getting unwieldy. I think I will defer it for documentation or in-person discussion, it's a good one. I'm not entirely sure if and under what conditions this can be made to be safe at the moment. =)

That said, I also want to mention a feature I've been considering that I think is may help with your use case. The idea would be to make it easy to have a value that carries a memory arena and references into that arena. This is meant to model things like MIR, where we have some data structure that represents a function, and to allow it to go through phases where it is changed and updated, but without requiring everything to be in vectors nor requiring everything to be cloned constantly. I'm not sure the ergonomics exactly but the idea is roughly that you can declare a struct with two lifetimes...

```rust
#[in_arena(AstRoot)]
struct Ast<'ast, 'db: 'ast> {
   data: AstData<'ast, 'db>,
   children: Vec<&'ast Ast<'ast, 'db>>,
}
```

...and the procedural macro will create a type `AstRoot` that "hides" the first one:

```rust
struct AstRoot<'db> {
    arena: Arc<MemoryArena>,
    root: &'static Ast<'static 'static>, // <-- the lifetimes here are obviously lies
}
```

Later you can do `root.open(|r| { .. })` to work with the data. One of the goals is that you can create new, derived values based on the same arena that have different pointers -- so e.g. it should be possible to extra subvalues from the tree. Each of them would carry a reference count to the same base arena.

book/src/plumbing/db_lifetime.md
nikomatsakis: Yes

nikomatsakis:
I am debating about how to proceed here. It's taking me far longer to write these docs than I hoped. And they are complicated. I'm somewhat inclined to merge the PR so that people can start trying it out and then continue to write the safety docs in parallel.

But I'm also eager to not overlook the safety docs. 

Thoughts from others?

nikomatsakis:
One thing I don't know 100% is how much the memory safety of the scheme relies on salsa correctly tracking dependencies. I'd prefer if it didn't, because I don't trust users not to poke holes in that system, but I feel like as I talked out the reasoning I found myself wanting to rely on "and we know that the fn will have been re-executed". But then I think I did add some judicious panics in there partly for this reason (i.e., to double check users and panic if they messed things up), I just don't know yet if that's enough.

nikomatsakis:
OK, I pushed an update to the safety documentation that I think covers all the key details in a readable way. It also identifies what I believe to be a flaw in the current setup -- I think it's possible for users to abuse salsa through leaked structs or nondeterminism and access freed memory. It's not easy to do, and I should make a test (will do later).

There are various ways to close this hole, and in particular I think one of the planned improvements I had in mind (adopting a sharded-slab like structure) would serve for it. But fundamentally we need some way to test if a pointer is still "in bounds". 

I'm inclined to land the PR in its current state and work on those improvements as follow-up.

book/src/overview.md
MichaReiser: > That said, I also want to mention a feature I've been considering that I think is may help with your use case. The idea would be to make it easy to have a value that carries a memory arena and references into that arena. 

Yeah, that sounds very similar to our "work-around", except that it is more flexible and the unsafety is handled by salsa instead of sprinkled through our code.

MichaReiser:
> I'm somewhat inclined to merge the PR so that people can start trying it out and then continue to write the safety docs in parallel.

I don't expect to have time to try out the new branch before the end of next week, but having something to play with certainly is nice (although that's also possible by pointing cargo to this PR's revision).

> I'm inclined to land the PR in its current state and work on those improvements as follow-up.


I'm supportive of this. I would find separate PR's useful for more focused discussions around specific improvements.

nikomatsakis:
OK, I'm going to land this change. We can discuss future developments on Zulip. One thing I plan to do is explore using sharded-slab in the implementation, which will make the safety concerns much simpler but (maybe, not clearly) cost a bit of performance. It'd be useful to have measurements at some point.

nikomatsakis:
That said, I do feel certain that the lifetimes and requirement to use the `Update` trait etc is good. It is a bit more restrictive but gives us a lot of room for future improvement.

ematipico:
I think you should merge and publish this version.

I want to start experimenting, so I can help filing possible issues and help with the developments, if they are accepted.

nikomatsakis:
I'm going to land it. I'm not inclined to publish though I would be, I think, ok with something like `3.0-alpha` or something

PR #499, opened 2024-06-17
book: Creating inputs no longer requires `&mut dyn Db`

camelid:
This was changed in #404.

### DIFF
--- a/book/src/overview.md
+++ b/book/src/overview.md
@@ -56,15 +56,17 @@ pub struct ProgramFile {
 ```
 
 You create an input by using the `new` method.
-Because the values of input fields are stored in the database, you also give an `&mut`-reference to the database:
+Because the values of input fields are stored in the database, you also give an `&`-reference to the database:
 
 ```rust
 let file: ProgramFile = ProgramFile::new(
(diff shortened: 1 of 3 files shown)


book/src/overview.md
```diff
@@ -140,7 +143,7 @@ The algorithm Salsa uses to decide when a tracked function needs to be re-execut
 Tracked functions have to follow a particular structure:
 
 - They must take a `&`-reference to the database as their first argument.
-  - Note that because this is an `&`-reference, it is not possible to create or modify inputs during a tracked function!
```
camelid, line 143: It does seem a little unfortunate that inputs can be created in tracked functions now. Although since you still need `&mut` to modify an input, I guess it doesn't really matter because they'll basically behave like tracked structs.

nikomatsakis: I think that was intentional -- so that you can have lazilly created inputs

nikomatsakis (approved):

book/src/overview.md
camelid: Ah, true.

Issue #504, opened 2024-06-19
Why do we use eyre crate?


nikomatsakis:
Salsa is using `eyre`, but that seems oddly opinionated to me. I forget why. Is there a good reason?

Veykril:
We only use this in the examples now so I think we can close this

PR #501, opened 2024-06-18
Remove the old Salsa, rename `salsa-2022` crate to `salsa`

mbrobbel:
Closes #493.

PR #505, opened 2024-06-19
Remove some unused dependencies

mbrobbel:
Follow-up of #501: removing some unused dependencies.

Relates to #504.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ smallvec = "1.0.0"
 
 [dev-dependencies]
 derive-new = "0.5.9"
-env_logger = "*"
 expect-test = "1.4.0"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -27,7 +26,6 @@ expect-test = "1.4.0"
 eyre = "0.6.8"
 notify-debouncer-mini = "0.2.1"
 ordered-float = "3.0"
-parking_lot = "0.12.1"
```
MichaReiser, line 30: I think `env_logger` can also be removed. It changes the output of one test because `cargo` wont suggest a type from the `env_logger` crate anymore but I'm not sure if it's work keeping a dependency just because of that.

MichaReiser (approved):

PR #507, opened 2024-06-19
Remove `bors.toml` and update GitHub Actions

mbrobbel:
Another follow-up of #501. 

Removes `bors.toml`, because it's no longer used/needed (it seems salsa is using a merge queue now). Also removes the empty `rustfmt.toml`.

Updates some GitHub Actions to the latest, removes the `bors` branch push triggers, and removes the use of the unmaintained `actions-rs` actions. Also bumps the `mdbook` versions to the latest and adds cacheing for the rust build jobs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,13 +4,11 @@ on:
   push:
     branches:
       - master
-      - staging
-      - trying
   pull_request:
     paths:
       - "**.rs"
       - "**/Cargo.*"
-      - ".github/workflows/**.yml"
+      - ".github/workflows/test.yml"
       - "tests/compile-fail/**.stderr"
   merge_group:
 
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -31,51 +29,62 @@ jobs:
...
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-${{ hashFiles('**/Cargo.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-${{ steps.rust-toolchain.outputs.cachekey }}-
+            ${{ runner.os }}-cargo-
```
MichaReiser, line 50: What's your experience for using github action's cache directly vs https://github.com/Swatinem/rust-cache. Or does this set up caching in addition to Swatinem because https://github.com/actions-rust-lang/setup-rust-toolchain enables caching by default.

MichaReiser (approved):
Thanks

.github/workflows/test.yml
mbrobbel: I'm using a similar setup for other crates and it works fine to reduce job times. This is not in addition to anything else, this PR proposes to use https://github.com/dtolnay/rust-toolchain/.

Checking in a `Cargo.lock` file would make this more efficient.

nikomatsakis: I don't really have any opinion on these things. Happy to trust your judgment here.

nikomatsakis (approved):

PR #508, opened 2024-06-20
Manual `Debug` impl for `salsa::Id`

MichaReiser:
I've often been confused by the fact that the IDs printed by `DebugWithDb` don't match with the ids printed when dumping a `salsa::Id` with `Debug`. 

This PR replaces the derived `Debug` implementation of `salsa::Id` with a manual `Debug` implementation that ensures that the IDs printed by `Debug` and `DebugWithdb` are consistent. 

## Open Questions

* Should `Id` use `f.debug_tuple` to preserve the name in the `Debug` output?
* Should `DebugWithDb` continue to use `as_u32` directly or should it just call the `Id`'s `Debug` implementation (It would change the output if we decide that `Debug` should output `Id(num)`.


src/id.rs
```diff
@@ -40,6 +40,12 @@ impl Id {
     }
 }
 
+impl std::fmt::Debug for Id {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.as_u32().fmt(f)
```
nikomatsakis, line 45: Hmm. I think it would be better to print *something* other than just a number -- e.g., `Id(N)`. I've usually found there is value in having some more "grep-able".

nikomatsakis (changes requested):
Thoughts?

nikomatsakis:
>     * Should `Id` use `f.debug_tuple` to preserve the name in the `Debug` output?

I think yes :)

>     * Should `DebugWithDb` continue to use `as_u32` directly or should it just call the `Id`'s `Debug` implementation (It would change the output if we decide that `Debug` should output `Id(num)`.

I like that it uses the debug impl, keep that I think.

tests/tracked_with_struct_db.rs
```diff
@@ -51,16 +51,24 @@ fn execute() {
...
         MyTracked {
-            [salsa id]: 1,
+            [salsa id]: Id(
+                1,
+            ),
             data: MyInput {
-                [salsa id]: 0,
+                [salsa id]: Id(
+                    0,
+                ),
```
MichaReiser, line 60: Okay, the fact that `debug_tuple` splits the ID across multiple lines in alternate mode is too verbose. I'm just gonna use a regular write call.

MichaReiser:
I updated the Debug implementation to write `Id({})`.

MichaReiser:
@nikomatsakis I addressed the feedback. Would you mind taking another look?

MichaReiser:
Ha, I probably need to start over. @nikomatsakis I don't mind rebasing (re-doing) if this is something we want to merge.

nikomatsakis:
I'm in favor of this but needs a rebase

MichaReiser:
> I'm in favor of this but needs a rebase

Done

nikomatsakis (approved):

PR #513, opened 2024-06-28
Allow `clippy::unused_unit` in `salsa::Update` macro

Y-Nak:
Clippy emits an `unused_unit` warning when the `salsa::Update` macro is applied to an enum that contains a unit variant.
This PR suppresses the warning.


mbrobbel (approved):

nikomatsakis:
Thanks! I think I included this fix already in my massive PR though (wish I'd seen this earlier). Going to close this.

PR #514, opened 2024-07-01
Update `dashmap`, `hashlink`, and `rustc-hash`

MichaReiser:
Update `dashmap`, `hashlink` and `rustc-hash` to the most recent major version.

**dashmap**

> This release contains performance optimizations, most notably 10-40% gains on Apple Silicon but also 5-10% gains when measured in Intel Sapphire Rapids. This work was accomplished in:

It's a major because there's a slight API change. 

**hashlink**

> API incompatible change: Don't panic in reserve and try_reserve in cases where a rehash is needed. Previously would panic, adds the proper bounds on reserve methods to do a rehash (previously bounds were missing, and reserve would result in panics). (Thank you cuviper!)

I think that's fine

**rustc-hash**

> Replace hash with faster and better finalized hash. This replaces the previous "fxhash" algorithm originating in Firefox with a custom hasher designed and implemented by Orson Peters ([orlp](https://github.com/orlp)). It was measured to have slightly better performance for rustc, has better theoretical properties and also includes a signficantly better string hasher.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,13 +11,13 @@ description = "A generic framework for on-demand, incrementalized computation (e
 arc-swap = "1.6.0"
 boomphf = "0.6.0"
 crossbeam = "0.8.1"
-dashmap = "5.3.4"
-hashlink = "0.8.0"
+dashmap = "6.0.1"
+hashlink = "0.9.1"
 indexmap = "2"
 log = "0.4.5"
 orx-concurrent-vec = "1.10.0"
 parking_lot = "0.12.1"
-rustc-hash = "1.1.0"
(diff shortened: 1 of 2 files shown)


mbrobbel (approved):

davidbarsky:
Using this commit against https://github.com/davidbarsky/salsa-benchmarks/, I saw an 11-12% improvement (418.66ns versus 466.92ns on `compare/new_salsa_length/1`).

nikomatsakis (approved):

nikomatsakis:
Needs rebase but r=me

MichaReiser:
@nikomatsakis done

PR #516, opened 2024-07-05
chore: use tracing instead of log

davidbarsky:
Besides my well-known biases, rust-analyzer uses `tracing`, and given how big of a difference Salsa 3.0 it, I figured it makes sense to switch over.

(We switched over a few months back in rust-analyzer's vendored fork of Salsa.)


nikomatsakis:
@davidbarsky I'm happy-ish to merge this PR (I'm a bit grumpy that we don't have a clearer standard in Rust land, but that's neither your fault nor tracing's, really) but it needs a rebase =)

davidbarsky:
> @davidbarsky I'm happy-ish to merge this PR (I'm a bit grumpy that we don't have a clearer standard in Rust land, but that's neither your fault nor tracing's, really) but it needs a rebase =)

Rebased! And yeah: I think they occupy their own, slightly distinct niches at this point (particularly from an FFI perspective), but I wouldn't mind resolving those gaps.

nikomatsakis (approved):

MichaReiser:
I'm not sure if it is intentional but this log can blow up, depending on the value (in our case, it logs the entire AST :laughing:).

https://github.com/MichaReiser/salsa/blob/2f4f80fe23cfb968ffe58fdf99aa77eacf6dbf9a/src/function/maybe_changed_after.rs#L109

Should we exclude the value from the `Memo`'s `Debug` impl?

davidbarsky:
> I'm not sure if it is intentional but this log can blow up, depending on the value (in our case, it logs the entire AST 😆).
> 
> https://github.com/MichaReiser/salsa/blob/2f4f80fe23cfb968ffe58fdf99aa77eacf6dbf9a/src/function/maybe_changed_after.rs#L109
> 
> Should we exclude the value from the `Memo`'s `Debug` impl?

@MichaReiser Yes, it's probably a good idea to exclude it. sorry about that!

Issue #521, opened 2024-07-19
no_eq


nikomatsakis:
The `no_eq` flag is ignored in tracked functions after #518. @MichaReiser commented on Zulip.

-----

There might also be an issue around interned or I'm using it incorrectly. But

```rust
#[salsa::tracked]
pub(crate) fn resolve_module_query<'db>(
    db: &'db dyn Db,
    module_name: ModuleNameIngredient<'db>,
) -> Option<Module> {
    let _span = tracing::trace_span!("resolve_module", ?module_name).entered();

    let name = module_name.name(db);

    let (search_path, module_file, kind) = resolve_name(db, name)?;

    let module = Module::new(name.clone(), kind, search_path, module_file);

    Some(module)
}

#[salsa::interned]
pub(crate) struct ModuleNameIngredient<'db> {
    #[return_ref]
    pub(super) name: ModuleName,
}
```

No longer compiles and fails with

```
error[E0277]: the trait bound `ModuleNameIngredient<'_>: LookupId<'_>` is not satisfied
  --> crates/red_knot_module_resolver/src/resolver.rs:30:1
   |
30 | #[salsa::tracked]
   | ^^^^^^^^^^^^^^^^^ the trait `FromId` is not implemented for `ModuleNameIngredient<'_>`, which is required by `ModuleNameIngredient<'_>: LookupId<'_>`
   |
   = note: required for `ModuleNameIngredient<'_>` to implement `LookupId<'_>`
   = note: this error originates in the macro `salsa::plumbing::setup_tracked_fn` which comes from the expansion of the attribute macro `salsa::tracked` (in Nightly builds, run with -Z macro-backtrace for more info)

```

nikomatsakis:
Mentoring tips:

We can use an approach similar to the [existing `is_specifiable` flag](https://github.com/salsa-rs/salsa/blob/master/components/salsa-macros/src/tracked_fn.rs#L71). Basically pass in a boolean flag indicating whether `no_eq` was set or not, and then in the expanded macro use `$zalsa::macro_if!`. We need to change what this code does...

https://github.com/salsa-rs/salsa/blob/1c69d3ba7b00b36fb1594edbb8b5de05d27b1419/components/salsa-macro-rules/src/setup_tracked_fn.rs#L161-L166


...so that it does something like

```rust
$zalsa::macro_if! {
    if $no_eq {
        false
    } else {
        $zalsa::should_backdate_value(old_value, new_value)
    }
}
```

[...]

PR #523, opened 2024-07-20
Allow interned values as tracked function arguments

MichaReiser:
This fixes the second part of https://github.com/salsa-rs/salsa/issues/521 where tracked functions didn't accept interned values

### DIFF
--- a/components/salsa-macro-rules/src/setup_interned_struct.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct.rs
@@ -93,6 +93,12 @@ macro_rules! setup_interned_struct {
                 }
             }
 
+            impl<$db_lt> $zalsa::LookupId<$db_lt> for $Struct<$db_lt> {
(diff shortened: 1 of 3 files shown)


components/salsa-macro-rules/src/setup_interned_struct.rs
```diff
@@ -93,6 +93,12 @@ macro_rules! setup_interned_struct {
                 }
             }
 
+            impl<$db_lt> $zalsa::LookupId<$db_lt> for $Struct<$db_lt> {
+                fn lookup_id(id: salsa::Id, db: &$db_lt dyn $zalsa::Database) -> Self {
+                    $Configuration::ingredient(db).interned_value(id)
```
MichaReiser, line 98: Not 100% sure about this but it compiles and the test passes ;)

nikomatsakis: Yeah, this looks good.

nikomatsakis (approved):

PR #527, opened 2024-07-24
re-enable parallel tests

nikomatsakis:
This branch re-enables the parallel tests. It does not yet add the nicer parallel APIs imagined for Salsa 3.0.

Changes:

* The per-thread state is now stored in thread-local variables.
* To create a per-thread 'snapshot', you store the database in a `salsa::Handle`, which can simply be cloned.
* When the `get_mut` method on the `Handle` is called, any outstanding clones will be cancelled (I added a test for this).


src/handle.rs
```diff
@@ -49,6 +49,12 @@ impl<Db: HasStorage> Handle<Db> {
         let storage = self.db.storage();
         storage.runtime().set_cancellation_flag();
 
+        self.db.salsa_event(Event {
```
MichaReiser, line 52: Nit: We could probably use an `AtomicUsize` for `clones`

MichaReiser (commented):
I like the idea of a handle. It also removes the need for implementing a snapshot function on any other field stored on `Db` (which normally involves wrapping that field in an `Arc`). 

The only thing that I'm slightly concerned about is the cost of using thread locals for every Salsa operation.

nikomatsakis:
> thread 'parallel_cancellation::execute' panicked at /home/runner/work/salsa/salsa/src/handle.rs:39:36:

interesting, I don't see this...

src/handle.rs
nikomatsakis: Yes, we could, we do need a mutex for signalling. This is simpler.

PR #530, opened 2024-07-25
Add benchmark for query with many tracked structs (pre-rewrite)

MichaReiser:


MichaReiser:
This is not intended to merge. See https://github.com/salsa-rs/salsa/pull/529

PR #532, opened 2024-07-26
Fix assertion for same DB in `DbGuard`

MichaReiser:
Working on https://github.com/salsa-rs/salsa/pull/529 I ran into the issue that the assertion in `DbGuard` asserting that the DB is unchanged consistently fails. 

The underlying problem is that `NonNull::eq` not only compares the pointer but also the metadata which is unreliable for trait object pointers

> When comparing wide pointers, both the address and the metadata are tested for equality. However, note that comparing trait object pointers (*const dyn Trait) is unreliable: pointers to values of the same underlying type can compare inequal (because vtables are duplicated in multiple codegen units), and pointers to values of different underlying type can compare equal (since identical vtables can be deduplicated within a codegen unit).
[source](https://doc.rust-lang.org/std/ptr/fn.eq.html)

This PR changes the assertion to use `std::ptr::addr_eq` which only compares the address without the metadata.

Fixes https://github.com/salsa-rs/salsa/issues/536

### DIFF
--- a/src/local_state.rs
+++ b/src/local_state.rs
@@ -83,12 +83,16 @@ impl LocalState {
         impl<'s> DbGuard<'s> {
             fn new(state: &'s LocalState, db: &dyn Database) -> Self {
                 if let Some(current_db) = state.database.get() {
+                    let new_db = NonNull::from(db);
+
                     // Already attached? Assert that the database has not changed.
-                    assert_eq!(
-                        current_db,
-                        NonNull::from(db),
-                        "cannot change database mid-query",
-                    );
+                    // NOTE: It's important to use `addr_eq` here because `NonNull` not only compares the address but also the type's metadata.
+                    if !std::ptr::addr_eq(current_db.as_ptr(), new_db.as_ptr()) {
+                        panic!(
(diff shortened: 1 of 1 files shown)


src/local_state.rs
```diff
@@ -83,12 +83,16 @@ impl LocalState {
...
                 if let Some(current_db) = state.database.get() {
+                    let new_db = NonNull::from(db);
+
                     // Already attached? Assert that the database has not changed.
-                    assert_eq!(
-                        current_db,
-                        NonNull::from(db),
-                        "cannot change database mid-query",
-                    );
+                    // NOTE: It's important to use `addr_eq` here because `NonNull` not only compares the address but also the type's metadata.
```
camelid, line 89:
```suggestion
                    // NOTE: It's important to use `addr_eq` here because `NonNull::eq` not only compares the address but also the type's metadata.
```

MichaReiser:
@nikomatsakis do you think we could merge this to mitigate the immediate panic. I know you're working on a more long-term fix, but it would unblock some short-term usage.

nikomatsakis (approved):
OK, I'm a bit confused why the metadata is different but... sounds good.

PR #535, opened 2024-07-26
Introduce `Input::builder` API

MichaReiser:
This PR introduces a new `Input::builder` API. 

```rust
let input_high = MyInput::builder(&db).durability(Durability::HIGH).new(2200);
```

It's intentionally limited for now but we could extend it in the future to e.g. allow setting per-field durability. 

## Design

~~I'm not super happy with the implementation. Mainly because it requires a private module to hide *most* implementation details of the `Builder`.~~
~~But even then, we don't get perfect encapsulation because the `Builder::create` method is still visible in the input's module. But I'm not sure what else we could do.~~

I had to revert implementing `Configuration` directly on the `$Struct` because self-referential types then stopped compiling. Example: 

https://github.com/salsa-rs/salsa/blob/6975a47690a20ee188c05ea2ddd772bbd985378b/tests/accumulate-reuse.rs#L16-L20

What I dislike about the current solution is that `Builder::inner` is accessible from the input's module. But not sure what we should do about it. I tried using a `trait` but that's a bit annoying to use because it must be explicitly imported from the ingredient's scope.  

## Tests

I did add a test but I haven't figured out a way to automatically verify if a tracked function was validated using deep or shallow comparison. 
I did do some manual testing by adding `dbg` statements that show that Salsa does use the shallow comparison for inputs with `Durability::HIGH`.


src/database.rs
MichaReiser, line 1: See https://github.com/salsa-rs/salsa/pull/534

nikomatsakis (commented):
What about a setup like this one?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=11175c08bb02381f293d7558dcdae52f

MichaReiser:
That sounds interesting. I'll give it a try

MichaReiser:
I'll re-open this PR from my own fork.

PR #537, opened 2024-07-26
fix: compare fat pointer leads to wrong panic in some condition

### LINKED ISSUE #536: Unexpected panic: `assertion 'left= right failed: cannot change database mid-query`
Hi, I recently updated my project's salsa to the newest version, and everything works well in debug mode. But as
soon as I switched to the release build, a weird panic occurred:
5fdfa4e06b2fa8fd121d20b940e58245

You can see in the pic above that Rust is saying that two pointers with the address `0×14660cbb0` are not equal.

After some digging around, I found this issue: https://github.com/rust-lang/rust/issues/107975

And this: https://github.com/rust-lang/rust/issues/107975

I believe there's a mistake in the mid-query check, instead of checking the equivalent of two fat pointers, we shall check only the address.

https://github.com/salsa-rs/salsa/blob/e4ce917f6eda0d775ef654742a6f3cdc710f7d1a/src/local_state.rs#L87-L91

Chronostasys:
Fix #536

### DIFF
--- a/src/local_state.rs
+++ b/src/local_state.rs
@@ -84,9 +84,10 @@ impl LocalState {
             fn new(state: &'s LocalState, db: &dyn Database) -> Self {
                 if let Some(current_db) = state.database.get() {
                     // Already attached? Assert that the database has not changed.
+                    // compare the thin pointer https://github.com/salsa-rs/salsa/issues/536
                     assert_eq!(
-                        current_db,
-                        NonNull::from(db),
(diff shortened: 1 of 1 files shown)


MichaReiser:
I think that's the same fix as https://github.com/salsa-rs/salsa/pull/532

Chronostasys:
> I think that's the same fix as #532

Oh, my bad. Forget to search related pr. I'll close this one.

PR #538, opened 2024-07-27
New database design

nikomatsakis:
Status: Ready

Prototype new database design. See the final commit for a write-up. This is ready to land but I was planning to do a bit more work, in particular removing the use of thread-local state.

Update: I've removed the thread-local state now.


examples/lazy-input/main.rs
```diff
@@ -74,28 +74,34 @@ trait Db: salsa::Database {
...
         Self {
-            storage,
             logs: Default::default(),
             files: DashMap::new(),
             file_watcher: Mutex::new(new_debouncer(Duration::from_secs(1), tx).unwrap()),
         }
     }
 }
 
+impl UserData for LazyInput {
```
MichaReiser, line 93: What I understand is that the major change in this PR is to go from `Database` (application code) wrapping a `Storage` to `Database` (framework code) wrapping `UserData`. I don't understand the motivation for the change and I'm sure there are good reason for it. That's why my feedback here is primarily from a user perspective. 

To me, the old `Database` wraps a storage was easier to understand and use in the application code:

* I only need to name one unit, the database. With the new approach I have to both find a name for the `Database` alias AND the struct encapsulating `UserData`. I especially find naming `UserData` hard because it's just a collection of all the things not inside salsa. 
* I do find the name `UserData` confusing, especially when implementing in the application. Aren't all ingredients instances user data (it's not framework data)? 

I think this is probably mainly a naming problem, but I must admit that I fail to come up with a better name. Maybe `ExternalData`? But is the trait just about data? Isn't `salsa_event` an indicator that the trait does much more? 

I don't think it's blocking but we may want to improve the names of external concepts in follow up PRs.

nikomatsakis: Thanks for the feedback. You are correct that this is the major change. I agree that the old approach was "nice" from a user point-of-view (which is why I pursued it), but it had some pretty strong disadvantages too. I'll write a separate comment on the motivations for doing the inversion, it'd be good to document them.

Regarding naming, I agree, `UserData` is kind of leaking my "salsa implementor" perspective. Perhaps `CustomData` or `UntrackedData`? I'm going to run with the latter.

I also think we need to document the pattern(s) we expect users to follow. In this PR I think there are three possibilities, from simplest to most complex:

* Simplest: if you don't need any custom fields, just use `DatabaseImpl::new` and `dyn Database` in your functions.
* More complex: If you have one set of data that you need across your entire application, you can define it in one place (perhaps a core crate) and then do `trait MyDatabase: Database`, `struct MyUntrackedData`, and implement `MyDatabase` for `DatabaseImpl<MyUntrackedData>` and `UntrackedData` for `MyUntrackedData.`
* Even more complex: If you have many crates and you don't want them all to depend on one base crate that defines your user data, you can make two traits, `MyDatabase1: Database` and `MyUntrackedData1: UntrackedData` and then implement `MyDatabase1` for all `DatabaseImpl<U>` where `U: MyUntrackedData1`. You can then define in some final crate a `MyUntrackedData` that implements `MyUntrackedData1` (along with `MyUntrackedData2..N` from other crates).

Overall I found it nicer to use for most tests (less boilerplate) but I agree that it's less "obvious' somehow. I'm thinking about it, maybe there's a way to reshuffle it.

It's *definitely* nice from the implementor perspective,_ because before it was always this "dance" where we didn't know the *actual* type of the database because it was embedded into something else.

nikomatsakis: The more I think about it, the more I like `UntrackedData`. The idea then is that you instantiate a salsa database and it stores the tracked data; if you need other data, you can create it and get access to it yourself.

One other thought I had is that I think we could define a `DatabaseWith<U>` trait that you can use, so that the middle pattern (one but of untracked data) because like so:

* Simple: as above
* Medium: define your untracked data `MyUntrackedData` type in some core crate. Right all your code as `&dyn DatabaseWith<MyUntrackedData>` (you may want to create an alias), which is what I forced you to do before.
* Complex: define multiple untracked data pieces and code against them, basically unchanged.

nikomatsakis: Oh, the last thought is that we can make `DatabaseImpl` crate-local, so you only work with a `Box<dyn Database>` or ` Box<dyn DatabaseWith>`, essentially. You create the database with `salsa::database()` and `salsa::database_with(u)`.

nikomatsakis: (Hmm, maybe that doesn't *quite* work with the handle cloning...well, something like that.)

MichaReiser: The name `UntrackedData` eliminates most of my concerns. I think the name captures well what data is stored inside the struct. 

I have to play with the traits a bit to get a better feeling on how complicated they feel. Our setup currently uses multiple databases but we may get away with a single database and just two untracked (one for production and one for testing) the more a I think about it.

Cargo.toml
```diff
@@ -21,6 +21,7 @@ rustc-hash = "2.0.0"
 salsa-macro-rules = { version = "0.1.0", path = "components/salsa-macro-rules" }
 salsa-macros = { path = "components/salsa-macros" }
 smallvec = "1.0.0"
+lazy_static = "1.5.0"
```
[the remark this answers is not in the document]
nikomatsakis: I suppose this could just be "1.4". I wonder why cargo-insta is so particular?

src/function/accumulated.rs
```diff
@@ -14,41 +12,41 @@ where
...
-                    stack.extend(
-                        inputs
-                            .flat_map(|input| {
-                                TryInto::<DatabaseKeyIndex>::try_into(input).into_iter()
-                            })
-                            .rev(),
-                    );
-                }
+        let zalsa = db.zalsa();
+        let zalsa_local = db.zalsa_local();
```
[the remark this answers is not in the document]
nikomatsakis: zalsa = plumbing, not be used by end-users, is the intention.

examples/lazy-input/main.rs
MichaReiser: > This use case now feels significantly more complicated because it

It sounds as if I feel strongly about it. I don't. It's just something I noticed, and I strongly favor better performance over avoiding boilerplate at all costs (when setting up the database) . 

I haven't yet figured out how to implement `Upcast` with the new design. Could we use the `views` API instead?

nikomatsakis:
@MichaReiser I've been AFK for a bit...hopefully will be getting more reliable internet access soon. I've been thinking it over and I think I can produce a variation of this branch where we have the old setup (i.e., you define a struct that contains the Storage), which should address the various concerns you've raised.

MichaReiser:
No worries. I hope you're having a great time. I don't have a strong preference over either approach. I think both is fine. The only real blocking item right now is that I don't know how to implement `Upcast`.

nikomatsakis:
@MichaReiser I retooled it now, take a look. I think you convinced me that the old approach was better. We now have:

* For simple uses, the built-in `Database` trait and `DatabaseImpl` struct.
* For more complex uses, you can define your own database trait and structs using `#[salsa::db]`; the struct must have `storage: salsa::Storage<Self>` as one of its fields.

nikomatsakis:
I went ahead and pushed some other commits I had that make things miri-safe again (at least I think they do, let's see what CI has to say about it).

src/zalsa.rs
```diff
@@ -0,0 +1,279 @@
...
+                }
+                index
+            })
+        }
+    }
+
+    pub(crate) fn lookup_ingredient(&self, index: IngredientIndex) -> &dyn Ingredient {
+        &**self.ingredients_vec.get(index.as_usize()).unwrap()
+    }
+
```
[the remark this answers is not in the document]
nikomatsakis: I will just add an API like `fn debug_name(&self, ingredient: IngredientIndex) -> Cow<'_, String>`, how's that? (That will give us some room to move away from static strings and be more future proof).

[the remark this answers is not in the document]
nikomatsakis: Done.

src/database.rs
```diff
@@ -21,31 +37,236 @@ pub trait Database: DatabaseGen {
...
+}
+
+impl<U: UserData> std::ops::Deref for DatabaseImpl<U> {
+    type Target = U;
+
+    fn deref(&self) -> &U {
+        self.zalsa_impl().user_data().downcast_ref::<U>().unwrap()
+    }
+}
+
```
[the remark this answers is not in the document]
nikomatsakis: No longer relevant.

Cargo.toml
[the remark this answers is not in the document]
nikomatsakis: No longer relevant, I think? At least I just see `= "1"`

examples/lazy-input/main.rs
nikomatsakis: Resolved and switched it over. I think you should be able to do whatever you were doing before.

src/zalsa.rs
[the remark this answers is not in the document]
MichaReiser: That's even better. Thank you

Issue #476, opened 2024-03-17
Generate builders methods


gavrilikhin-d:
It seems a little inconvenient to pass all fields through `XXX::new` call. 
Can salsa generate builder methods for `interned`, `tracked` and `input` structs?

E.g:
```rs
#[salsa::input]
pub struct SourceProgram {
    #[return_ref]
    pub text: String,
    #[return_ref]
    pub path: PathBuf,
}

SourceProgram::build().with_text("text").with_path("foo.rs").finish()
```

P.S: Names for building functions are just for example, you may pick the ones you like

gavrilikhin-d:
Related to https://github.com/salsa-rs/salsa/issues/385

nikomatsakis:
Agreed! I would like to generate builders. This seems like a good starter issue, @gavrilikhin-d

nikomatsakis:
See also https://github.com/salsa-rs/salsa/pull/418 -- I forget exactly what that PR is doing, maybe same general thing?

PR #543, opened 2024-07-29
Introduce `Input::builder` API

MichaReiser:
This PR introduces a new `Input::builder` API.

```rust
#[salsa::input]
struct MyInput {
    required_field: bool,

    #[default]
    optional_field: usize,
}

let input = MyInput::builder(true)
    .optional_field(20)
    .required_field_durability(Durability::HIGH)
    .new(&db);
```

It also adds support for the new input-field-attribute `#[default]`. It specifies that the value should be initialized with `$ty::default()` unless it is specified when using the `builder` API. Optional fields don't need to be specified when calling `Input::new`. I intentionally kept the default attribute simple for now. We can add support for calling a function, or initializing with a const value in the future.

One downside of the macro-generated builder is that at least RustRover (IntelliJ) completely breaks down and fails to provide any auto-completion. I yet have to try if rust analyzer does any better.

Edit: Ra does better! It does suggest the different builder methods, except `new`. `new` doesn't show up for some reason.

Fixes https://github.com/salsa-rs/salsa/issues/385
Fixes https://github.com/salsa-rs/salsa/issues/476 (not in its entirety but a start)

## Design

This API inverts the original design where `db` was passed to `builder` and the required fields to `new`. There are a few reasons for this:

* Passing `db` in `new` requires that the `builder` stores a reference to `IngredientImpl<C>`. 
  * It requires passing around some more lifetimes that aren't necessary with the proposed design
  * `Configuration` isn't referenceable in the `builder` module. That means, the generated `Builder` would need to be generic over `Configuration` (or use some other trickery to avoid naming `Configuration`
  * It should allow for `builder(required_fields).optional(some_query(db)).new(&mut db)` without having to store the `optional` value in a temporary variable (once we support optional fields). 
* Enforcing that all mandatory fields are passed when constructing the `builder` should simplify implementing support for optional fields where the macro code can generate $fields` using either the default or provided value. 

## Alternative Designs

I'm not too fond of the `*_durability` methods. An alternative could be introducing a `#[durability=Durability::HIGH]` field and the builders and setters would respect that durability. The downside of this is that it is less flexible and it is unclear whether `builder.durability` should override the durability for those fields (I see reasons for and against it).

## Why not use type-states?

I thought about it but it felt overly-complicated to me and goes beyond Salsa's scope. If someone wants more advanced builders than using any other builder crate on top of salsa should be possible (create a single function that takes all input values and generate a builder for that function). 

## Tests

I did add a test but I haven't figured out a way to automatically verify if a tracked function was validated using deep or shallow comparison. 
I did do some manual testing by adding `dbg` statements that show that Salsa does use the shallow comparison for inputs with `Durability::HIGH`.


components/salsa-macro-rules/src/setup_input_struct.rs
```diff
@@ -121,14 +127,33 @@ macro_rules! setup_input_struct {
...
                     // FIXME(rust-lang/rust#65991): The `db` argument *should* have the type `dyn Database`
                     $Db: ?Sized + salsa::Database,
                 {
-                    let current_revision = $zalsa::current_revision(db);
-                    let stamps = $zalsa::Array::new([$zalsa::stamp(current_revision, Default::default()); $N]);
-                    $Configuration::ingredient(db.as_salsa_database()).new_input(($($field_id,)*), stamps)
+                    Self::builder($($required_field_id,)*).new(db)
+                }
+
+                pub fn builder($($required_field_id: $required_field_ty),*) -> <Self as $zalsa_struct::HasBuilder>::Builder
```
MichaReiser, line 139: It would be nice if we could avoid the extra `required_fields` macro argument but I couldn't figure out how to use `macro_if` in a function declaration.

components/salsa-macro-rules/src/setup_input_struct.rs
```diff
@@ -204,6 +229,72 @@ macro_rules! setup_input_struct {
...
+
+                    $(
+                        /// Sets the value of the field `$field_id`.
+                        #[must_use]
+                        pub fn $field_id(mut self, value: $field_ty) -> Self
+                        {
+                            self.fields.$field_index = value;
+                            self
+                        }
+                    )*
```
MichaReiser, line 285: This also generates setters for required fields which is redundant because they were already provided when calling `builder`. However, not adding them is slightly more complicated and probably requires explicitly passing an `optional_fields` structure to the macro with the `id`, `ty` and `index` of each optional field.

components/salsa-macro-rules/src/setup_input_struct.rs
```diff
@@ -204,6 +229,72 @@ macro_rules! setup_input_struct {
...
+
+                    (builder.fields, stamps)
+                }
+
+                pub struct $Builder {
+                    /// The field values.
+                    fields: ($($field_ty,)*),
+
+                    /// The durabilities per field.
+                    durabilities: [$zalsa::Durability; $N],
```
MichaReiser, line 265: I'm a bit confused. For some reason, this compiles in the test code:

```rust
    let input = MyInput::builder(true).durabilities;
```

I would have expected this to fail because the `Builder` is defined in its own module and the field isn't `pub(super)`.

MichaReiser:
I plan to rebase this PR tomorrow

components/salsa-macro-rules/src/setup_input_struct.rs
nikomatsakis: yeah, you'd have to wrap the whole function. seems fine this way.

components/salsa-macro-rules/src/setup_input_struct.rs
```diff
@@ -121,14 +128,35 @@ macro_rules! setup_input_struct {
...
                 {
-                    let current_revision = $zalsa::current_revision(db);
-                    let stamps = $zalsa::Array::new([$zalsa::stamp(current_revision, Default::default()); $N]);
-                    $Configuration::ingredient(db.as_salsa_database()).new_input(($($field_id,)*), stamps)
+                    Self::builder($($required_field_id,)*).new(db)
+                }
+
+                pub fn builder($($required_field_id: $required_field_ty),*) -> <Self as $zalsa_struct::HasBuilder>::Builder
+                {
+                    // Implement `new` here instead of inside the builder module
```
nikomatsakis, line 142: Shouldn't this be out in the `const _: () = { ... }` section, and not nested inside this `build` function? We're trying to crack down on "hidden" methods like that.

tests/tracked_fn_on_input_with_high_durability.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    db.synthetic_write(Durability::LOW);
+
+    assert_eq!(tracked_fn(&db, input_low), 44);
+    assert_eq!(tracked_fn(&db, input_high), 4400);
+
+    // There's currently no good way to verify whether an input was validated using shallow or deep comparison.
+    // All we can do for now is verify that the values were validated.
+    // Note: It maybe confusing why it validates `input_high` when the write has `Durability::LOW`.
+    // This is because all values must be validated whenever a write occurs. It doesn't mean that it
```
nikomatsakis, line 68: I'll think on how to test this...but it's ok to open a FIXME...

nikomatsakis (approved):
Looks great! I'd be happy to merge.

components/salsa-macro-rules/src/setup_input_struct.rs
MichaReiser: That makes sense and it may help rust analyzer to find the impl block.

PR #544, opened 2024-07-30
Add test tracing, fix up comments

carljm:
This PR does two things:

1) Add the `trace` feature for the `test-log` dependency, which I've found very useful in trying to understand the cycle-handling code. It means that I can change `#[test]` to `#[test_log::test]` above a failing test of interest, and then if I run `cargo test` with e.g. `RUST_LOG=debug`, the failing test output will include all the debug traces from that test.

2) Fix up a bunch of comments and some debug traces I ran across in my exploration that looked out of date.

Happy to split this into two PRs if that's preferred.

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -65,7 +65,7 @@ pub trait Configuration: Any {
     /// This invokes the function the user wrote.
     fn execute<'db>(db: &'db Self::DbView, input: Self::Input<'db>) -> Self::Output<'db>;
 
-    /// If the cycle strategy is `Recover`, then invoked when `key` is a participant
+    /// If the cycle strategy is `Fallback`, then invoked when `key` is a participant
     /// in a cycle to find out what value it should have.
     ///
     /// This invokes the recovery function given by the user.
--- a/tests/cycles.rs
+++ b/tests/cycles.rs
@@ -49,9 +49,6 @@ use salsa::Durability;
 // | Cross  | Fallback | N/A      | Tracked   | both     | parallel/parallel_cycle_mid_recover.rs |
 // | Cross  | Fallback | N/A      | Tracked   | both     | parallel/parallel_cycle_all_recover.rs |
 
-// TODO: The following test is not yet ported.
-// | Intra  | Fallback | Old      | Tracked   | direct   | cycle_disappears_durability |
-
 #[derive(PartialEq, Eq, Hash, Clone, Debug)]
 struct Error {
     cycle: Vec<String>,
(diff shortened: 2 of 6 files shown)


src/function.rs
```diff
@@ -69,7 +69,7 @@ pub trait Configuration: Any {
     /// This invokes the function the user wrote.
     fn execute<'db>(db: &'db Self::DbView, input: Self::Input<'db>) -> Self::Output<'db>;
 
-    /// If the cycle strategy is `Recover`, then invoked when `key` is a participant
+    /// If the cycle strategy is `Fallback`, then invoked when `key` is a participant
```
MichaReiser, line 72:
```suggestion
    /// If the cycle strategy is [`Fallback`](CycleRecoveryStrategy::Fallback), then invoked when `key` is a participant
```

tests/cycles.rs
```diff
@@ -48,9 +48,6 @@ use salsa::Durability;
 // | Cross  | Fallback | N/A      | Tracked   | both     | parallel/parallel_cycle_mid_recover.rs |
 // | Cross  | Fallback | N/A      | Tracked   | both     | parallel/parallel_cycle_all_recover.rs |
 
-// TODO: The following test is not yet ported.
-// | Intra  | Fallback | Old      | Tracked   | direct   | cycle_disappears_durability |
-
```
MichaReiser, line 53: Did you remvoe the TODO because the test is now ported?

examples/calc/parser.rs
MichaReiser, line 1: You probably want to revert this change, considering that it is part of your other PR

MichaReiser (approved):

tests/cycles.rs
carljm: Yes; the test exists in the file and isn't commented out and runs and passes, so it doesn't seem like anything more is needed.

examples/calc/parser.rs
carljm: I rebased this PR on that one to get the tests passing here, then noticed that since I pushed the branch to my own fork I couldn't actually set the PR base branch to the other PRs branch, so the changes are included here. Once the other PR is merged I can rebase this one. (Or I can rebase it sooner and just let the tests fail.)

nikomatsakis (approved):

nikomatsakis:
@carljm looks good to me but needs to be rebased. feel free to merge.

carljm:
@nikomatsakis Rebased, but I can't merge; it says "Merging is blocked: The base branch restricts merging to authorized users."

PR #545, opened 2024-07-30
Update expect test output to account for OrderedFloat changes

carljm:
The `expect!` tests for the parser example are failing in main branch. This failure started with https://github.com/salsa-rs/salsa/commit/c50cefa71e24ccdd108576938dfa44139fc72c00 which upgraded the `ordered-float` dependency. It looks like the new version of ordered-float implements `Display` as a pass-through to the underlying float, and the old version did not, causing these test failures.

Generated by running `env UPDATE_EXPECT=1 cargo test --workspace --all-features --all-targets`.

### DIFF
--- a/examples/calc/parser.rs
+++ b/examples/calc/parser.rs
@@ -399,9 +399,7 @@ fn parse_print() {
                                             end: 7,
                                         },
                                         data: Number(
-                                            OrderedFloat(
-                                                1.0,
-                                            ),
+                                            1.0,
                                         ),
                                     },
                                     Add,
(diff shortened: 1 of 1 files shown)


carljm:
Is it expected that Miri is currently unhappy in main?

MichaReiser:
> Is it expected that Miri is currently unhappy in main?

Yes, MIRI isn't happy with the concurrent vec dependency.

MichaReiser (approved):

MichaReiser:
We should consider running examples as part of CI

carljm:
> We should consider running examples as part of CI

They are run as part of CI; I discovered this failure because this test failed on my other PR.

Issue #546, opened 2024-08-01
`Handle` doesn't allow reads after `get_mut`


obsgolem:
The `Handle` type sets the cancellation flag but doesn't clear it before handing the mutable access back to the caller. You can get around this with a `synthetic_write`, but otherwise any attempts to perform an immutable access to the database will panic.

MichaReiser:
Closing because `Handle` is gone ;)

PR #548, opened 2024-08-02
Fix bug with Handle::get_mut not allowing non-mutable access to the database

obsgolem:
Resolves https://github.com/salsa-rs/salsa/issues/546


src/handle.rs
```diff
@@ -86,10 +86,18 @@ impl<Db: HasStorage> Handle<Db> {
...
+            while *clones != 1 {
+                self.coordinate.cvar.wait(&mut clones);
+            }
         }
+
+        // Force a new revision to clear the cancellation flag. 
+        // This is necessary to ensure non-mutable operations work after this.
+        let db =
+            Arc::get_mut(self.db_mut()).expect("other threads remain active despite cancellation");
+        db.runtime_mut().new_revision();
```
MichaReiser, line 100: I would prefer adding a `reset_cancellation_flag()` method to `db`. Creating a new revision comes at the cost that all queries need to be verified and avoiding the `get_mut` call in user-code when there are no "write" changes might not always be possible (or complicates the code a lot)

nikomatsakis (changes requested):
Hmm, this is probably still relevant-- can you rebase this atop master? The handle code has gone away but you should see equivalent code in storage.rs

Also, can you add a test?

Thanks! 💜

MichaReiser:
I merged in the latest changes. We only need to add some tests

MichaReiser:
I don't think this PR is still needed now that `Handle` has been removed because the only call-site of `cancel_others` is inside `zalza_mut` which always bumps the revision. 

https://github.com/salsa-rs/salsa/blob/5d3b75631bcf07e0ca62d433c53a22d6f5414c06/src/storage.rs#L99-L108

nikomatsakis:
Ah, ok, I'll close it. It may be relevant if we add some way to explicitly cancel things. Thanks @obsgolem !

PR #552, opened 2024-08-02
Require Debug for C::Stamps

carljm:
I found this really useful in debugging durability issues; it allows sprinkling a `dbg!(&data.stamps)` in some key places to see what the field durabilities and changed-at values are for the fields.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -1,6 +1,7 @@
+use std::fmt::Debug;
 use std::ops::{Deref, DerefMut};
 
-#[derive(Copy, Clone)]
+#[derive(Copy, Clone, Debug)]
 pub struct Array<T, const N: usize> {
     data: [T; N],
 }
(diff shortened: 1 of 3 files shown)


src/array.rs
```diff
@@ -1,25 +1,26 @@
+use std::fmt::Debug;
 use std::ops::{Deref, DerefMut};
 
-#[derive(Copy, Clone)]
-pub struct Array<T, const N: usize> {
+#[derive(Copy, Clone, Debug)]
+pub struct Array<T: Debug, const N: usize> {
```
MichaReiser, line 5: What's the reason for requiring Debug on Array. Isn't the automatic derive implementation enough. If not, then only require the debug type constraint when implementing debug for array

carljm: Ah yeah, not needed, I went overboard here :)

MichaReiser (approved):

nikomatsakis (approved):

PR #554, opened 2024-08-05
Replace `orx-concurrent-vec` with `boxcar`

MichaReiser:
This PR replaces `orx-concurrent-vec` with `boxcar` because:

* `boxcar` is a much smaller dependency 11.1KB without dependencies vs 90.9KB with a handful of dependencies
* `orx-concurrent-vec` [fails to build on 32bit](https://github.com/orxfun/orx-split-vec/pull/45) platforms (may be fixed soon)
* `boxcar` shows better performance

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,11 +10,11 @@ description = "A generic framework for on-demand, incrementalized computation (e
 [dependencies]
 arc-swap = "1"
 boomphf = "0.6"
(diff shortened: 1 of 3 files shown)


MichaReiser:
I close this in favor of https://github.com/salsa-rs/salsa/pull/555 which shows even better performance

PR #555, opened 2024-08-05
Replace `orx-concurrent-vec` with `append-only-vec`

MichaReiser:
An alternative to https://github.com/salsa-rs/salsa/pull/554

`append-only-vec` performs even better than `boxcar` and also helps to reduce the overall dependencies (and runs on 32bit!).

### DIFF
--- a/src/zalsa.rs
+++ b/src/zalsa.rs
@@ -167,7 +167,7 @@ impl Zalsa {
     }
 
     pub(crate) fn lookup_ingredient(&self, index: IngredientIndex) -> &dyn Ingredient {
-        &**(&self.ingredients_vec[index.as_usize()])
+        &*self.ingredients_vec[index.as_usize()]
     }
 
     /// **NOT SEMVER STABLE**
@@ -176,7 +176,7 @@ impl Zalsa {
         index: IngredientIndex,
     ) -> (&mut dyn Ingredient, &mut Runtime) {
         (
-            &mut **(&mut self.ingredients_vec[index.as_usize()]),
+            &mut *self.ingredients_vec[index.as_usize()],
             &mut self.runtime,
         )
     }


nikomatsakis:
Seems like a winner!

Cargo.toml
```diff
@@ -14,7 +14,7 @@ crossbeam = "0.8"
 dashmap = "6"
 hashlink = "0.9"
 indexmap = "2"
-orx-concurrent-vec = "2"
+append-only-vec = "0.1"
```
nikomatsakis, line 17:
```suggestion
append-only-vec = "0.1.5"
```

we actually need 0.1.5 since it contains [this change](https://github.com/droundy/append-only-vec/pull/4)

nikomatsakis (changes requested):

MichaReiser:
Thanks for fixing the version. I'll merge this tomorrow unless you wanted me to address something else.

MichaReiser:
Okay I can't merge this while changes are requested. @nikomatsakis would you mind having another look.

nikomatsakis (approved):

ibraheemdev:
FWIW `append-only-vec` isn't a great concurrent vector implementation as it relies heavily on spin locks. In particular, it holds a [spin lock when allocating](https://github.com/droundy/append-only-vec/blob/main/src/lib.rs#L131) which is almost always a bad idea, but it also uses a spin lock [to serialize insertion order](https://github.com/droundy/append-only-vec/blob/main/src/lib.rs#L158). These not only [hurt scalability](https://github.com/orxfun/orx-concurrent-vec?tab=readme-ov-file#growth-with-push) but can also lead to unfortunate tail latency scenarios that are common with spin locks.

Of course, the upside is that array access is extremely cheap due to the extra ad-hoc synchronization, so if you expect writes to be relatively rare this may not be an issue.. but something to look out for.

MichaReiser:
Thanks @ibraheemdev for the in-depth explanation. 

I do think the spin-lock is fine the way we use it (at least for ingredients) because all write paths go through a lock anyway.

ibraheemdev:
Interesting. A single-writer concurrent vector could be implemented more efficiently in general.. but probably not worth it for Salsa.

PR #556, opened 2024-08-05
remove unused `boomphf` dependency

gvozdvmozgu:


MichaReiser (approved):

nikomatsakis:
I was going to use this to optimize the lookup for "view casts" etc...

nikomatsakis:
But never got around to it :)

PR #557, opened 2024-08-05
Fix query invalidation when high durability input changes

MichaReiser:
This PR fixes a bug in Salsa where changing the value of a high durability field did not invalidate any queries with `durability > LOW` 

I had to re-expose `&mut Runtime` again when calling `ingredient_mut` to call `runtime.report_tracked_write` in `SetterImpl`


MichaReiser:
Hmm, not sure what the issue is with miri.

nikomatsakis (approved):

MichaReiser:
I can't reproduce the miri failure locally but it also happens on master.

PR #567, opened 2024-08-22
Reduce cloning of sets in `ActiveQuery` and `QueryRevisions`

MichaReiser:
A small performance improvement to avoid cloning hash sets where we can consume the value instead.

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -1,4 +1,5 @@
 use rustc_hash::FxHashMap;
+use std::sync::Arc;
 
 use crate::{
     durability::Durability,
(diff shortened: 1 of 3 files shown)


src/active_query.rs
```diff
@@ -117,7 +118,7 @@ impl ActiveQuery {
             changed_at: self.changed_at,
             origin,
             durability: self.durability,
-            tracked_struct_ids: self.tracked_struct_ids.clone(),
+            tracked_struct_ids: self.tracked_struct_ids,
```
MichaReiser, line 121: There's one remaining place where `QueryRevisions` are deep cloned but I'm not sure if it justifies wrapping `tracked_struct_ids` in an `Arc`.

It might be cheaper to wrap `QueryRevisions` in an `Arc` instead (and remove the Arc around `QueryEdges``)?

MichaReiser:
Okay, that's not what I expected. I do see a 1-2% perf improvement in local benchmarks

MichaReiser:
> Okay, that's not what I expected. I do see a 1-2% perf improvement in local hyperfine benchmarks

Using `IndexSet::into_boxed_slice` seems to really hurt performance (at least in codspeed). I reverted that change.

MichaReiser:
I see about a 1% performance improvement on the [Ruff benchmarks](https://codspeed.io/astral-sh/ruff/branches/salsa-reduced-cloning) (second last benchmark). I hoped for more, but that's how it is with perf ;) 

It seems that codspeed struggles a bit with matching the salsa frames. I suspect it doesn't help that too many frames have the same name :laughing:

nikomatsakis (approved):

PR #568, opened 2024-08-23
introduce parallel salsa

davidbarsky:
Initial pass; very much not ready. Putting it up for initial feedback; I'm still figuring out how to wrap `&Db` reasonably.


examples/lazy-input/main.rs
```diff
@@ -77,11 +81,12 @@ trait Db: salsa::Database {
 }
 
 #[salsa::db]
+#[derive(Clone)]
```
MichaReiser, line 84: Nit: I kind of preferred the old salsa's `ParallelDatabase` trait rather than relying on `Clone`. It makes the contract more explicit.

src/storage.rs
```diff
@@ -108,6 +108,10 @@ unsafe impl<T: HasStorage> ZalsaDatabase for T {
     fn zalsa_local(&self) -> &ZalsaLocal {
         &self.storage().zalsa_local
     }
+
+    fn fork_db(&self) -> Box<dyn Database> {
```
MichaReiser, line 112: Could we return `T` instead of a `Box<dyn Database>`?

src/par_map.rs
```diff
@@ -0,0 +1,25 @@
...
+    inputs: impl IntoParallelIterator<Item = D>,
+    op: fn(&Db, D) -> E,
+) -> C
+where
+    Db: Database,
+    D: Send,
+    E: Send + Sync,
+    C: FromParallelIterator<E>,
+{
+    dbg!(db.zalsa().views());
```
davidbarsky, line 16: i've been doing a bunch of digging to figure out why this panics when used with `tests/parallel/parallel_map.rs` test. i initially saw that the underling `Zalza::views` was empty, so the test would naturally panic—there is nothing to downcast to! I've since added calls to `tracked_fn` in the test, and now, I'm seeing a downcast-able database.

```
[src/par_map.rs:16:5] db.zalsa().views() = DynDowncasts {
    vec: [
        DynDowncast(
            "dyn parallel_map::common::LogDatabase",
        ),
    ],
```

This makes sense! databases are discovered upon first use, but for `par_map` to work, I think the list of databases to cast to need to be reflexive: that is, they need to include themselves. I'll try making this change.

examples/lazy-input/main.rs
davidbarsky: At least in rust-analyzer's case, I can't really think of a situation where it _wouldn't_ make sense to have every `Database` be a `ParallelDatabase`. I don't think we'd want to make everything parallel, but having the option to do so without extensive refactors is really valuable, in my view.

src/par_map.rs
```diff
@@ -0,0 +1,25 @@
...
+    let fork_db_a: Box<dyn Database> = db.fork_db();
+    let db: &Db = fork_db_a.as_view::<Db>();
+
+    todo!()
+    // inputs
+    //     .into_par_iter()
+    //     .map_with(wrapper, |db, element| op(db, element))
+    //     .collect()
+}
```
nikomatsakis, line 25: This is what I imagined:

```rust

pub fn par_map<Db, D, E, C>(
    db: &Db,
    inputs: impl IntoParallelIterator<Item = D>,
    op: fn(&Db, D) -> E,
) -> C
where
    Db: Database,
    D: Send,
    E: Send + Sync,
    C: FromParallelIterator<E>,
{
    let parallel_db = ParallelDb::Ref(db);
    inputs
        .into_par_iter()
        .map_with(parallel_db, |parallel_db, element| {
            let db = parallel_db.as_view::<Db>();
            op(db, element)
        })
        .collect()
}

enum ParallelDb<'db> {
    Ref(&'db dyn Database),
    Fork(Box<dyn Database>),
}

impl Deref for ParallelDb {
    type Target = dyn Database;

    fn deref(&self) -> &Self::Target {
        match self { ... }
    }
}

impl Clone for ParallelDb<'_> {
    fn clone(&self) -> Self {
        ParallelDb::Fork(self.fork())
    }
}
```

nikomatsakis (changes requested):

src/par_map.rs
davidbarsky: Oh man, that's way easier. Switched over to this.

src/table.rs
```diff
@@ -90,6 +90,10 @@ impl<T: Slot> RefUnwindSafe for Page<T> {}
 #[derive(Copy, Clone, Debug)]
 pub struct PageIndex(usize);
 
+unsafe impl Send for PageIndex {}
+
+unsafe impl Sync for PageIndex {}
```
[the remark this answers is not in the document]
davidbarsky: removed!

nikomatsakis:
Does miri have a way to add 'known failures'...

nikomatsakis:
@davidbarsky I'd be ok moving the miri tests to "warn only" for the time being and then trying to investigate the use of crossbeam-epoch more closely.

davidbarsky:
> @davidbarsky I'd be ok moving the miri tests to "warn only" for the time being and then trying to investigate the use of crossbeam-epoch more closely.

Lemme see about doing that. I think I still need to port over a test or two over as well.

davidbarsky:
Added a cancellation test.

MichaReiser:
@davidbarsky what's the status of this PR? We now have a use case where the hacky hand rolled concurrency no longer works :laughing:

MichaReiser:
I tried to use the `salsa::par_map` function in https://github.com/astral-sh/ruff/pull/14760 but I always received an empty list of accumulated values. I haven't yet looked into what goes wrong when using `par_map` but I suspect that some query dependencies aren't correctly recorded?

PR #569, opened 2024-08-28
add test for high-durability dependency validation

carljm:
This brings over just the test from https://github.com/salsa-rs/salsa/pull/550, which was already fixed in https://github.com/salsa-rs/salsa/pull/564


davidbarsky (approved):
I feel comfortable approving this!

PR #573, opened 2024-09-13
Fix panic when recreating tracked struct that was deleted in previous revision

dhruvmanila:
Edited by @MichaReiser 

This is an error that surfaced when working on Ruff's LSP. The scenario is:

* Revision 0: Query `Q` creates a tracked struct `S`
* Revision 1: Query `Q` no longer creates a tracked struct
* Revision 2: Query `Q` again creates a tracked struct `S`

The real world scenario is Ruff's semantic index that creates a tracked struct `Scope` for every function. Deleting a function and then creating a new function resulted in a panic `two concurrent writers to {id:?}, should not be possible`.

The root cause of the panic is that a removed entity is added to an ingredients-free list, but the ID is never removed from the `tracked_struct_ids`, resulting in Revision 2 re-using the struct from the free list.

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -6,7 +6,7 @@ use crate::{
     key::{DatabaseKeyIndex, DependencyIndex},
     tracked_struct::{Disambiguator, KeyStruct},
     zalsa_local::EMPTY_DEPENDENCIES,
-    Cycle, Id, Revision,
+    Cycle, Revision,
 };
 
 use super::zalsa_local::{EdgeKind, QueryEdges, QueryOrigin, QueryRevisions};
@@ -48,7 +48,7 @@ pub(crate) struct ActiveQuery {
 
     /// Map from tracked struct keys (which include the hash + disambiguator) to their
     /// final id.
-    pub(crate) tracked_struct_ids: FxHashMap<KeyStruct, Id>,
+    pub(crate) tracked_struct_ids: FxHashMap<KeyStruct, DatabaseKeyIndex>,
 }
 
 impl ActiveQuery {
--- a/src/function.rs
+++ b/src/function.rs
@@ -165,7 +165,7 @@ where
         zalsa: &'db Zalsa,
         id: Id,
         memo: memo::Memo<C::Output<'db>>,
-    ) -> Option<&C::Output<'db>> {
+    ) -> Option<&'db C::Output<'db>> {
         let memo = Arc::new(memo);
         let value = unsafe {
             // Unsafety conditions: memo must be in the map (it's not yet, but it will be by the time this
--- a/src/function/diff_outputs.rs
+++ b/src/function/diff_outputs.rs
@@ -1,5 +1,4 @@
 use super::{memo::Memo, Configuration, IngredientImpl};
-use crate::tracked_struct::KeyStruct;
 use crate::{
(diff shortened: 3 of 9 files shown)


src/active_query.rs
```diff
@@ -48,7 +48,7 @@ pub(crate) struct ActiveQuery {
 
     /// Map from tracked struct keys (which include the hash + disambiguator) to their
     /// final id.
-    pub(crate) tracked_struct_ids: FxHashMap<KeyStruct, Id>,
+    pub(crate) tracked_struct_ids: FxHashMap<KeyStruct, DatabaseKeyIndex>,
```
MichaReiser, line 51: I'm not 100% sure if storing the `DatabaseKeyIndex` is necessary, but we now need a way to remove elements, and the `Id` itself isn't unique enough. Two tracked struct from different ingredients can both have the id 0. The ambiguity hasn't been a problem up to now because we only did lookups by key.

MichaReiser:
Nice. This seems to improve performance by a few percent.

MichaReiser:
@nikomatsakis any thoughts on this. I would expect that @davidbarsky will run into this very soon when starting to port over rust analyzer.

nikomatsakis (approved):
Good catch!

nikomatsakis:
I see that clippy is grouchy

nikomatsakis:
@MichaReiser I'd be happy to merge once clippy warnings are fixed

nikomatsakis:
I'm not sure what's up with miri

src/active_query.rs
nikomatsakis: Yeah, it seems necessary if we take this approach of filtering out afterwards. The alternative would be to have two hashmaps, one with the `tracked_struct_ids` from the previous revision, and one that you are building up with the new values. You would first check the old map and, if something is found, move it to the new map. Then you don't have to remove anything after the fact.

I thought about suggesting this but I think this retain-based design is probably mildly more efficient.

nikomatsakis (changes requested):
Actually, I have one ask:

This is a fairly subtle interaction, so can we add some comments? I would first comment `diff_outputs` to explain why it takes an `&mut` and also put a comment on the `tracked_struct_ids` field of `ActiveQuery`, explaining that it is sometimes seeded with the results from previous revision and that (after the fact) anything that wasn't used will be removed by `diff_outputs`.

MichaReiser:
Eh, not sure what's up with clippy. I can't reproduce locally. Let's see if it passes once I pushed a new revision.

src/active_query.rs
MichaReiser: > I thought about suggesting this but I think this retain-based design is probably mildly more efficient.

Yeah, I think so too. At least for functions that create many tracked structs. It also means that there's one extra allocation for every tracked function.

MichaReiser:
The miri errors seem unrelated. I don't understand the `ViewCaster` code well enough but isn't it storing an object with its vtable? Maybe we could try something closer to what [`anyhow` does](https://github.com/dtolnay/anyhow/blob/master/src/error.rs) and it hopefully makes miri happy

nikomatsakis:
Ah, yeah, I was told that the technique I was using was invalid at some point and there was a suggested better way. 

Oops, it appears that this PR needs a rebase now :(

nikomatsakis (approved):

Saddamabdullh (commented):
Print:(saddam)

PR #588, opened 2024-10-03
Rewrite `Views` to be Miri compatible

ibraheemdev:
Rewrite the `Views` type to use a manual VTable instead of fat pointer manipulation. I find this pattern to be much clearer than the existing code.

The current implementation is also aliasing `Box` through the `OpaqueBoxDyn` and `Box<dyn Free>` pointers, which is dubious, although Miri seemed to be pleased with the small diff:
```diff
- type OpaqueBoxDyn = [u8; std::mem::size_of::<Box<dyn CastTo<Dummy>>>()];
+ type OpaqueBoxDyn = [MaybeUninit<u8>; std::mem::size_of::<Box<dyn CastTo<Dummy>>>()];
```

### DIFF
--- a/src/views.rs
+++ b/src/views.rs
@@ -24,65 +24,54 @@ use std::{
 #[derive(Clone)]
 pub struct Views {
     source_type_id: TypeId,
-    view_casters: Arc<AppendOnlyVec<ViewCaster>>,
+    view_casters: Arc<AppendOnlyVec<DynViewCaster>>,
 }
 
-/// A ViewCaster contains a trait object that can cast from the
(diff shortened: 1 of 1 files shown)


nikomatsakis (commented):
I gave this a quick look and it *seems* reasonable. I know I considered this approach and decided against it; I'm trying to remember why. I'll look it over more closely later today.

nikomatsakis (approved):
Read it over again. I like it! Nice work.

nikomatsakis:
I'd merge this but clippy is grouchy.

MichaReiser:
I'll fix the clippy issue and merge the PR afterward.

PR #589, opened 2024-10-04
Add a way to use reference-types in InternedStruct::new

### LINKED ISSUE #583: Proposal: in `new` methods on intern structs, leverage `Borrow` or `ToOwned` traits or something
I often wind up with an `Identifier` interned struct:

```rust
#[salsa::interned]
struct Identifier<'db> { text: String }
```

and then I want to create an instance from an `&str`. It's silly that I have to allocate an owned string for that. The hashmap methods already support the ability to start with a borrowed value and "promote" to an owned value if needed, we should leverage that.

puuuuh:
Closes #583

### DIFF
--- a/components/salsa-macro-rules/src/setup_interned_struct.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct.rs
@@ -67,7 +67,12 @@ macro_rules! setup_interned_struct {
 
             type StructData<$db_lt> = ($($field_ty,)*);
 
-            struct StructKey<$db_lt, $($indexed_ty: $zalsa::interned::Lookup<$field_ty>),*>($($indexed_ty,)* std::marker::PhantomData<&$db_lt ()>,);
+            /// Key to use during hash lookups. Each field is some type that implements `Lookup<T>`
+            /// for the owned type. This permits interning with an `&str` when a `String` is required and so forth.
+            struct StructKey<$db_lt, $($indexed_ty: $zalsa::interned::Lookup<$field_ty>),*>(
+                $($indexed_ty,)*
+                std::marker::PhantomData<&$db_lt ()>,
+            );
 
             impl<$db_lt, $($indexed_ty: $zalsa::interned::Lookup<$field_ty>),*> $zalsa::interned::Lookup<StructData<$db_lt>>
                 for StructKey<$db_lt, $($indexed_ty),*> {
--- a/components/salsa-macro-rules/src/setup_interned_struct.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct.rs
@@ -85,6 +85,7 @@ macro_rules! setup_interned_struct {
                     ($($zalsa::interned::Lookup::eq(&self.$field_index, &data.$field_index) && )* true)
                 }
 
+                #[allow(unused_unit)]
                 fn into_owned(self) -> StructData<$db_lt> {
                     ($($zalsa::interned::Lookup::into_owned(self.$field_index),)*)
                 }
--- a/components/salsa-macro-rules/src/setup_interned_struct.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct.rs
@@ -32,6 +32,9 @@ macro_rules! setup_interned_struct {
         // Indices for each field from 0..N -- must be unsuffixed (e.g., `0`, `1`).
         field_indices: [$($field_index:tt),*],
 
+        // Indexed types for each field (T0, T1, ...)
+        field_indexed_tys: [$($indexed_ty:ident),*],
+
         // Number of fields
         num_fields: $N:literal,
 
@@ -62,10 +65,30 @@ macro_rules! setup_interned_struct {
 
             type $Configuration = $Struct<'static>;
 
+            type StructData<$db_lt> = ($($field_ty,)*);
+
+            struct StructKey<$db_lt, $($indexed_ty: $zalsa::interned::Lookup<$field_ty>),*>($($indexed_ty,)* std::marker::PhantomData<&$db_lt ()>,);
+
+            impl<$db_lt, $($indexed_ty: $zalsa::interned::Lookup<$field_ty>),*> $zalsa::interned::Lookup<StructData<$db_lt>>
+                for StructKey<$db_lt, $($indexed_ty),*> {
+
+                fn hash<H: std::hash::Hasher>(&self, h: &mut H) {
+                    $($zalsa::interned::Lookup::hash(&self.$field_index, &mut *h);)*
+                }
+
+                fn eq(&self, data: &StructData<$db_lt>) -> bool {
+                    ($($zalsa::interned::Lookup::eq(&self.$field_index, &data.$field_index) && )* true)
+                }
+
+                fn into_owned(self) -> StructData<$db_lt> {
+                    ($($zalsa::interned::Lookup::into_owned(self.$field_index),)*)
+                }
+            }
+
             impl $zalsa_struct::Configuration for $Configuration {
                 const DEBUG_NAME: &'static str = stringify!($Struct);
-                type Data<$db_lt> = ($($field_ty,)*);
-                type Struct<$db_lt> = $Struct<$db_lt>;
+                type Data<'a> = StructData<'a>;
+                type Struct<'a> = $Struct<'a>;
                 fn struct_from_id<'db>(id: salsa::Id) -> Self::Struct<'db> {
                     $Struct(id, std::marker::PhantomData)
                 }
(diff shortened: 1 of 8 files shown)


src/interned.rs
```diff
@@ -288,3 +306,58 @@ where
...
+    }
+
+    fn eq(&self, data: &String) -> bool {
+        self == data
+    }
+
+    fn into_owned(self) -> String {
+        self.to_owned()
+    }
+}
```
nikomatsakis, line 363: We should add `Vec` and `[]` too... and maybe `PathBuf` and `Path`?

src/interned.rs
```diff
@@ -137,7 +137,25 @@ where
 
         // Optimisation to only get read lock on the map if the data has already
         // been interned.
```
nikomatsakis, line 139: We should add a comment here about how we have to use the raw API because we want to avoid creating an owned copy of the data if not necessary and how the `Borrow` trait isn't flexible enough for this use case (it requires that an `&self` be converted to `&Borrow` but we have a struct that wraps a number of fields).

src/interned.rs
```diff
@@ -137,7 +137,25 @@ where
...
+        let mut hasher = self.key_map.hasher().build_hasher();
+        data.hash(&mut hasher);
+        let data_hash = hasher.finish();
+        let shard = self.key_map.determine_shard(data_hash as _);
+        {
+            let lock = self.key_map.shards()[shard].read();
+            if let Some(bucket) = lock.find(data_hash, |(a, _)| {
+                // lifetime shrink
+                let a: &C::Data<'db> = unsafe { std::mem::transmute(a) };
```
nikomatsakis, line 148:
```suggestion
                let a: &C::Data<'db> = unsafe { self.to_internal_data(a) };
```

src/interned.rs
```diff
@@ -137,7 +137,25 @@ where
...
+                Lookup::eq(&data, a)
+            }) {
+                return C::struct_from_id(unsafe { *bucket.as_ref().1.get() });
```
nikomatsakis, line 151:
```suggestion
                // SAFETY: Read lock on map is held during this block
                return C::struct_from_id(unsafe { *bucket.as_ref().1.get() });
```

Is this correct?

src/interned.rs
```diff
@@ -137,7 +137,25 @@ where
...
+            }
+        };
+
+        let data = data.into_owned();
+
         let internal_data = unsafe { self.to_internal_data(data) };
+
         if let Some(guard) = self.key_map.get(&internal_data) {
```
nikomatsakis, line 159: This `if let` can be removed

components/salsa-macro-rules/src/setup_interned_struct.rs
```diff
@@ -62,10 +65,30 @@ macro_rules! setup_interned_struct {
 
             type $Configuration = $Struct<'static>;
 
+            type StructData<$db_lt> = ($($field_ty,)*);
+
+            struct StructKey<$db_lt, $($indexed_ty: $zalsa::interned::Lookup<$field_ty>),*>($($indexed_ty,)* std::marker::PhantomData<&$db_lt ()>,);
```
nikomatsakis, line 70:
```suggestion
            /// Key to use during hash lookups. Each field is some type that implements `Lookup<T>`
            /// for the owned type. This permits interning with an `&str` when a `String` is required and so forth.
            struct StructKey<$db_lt, $($indexed_ty: $zalsa::interned::Lookup<$field_ty>),*>(
                $($indexed_ty,)*
                std::marker::PhantomData<&$db_lt ()>,
            );
```

src/interned.rs
```diff
@@ -288,3 +306,58 @@ where
         &self.syncs
     }
 }
+
+pub trait Lookup<O>
```
nikomatsakis, line 310:
```suggestion
/// The `Lookup` trait is a more flexible variant on [`std::borrow::Borrow`]
/// and [`std::borrow::ToOwned`]. It is implemented by "some type that can
/// be used as the lookup key for `O`". This means that `self` 
/// can be hashed and compared for equality with values of type `O`
/// without actually creating an owned value. It `self` needs to be interned,
/// it can be converted into an equivalent value of type `O`.
///
/// The canonical example is `&str: Lookup<String>`. However, this example
/// alone can be handled by [`std::borrow::Borrow`][]. In our case, we may have
/// multiple keys accumulated into a struct, like `ViewStruct: Lookup<(K1, ...)>`,
/// where `struct ViewStruct<L1: Lookup<K1>...>(K1...)`. The `Borrow` trait
/// requires that `&(K1...)` be convertible to `&ViewStruct` which just isn't
/// possible. `Lookup` instead offers direct `hash` and `eq` methods.
pub trait Lookup<O>
```

tests/interned-struct-with-lifetime.rs
```diff
@@ -29,3 +37,24 @@ fn execute() {
...
+}
+#[test]
+fn interning_returns_equal_keys_for_equal_data_multi_field() {
+    let db = salsa::DatabaseImpl::new();
+    let s1 = InternedTwoFields::new(&db, "Hello, ".to_string(), "World");
+    let s2 = InternedTwoFields::new(&db, "World, ", "Hello".to_string());
+    let s1_2 = InternedTwoFields::new(&db, "Hello, ", "World");
+    let s2_2 = InternedTwoFields::new(&db, "World, ", "Hello");
+    assert_eq!(s1, s1_2);
+    assert_eq!(s2, s2_2);
```
nikomatsakis, line 59: Let's also test that `"Hello World", ""` and `"Hello ", "World"` does NOT intern to the same thing =)

src/interned.rs
```diff
@@ -137,7 +137,25 @@ where
 
         // Optimisation to only get read lock on the map if the data has already
         // been interned.
+        let mut hasher = self.key_map.hasher().build_hasher();
+        data.hash(&mut hasher);
+        let data_hash = hasher.finish();
```
nikomatsakis, line 142:
```suggestion
        // We need to use the raw API for this lookup. See the [`Lookup`][] trait definition for an explanation of why.
        let data_hash = {
            let mut hasher = self.key_map.hasher().build_hasher();
            data.hash(&mut hasher);
            hasher.finish()
        };
```

nikomatsakis (changes requested):
This is nice! Some suggestions, comments, and nits.

src/interned.rs
puuuuh: I need from_internal_data here because i don't want to use GAT for Owned type in Lookup trait, what is necessary to express eq(&Key<'db>, &Data<'any_lifetime>), but it requires &'db C::Data and we have only &'1 C::Data here

nikomatsakis: (Too annoying?)

nikomatsakis: Also, yeah, I had it wrong, I see that it should be `from_internal_data`....do we not have this method though?

src/interned.rs
```diff
@@ -114,19 +114,19 @@ where
         unsafe { std::mem::transmute(data) }
     }
 
-    pub fn intern_id<'db>(
+    pub fn intern_id<'db, T: Lookup<C::Data<'db>>>(
         &'db self,
         db: &'db dyn crate::Database,
-        data: C::Data<'db>,
+        data: T,
```
nikomatsakis, line 120:
```suggestion
        data: impl Lookup<C::Data<'db>>,
```

can we do this?

src/interned.rs
```diff
@@ -114,19 +114,19 @@ where
...
         C::deref_struct(self.intern(db, data)).as_id()
     }
 
     /// Intern data to a unique reference.
-    pub fn intern<'db>(
+    pub fn intern<'db, T: Lookup<C::Data<'db>>>(
         &'db self,
         db: &'db dyn crate::Database,
-        data: C::Data<'db>,
+        data: T,
```
nikomatsakis, line 129:
```suggestion
        data: impl Lookup<C::Data<'db>>,
```

as above

nikomatsakis (changes requested):

src/interned.rs
puuuuh: Yes, GAT bring us to more annoying bounds. Self::from_internal_data requires &'db C::Data<'db> and idk how to express this bound on eq callback

nikomatsakis:
Hi @puuuuh -- can you rebase + fix the clippy warnings? The miri failure should go away then.

nikomatsakis:
I'll take a look a bit later at your other comments, are there any outstanding questions? I'd like to merge this.

nikomatsakis (approved):

nikomatsakis:
This looks great! Let's merge it, we can probably refine the unsafe code a bit but overall looks good.

PR #591, opened 2024-10-12
Add IngredientIndex to KeyStruct

### LINKED ISSUE #590: Input change cause panic on next queries
From Zulip (thanks Andrey for the minimal repro) :

```rust
#[test]
fn hello() {
    use salsa::{Database, DatabaseImpl, Setter};

    #[salsa::input]
    struct Bool {
        value: bool,
    }

    #[salsa::tracked]
    struct True<'db> {}

    #[salsa::tracked]
    struct False<'db> {}

    #[salsa::tracked]
    fn hello(db: &dyn Database, bool: Bool) {
        if bool.value(db) {
            True::new(db);
        } else {
            False::new(db);
        }
    }

    let mut db = DatabaseImpl::new();
    let input = Bool::new(&db, false);
    hello(&mut db, input);
    input.set_value(&mut db).to(true);
    hello(&mut db, input);
}
```

Output :

```
assertion `left == right` failed: page has hidden type `"salsa::table::Page<salsa::tracked_struct::Value<mitki_parse::hello::False>>"` but `"salsa::table::Page<salsa::tracked_struct::Value<mitki_parse::hello::True>>"` was expected
```

Stacktrace in `table.rs` : 

```rust
impl dyn TablePage {
    fn assert_type<T: Any>(&self) -> &T {
        assert_eq!( // <-- Assertion failure here
            Any::type_id(self),
            TypeId::of::<T>(),
            "page has hidden type `{:?}` but `{:?}` was expected",
            self.hidden_type_name(),
            std::any::type_name::<T>(),
        );

        // SAFETY: Assertion above
        unsafe { transmute_data_ptr::<dyn TablePage, T>(self) }
    }
}
```

puuuuh:
Closes #590

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -7,7 +7,7 @@ use crate::{
     key::{DatabaseKeyIndex, DependencyIndex},
     tracked_struct::{Disambiguator, KeyStruct},
     zalsa_local::EMPTY_DEPENDENCIES,
-    Cycle, Revision,
+    Cycle, IngredientIndex, Revision,
 };
 
 use super::zalsa_local::{EdgeKind, QueryEdges, QueryOrigin, QueryRevisions};
@@ -45,7 +45,7 @@ pub(crate) struct ActiveQuery {
     /// This table starts empty as the query begins and is gradually populated.
     /// Note that if a query executes in 2 different revisions but creates the same
     /// set of tracked structs, they will get the same disambiguator values.
-    disambiguator_map: FxHashMap<u64, Disambiguator>,
+    disambiguator_map: FxHashMap<(IngredientIndex, u64), Disambiguator>,
 
     /// Map from tracked struct keys (which include the hash + disambiguator) to their
     /// final id.
@@ -155,10 +155,14 @@ impl ActiveQuery {
         self.input_outputs.clone_from(&cycle_query.input_outputs);
     }
 
-    pub(super) fn disambiguate(&mut self, hash: u64) -> Disambiguator {
+    pub(super) fn disambiguate(
+        &mut self,
+        ingredient_index: IngredientIndex,
+        hash: u64,
+    ) -> Disambiguator {
         let disambiguator = self
             .disambiguator_map
-            .entry(hash)
+            .entry((ingredient_index, hash))
             .or_insert(Disambiguator(0));
         let result = *disambiguator;
         disambiguator.0 += 1;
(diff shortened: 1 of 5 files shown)


src/active_query.rs
```diff
@@ -45,7 +45,7 @@ pub(crate) struct ActiveQuery {
     /// This table starts empty as the query begins and is gradually populated.
     /// Note that if a query executes in 2 different revisions but creates the same
     /// set of tracked structs, they will get the same disambiguator values.
-    disambiguator_map: FxHashMap<u64, Disambiguator>,
+    disambiguator_map: FxHashMap<(IngredientIndex, u64), Disambiguator>,
```
MichaReiser, line 48: Does this change allow us to remove the DatabaseKeyIndex from the id map below because the index can now be queried from the key?

puuuuh: No, its different index, afaik. IngredientIndex is kinda TypeId for tracked struct

nikomatsakis (changes requested):
Nice fix! Left some cleanup suggestions.

src/tracked_struct.rs
```diff
@@ -148,15 +148,34 @@ where
...
+    }
+}
+
+/// Stores the data that (almost) uniquely identifies a tracked struct.
+/// This includes the ingredient index of that struct type plus the hash of its id fields.
+/// This is mapped to a disambiguator -- a value that starts as 0 but increments each round,
+/// allowing for multiple tracked structs with the same hash and ingredient_index
+/// created within the query to each have a unique id.
+#[derive(Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Copy, Clone)]
+pub struct DisambiguationKey {
```
nikomatsakis, line 171: It seems to me that the naming of `DisambiguationKey` is not ideal, given e.g. `KeyStruct`. I'd like some names that make the relationship clearer. Here are some possibilities -- hopefully it's clear which is meant to be which in each case!

* `HashStruct` and `KeyStruct` (close to today)
* `IdentityHash` and `Identity`
* `IdentityHash` and `DisambiguatedIdentityHash`
* `IngredientAndHash` and `DisambiguatorIngredientAndHash` (just name it after the contents)

What do you think @puuuuh?

nikomatsakis (changes requested):
Nice! I'm still toying with the name. Thoughts?

src/tracked_struct.rs
puuuuh: IdentityHash and Identity is the best variant, imo

puuuuh: Also something like IdentityHash-StructCompoundId?

nikomatsakis (approved):

PR #593, opened 2024-10-13
fix: Replace `SelfTy` with actual type in tracked methods

### LINKED ISSUE #577: `#[salsa::tracked]` on methods in impls doesn't work with associated types
If you have ...

```rust
trait Test {
    type Type;

    fn get(self) -> Self::Type;
}

#[salsa::tracked]
impl Test for () {
    type Type = ();

    #[salsa::tracked]
    fn get(self) -> Self::Type {
        ()
    }
}
```

you get errors about `Self::Type` not being accessible in the definition of `#[salsa::tracked]`. Not sure best way to deal with this, it's annoying, probably have to massage the generated code somehow or another.

ShoyuVanilla:
Fixes #577

### DIFF
--- a/components/salsa-macros/src/debug.rs
+++ b/components/salsa-macros/src/debug.rs
@@ -33,7 +33,10 @@ pub(crate) fn dump_tokens(input_name: impl ToString, tokens: TokenStream) -> Tok
                 rustfmt.wait_with_output()
             })
             .map(|output| eprintln!("{}", String::from_utf8_lossy(&output.stdout)))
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):
This looks good! We could probably use a few more tests to cover the edge cases, but the logic looks right, and I'd be ok letting people report bugs and fixing them...

nikomatsakis:
I enabled CI. Presuming it passes, I say we merge.

PR #594, opened 2024-10-15
Add `Lookup<Box<T>>` for `&T`

gvozdvmozgu:


MichaReiser (approved):

src/interned.rs
```diff
@@ -374,6 +374,20 @@ impl Lookup<String> for &str {
...
+
+    fn eq(&self, data: &Box<str>) -> bool {
+        **self == **data
+    }
+
+    fn into_owned(self) -> Box<str> {
+        self.into()
+    }
+}
+
```
MichaReiser, line 390: Wdyt of implementing `Lookup` for all `Box<T>`:

```suggestion
impl<'a, T> Lookup<Box<T>> for &'a T
where
    T: Hash + Eq,
    Box<T>: From<&'a T>,
{
    fn hash<H: Hasher>(&self, h: &mut H) {
        Hash::hash(self, &mut *h)
    }

    fn eq(&self, data: &Box<T>) -> bool {
        **self == **data
    }

    fn into_owned(self) -> Box<T> {
        Box::from(self)
    }
}

```

Veykril: *`T: ?Sized + Hash + Eq,`

PR #596, opened 2024-10-15
fix: Inconsistent behaviour with lifetime elision on tracked fn

### LINKED ISSUE #519: inconsistent behavior with respect to lifetime elision
In #518 the behavior with respect to lifetime elision for tracked functions is kind of inconsistent:

* For the `&dyn Db` parameter, we always infer the elided lifetime to be `'db`
* For other random inputs, we basically disallow elision: if you use a `'_`, it will wind up in the value of an associated type, and give an error.
* We'll also do one weird thing which is if you write `fn foo<'db>(db: &dyn Db, x: Foo<'db>)`, I believe the `db` becomes `&'db dyn Db`.

In reality, there is only one lifetime you can correctly use, which is `'db` -- salsa tracked functions must not take references except for a case like `TrackedStruct<'db>` or `InternedStruct<'db>`. 

The current setup seems weird but there are two different choices we could make:

* Introduce a `'db` if it's not already there and replace all elided lifetimes with `'db`.
* Forbid elided lifetimes in the inputs unless there is exactly one (i.e., on the database). This is ~the current behavior, actually, except for the third bullet, which we should fix.

The former seems more convenient: there's only one thing you could want, so let's do it.

The latter is more consistent with non-tracked-functions, and it means if you remove the `#[salsa::tracked]`, your function keeps compiling. 

I'm inclined towards the latter for now.

ShoyuVanilla:
Fixes #519

I think that there would be roughly two options for fixing this;

1. Returning error while proc macro expansion
2. Inserting a type alias like `type __T<'db> = DbInputArgType;` inside the expanded function's body so that the expanded code fails to compile

In this PR, I chose the first one because the second is little bit non-intuitive and the expanded code is slightly more bloated.
But the second one also has its own merit; It can utilize rustc's more detailed error message and is more consistent with the case that the other input parameter has an elided lifetime.
So, I'm okay to switch to the second one if I have to 😄

### DIFF
--- a/tests/compile-fail/tracked_fn_incompatibles.rs
+++ b/tests/compile-fail/tracked_fn_incompatibles.rs
@@ -34,4 +34,38 @@ fn tracked_fn_with_too_many_arguments_for_specify(
 ) -> u32 {
 }
 
+#[salsa::interned]
+struct MyInterned<'db> {
+    field: u32,
+}
+
+#[salsa::tracked]
+fn tracked_fn_with_lt_param_and_elided_lt_on_db_arg1<'db>(
+    db: &dyn Db,
+    interned: MyInterned<'db>,
+) -> u32 {
+    interned.field(db) * 2
+}
+
+#[salsa::tracked]
+fn tracked_fn_with_lt_param_and_elided_lt_on_db_arg2<'db_lifetime>(
+    db: &dyn Db,
+    interned: MyInterned<'db_lifetime>,
+) -> u32 {
+    interned.field(db) * 2
+}
+
+#[salsa::tracked]
+fn tracked_fn_with_lt_param_and_elided_lt_on_input<'db>(
+    db: &'db dyn Db,
+    interned: MyInterned,
+) -> u32 {
+    interned.field(db) * 2
+}
+
+#[salsa::tracked]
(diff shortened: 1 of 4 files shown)


nikomatsakis (approved):
I like it.

tests/compile-fail/tracked_fn_incompatibles.rs
```diff
@@ -34,4 +34,30 @@ fn tracked_fn_with_too_many_arguments_for_specify(
 ) -> u32 {
 }
 
+#[salsa::interned]
+struct MyInterned<'db> {
+    field: u32,
+}
+
+#[salsa::tracked]
+fn tracked_fn_with_lt_param_and_elided_lt_on_db_arg<'db>(
```
nikomatsakis, line 43: Can you add a variant of this test where the lifetime is named `'db1` or something like that, to show how that is reflected in the error message?

nikomatsakis (changes requested):
Nit: one test

tests/compile-fail/tracked_fn_incompatibles.rs
ShoyuVanilla: Sure

nikomatsakis (approved):

Issue #597, opened 2024-10-16
LRU ("garbage collection") for interned values


nikomatsakis:
As described in [this hackmd](https://hackmd.io/@salsa/Hks2VOnj0), we would like to support LRU-style collection for interned values so that we can cap maximum memory consumption.

nikomatsakis:
# Mentoring instructions

Here are some of the steps required

* Remove the [`reset_at` field](https://github.com/salsa-rs/salsa/blob/2af849b9a911a72da699df21c48e2f12c211d0e7/src/interned.rs#L62-L66) and [`reset` method](https://github.com/salsa-rs/salsa/blob/2af849b9a911a72da699df21c48e2f12c211d0e7/src/interned.rs#L188-L192). These were used to record when the intern table was completed emptied. I don't think it's useful and nobody uses it, so let's just remove that capability.
    * `reset_at` was also used in [`maybe_changed_after`](https://github.com/salsa-rs/salsa/blob/2af849b9a911a72da699df21c48e2f12c211d0e7/src/interned.rs#L203-L210); for now, we can just make this method return false, since interned data never changes. We'll revisit this later on.
* Today a call to `intern` [records a dependency index on the *entire intern ingredient*](https://github.com/salsa-rs/salsa/blob/2af849b9a911a72da699df21c48e2f12c211d0e7/src/interned.rs#L132-L136)
    * This concept (a dependency on the entire table) will be removed entirely when we are done. That can be a nice follow-up refactoring, since we can delete the distinction between a `DatabaseKeyIndex` (a dependency on a specific id) and `DependencyIndex` (might depend on the ingredient as a whole).
    * We want to modify this to record a read of the specific interned id:
        * So we have to move that call to `report_tracked_read` down to the places where the fn returns, e.g., [here](https://github.com/salsa-rs/salsa/blob/2af849b9a911a72da699df21c48e2f12c211d0e7/src/interned.rs#L144), [here](https://github.com/salsa-rs/salsa/blob/2af849b9a911a72da699df21c48e2f12c211d0e7/src/interned.rs#L152), and [here](https://github.com/salsa-rs/salsa/blob/2af849b9a911a72da699df21c48e2f12c211d0e7/src/interned.rs#L169). We will report a tracked read of `DependencyIndex { ingredient_index: self.ingredient, key_index: Some(id) }` instead of the `for_table` method (which should be deadcode I believe).
* **Optional:** refactor to remove `DependencyIndex`
[...]

PR #602, opened 2024-10-23
Enable Garbage Collection for Interned Values

ibraheemdev:
Per the mentoring instructions in https://github.com/salsa-rs/salsa/issues/597.

Still needs some tests. Note this does not actually implement garbage collection but adds the necessary plumbing to do so.

### DIFF
--- a/src/function/specify.rs
+++ b/src/function/specify.rs
@@ -41,8 +41,7 @@ where
         //
         // Now, if We invoke Q3 first, We get one result for Q2, but if We invoke Q4 first, We get a different value. That's no good.
         let database_key_index = <C::Input<'db>>::database_key_index(db.as_dyn_database(), key);
-        let dependency_index = database_key_index.into();
-        if !zalsa_local.is_output_of_active_query(dependency_index) {
+        if !zalsa_local.is_output_of_active_query(database_key_index) {
             panic!("can only use `specify` on salsa structs created during the current tracked fn");
         }
 
@@ -105,7 +104,7 @@ where
 
         // Record that the current query *specified* a value for this cell.
         let database_key_index = self.database_key_index(key);
-        zalsa_local.add_output(database_key_index.into());
+        zalsa_local.add_output(database_key_index);
     }
 
     /// Invoked when the query `executor` has been validated as having green inputs
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -3,7 +3,6 @@ use dashmap::SharedValue;
 use crate::durability::Durability;
 use crate::function::VerifyResult;
 use crate::ingredient::fmt_index;
-use crate::key::InputDependencyIndex;
 use crate::plumbing::{IngredientIndices, Jar};
 use crate::table::memo::MemoTable;
 use crate::table::sync::SyncTable;
@@ -61,12 +60,6 @@ pub struct IngredientImpl<C: Configuration> {
     ///
     /// Deadlock requirement: We access `value_map` while holding lock on `key_map`, but not vice versa.
     key_map: FxDashMap<C::Fields<'static>, Id>,
-
-    /// Stores the revision when this interned ingredient was last cleared.
-    /// You can clear an interned table at any point, deleting all its entries,
-    /// but that will make anything dependent on those entries dirty and in need
-    /// of being recomputed.
-    reset_at: Revision,
 }
 
 /// Struct storing the interned fields.
@@ -120,7 +113,6 @@ where
         Self {
             ingredient_index,
             key_map: Default::default(),
-            reset_at: Revision::start(),
         }
     }
 
@@ -178,11 +170,6 @@ where
         C::Fields<'db>: HashEqLike<Key>,
     {
         let zalsa_local = db.zalsa_local();
-        zalsa_local.report_tracked_read_simple(
-            InputDependencyIndex::for_table(self.ingredient_index),
-            Durability::MAX,
-            self.reset_at,
-        );
 
         // Optimization to only get read lock on the map if the data has already been interned.
         let data_hash = self.key_map.hasher().hash_one(&key);
@@ -198,7 +185,17 @@ where
             let lock = shard.read();
             if let Some(bucket) = lock.find(data_hash, eq) {
                 // SAFETY: Read lock on map is held during this block
-                return unsafe { *bucket.as_ref().1.get() };
+                let id = unsafe { *bucket.as_ref().1.get() };
+
+                // Record a dependency on this value.
+                let index = self.database_key_index(id);
+                zalsa_local.report_tracked_read_simple(
+                    index,
+                    Durability::MAX,
+                    Revision::start(),
+                );
+
+                return id;
             }
         }
 
@@ -207,35 +204,62 @@ where
             self.key_map.hasher().hash_one(element)
         }) {
             // Data has been interned by a racing call, use that ID instead
-            Ok(slot) => unsafe { *slot.as_ref().1.get() },
+            Ok(slot) => {
+                let id = unsafe { *slot.as_ref().1.get() };
+                drop(slot);
+
+                // Record a dependency on this value.
+                let index = self.database_key_index(id);
+                zalsa_local.report_tracked_read_simple(
+                    index,
+                    Durability::MAX,
+                    Revision::start(),
+                );
+
+                return id;
+            }
+
             // We won any races so should intern the data
             Err(slot) => {
                 let zalsa = db.zalsa();
                 let table = zalsa.table();
+
                 let id = zalsa_local.allocate(table, self.ingredient_index, |id| Value::<C> {
                     fields: unsafe { self.to_internal_data(assemble(id, key)) },
                     memos: Default::default(),
                     syncs: Default::default(),
                 });
-                unsafe {
-                    lock.insert_in_slot(
-                        data_hash,
-                        slot,
-                        (
-                            table.get::<Value<C>>(id).fields.clone(),
-                            SharedValue::new(id),
-                        ),
-                    )
-                };
+
+                let value = (
+                    table.get::<Value<C>>(id).fields.clone(),
+                    SharedValue::new(id),
+                );
+
+                unsafe { lock.insert_in_slot(data_hash, slot, value) };
+
                 debug_assert_eq!(
                     data_hash,
                     self.key_map
                         .hasher()
                         .hash_one(table.get::<Value<C>>(id).fields.clone())
                 );
+
+                // Record a dependency on this value.
+                let index = self.database_key_index(id);
+                zalsa_local.report_tracked_read_simple(
+                    index,
+                    Durability::MAX,
+                    Revision::start(),
+                );
+
                 id
             }
         }
+    }
+
+    /// Returns the database key index for an interned value with the given id.
+    pub fn database_key_index(&self, id: Id) -> DatabaseKeyIndex {
+        DatabaseKeyIndex::new(self.ingredient_index, id)
     }
 
     /// Lookup the data for an interned value based on its id.
@@ -265,12 +289,6 @@ where
             .filter_map(|(_, page)| page.cast_type::<crate::table::Page<Value<C>>>())
             .flat_map(|page| page.slots())
     }
-
-    pub fn reset(&mut self, revision: Revision) {
-        assert!(revision > self.reset_at);
-        self.reset_at = revision;
-        self.key_map.clear();
-    }
 }
 
 impl<C> Ingredient for IngredientImpl<C>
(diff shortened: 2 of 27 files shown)


src/function/specify.rs
```diff
@@ -39,7 +39,7 @@ where
         //
         // Now, if We invoke Q3 first, We get one result for Q2, but if We invoke Q4 first, We get a different value. That's no good.
         let database_key_index = <C::Input<'db>>::database_key_index(db.as_dyn_database(), key);
-        let dependency_index = database_key_index.into();
+        let dependency_index = database_key_index;
         if !zalsa_local.is_output_of_active_query(dependency_index) {
```
MichaReiser, line 43: Nit
```suggestion
        if !zalsa_local.is_output_of_active_query(database_key_index) {
```

src/interned.rs
```diff
@@ -74,6 +68,10 @@ where
     data: C::Data<'static>,
     memos: MemoTable,
     syncs: SyncTable,
+    // The revision the value was first interned in.
+    first_interned_at: Revision,
+    // The most recent interned revision.
+    last_interned_at: AtomicRevision,
```
MichaReiser, line 74:
```suggestion
    /// The revision the value was first interned in.
    first_interned_at: Revision,
    /// The most recent interned revision.
    last_interned_at: AtomicRevision,
```

nikomatsakis (approved):
Yes, looks good to me!

nikomatsakis:
I see there is a test failure....have you investigated it at all?

ibraheemdev:
@nikomatsakis I'm also hitting the assertion error in this test:
```rust
#[salsa::tracked]
fn function<'db>(db: &'db dyn Database, input: Input) -> Interned<'db> {
    Interned::new(db, 0)
}

let input = Input::new(&db, 0);
function(&db, input);
input.set_field1(&mut db).to(1);
function(&db, input);
```

The function does not read the input so the interned value is memoized, meaning it is not re-interned in R2. Is `value.last_interned_at.load() >= db.zalsa().current_revision()` the correct assertion that I should be making?

MichaReiser:
> I'm also a little unclear about the testing instructions. Should I add a log event for when values are re-interned, or is there another way I can test that?

That would sound reasonable to me, similar to the backdate event. We could also consider adding an event when a value is garbage collected.

MichaReiser:
If I understand it correctly the problem is that interned ingredients created outside of a query are never revalidated because there's no corresponding query that creates them. 

I see two options:

1. We forbid creating interned values outside of queries. That would match the behavior with tracked structs. The downside is that it technically becomes impossible to call a root salsa query with more than one argument because multi-argument functions intern the arguments (outside the query). 
2. We disable garbage collection for ingredients created outside of queries because salsa can't track their creation. Salsa has to consider them as always referenced. That's a bit of a footgun but probably fine?

MichaReiser:
I took a closer look at `test_leaked_inputs_ignored` and this is a slightly different variation of the above problem in the sense that the multi-argument-query is called as part of another query, but said query never re-executes and `id_to_input` panics. 

I think the problem here is that `function::maybe_changed_after` never calls `id_to_input` when it short-circuits where it probably should, to at least mark the input still as used. However, it's unclear to me how that would work when the query using an interner is deeper in the tree (but the outermost query short-circuits, e.g. because the durability is higher)

nikomatsakis:
We had a brief discussion in our meeting today--- @ibraheemdev I think values that are interned outside of any salsa query have to be concerned immortal.

src/interned.rs
```diff
@@ -213,13 +244,19 @@ where
...
+        let value = db.zalsa().table().get::<Value<C>>(input);
+        if value.first_interned_at > revision {
+            // The slot was reused.
+            return true;
+        }
+
+        if value.first_interned_at < revision {
+            // The slot is valid in this revision but we have to sync the value's revision.
+            value.last_interned_at.store(db.zalsa().current_revision());
+        }
```
MichaReiser, line 257: I already discussed this with @ibraheemdev but I document it here in case anyone else ends up reviewing this PR or tries to fix the failing test. 

The fix here is to always set `last_interned_at` even then the `first_interned_at` is equal to `revision`
```suggestion
        if value.first_interned_at > revision {
            // The slot was reused.
            true
        } else {
            // The slot is valid in this revision but we have to sync the value's revision.
            value.last_interned_at.store(db.zalsa().current_revision());
            false
        }
```

ibraheemdev:
The `test_leaked_inputs_ignored` test is now passing, but I added [a test that still fails](https://github.com/salsa-rs/salsa/blob/650545363ef1508be1b7e84a2d6db976c9eb4455/tests/interned-revisions.rs#L19).

This one seems trickier. The query does not have a dependency on its input. However, changing the input still causes the database revision to be advanced. I'm not exactly sure why, but I believe this leads to the memoization check only being shallow, meaning that `maybe_changed_after` is never called for the interned value, so it's revision is never updated, and the read panics.

src/interned.rs
```diff
@@ -153,7 +146,17 @@ where
...
-                return C::struct_from_id(unsafe { *bucket.as_ref().1.get() });
+                let id = unsafe { *bucket.as_ref().1.get() };
+
+                // Sync the value's revision.
+                let value = zalsa.table().get::<Value<C>>(id);
+                value.last_interned_at.store(current_revision);
+
+                // Record a dependency on this value.
+                let index = self.database_key_index(id);
+                zalsa_local.report_tracked_read(index, Durability::MAX, current_revision);
```
ibraheemdev, line 157: Just making sure, is `current_revision` the correct argument to pass here?

MichaReiser:
Oh this is interesting. I have to take a closer look. We probably have a similar problem for when the input has a higher durability and Salsa skips the deep verification because of it.

MichaReiser:
Okay, I think I now understand the underlying problem. I'm not yet sure what the right fix is. I'll reach out in https://salsa.zulipchat.com/#narrow/channel/333573-Contributing-to-Salsa/topic/LRU.20.28GC.29.20for.20interned.20values.20.23597

MichaReiser:
I thought more about this and I feel somewhat confident that we should store the `Durability` on the interned value, similar to what's done for tracked structs. 

https://github.com/MichaReiser/salsa/blob/3e2d7f351894ec5b75df5bc1181cdf22b3243e95/src/tracked_struct.rs#L191-L196

What's different for interned values is that multiple queries can create the same tracked struct and the queries might also change their durability over time. For now, I'd suggest that we store the highest durability across all reads because we have to be pessimistic and assume the interned value only gets recreated after a HIGH durability change if we have one HIGH and two low durability queries that both create the same interned value. 

The read path should compare against the most recent revision for the interned value's durability instead of using the current revision. 

This should unblock this PR. What's unclear is how garbage collection would work because there's no guarantee that a query has to run in every revision. That means it's possible for interned values never to have been read in the last 1000 revisions, and they might get garbage collected because of it. However, the result of a cached query might still depend on the interned value (or even reference it in the output).

nikomatsakis:
@MichaReiser @ibraheemdev -- just to confirm, this is still pending the changes we discussed in our last sync meeting, correct?

ibraheemdev:
@nikomatsakis yes, I'll be working on it this week.

MichaReiser:
Hmm, the accumulator regression is surprising. Did we accidentally break the optimization that skips traversing queries that are known not to have any accumulated values?

MichaReiser:
I don't see anything obvious that should have caused the regression but I suspect it's simply the fact that we now need to look up the revision from the active query? 

This otherwise looks good to me. Do we need to remove any code? I remember you mentioned that we don't need the "immortal" handling.

src/interned.rs
```diff
@@ -208,42 +241,105 @@ where
...
                 let table = zalsa.table();
+
+                let (durability, last_interned_at) = match zalsa_local.active_query() {
+                    // Record the durability of the current query, along with the revision
+                    // we are interning in.
+                    Some((_, stamp)) => (stamp.durability, current_revision),
+
+                    // An interned value created outside of a query is considered immortal.
+                    // The durability in this case doesn't really matter.
+                    None => (Durability::MAX, Revision::from(usize::MAX)),
```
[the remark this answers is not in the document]
MichaReiser: Let's remove this then.

src/interned.rs
```diff
@@ -199,7 +205,34 @@ where
...
+                let value = zalsa.table().get::<Value<C>>(id);
+
+                // Sync the value's revision.
+                value.last_interned_at.store(current_revision);
+
+                let durability = if let Some((_, stamp)) = zalsa_local.active_query() {
+                    // Record the maximum durability across all queries that intern this value.
+                    let previous_durability = value
+                        .durability
+                        .fetch_max(stamp.durability.as_u8(), Ordering::AcqRel);
```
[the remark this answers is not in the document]
MichaReiser: I think `max` is correct here. It's the maximum durability over all queries that created this interned value (where the query's durability is the lowest read dependency).

The reason why I think `max` is correct is that if you have two queries a and b. Both create the same interned struct but `a` has durability `HIGH` and `b` has durability `LOW`. 

Unlike `b`, `a` is only guaranteed to re-run if an input with durability `HIGH` changes. That means, a low durability query using `a` (that doesn't go through `b`) might access a tracked struct field without Salsa ever having called `maybe_changed_after` for `a` (and, in turn, for the tracked struct) because salsa short circuits if the last revision for the query's lowest durability hasn't changed.

src/interned.rs
ibraheemdev: This feels incorrect, should it be the `last_interned_at` value?

ibraheemdev:
I'm going to try to add some more tests before this gets merged.

src/interned.rs
```diff
@@ -208,42 +241,103 @@ where
...
+                    current_revision,
+                    InputAccumulatedValues::Empty,
+                );
+
+                id
+            }
+
             // We won any races so should intern the data
             Err(slot) => {
                 let zalsa = db.zalsa();
```
Veykril, line 276: nit, we already have zalsa in scope here
```suggestion
```

src/interned.rs
```diff
@@ -208,42 +241,103 @@ where
...
+    pub fn database_key_index(&self, id: Id) -> DatabaseKeyIndex {
+        DatabaseKeyIndex::new(self.ingredient_index, id)
+    }
+
     /// Lookup the data for an interned value based on its id.
     /// Rarely used since end-users generally carry a struct with a pointer directly
     /// to the interned item.
     pub fn data<'db>(&'db self, db: &'db dyn Database, id: Id) -> &'db C::Fields<'db> {
         let internal_data = db.zalsa().table().get::<Value<C>>(id);
+        let last_changed_revision = db.zalsa().last_changed_revision(internal_data.durability());
```
Veykril, line 334:
```suggestion
        let zalsa = db.zalsa();
        let internal_data = zalsa.table().get::<Value<C>>(id);
        let last_changed_revision = zalsa.last_changed_revision(internal_data.durability());
```
to reduce the amount of dynamically dispatched calls

src/interned.rs
```diff
@@ -253,6 +347,12 @@ where
         self.data(db, C::deref_struct(s))
     }
 
+    pub fn reset(&mut self, db: &mut dyn Database) {
+        // Trigger a new revision.
+        let _zalsa_mut = db.zalsa_mut();
```
Veykril, line 352:
```suggestion
        // Trigger a new revision.
        db.zalsa_mut().new_revision();
```
The behavior here has changed on master.

src/key.rs
```diff
@@ -1,41 +1,54 @@
...
 /// An integer that uniquely identifies a particular query instance within the
-/// database. Used to track output dependencies between queries. Fully ordered and
-/// equatable but those orderings are arbitrary, and meant to be used only for
-/// inserting into maps and the like.
+/// database. Used to track input and output dependencies between queries. Fully
+/// ordered and equatable but those orderings are arbitrary, and meant to be used
+/// only for inserting into maps and the like.
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
-pub struct OutputDependencyIndex {
+pub struct DatabaseKeyIndex {
```
Veykril, line 16: What is the motivation behind discarding the type level differentiation of input and output keys btw?

MichaReiser: I'd have to take a closer look but I think the reason is simply that they're the same now? see https://github.com/salsa-rs/salsa/issues/597

ibraheemdev:
@MichaReiser this should be good to merge now, I'm not sure what the CI failures are about.

Veykril:
We don't have the lockfile committed so one of our transitive dependencies bumped their msrv in a point release which CI now pulls

MichaReiser (approved):

MichaReiser:
I can reproduce the perf regression locally but I think it's inherent to the change. The main reasons for the regression are:

* Salsa now records a dependency for each interned value, whereas before, it recorded a dependency on the ingredient. That means that salsa has to track more fine-grained dependencies than before and that book keeping comes at a cost
* Similar to recording dependencies, `maybe_changed_after` now has to load the value where it previously was sufficient to check when the table was last reset. That comes at an extra cost.

I could see how these numbers might motivate us to support interned structs with and without garbage collection. For now, I think what we have here is good enough and we can iterate on the details in separate PRs.

Issue #193, opened 2019-09-19
explore alternative impl for cycle handling


nikomatsakis:
In #147, I noted an alternative impl strategy for cycle handling that I wanted to explore:

> Anyway, I also started a branch that was implemented a bit differently. What I was trying to do was to have each active query have an associated `Arc` with a pointer to the next active query on the stack (i.e., what it is blocked on). This way you could follow a chain of queries forward across threads pretty cleanly and lazily. I seem to recall that the code in this PR was a bit more complex -- but then I don't think I've looked at it since you rebased. 

Filing this issue so I don't forget. Still hoping to explore this, but I'd also be happy to work with someone else on it.

nikomatsakis:
Motivation is at least in part to [avoid instantiating a vector for cross-thread calls](https://github.com/salsa-rs/salsa/pull/147#pullrequestreview-290450757):

https://github.com/salsa-rs/salsa/blob/a9860bf37f08c2b8029d62aacca60ea43c35d767/src/runtime.rs#L716-L720

But also to simplify the handling of cross-thread cycle collection generally by having a more standard graph.

carljm:
@nikomatsakis Is this still relevant after #603 ?

Veykril:
I believe we can close this now, the motivating comment is no longer relevant either

Issue #9, opened 2018-10-01
Fixed-point computations and other desirable cycles


nikomatsakis:
This is a sister issue to #6 to handle the case of "fixed-point computations" or other cases where a cycles are expected and do **not** represent user error.

Some examples that have arisen in the context of rustc:

- Inlining when tracing across call graphs
- Trait resolution
- ...

I think we should punt on this issue for a while. It's an area of active research in the incremental computation community, to start, and I've found that in most instances, it's desirable to reframe the issue in a different way. e.g., to handle inlining, you can have one query that finds the call graph, then other queries that traverse it, avoiding cycles.

That said, if we did want some semantics, we could probably implement a Prolog-like tabling semantics fairly easily. This is specific to a "fixed-point" computation where you are trying to find a set of things. The idea would be that when we detect a cycle we return an empty set to start, but we mark the cycle participants. Then, when the last participant in the cycle finishes, we save the result R we got and re-execute the cycle. This time, when we return that last result R. We again mark the participants in the cycle though and we keep re-executing as long as the result is changing (it should be growing each time).

PR #603, opened 2024-10-23
fixed-point iteration support

carljm:
This PR removes the existing unwind-based cycle fallback support (a plus for WASM compatibility), and replaces it with support for fixpoint iteration of cycles.

To opt in to fixpoint iteration, provide two additional arguments to `salsa::tracked` on the definition of a tracked function: `cycle_initial` and `cycle_fn`. The former is a function which should provide a provisional starting value for fixpoint iteration on this query, and the latter is a function which has the opportunity, after each iteration that failed to converge, to decide whether to continue iterating or fallback to some fixed value. See the added tests and docs for details.

Usability points that should be covered in the documentation:

- With the old cycle fallback, it was sufficient to avoid panic for at least one query in a cycle to define a cycle fallback. With fixpoint iteration, to avoid cycle panics you must define `cycle_fn` and `cycle_initial` on every query that might end up as the "head" of a cycle (that is, queried for its value while it is already executing.)
- It is entirely possible to define `cycle_fn` and `cycle_initial` so as to cause iteration to diverge and never terminate; it's up to the user to avoid this. Techniques to avoid this include a) ensuring that cycles will converge, by defining the initial value and the queries themselves monotonically (for example, in a type-inference scenario, the initial value is the bottom, or empty, type, and types will only widen, never narrow, as the cycle iterates -- thus the cycle must eventually converge to the top type, if nowhere else), and/or b) with a larger hammer, by ensuring that `cycle_fn` respects the iteration count it is given, and always halts iteration with a fallback value if the count reaches some "too large" value.
- It's also entirely possible to define `cycle_fn` and `cycle_initial` such that memoized results can vary depending only on the order in which queries occur. Avoid this by minimizing the number of tracked functions that support fixpoint iteration and ensuring initial values and fallback values are consistent among tracked functions that may occur in a cycle together.
- You can call Salsa queries from within your `cycle_fn` and `cycle_initial` queries, but if the query you call re-enters the same cycle, it could lead to unexpected behavior. Take care what queries you call inside cycle recovery functions.


nikomatsakis:
This is very cool! (Admittedly, I say this pre-review.)

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -179,12 +182,16 @@ macro_rules! setup_tracked_fn {
                     $inner($db, $($input_id),*)
                 }
 
+                fn cycle_initial<$db_lt>(db: &$db_lt dyn $Db) -> Self::Output<$db_lt> {
+                    $($cycle_recovery_initial)*(db)
```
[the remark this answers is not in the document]
carljm: Hmm. In the use cases I have in mind (e.g. for red-knot) there should not be any reason to call a query from within one of these functions. And it seems like this could be quite difficult to deal with. So unless we have clear use cases, I would rather consider this out of scope. We could just document "don't do that," or we could add some kind of explicit prevention.

src/function/execute.rs
```diff
@@ -38,53 +38,87 @@ where
...
+                            match C::recover_from_cycle(db, &new_value, iteration_count) {
+                                crate::CycleRecoveryAction::Iterate => {
+                                    iteration_count += 1;
+                                    revisions.cycle_ignore = false;
+                                    opt_last_provisional = Some(self.insert_memo(
+                                        zalsa,
+                                        id,
+                                        Memo::new(Some(new_value), revision_now, revisions),
+                                    ));
+                                    continue;
```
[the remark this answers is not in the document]
carljm: Yes, this is one of the areas where we may want to spend some more time on performance optimization. Reusing memos is tricky because in general once they are in the memo map, unless we have a mutable reference to the db (which we don't have here, we only have in between revisions), we have to assume someone else may have a reference to that memo.

src/function/execute.rs
```diff
@@ -38,53 +38,87 @@ where
...
+            if let Some(old_memo) = &opt_old_memo {
+                self.backdate_if_appropriate(old_memo, &mut revisions, &new_value);
+                self.diff_outputs(db, database_key_index, old_memo, &mut revisions);
+            }
+
+            // Did the new result we got depend on our own provisional value, in a cycle?
+            if revisions.cycle_heads.contains(&database_key_index) {
+                if let Some(last_provisional) = opt_last_provisional {
+                    if let Some(provisional_value) = &last_provisional.value {
+                        // If the new result is equal to the last provisional result, the cycle has
```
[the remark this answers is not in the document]
carljm: I think it is only possible for a memo value to be `None` if its value has been evicted by LRU. I suspect that's practically impossible here, but maybe I need to add some explicit guarantee that LRU eviction of a provisional memo never occurs.

tests/cycle_fixpoint.rs
```diff
@@ -0,0 +1,207 @@
...
+                    set.extend(b_ints);
+                    Type::Values(set.into_iter().collect())
+                }
+            }
+        }
+        result
+    }
+}
+
+#[salsa::tracked(cycle_fn=cycle_recover, cycle_initial=cycle_initial)]
```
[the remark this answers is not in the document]
carljm: Yes, I thought about this as well. It's not clear to me what is actually the better UX. Implementing a trait is somewhat more boilerplate, and in practice there's not much difference in the experience if you fail to implement one of the methods. Either way the compiler catches the error for you, because its an error in macro expansion if you don't provide both `cycle_fn` and `cycle_initial`, with a trait it would be a compiler error that you didn't fully implement the trait.

components/salsa-macro-rules/src/setup_tracked_fn.rs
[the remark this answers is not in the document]
MichaReiser: Yeah. I'm not suggesting that users should do that. I only wonder about what happens if a user does it regardless. 

Ideally, we wouldn't provide a `Db` but we can't do that because a user might want to create a tracked struct.

src/function/execute.rs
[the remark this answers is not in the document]
MichaReiser: Yeah, that's true. I wonder if we can make some use of the knowledge that these memo's can only be reached by tracked functions participating in a cycle. They would have to store the memo value somewhere globally for it to escape and that would require casting the value to `'static`?

tests/cycle_fixpoint.rs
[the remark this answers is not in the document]
carljm: I guess with a trait your IDE might give you the right signatures of the methods for free, which is kind of nice...

src/function/execute.rs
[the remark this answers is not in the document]
carljm: I suspect this comment was intended for a different thread (the one about creating lots of memos)? Maybe repost it there and delete it here, to make it easier to follow the discussion?

src/zalsa_local.rs
```diff
@@ -377,9 +352,31 @@ pub(crate) struct QueryRevisions {
...
+
+    /// True if this result is based on provisional results, and is not itself a cycle head; this
+    /// should not be used as a cached result.
+    pub(super) cycle_ignore: bool,
 }
 
 impl QueryRevisions {
+    pub(crate) fn fixpoint_initial(query: DatabaseKeyIndex) -> Self {
+        let mut cycle_heads = FxHashSet::default();
+        cycle_heads.insert(query);
```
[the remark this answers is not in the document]
carljm: This was the first thing I had tried, but it doesn't work, it results in the compiler error

```
   Compiling salsa v0.18.0 (/Users/carlmeyer/projects/salsa)
error[E0308]: mismatched types
   --> src/zalsa_local.rs:375:47
    |
375 |         let mut cycle_heads = FxHashSet::from([query]);
    |                               --------------- ^^^^^^^ expected `HashSet<_, ...>`, found `[DatabaseKeyIndex; 1]`
    |                               |
    |                               arguments to this function are incorrect
    |
    = note: expected struct `HashSet<_, FxBuildHasher>`
                found array `[DatabaseKeyIndex; 1]`
note: associated function defined here
```

I couldn't find a better way to do this (but admittedly I didn't spend very long looking.)

src/function/execute.rs
```diff
@@ -38,53 +38,87 @@ where
...
+                        // converged and we are done.
+                        if !C::values_equal(&new_value, provisional_value) {
+                            // We are in a cycle that hasn't converged; ask the user's
+                            // cycle-recovery function what to do:
+                            match C::recover_from_cycle(db, &new_value, iteration_count) {
```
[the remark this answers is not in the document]
carljm: Added several debug traces in this area; perhaps it's too much, but it seems the tracing output could be pretty confusing if it doesn't include this information to understand what is happening with the cycle.

src/function/execute.rs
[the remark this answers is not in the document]
MichaReiser: Yeah, that's true. I wonder if we can make some use of the knowledge that these memo's can only be reached by tracked functions participating in a cycle. They would have to store the memo value somewhere globally for it to escape and that would require casting the value to 'static?

components/salsa-macro-rules/src/setup_tracked_fn.rs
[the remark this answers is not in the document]
carljm: I think the simplest approach here would be to add some state on `Runtime` that causes an error if you try to push a new active query. I'll wait for Niko review before doing that, though.

src/function/execute.rs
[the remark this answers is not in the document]
carljm: Another thing to try here would be to avoid pre-emptively pushing a provisional memo for every query that supports fixpoint, and instead lazily generate and insert that provisional memo only if/when a cycle is encountered. This would really reduce the overhead in the scenario "queries support fixpoint but relatively few cycles actually encountered," which I think is the scenario whose performance we should be most concerned about here.

I think this might be doable; I'll experiment with it once I have more thorough tests in place as guardrails.

src/function/execute.rs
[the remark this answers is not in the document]
carljm: It's clearly wrong to consider the cycle converged just because our provisional result got LRU evicted, so I've changed this to an `unwrap` to fail fast in that case, with an explanation comment. Will wait for Niko review before doing more here.

tests/cycle_fixpoint.rs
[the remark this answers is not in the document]
davidbarsky: I normally don't like traits, but I feel like a trait would be ideal preferable because the annotation seems pretty darn wordy.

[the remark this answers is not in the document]
carljm: Ok, going to still wait for Niko's feedback on this point before updating, but the trait idea makes sense to me.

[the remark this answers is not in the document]
davidbarsky: Actually, I'm okay with a non-trait based approach. rust-analyzer uses a `#[salsa::cycle(path::to::function)]`-style annotation, so I think to ease the transition, avoiding a trait would be nice.

carljm:
In writing more comprehensive tests for this, I realized that it needs some changes to correctly handle multi-revision scenarios; taking it to Draft mode until I get that fixed.

tests/cycle_fixpoint.rs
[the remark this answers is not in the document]
MichaReiser: >  rust-analyzer uses a #[salsa::cycle(path::to::function)]-style annotation, so I think to ease the transition, avoiding a trait would be nice.

Is cycle handling common in ra? I do see how it reduces the diff size because it isn't necessary to make the function a trait-method but you would still have to change every use because you now have to specify both the cycle and cycle initial functions.

src/function/execute.rs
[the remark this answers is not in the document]
carljm: This is now done; it turned out to be the same thing that I needed in order to fix multiple-revisions.

carljm:
Ok, multiple-revision cases are now fixed, and we now populate the initial provisional value only lazily, in case a cycle is actually encountered, which should reduce the number of memos created by quite a lot.

Also added a bunch of tests, including multiple-revision cases and one test involving durability. Still need to add cross-thread cycle tests.

tests/cycle/main.rs
```diff
@@ -0,0 +1,755 @@
...
+// - `Np` for `min_panic`
+// - `Xp` for `max_panic`
+//
+// and `ii` is the inputs for that query, represented as a comma-separated list, with each
+// component representing an input:
+// - `a`, `b`, `c`... where the input is another node,
+// - `uXX` for `UntrackedRead(XX)`
+// - `vXX` for `Value(XX)`
+// - `sY` for `Successor(Y)`
+//
```
carljm, line 139: These are admittedly obscure-looking chicken scratches. I'm not claiming they are super readable, but they are concise enough to put into an ASCII graph diagram, and (once you get familiar with them) give a lot of information about the behavior of the test. They were really helpful to me in writing and debugging the tests.

Open to feedback that I should do this differently for better readability by future maintainers...

src/function/execute.rs
```diff
@@ -38,53 +36,114 @@ where
...
+                    &opt_owned_last_provisional
+                        .as_deref()
+                        .expect(
+                            "{database_key_index:#?} is a cycle head, \
+                            but no provisional memo found",
+                        )
+                        .value
+                        .as_ref()
+                        .expect("provisional value evicted by LRU?")
+                };
```
MichaReiser, line 92: Nit
```suggestion
                    last_provisional
                        .value
                        .as_ref()
                        .expect("provisional value evicted by LRU?")
                } else {
                    // This is our first time around the loop; a provisional value must have been
                    // inserted into the memo table when the cycle was hit, so let's pull our
                    // initial provisional value from there.
                    opt_owned_last_provisional = self.get_memo_from_table_for(zalsa, id);
                    opt_owned_last_provisional
                        .as_deref()
                        .expect(
                            "{database_key_index:#?} is a cycle head, \
                            but no provisional memo found",
                        )
                        .value
                        .as_ref()
                        .expect("provisional value evicted by LRU?")
                };
```

Clever with declaring `opt_owned_last_provisional` outside the `if..let` to extend the lifetime. I didn't know that would work.

src/function/execute.rs
```diff
@@ -38,53 +36,114 @@ where
...
                 tracing::debug!(
-                    "{database_key_index:?}: caught cycle {cycle:?}, have strategy {:?}",
-                    C::CYCLE_STRATEGY
+                    "{database_key_index:?}: execute: \
+                            I am a cycle head, comparing last provisional value \
+                            {last_provisional_value:#?} with new value {new_value:#?}"
```
MichaReiser, line 96: is value the user provided value? If so, you might not want to log that with tracing because it could be huge (e.g. an entire AST).

src/function/execute.rs
```diff
@@ -38,53 +36,114 @@ where
...
+                    opt_owned_last_provisional = self.get_memo_from_table_for(zalsa, id);
+                    &opt_owned_last_provisional
+                        .as_deref()
+                        .expect(
+                            "{database_key_index:#?} is a cycle head, \
+                            but no provisional memo found",
+                        )
+                        .value
+                        .as_ref()
+                        .expect("provisional value evicted by LRU?")
```
MichaReiser, line 91: Hehe, this `expect` message is more of a question than an explanation why the value shouldn't be `None`. Maybe rephrase to `value created in this revision should not have been evicted by LRU`

src/function/execute.rs
```diff
@@ -38,53 +36,114 @@ where
...
-                            cycle.throw()
+                // If the new result is equal to the last provisional result, the cycle has
+                // converged and we are done.
+                if !C::values_equal(&new_value, last_provisional_value) {
+                    // We are in a cycle that hasn't converged; ask the user's
+                    // cycle-recovery function what to do:
+                    match C::recover_from_cycle(db, &new_value, iteration_count) {
+                        crate::CycleRecoveryAction::Iterate => {
+                            tracing::debug!("{database_key_index:?}: execute: iterate again");
+                            iteration_count += 1;
```
MichaReiser, line 106: Let's panic with a meaningful message if the algorithm doesn't converage in `u32::MAX` steps


```suggestion
                            iteration_count = iteration_count.checked_add(1).expect("Cycle recovery to converge after u32::MAX iterations");
```

MichaReiser (commented):
The lazy creation of the initial value is a neat improvement. Nice for taking the time to work on it !

MichaReiser:
The benchmarks show a 4-5% regression. Could you take a look what's causing it?

src/function/execute.rs
carljm: > Clever with declaring `opt_owned_last_provisional` outside the `if..let` to extend the lifetime. I didn't know that would work.

Not sure if clever or ugly, I don't love it, but couldn't find a better way to make this work.

src/function/execute.rs
carljm: Yeah makes sense. On the flip side, it can be super useful for debugging cycle iteration. I'll remove it for now and we'll see.

carljm:
Initial experiments using this in the red-knot type checker are promising: https://github.com/astral-sh/ruff/pull/14029

Not yet using it for loopy control flow in that PR, but there are cycles in the core type definitions of Python builtins and standard library, which we previously had a hacky fallback in place for using Salsa's previous cycle fallback support. Moving over to fixpoint iteration just worked, and fixed the type of a builtin impacted by the cycle.

On the downside, it is a performance regression. Need to do more work there.

components/salsa-macro-rules/src/setup_tracked_fn.rs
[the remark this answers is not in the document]
carljm: Added a TODO comment on this to get Niko's feedback.

tests/cycle_fixpoint.rs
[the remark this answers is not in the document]
carljm: Added a TODO comment in the code for this question as well.

src/function/maybe_changed_after.rs
```diff
@@ -73,22 +114,29 @@ where
         );
 
         // Check if the inputs are still valid and we can just compare `changed_at`.
-        if self.deep_verify_memo(db, &old_memo, &active_query) {
-            return Some(old_memo.revisions.changed_at > revision);
+        let active_query = zalsa_local.push_query(database_key_index);
```
MichaReiser, line 117: From looking at the red knot benchmarks, the regression mainly comes from the extra `push_query` calls here (that probably also applies for queries not participating in cycles) and the constructed hash set in `deep_verify_memo` (specific to red knot?)

src/function/maybe_changed_after.rs
```diff
@@ -148,79 +201,95 @@ where
...
-                let last_verified_at = old_memo.verified_at.load();
-                for &(edge_kind, dependency_index) in edges.input_outputs.iter() {
-                    match edge_kind {
-                        EdgeKind::Input => {
-                            if dependency_index
-                                .maybe_changed_after(db.as_dyn_database(), last_verified_at)
-                            {
-                                return false;
+        loop {
+            let mut cycle_heads = FxHashSet::default();
```
MichaReiser, line 208: Is it intentional that we create a new `cycle_heads` in every iteration? Could we re-use the `cycle_heads` and instead call `clear` to avoid re-allocating the hash set on every iteration?

src/function/maybe_changed_after.rs
MichaReiser: Could we move the `push_query` call into `deep_verify_memo` after the second `shallow_verify_memo` or does that result in deadlocks? Just so that we can avoid pushing queries unless it's absolutely necessary

src/ingredient.rs
```diff
@@ -38,7 +39,7 @@ pub trait Ingredient: Any + std::fmt::Debug + Send + Sync {
         db: &'db dyn Database,
         input: Option<Id>,
         revision: Revision,
-    ) -> bool;
+    ) -> VerifyResult;
```
MichaReiser, line 42: I think this change will also come handy for the "faster accumulator" work because we'll need to also return whether the ingredient had any accumulated values.

components/salsa-macro-rules/src/setup_tracked_fn.rs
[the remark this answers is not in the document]
Veykril: If this replaces the old cycle handling, we (rust-analyzer) actually needs to be able to call queries here as we already do that in some of our cycle recovery functions today. I think most of those are in part because of assumptions else where in the code, an example being `generic_defaults` always returning a list with length matching the number of generics, and to compute the number we need to call a query. That query can run into a cycle though, so we still need to figure out the number of generics in the recovery function which means we need to call the corresponding query.

tests/cycle_fixpoint.rs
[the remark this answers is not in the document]
Veykril: its common but not common enough that it would matter (there are ~11 cycle handlers and the majority of them return a dummy result)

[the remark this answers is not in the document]
Veykril: One theoreitcal nice thing about taking the functions here directly would be that we could allow supporting closure expressions in that position inline instead of having to create a whole function (as cycles that are not going to be fixpoint are likely not gonna have complex functions here), a trait would be quite wordy imo.

src/function/execute.rs
```diff
@@ -38,53 +38,122 @@ where
...
-        let id = database_key_index.key_index;
-        let value = match Cycle::catch(|| C::execute(db, C::id_to_input(db, id))) {
-            Ok(v) => v,
-            Err(cycle) => {
+        // Our provisional value from the previous iteration, when doing fixpoint iteration.
+        // Initially it's set to None, because the initial provisional value is created lazily,
+        // only when a cycle is actually encountered.
+        let mut opt_last_provisional: Option<&Memo<<C as Configuration>::Output<'db>>> = None;
+
+        loop {
```
Veykril, line 48: Can we branch on the cycle strategy here? There is a lot of work that I don't think needs to happen for `Panic` strategies that non cycle supporting queries currently pay (and given the branching would happen on a `const` it should optimized out)

src/function/maybe_changed_after.rs
```diff
@@ -1,12 +1,41 @@
...
+pub enum VerifyResult {
+    /// Memo has changed and needs to be recomputed.
+    Changed,
+
+    /// Memo remains valid.
+    ///
+    /// Database keys in the hashset represent cycle heads encountered in validation; don't mark
+    /// memos verified until we've iterated the full cycle to ensure no inputs changed.
+    Unchanged(FxHashSet<DatabaseKeyIndex>),
+}
```
Veykril, line 23: This change (`bool` being replaced by `VerifyResult`) also seems like it will make non cycle queries pay a hefty perf price. That type has a destructor and is 40 bytes where as the previous return value could fit into a register, this can't.

src/cycle.rs
```diff
@@ -1,92 +1,16 @@
...
-                .finish()
-        })
-        .unwrap_or_else(|| {
-            f.debug_struct("Cycle")
-                .field("participants", &self.participants)
-                .finish()
-        })
-    }
+    /// Cut off iteration and use the given result value for this query.
+    Fallback(T),
```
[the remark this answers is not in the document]
carljm: In the current implementation, the initial function will still be called. If query A(0) is executing, it calls B(0), which in turn calls A(0), the second call to A(0) will call the initial function to get an initial value to return to B(0), which will compute its provisional value and return that to A(0), which will then call the cycle recovery function. If that cycle recovery function immediately returns `Fallback`, then that fallback value will be used for A(0). (I think we also need to then iterate the cycle one more time in order to populate a correct value for B(0) based on the fallback value for A(0), but that's not currently implemented.)

It seems like it could be possible to instead call the cycle recovery function first thing when we encounter the cycle, and allow it to return `Fallback` immediately, before the initial function is ever called. That seems worth experimenting with.

src/function/maybe_changed_after.rs
carljm: I'll focus first on getting consistently right behavior (including across threads) and then revisit this and other possible ways to claw back perf. Off the top of my head I'm not sure how to improve this, will need to think about it.

src/cycle.rs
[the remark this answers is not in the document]
carljm: A key difference in this PR from the current cycle handling is that in the current cycle handling _all_ cycle participants that have a cycle fallback receive a fallback value; this ensures determinism regardless of query order, but it means that you can't have some participants in the cycle compute a real value based on a fallback for some other cycle participant.

In this PR, if your recovery function falls back to some value, that fallback only applies to the individual query, and the rest of the cycle will then resolve normally using that fallback value.

src/function/maybe_changed_after.rs
Veykril: Yes of course, I am mainly noting down what I saw, no need to tackle perf nits/reviews until things are working properly.

src/cycle.rs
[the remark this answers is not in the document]
Veykril: I see, thanks, that explanation made this very clear!

Veykril:
On a related perf note, would it make sense to move the dataflow test (or a maybe slightly more involved version) into a benchmark?

src/function/maybe_changed_after.rs
Veykril: That seems weird, this isnt an extra `push_query` call after all. This merely moved down a couple lines

carljm: Yeah, we no longer pass an active query to `execute`, so we push twice instead of once.

Once the behavior is right I can look at whether we can eliminate this. Maybe we don't have to push and pop the cycle-head active query for each iteration loop.

src/ingredient.rs
carljm: I suspect this comment might be out of date, since the faster accumulator PR landed without needing this?

src/cycle.rs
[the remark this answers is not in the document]
carljm: > (I think we also need to then iterate the cycle one more time in order to populate a correct value for B(0) based on the fallback value for A(0), but that's not currently implemented.)

I was wrong, I had already implemented this.

There is a question about what should happen if the `cycle_fn` says to fallback to some value, and we provisionally insert the fallback value and iterate the cycle one more time, and the cycle still fails to converge. It would be nice if we can require that it does, because that means we still leave the cycle in a consistent converged state. But maybe we will have to just return the fallback value and leave the cycle in an inconsistent state.

I'm not sure if real-world use cases for fallback will require that we leave the cycle in an inconsistent state. For now I've added a panic, but we can change this.

components/salsa-macro-rules/src/setup_tracked_fn.rs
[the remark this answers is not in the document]
carljm: I think that calling queries in a cycle recovery function should be fine as long as you don't call something that might re-enter the same cycle. So it requires some care in what queries you call, which would need documenting. If we need to allow this, then I can remove this TODO and instead add a test showing that calling an unrelated query works fine.

src/function/maybe_changed_after.rs
carljm: We do need a separate active-query per actual execution when we iterate, because the active-query collects information from the execution. But I adjusted this so in the no-iteration case we only push one active-query (like before) instead of two, and same in maybe-changed-after.

src/function/execute.rs
carljm: Unless there's overhead simply from entering a `loop` (that we then return from without ever iterating), I think I've removed all the extra overhead for non-cyclic queries now, and we do the static check on cycle strategy before actually checking for ourselves in the cycle heads, since we can't appear in the cycle heads if we have a cycle strategy of panic.

components/salsa-macro-rules/src/setup_tracked_fn.rs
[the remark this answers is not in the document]
Veykril: Supporting it (if it doesn't add extra impl burden) would be preferable in my eyes (for now), as I am fairly sure r-a currently needs it (it might not in the future, but that's an answer we get once migration is fully finished).

nikomatsakis:
@carljm -- to confirm, not yet ready for my review, right? (or did I forget something)

carljm:
@nikomatsakis I commented last week at https://salsa.zulipchat.com/#narrow/channel/333573-Contributing-to-Salsa/topic/fixed.20point.20cycles that this was ready for review.

Two things have come up since:
1. A test failure here in CI (on the `cycle_t1_a_t2_b` test) that I have never seen locally and so far have not been able to reproduce; it suggests some kind of race condition that I need to track down and fix.
2. The PR that landed on Friday to fix accumulators conflicts badly with this PR, and I haven't completed that rebase yet (thus all the conflicts shown here.)

I think a review from you of what's here could still be useful (you might have an idea where the race condition is that is causing the test failure?) but it's also fine if you want to wait until I sort the above items.

carljm:
I may try and see if https://github.com/awslabs/shuttle can track down the race condition. It's a bit invasive to use it, since it requires that all our uses from `std::thread` and `std::sync` throughout all of Salsa go through a shim module that can `cfg` redirect them to shuttle in tests.

nikomatsakis:
Ok. I will try to find time to review just in case. Im busy this week with Things but will do my best.

carljm:
Ok, problem (2) is addressed; I've merged in latest master and resolved all conflicts. Now back to problem (1).

src/function/execute.rs
```diff
@@ -37,53 +41,132 @@ where
...
+                    // We have a last provisional value from our previous time around the loop.
+                    last_provisional
+                        .value
+                        .as_ref()
+                        .expect("provisional value should not be evicted by LRU")
+                } else {
+                    // This is our first time around the loop; a provisional value must have been
+                    // inserted into the memo table when the cycle was hit, so let's pull our
+                    // initial provisional value from there.
+                    opt_owned_last_provisional = self.get_memo_from_table_for(zalsa, id);
```
Veykril, line 79: Given we expect this to be a provisional value, it might be good to add a 
```rs
                    debug_assert!(opt_owned_last_provisional
                        .as_ref()
                        .unwrap()
                        .may_be_provisional());
```
for good measure?

carljm:
Failing tests were Linux only; there are apparently some thread interleavings that don't occur on the MacOS thread scheduler. (It seems be tied to OS, not architecture.) Figured out the logic bugs that could result on those failures and pushed fixes.

I will work on some additional tests here, and then on performance. But this PR is back in a state where tests should be passing and I think it is ready for review.

components/salsa-macro-rules/src/setup_tracked_fn.rs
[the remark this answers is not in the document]
carljm: Added tests showing that it is possible to call a Salsa query from within a cycle recovery function.

If you call a query that can re-enter the same cycle you are recovering from, that can lead to unbounded recursion. Added this to the list of caveats to document.

src/cycle.rs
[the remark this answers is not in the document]
carljm: > It seems like it could be possible to instead call the cycle recovery function first thing when we encounter the cycle, and allow it to return `Fallback` immediately, before the initial function is ever called. That seems worth experimenting with.

I did experiment with this, and I think it's doable, but it has some costs. When you do want to fixpoint iterate, your cycle recovery function will be called one extra time. (Small cost.) Cycle recovery functions have to take `value: Option<&T>` instead of `value: &T` as they do now, because on the first call there is no value to provide.

If we keep the current rule I have in place that even a fallback value must converge on one iteration, I'm not sure there is a strong use case for this change. You can just as easily return the "fallback" value you want as your "initial" value instead, and it will converge without ever needing to fallback.

If it's necessary to allow fallback values that don't converge, and we want to allow "immediate" fallback, we could revisit this. I'd rather let that be a separate follow-up driven by real use cases.

carljm:
> On a related perf note, would it make sense to move the dataflow test (or a maybe slightly more involved version) into a benchmark?

I added a benchmark based on the dataflow test. I think it's already a useful benchmark as-is, but if you have suggestions of how to make it more involved in such a way that would improve its usefulness as a benchmark, I'm open to suggestions!

src/table/sync.rs
```diff
@@ -30,27 +30,36 @@ struct SyncState {
     anyone_waiting: AtomicBool,
 }
 
+pub(crate) enum ClaimResult<'a> {
+    Retry,
+    Cycle,
+    Claimed(ClaimGuard<'a>),
+}
```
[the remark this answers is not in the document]
carljm: Thanks for the suggestion! I did this; it seemed to improve results on some benchmarks locally, not sure if real or noise; we'll see what codspeed says.

src/function/fetch.rs
```diff
@@ -42,8 +44,50 @@ where
...
+                            // We hit a cycle blocking on the cycle head; this means it's in
+                            // our own active query stack and we are responsible to resolve the
+                            // cycle, so go ahead and return the provisional memo.
+                            return memo;
+                        }
+                    }
+                    // All our cycle heads are verified final; re-fetch and we should get a
+                    // non-provisional memo.
+                    if has_verified_cycle_heads {
+                        continue 'outer;
```
MichaReiser, line 88: Would you mind adding a comment here explaining why it's okay to return the provisional value if there's no verified cycle head?

src/function/maybe_changed_after.rs
```diff
@@ -1,14 +1,45 @@
...
+    Changed,
+
+    /// Memo remains valid.
+    ///
+    /// The first inner value tracks whether the memo or any of its dependencies have an
+    /// accumulated value.
+    ///
+    /// Database keys in the hashset represent cycle heads encountered in validation; don't mark
+    /// memos verified until we've iterated the full cycle to ensure no inputs changed.
+    Unchanged(InputAccumulatedValues, FxHashSet<DatabaseKeyIndex>),
```
MichaReiser, line 26: As discussed in our sync meeting. It might be worthwhile to use an `Option<Box<FxHashSet<...>>>` here to reduce the size of `VerifyResult`. It would be even better if we could store the cycle heads somewhere else to avoid `VerifyResult` implementing `Drop` but that could be tricky and might not be worth it just to safe one extra condition.

src/function/maybe_changed_after.rs
```diff
@@ -118,14 +171,23 @@ where
         zalsa: &Zalsa,
         database_key_index: DatabaseKeyIndex,
         memo: &Memo<C::Output<'_>>,
+        allow_provisional: bool,
```
MichaReiser, line 174: I'd find it helpful to have some documentation around `allow_provisional`. When are provisional values allowed?

src/function/maybe_changed_after.rs
```diff
@@ -170,88 +251,105 @@ where
...
+
+            if cycle_heads.is_empty() {
+                old_memo.mark_as_verified(
+                    db.as_dyn_database(),
+                    zalsa.current_revision(),
+                    database_key_index,
+                    inputs,
+                );
+            }
+            if in_heads {
```
MichaReiser, line 348: Would you mind adding a comment explaining why we clear the cycle heads if the query itself is in the cycle heads?

src/function/memo.rs
```diff
@@ -119,23 +134,42 @@ pub(super) struct Memo<V> {
     /// as the current revision.
     pub(super) verified_at: AtomicCell<Revision>,
 
+    /// Is this memo verified to not be a provisional cycle result?
+    pub(super) verified_final: AtomicCell<bool>,
```
MichaReiser, line 138: Nit: 
```suggestion
    pub(super) verified_final: AtomicBool,
```

src/active_query.rs
```diff
@@ -57,10 +55,13 @@ pub(crate) struct ActiveQuery {
     /// Stores the values accumulated to the given ingredient.
     /// The type of accumulated value is erased but known to the ingredient.
     pub(crate) accumulated: AccumulatedMap,
-
+    ///
```
MichaReiser, line 58:
```suggestion

```

src/function.rs
```diff
@@ -67,15 +69,16 @@ pub trait Configuration: Any {
...
     fn execute<'db>(db: &'db Self::DbView, input: Self::Input<'db>) -> Self::Output<'db>;
 
-    /// If the cycle strategy is `Fallback`, then invoked when `key` is a participant
-    /// in a cycle to find out what value it should have.
-    ///
-    /// This invokes the recovery function given by the user.
+    /// Get the cycle recovery initial value.
+    fn cycle_initial<'db>(db: &'db Self::DbView, input: Self::Input<'db>) -> Self::Output<'db>;
+
+    /// Decide whether to iterate a cycle again or fallback.
```
MichaReiser, line 75: Can we document the meaning of `value` and `count`

src/ingredient.rs
```diff
@@ -61,7 +62,13 @@ pub trait Ingredient: Any + std::fmt::Debug + Send + Sync {
         db: &'db dyn Database,
         input: Id,
         revision: Revision,
-    ) -> MaybeChangedAfter;
+    ) -> VerifyResult;
+
+    /// Is the value for `input` in this ingredient marked as possibly a provisional cycle value?
+    fn is_verified_final<'db>(&'db self, db: &'db dyn Database, input: Id) -> bool;
+
+    /// Invoked when the current thread needs to wait for a result for the given `key_index`.
```
MichaReiser, line 70: Would you mind documenting the meaning of the return value. What does `true` indicate?

src/input.rs
```diff
@@ -204,10 +206,18 @@ impl<C: Configuration> Ingredient for IngredientImpl<C> {
...
-    ) -> MaybeChangedAfter {
+    ) -> VerifyResult {
         // Input ingredients are just a counter, they store no data, they are immortal.
         // Their *fields* are stored in function ingredients elsewhere.
-        MaybeChangedAfter::No(InputAccumulatedValues::Empty)
+        VerifyResult::unchanged()
+    }
+
+    fn is_verified_final<'db>(&'db self, _db: &'db dyn Database, _input: Id) -> bool {
+        false
```
MichaReiser, line 216: What's the reason that `Input`, `Tracked`, etc. return `false`. Aren't they always considered final?

src/zalsa_local.rs
```diff
@@ -339,12 +314,36 @@ pub(crate) struct QueryRevisions {
...
+
+    /// This result was computed based on provisional values from
+    /// these cycle heads. The "cycle head" is the query responsible
+    /// for managing a fixpoint iteration. In a cycle like
+    /// `--> A --> B --> C --> A`, the cycle head is query `A`: it is
+    /// the query whose value is requested while it is executing,
+    /// which must provide the initial provisional value and decide,
+    /// after each iteration, whether the cycle has converged or must
+    /// iterate again.
+    pub(super) cycle_heads: FxHashSet<DatabaseKeyIndex>,
```
MichaReiser, line 330: I think we should consider using a `Option<Box<FxHashSet<..>>>` here similar to `accumulated (https://github.com/salsa-rs/salsa/pull/636). Doing the same for accumulated helped improve performance by a few percent and this might be part of the reason why we see an overall regression.
```suggestion
    pub(super) cycle_heads: Option<Box<FxHashSet<DatabaseKeyIndex>>>,
```

It may also be worth considering to create a `CycleHeads` new type wrapper around `Option<Box<FxHashSet>>`

tests/parallel/cycle_a_t1_b_t2.rs
```diff
@@ -0,0 +1,79 @@
...
+            let r = query_a(&db_t1);
+            assert_eq!(r, MAX);
+        });
+
+        // Thread 2:
+        scope.spawn(move || {
+            let r = query_b(&db_t2);
+            assert_eq!(r, MAX);
+        });
+    });
```
MichaReiser, line 78: Nit: I think rewriting the test to the following will result in better error messages if only one thread converges to the final value
```suggestion
        let db_t1 = Knobs::default();
        let db_t2 = db_t1.clone();

        // Thread 1:
        let t1 = scope.spawn(move || query_a(&db_t1));

        // Thread 2:
        let t2 = scope.spawn(move || query_b(&db_t2));

        let (r_t1, r_t2) = (t1.join().unwrap(), t2.join().unwrap());

        assert_eq!((r_t1, r_t2), (MAX, MAX));
    });
```

tests/cycle.rs
```diff
@@ -0,0 +1,881 @@
...
+    a.assert_value(&db, 45);
+}
+
+/// a:Np(v59, b) -> b:Ni(v60, c) -> c:Np(b)
+///                 ^                     |
+///                 +---------------------+
+///
+/// If nothing in a cycle changed in the new revision, no part of the cycle should re-execute.
+#[test]
+fn cycle_unchanged() {
```
MichaReiser, line 851: Not required but it could be nice to have a multi threaded tests where:

* the recovery function panics
* The cycle head panics while iterating

Mainly to show that we don't dead-lock the other thread.

tests/cycle_initial_call_query.rs
```diff
@@ -0,0 +1,35 @@
+//! It's possible to call a Salsa query from within a cycle initial fn.
```
MichaReiser, line 1: Another test idea: A recovery or initial function that calls into the cycle itself.

MichaReiser (approved):
This is excellent. Most of my comments are around tests and documentation. 

I noticed that there's a comment about how this all interacts with accumulators. That's why I think it would be great to add a test where a query participating in a cycle emits accumulated values. 

Regarding documentation: I'm not sure what's the best place for it (maybe `cycle.rs`?) but I'd find a high level explanation of the implementation useful (e.g. that memo validation happens lazily etc)

src/function/maybe_changed_after.rs
Veykril: Ah, Micha already proposed a wrapper type a couple reviews below https://github.com/salsa-rs/salsa/pull/603#discussion_r1946252725

tests/cycle_initial_call_query.rs
carljm: Added some tests for this. I'm not sure to what extent we want to promise any specific behavior here; in the docs I expect to just say "don't do this." But the behavior that falls out is actually somewhat reasonable, and it makes sense to have tests for that case.

src/input.rs
carljm: Yeah, good point, I should make these return `true` instead. It doesn't really matter because in practice `is_verified_final` will never be called on any of these, but since we have to put the method on the `Ingredient` trait, it should return sensible values for all ingredient types.

src/function/memo.rs
carljm: This change forced me to think about ordering. I'm pretty sure Relaxed is actually fine here, because `verified_final` for a given memo is either always true, or starts false and later becomes true. The latter case is purely an optimization; reading an out-of-date false value just means we might re-validate it unnecessarily; there are no other reads whose validity requires that we get an up-to-date value. I tried to capture this in some added comments; let me know if it's clear enough (or even more so if you think my conclusion is wrong!)

src/function/memo.rs
```diff
@@ -119,23 +135,46 @@ pub(super) struct Memo<V> {
...
+
     /// Revision information
     pub(super) revisions: QueryRevisions,
 }
 
 // Memo's are stored a lot, make sure their size is doesn't randomly increase.
 // #[cfg(test)]
 const _: [(); std::mem::size_of::<Memo<std::num::NonZeroUsize>>()] =
-    [(); std::mem::size_of::<[usize; 12]>()];
+    [(); std::mem::size_of::<[usize; 17]>()];
```
Veykril, line 148: That's quite the regression, boxing the cycle heads as said in another review should bring this down to 14 I believe at least. Then the two usizes mainly come from the two fields being added.

src/function/maybe_changed_after.rs
carljm: The surface reason is pretty simple: we are going to do a fresh traversal of this memo's dependencies, so we need to start with a clean slate. I previously created a fresh hashset each loop, and in an earlier review you wisely suggested to just reuse one hashset by clearing it.

But thanks for suggesting I document this! Because trying to write this comment led me down a rabbit hole of finding and fixing several bugs, and commenting all of this `deep_verify_memo` logic much more thoroughly.

src/function/fetch.rs
carljm: Done.

src/function/maybe_changed_after.rs
carljm: Done

src/zalsa_local.rs
carljm: Done

src/function/memo.rs
carljm: Done

carljm:
Using an option box for cycle heads doesn't seem to have improved the overall picture in codspeed; slightly worse if anything. It did of course reduce memo size. Will look into perf more tomorrow.

MichaReiser:
> Using an option box for cycle heads doesn't seem to have improved the overall picture in codspeed; slightly worse if anything. It did of course reduce memo size. Will look into perf more tomorrow. 

That's surprising, considering that no Benchmark uses cycle handling

carljm:
I did add one new benchmark that uses cycle handling. But that of course isn't showing any increase or decrease since it's new in this PR.

carljm:
I realized one mistake I made that could be increasing cost; my `extend` implementation always allocates a hashset even if we are extending with nothing. Will fix this tomorrow and see how much it helps.

MichaReiser:
Yes. We could also use `None` to signify "no cycle heads" vs dereferencing the Box to see if it is empty. 

I also noticed that we now have some `Option<CycleHeads>` usages which boils down to a `Option<Option<Box<FxHashSet>>>`: Do we need the double-option or would one level of `Option` (just knowing whether it is empty or not) be sufficient?

carljm:
> We could also use `None` to signify "no cycle heads" vs dereferencing the Box to see if it is empty.

I think that's what I'm already doing? Where do you see an unnecessary dereference of the Box?

> I also noticed that we now have some `Option<CycleHeads>` usages which boils down to a `Option<Option<Box<FxHashSet>>>`: Do we need the double-option

Yeah I was going to ask about this / play with it more. This is in a case where I want to use a reference to a memo's `CycleHeads` so I don't have to clone it. But if the memo is `verified_final` I need to pass something to signify "no cycle heads", and I can't pass a reference to the memo's actual cycle heads since it will not be empty (we could maybe empty it when we set `verified_final` but that would require putting it behind a lock.) So I couldn't immediately find a better way than `Option<&CycleHeads>`, but open to suggestions.

src/cycle.rs
```diff
@@ -1,109 +1,95 @@
...
+#[derive(Clone, Debug, Default)]
+pub(crate) struct CycleHeads(Option<Box<FxHashSet<DatabaseKeyIndex>>>);
 
-    pub(crate) fn throw(self) -> ! {
-        tracing::debug!("throwing cycle {:?}", self);
-        std::panic::resume_unwind(Box::new(self))
+impl CycleHeads {
+    pub(crate) fn is_empty(&self) -> bool {
+        !self.0.as_ref().is_some_and(|heads| !heads.is_empty())
     }
```
MichaReiser, line 48: I think this is a case where we could only test if `self.0` is `Some`. It does require a slightly more complicated `remove` implementation.

carljm:
It looks like fixing the `extend` issue brought the benchmarks roughly back in line with where they were before, ranging from 0 (or slight positive on one benchmark) to -6%, with most benchmarks in the -3% to -6% range.

src/cycle.rs
carljm: I can do that but I wouldn't expect it to impact non-cycle-using benchmarks, those should always have `None` already.

src/function.rs
```diff
@@ -196,11 +201,32 @@ where
...
         let db = db.as_view::<C::DbView>();
         self.maybe_changed_after(db, input, revision)
     }
 
+    fn is_provisional_cycle_head<'db>(&'db self, db: &'db dyn Database, input: Id) -> bool {
+        self.get_memo_from_table_for(db.zalsa(), input)
+            .is_some_and(|memo| {
+                memo.cycle_heads()
+                    .is_some_and(|heads| heads.contains(&self.database_key_index(input)))
+            })
```
MichaReiser, line 214: Does `memo.cycle_heads` return an `Option`? Could it return a `CycleHeads` struct directly to simplify this call site?

carljm: It returns `Option<&CycleHeads>` for the reason described above. I don't want to clone the `CycleHeads`, and in the `verified_final` case I need to return something signifying "no cycle heads" but I don't have an empty `CycleHeads` available to return a reference to.

Maybe I could have a single static const empty `CycleHeads` I return a reference to here?

src/cycle.rs
carljm: Done

src/function.rs
carljm: Yep, the static empty `CycleHeads` works, pushed that

carljm:
Ok benchmarks are now slightly improved (-2% to -5% instead of -3% to -6%). Would still like to improve more than that if we can.

nikomatsakis (approved):
This looks pretty good. I think we gotta get it landed and start testing it.

carljm:
Oops, looks like we're hitting some GitHub CI job limits.

src/function/maybe_changed_after.rs
carljm: With the option boxing we've now significantly reduced the size of `VerifyResult`, at least.

davidbarsky:
No objections to the design or approach of this, but I was wondering if we could get https://github.com/rust-lang/rust-analyzer/pull/18964 landed first. For the corresponding PR to land in rust-analyzer, we'll need an `Salsa 1.0 alpha-1` release out with—hopefully—the ability dump the contents of tracked functions (as described in [here](https://hackmd.io/nNmrVvHfRYO_EpwmsQyYTQ#Dumping-fn-contents)). Those two together should should take roughly a a week or so of work, at which point `salsa 1.0 alpha-2` could contain _this_ change?

carljm:
I'm not in a big rush to merge this on my side, though the merge conflicts do get exhausting to deal with. I'll add some more tests and some documentation, and work on using it in red-knot in the mean time. We may learn some new things from that.

davidbarsky:
> I'm not in a big rush to merge this on my side, though the merge conflicts do get exhausting to deal with. I'll add some more tests and some documentation, and work on using it in red-knot in the mean time. We may learn some new things from that.

Yeah, I'm familiar with the rebase pains. _Hopefully_ you won't have too many conflicts with the remaining, in-flight changes, but in the event you do, I want to set  an upper bound for how long this situation exists. Hopefully it's tolerably short.

carljm:
The current code in this branch requires Rust 1.83 (for `Default` implementations of hashset iterators). Are we OK with this, or do we need to stay compatible with older Rust versions? If so, how old?

`Cargo.toml` still says `rust-version = 1.76`, but the tests only pass with 1.84, and main branch doesn't even compile with 1.76., so we clearly don't have any CI set up to enforce our declared MSRV.

carljm:
Ok, I added the necessary boilerplate here to maintain compatibility with Rust 1.80 for now.

Veykril:
Stumbled upon this issue https://github.com/salsa-rs/salsa/issues/193, might be good to check whether that is still relevant after this PR rewriting most of the cycle handling

Veykril:
Another related issue https://github.com/salsa-rs/salsa/issues/9

carljm:
I added some more tests (for tracked-struct and accumulator output from queries in cycles, panics in cycle handling in a cross-thread cycle), and fixed a bug in `diff_outputs` for outputs in cycles.

I also updated the book sections on cycle handling. I'm afraid I didn't reproduce the detail of the previous "plumbing" section on the old cycle handling, but I did add a doc comment about the new cycle handling in `cycle.rs`.

I now consider this branch "done" and ready for merge, pending any further review comments (cc @nikomatsakis, @MichaReiser, @Veykril). I'll wait to merge it until @davidbarsky and @Veykril give the green light from the rust-analyzer side.

davidbarsky:
Feel free to land this! We—rust-analyzer—got the changes we needed in https://github.com/salsa-rs/salsa/releases/tag/salsa-v0.19.0 and will update at our leisure to this when we can.

carljm:
Squashed to a single commit in preparation for landing; the commit history here has a lot of junk and false starts, as well as many merges from main branch.

Issue #586, opened 2024-09-30
Assorted macro/rust-analyzer woes


davidbarsky:
I noticed a few issues with Salsa's macro that impact the accuracy of rust-analyzer's semantic highlighting. While the [syntax theme](https://github.com/matklad/pale-fire) that I'm using has support for semantic highlighting (so the syntax highlighting looks slightly off), I confirmed the weirdness using a VS Code command called "Developer: Inspect Editor Tokens and Scopes" to determine just _exactly_ how slightly off the different tokens are. Using [`tests/is_send_sync.rs`](https://github.com/salsa-rs/salsa/blob/a20b894cc2318c5a6077659e2d8aef7c784c38e7/tests/is_send_sync.rs) as an example:

1. The `struct` keyword in `struct MyInput`, `struct MyTracked<'db>`, and `struct MyInterned<'db>` is reported as function, not a keyword.
    a. The lifetimes in the tracked and interned structs are reported as functions, not lifetimes.
    b. the field named `field` on `MyInput` is reported to be variable, not a property.
    c. The fields on `MyTracked` and `MyInterned` are reported to be methods, not fields. Arguably, this is the correct decision!
2. the tracked function `fn test(db: &dyn Database, input: MyInput)`'s name ("test") is reported as a struct, not a function.

In [`tests/input_field_durability.rs`](https://github.com/salsa-rs/salsa/blob/a20b894cc2318c5a6077659e2d8aef7c784c38e7/tests/input_field_durability.rs) on `struct MyInput`, the field types (`bool` and `usize`) are reported as unresolved references, not primitives.

I think it'd be good to resolve these either in Salsa or rust-analyzer since it's part of Salsa's fit-and-finish and these token mismatches might larger, more subtle issues in how rust-analyzer handles Salsa.

On the plus side, go-to-def/find-references on Salsa-annotated functions and structs work!

nikomatsakis:
Yes, I've noticed this. Presumably we need to adjust our spans somewhat.

Veykril:
Fixing 1 is easy -> https://github.com/salsa-rs/salsa/pull/606#issuecomment-2437493608
2 is more tricky, r-a currently ranks tokens by similarity and (I believe) in case of equal ranking it picks the last usage of it (which in this case is from the impl within the function). So we'd need the function definitions name to be moved to the end which isn't really doable with the current scoping tricks used. (arguably r-a should pick the first in this case I think, so I'll look into that)

PR #606, opened 2024-10-25
Improve the span used for salsa struct constructors

Veykril:
Fixes the first point of https://github.com/salsa-rs/salsa/issues/586


components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -204,7 +204,7 @@ macro_rules! setup_tracked_fn {
                     aux: &dyn $zalsa::JarAux,
                     first_index: $zalsa::IngredientIndex,
                 ) -> Vec<Box<dyn $zalsa::Ingredient>> {
-                    let mut fn_ingredient = <$zalsa::function::IngredientImpl<$Configuration>>::new(
```
Veykril, line 207: This is unused (r-a complains about it, whereas rustc does not, probably because r-a does not take into account cross crate diagnostic-ity)

nikomatsakis (approved):

PR #612, opened 2024-11-02
Emit struct in `#[tracked]` fn last for better IDE support

Veykril:
This + a change in rust-analyzer resolves the last thing in https://github.com/salsa-rs/salsa/issues/586

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -283,7 +283,9 @@ macro_rules! setup_tracked_fn {
                 }
             })
         }
-
+        // The struct needs be last in the macro expansion in order to make the tracked
+        // function's ident be identified as a function, not a struct, during semantic highlighting.
+        // for more details, see https://github.com/salsa-rs/salsa/pull/612.
         #[allow(non_camel_case_types)]
         $vis struct $fn_name {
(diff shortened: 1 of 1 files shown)


components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -288,5 +283,10 @@ macro_rules! setup_tracked_fn {
                 }
             })
         }
+
+        #[allow(non_camel_case_types)]
```
MichaReiser, line 287: I would consider adding a comment explaining the importance of why this has to come last to avoid accidentally breaking this in the future.

MichaReiser (approved):

components/salsa-macro-rules/src/setup_tracked_fn.rs
davidbarsky, line 287:
```suggestion
        // The struct needs be last in the macro expansion in order to make the tracked
        // function's ident be identified as a function, not a struct, during semantic highlighting.
        // for more details, see https://github.com/salsa-rs/salsa/pull/612.
        #[allow(non_camel_case_types)]
```

davidbarsky: Added a description; I'll land this change as Lukas is on vacation for the next 3 weeks.

Issue #599, opened 2024-10-16
Remove the global hashtable for tracked fns with >1 argument


nikomatsakis:
As described in [this hackmd](https://hackmd.io/@salsa/rkC-2qWCR), we currently create an interned ingredient for every tracked function unless it has exactly 1 argument. This is wasteful. The reason we do this is because the memos for tracked functions are stored on salsa structs[^ss] and so we need to have a 1:1 relationship between a (salsa struct, tracked fn) pair and a memo. 

[^ss]: (Salsa struct = input | tracked | interned)

We should refactor things so that:

* Every tracked function has as its first argument a salsa struct. If there are *no* arguments, we can add some kind of dummy interned struct in `salsa` like `#[salsa::intern] struct Placeholder<'db>{ data: () }` and redefine the argument as being called on a `Placeholder<'db>`, which we can trivially create (it'd also be nice to have a more optimized "global struct", maybe that's worth thinking about; even interning unit is more work than it really ought to be for this case). 
* For tracked functions with >1 argument, when you index into the memo-table for a tracked struct, you would get back a `DashMap` instead of a single memo; the map would be keyed by the additional arguments.

nikomatsakis:
# Mentoring instructions

To start, https://github.com/salsa-rs/salsa/issues/600 might be a good first step.

Beyond that, but I can get you started with some pointers. Memos are stored in a`MemoTable`, defined here:

https://github.com/salsa-rs/salsa/blob/710691d6070cf7cf3638f50ac74f0ac8bce9c48c/src/table/memo.rs#L16-L18

Right now, each tracked function gets a `MemoEntry`:

https://github.com/salsa-rs/salsa/blob/710691d6070cf7cf3638f50ac74f0ac8bce9c48c/src/table/memo.rs#L30

...which has an option, and thus can store 0 or 1 memos. The actual data is a `MemoEntryData` which uses some type-id tricks. The idea would be to change this *option* to something like a map.

The tricky part is that the keys of the map are the additional arguments and the type of memos is independent of those details. So you could package them up as an `Arc<dyn Hash+Eq>` or something like that (you'd have to define a trait like `dyn MemoKey: Hash + Eq`). The other part is that many tracked fns don't *really* want a map, because they only have exactly 1 argument.

[...]

PR #614, opened 2024-11-13
Assign memo ingredients per salsa-struct-ingredient

### LINKED ISSUE #600: Assign memo ingredients per salsa-struct-ingredient, not globally
Today every tracked function gets assigned a distinct [`MemoIngredientIndex`](https://github.com/salsa-rs/salsa/blob/c6c51a0ea0bb11ce5a449c4b9256c66c5e484fee/src/zalsa.rs#L99-L102). These are used to index into the memo table attached to a salsa struct. But every function can be attached to salsa structs of exactly one kind, so indices only have to be distinct per salsa struct, not globally. This would save a lot of space in larger projects.

ShoyuVanilla:
Closes #600

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -199,7 +199,22 @@ macro_rules! setup_tracked_fn {
                     aux: &dyn $zalsa::JarAux,
                     first_index: $zalsa::IngredientIndex,
                 ) -> Vec<Box<dyn $zalsa::Ingredient>> {
+                    let struct_index = $zalsa::macro_if! {
+                        if $needs_interner {
+                            first_index.successor(0)
+                        } else {
+                            aux
+                                .lookup_struct_ingredient_index(
+                                    core::any::TypeId::of::<$InternedData<'static>>()
+                                )
+                                .expect(
+                                    "Salsa struct is passed as an argument of a tracked function, but its ingredient hasn't been added!"
+                                )
+                        }
+                    };
+
                     let fn_ingredient = <$zalsa::function::IngredientImpl<$Configuration>>::new(
+                        struct_index,
                         first_index,
                         aux,
                     );
@@ -218,6 +233,10 @@ macro_rules! setup_tracked_fn {
                             ]
                         }
                     }
+                }
+
+                fn salsa_struct_type_id(&self) -> Option<core::any::TypeId> {
+                    None
                 }
             }
 
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -99,6 +99,9 @@ macro_rules! setup_tracked_fn {
                         $zalsa::IngredientCache::new();
 
                     impl $zalsa::SalsaStructInDb for $InternedData<'_> {
+                        fn lookup_ingredient_index(_aux: &dyn $zalsa::JarAux) -> core::option::Option<$zalsa::IngredientIndex> {
+                            None
+                        }
                     }
 
                     impl $zalsa::interned::Configuration for $Configuration {
@@ -203,10 +206,7 @@ macro_rules! setup_tracked_fn {
                         if $needs_interner {
                             first_index.successor(0)
                         } else {
-                            aux
-                                .lookup_struct_ingredient_index(
-                                    core::any::TypeId::of::<$InternedData<'static>>()
-                                )
+                            <$InternedData as $zalsa::SalsaStructInDb>::lookup_ingredient_index(aux)
                                 .expect(
                                     "Salsa struct is passed as an argument of a tracked function, but its ingredient hasn't been added!"
                                 )
--- a/src/ingredient.rs
+++ b/src/ingredient.rs
@@ -28,9 +28,23 @@ pub trait Jar: Any {
     fn salsa_struct_type_id(&self) -> Option<TypeId>;
 }
 
+/// Methods on the Salsa database available to jars while they are creating their ingredients.
 pub trait JarAux {
+    /// Return index of first ingredient from `jar` (based on the dynamic type of `jar`).
+    /// Returns `None` if the jar has not yet been added. 
+    /// Used by tracked functions to lookup the ingredient index for the salsa struct they take as argument.
     fn lookup_jar_by_type(&self, jar: &dyn Jar) -> Option<IngredientIndex>;
 
+    /// Returns the memo ingredient index that should be used to attach data from the given tracked function
+    /// to the given salsa struct (which the fn accepts as argument).
+    ///
+    /// The memo ingredient indices for a given function must be distinct from the memo indices
+    /// of all other functions that take the same salsa struct.
+    ///
+    /// # Parameters
+    ///
+    /// * `struct_ingredient_index`, the index of the salsa struct the memo will be attached to
+    /// * `ingredient_index`, the index of the tracked function whose data is stored in the memo
     fn next_memo_ingredient_index(
         &self,
         struct_ingredient_index: IngredientIndex,
--- a/src/ingredient.rs
+++ b/src/ingredient.rs
@@ -29,7 +29,7 @@ pub trait Jar: Any {
 }
 
 pub trait JarAux {
-    fn lookup_struct_ingredient_index(&self, type_id: TypeId) -> Option<IngredientIndex>;
+    fn lookup_jar_by_type(&self, jar: &dyn Jar) -> Option<IngredientIndex>;
 
     fn next_memo_ingredient_index(
         &self,
--- a/src/ingredient.rs
+++ b/src/ingredient.rs
@@ -31,7 +31,7 @@ pub trait Jar: Any {
 /// Methods on the Salsa database available to jars while they are creating their ingredients.
 pub trait JarAux {
     /// Return index of first ingredient from `jar` (based on the dynamic type of `jar`).
-    /// Returns `None` if the jar has not yet been added. 
+    /// Returns `None` if the jar has not yet been added.
     /// Used by tracked functions to lookup the ingredient index for the salsa struct they take as argument.
     fn lookup_jar_by_type(&self, jar: &dyn Jar) -> Option<IngredientIndex>;
 
--- a/src/ingredient.rs
+++ b/src/ingredient.rs
@@ -23,10 +23,19 @@ pub trait Jar: Any {
         aux: &dyn JarAux,
         first_index: IngredientIndex,
     ) -> Vec<Box<dyn Ingredient>>;
+
+    /// If this jar's first ingredient is a salsa struct, return its `TypeId`
+    fn salsa_struct_type_id(&self) -> Option<TypeId>;
 }
 
 pub trait JarAux {
-    fn next_memo_ingredient_index(&self, ingredient_index: IngredientIndex) -> MemoIngredientIndex;
+    fn lookup_struct_ingredient_index(&self, type_id: TypeId) -> Option<IngredientIndex>;
+
+    fn next_memo_ingredient_index(
+        &self,
+        struct_ingredient_index: IngredientIndex,
+        ingredient_index: IngredientIndex,
+    ) -> MemoIngredientIndex;
 }
 
 pub trait Ingredient: Any + std::fmt::Debug + Send + Sync {
--- a/src/zalsa.rs
+++ b/src/zalsa.rs
@@ -189,6 +189,12 @@ impl Zalsa {
             let jar_type_id = jar.type_id();
             let mut jar_map = self.jar_map.lock();
             let mut should_create = false;
+            // First record the index we will use into the map and then go and create the ingredients.
+            // Those ingredients may invoke methods on the `JarAux` trait that read from this map
+            // to lookup ingredient indices for already created jars.
+            //
+            // Note that we still hold the lock above so only one jar is being created at a time and hence
+            // ingredient indices cannot overlap.
             let index = *jar_map.entry(jar_type_id).or_insert_with(|| {
                 should_create = true;
                 IngredientIndex::from(self.ingredients_vec.len())
--- a/src/zalsa.rs
+++ b/src/zalsa.rs
@@ -119,10 +119,10 @@ pub struct Zalsa {
 
     nonce: Nonce<StorageNonce>,
 
-    /// Map from the [`IngredientIndex`][] of a salsa struct to a list of
-    /// [ingredient-indices](`IngredientIndex`)for tracked functions that have this salsa struct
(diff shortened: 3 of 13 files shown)


src/zalsa.rs
```diff
@@ -130,6 +132,9 @@ pub struct Zalsa {
     /// adding new kinds of ingredients.
     jar_map: Mutex<FxHashMap<TypeId, IngredientIndex>>,
 
+    /// Map from the type-id of a salsa struct to the index of its first ingredient.
+    salsa_struct_map: Mutex<FxHashMap<TypeId, IngredientIndex>>,
+
```
ShoyuVanilla, line 137: I don't like this new table for `SalsaStruct` -> `IngredientIndex` but I couldn't find any other way to pass the salsa-struct-ingredient to its tracked function when creating the function's ingredient, as salsa-struct-ingredient is accessible only with its `Configuration`, that is private inside the macro-expanded code of that salsa struct.
Would there be some good alternatives?

MichaReiser (commented):
I don't think we currently have a test for a tracked fn that has more than one argument (other than db) and thus uses the internet. Could you double check if that's indeed the case and, if so, add a test for it?

src/zalsa.rs
MichaReiser: I would be inclined to add a method `ingredient_index` to `SalsaStructInDb` but best to hear what @nikomatsakis thinks

ShoyuVanilla: We need to get the `IngredientIndex` for salsa struct here;

https://github.com/salsa-rs/salsa/blob/e4d36daf2dc4a096009753b6e326cf44e6c6e18a/components/salsa-macro-rules/src/setup_tracked_fn.rs#L196-L205

and we can't get the actual instance of salsa struct here. All we have is `JarAux` - actually, this is `Zalsa` -, so that function should be a static method, I guess. (I'm afraid that this makes `SalsaStructInDb` dyn-incompatible but we are not using it as a trait object anywhere in salsa)

Thus, the function would be implemented inside our `setup_*` macros like;

```rust
fn ingredient_index(aux: &dyn JarAux) -> IngredientIndex {
    aux.lookup_jar_by_type(JarImpl::<$Configuration>::default())
}
```

and we need to expose some method like `lookup_jar_by_type` in `JarAux`, I think.

ShoyuVanilla:
> I don't think we currently have a test for a tracked fn that has more than one argument (other than db) and thus uses the internet. Could you double check if that's indeed the case and, if so, add a test for it?

Sure! I'll add tests and edit things you commented when I get home. Thanks for the review

src/zalsa.rs
```diff
@@ -290,15 +293,39 @@ impl Zalsa {
...
-        self.memo_ingredients.lock()[memo_ingredient_index.as_usize()]
+        self.memo_ingredient_indices.read()[struct_ingredient_index.as_usize()]
+            [memo_ingredient_index.as_usize()]
     }
 }
 
-impl JarAux for Zalsa {
-    fn next_memo_ingredient_index(&self, ingredient_index: IngredientIndex) -> MemoIngredientIndex {
-        let mut memo_ingredients = self.memo_ingredients.lock();
+struct JarAuxImpl<'a>(&'a Zalsa, &'a FxHashMap<TypeId, IngredientIndex>);
```
ShoyuVanilla, line 304: I changed salsa struct ingredient lookup as I said [here](https://github.com/salsa-rs/salsa/pull/614#discussion_r1841483937) but I had to change this to avoid deadlocks 😢

src/zalsa.rs
```diff
@@ -186,21 +188,22 @@ impl Zalsa {
...
-                let index = IngredientIndex::from(self.ingredients_vec.len());
-                let ingredients = jar.create_ingredients(self, index);
+            let mut should_create = false;
+            let index = *jar_map.entry(jar_type_id).or_insert_with(|| {
+                should_create = true;
+                IngredientIndex::from(self.ingredients_vec.len())
+            });
+            if should_create {
+                let aux = JarAuxImpl(self, &jar_map);
+                let ingredients = jar.create_ingredients(&aux, index);
```
MichaReiser, line 198: There's a comment on `jar_map` that suggests that it's important to only access `ingredients_vec` while the lock on `jar_map` is held

https://github.com/MichaReiser/salsa/blob/38ad6555ebe4374bbc69022e005f4136a1839ad9/src/zalsa.rs#L127-L133

So moving it out to avoid the deadlock might not be entirely safe

MichaReiser: For example. What could happen now is that two calls end up with the same index because the new ingredient isn't pushed on the `ingredients_vec` until line 206, so that `self.ingredients_vec.len()` returns the same length.

ShoyuVanilla: Okay, I'll fix this

nikomatsakis:
@ShoyuVanilla hey, cool! My apologies for radio silence. I've been overwhelmed. I'm going to carve out more time for salsa though so let me take a look...

src/zalsa.rs
```diff
@@ -186,21 +188,22 @@ impl Zalsa {
         {
             let jar_type_id = jar.type_id();
             let mut jar_map = self.jar_map.lock();
-            *jar_map
-            .entry(jar_type_id)
-            .or_insert_with(|| {
-                let index = IngredientIndex::from(self.ingredients_vec.len());
-                let ingredients = jar.create_ingredients(self, index);
+            let mut should_create = false;
```
nikomatsakis, line 191: I don't quite follow this -- why move this logic out from the `or_insert_with` call?

src/ingredient.rs
```diff
@@ -23,10 +23,19 @@ pub trait Jar: Any {
...
         first_index: IngredientIndex,
     ) -> Vec<Box<dyn Ingredient>>;
+
+    /// If this jar's first ingredient is a salsa struct, return its `TypeId`
+    fn salsa_struct_type_id(&self) -> Option<TypeId>;
 }
 
 pub trait JarAux {
-    fn next_memo_ingredient_index(&self, ingredient_index: IngredientIndex) -> MemoIngredientIndex;
+    fn lookup_jar_by_type(&self, jar: &dyn Jar) -> Option<IngredientIndex>;
```
nikomatsakis, line 32: Pre-existing, but we should add comments to these methods. What is the reason people would invoke `lookup_jar_by_type` and `next_memo_ingredient_index`?

tests/tracked_fn_multiple_args.rs
```diff
@@ -0,0 +1,25 @@
...
+    field: u32,
+}
+
+#[salsa::interned]
+struct MyInterned<'db> {
+    field: u32,
+}
+
+#[salsa::tracked]
+fn tracked_fn<'db>(db: &'db dyn salsa::Database, input: MyInput, interned: MyInterned<'db>) -> u32 {
```
nikomatsakis, line 15: Why this particular test? Tracked functions with multiple arguments today compile to a tracked function on a hidden interned value -- is that what you were intending to test?

src/zalsa.rs
nikomatsakis: I guess it is so that ingredient construction can call `lookup_jar_by_type`?

nikomatsakis: It'd be nice to have a comment here explaining that interaction

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -199,7 +202,19 @@ macro_rules! setup_tracked_fn {
                     aux: &dyn $zalsa::JarAux,
                     first_index: $zalsa::IngredientIndex,
                 ) -> Vec<Box<dyn $zalsa::Ingredient>> {
+                    let struct_index = $zalsa::macro_if! {
+                        if $needs_interner {
+                            first_index.successor(0)
```
nikomatsakis, line 207: I...guess this is fine. It's a bit surprising to use the index of something that hasn't been made yet. An alternative would be to flip the order so that we create the interned struct *first* and then pass its index to the function. There'd probably be some minor adjustments made elsewhere in the file as a result but shouldn't be too bad.

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -199,7 +202,19 @@ macro_rules! setup_tracked_fn {
...
+                        } else {
+                            <$InternedData as $zalsa::SalsaStructInDb>::lookup_ingredient_index(aux)
```
nikomatsakis, line 209: OK, I see now the role of `lookup_ingredient_index`.

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -199,7 +202,19 @@ macro_rules! setup_tracked_fn {
...
+                                .expect(
```
nikomatsakis, line 210: I'm not 1000% percent sure that the struct ingredient will always have been created first. If this code doesn't run until the tracked functon is first called, I suppose that's true, but if it could run at some other time maybe not? Still, this seems like a good assertion for now to simplify our lives. If it's not true we can probably add some kind of dependency mechanism to force it to be true.

src/ingredient.rs
```diff
@@ -23,10 +23,19 @@ pub trait Jar: Any {
         aux: &dyn JarAux,
         first_index: IngredientIndex,
     ) -> Vec<Box<dyn Ingredient>>;
+
+    /// If this jar's first ingredient is a salsa struct, return its `TypeId`
+    fn salsa_struct_type_id(&self) -> Option<TypeId>;
 }
 
 pub trait JarAux {
```
nikomatsakis, line 31:
```suggestion
/// Methods on the Salsa database available to jars while they are creating their ingredients.
pub trait JarAux {
```

src/ingredient.rs
nikomatsakis:
```suggestion
    /// Return index of first ingredient from `jar` (based on the dynamic type of `jar`).
    /// Returns `None` if the jar has not yet been added. 
    /// Used by tracked functions to lookup the ingredient index for the salsa struct they take as argument.
    fn lookup_jar_by_type(&self, jar: &dyn Jar) -> Option<IngredientIndex>;
```

src/ingredient.rs
```diff
@@ -23,10 +23,19 @@ pub trait Jar: Any {
...
-    fn next_memo_ingredient_index(&self, ingredient_index: IngredientIndex) -> MemoIngredientIndex;
+    fn lookup_jar_by_type(&self, jar: &dyn Jar) -> Option<IngredientIndex>;
+
+    fn next_memo_ingredient_index(
```
nikomatsakis, line 34:
```suggestion
    /// Returns the memo ingredient index that should be used to attach data from the given tracked function
    /// to the given salsa struct (which the fn accepts as argument).
    ///
    /// The memo ingredient indices for a given function must be distinct from the memo indices
    /// of all other functions that take the same salsa struct.
    ///
    /// # Parameters
    ///
    /// * `struct_ingredient_index`, the index of the salsa struct the memo will be attached to
    /// * `ingredient_index`, the index of the tracked function whose data is stored in the memo
    fn next_memo_ingredient_index(
```

src/zalsa.rs
```diff
@@ -186,21 +188,22 @@ impl Zalsa {
...
+            let index = *jar_map.entry(jar_type_id).or_insert_with(|| {
```
nikomatsakis, line 192:
```suggestion
            // First record the index we will use into the map and then go and create the ingredients.
            // Those ingredients may invoke methods on the `JarAux` trait that read from this map
            // to lookup ingredient indices for already created jars.
            //
            // Note that we still hold the lock above so only one jar is being created at a time and hence
            // ingredient indices cannot overlap.
            let index = *jar_map.entry(jar_type_id).or_insert_with(|| {
```

src/zalsa.rs
```diff
@@ -186,21 +188,22 @@ impl Zalsa {
         {
             let jar_type_id = jar.type_id();
             let mut jar_map = self.jar_map.lock();
```
nikomatsakis, line 190:
```suggestion
            // Important: we hold the lock on `jar_map` during ingredient construction to
            // ensure all ingredients are created atomically.
            let mut jar_map = self.jar_map.lock();
```

nikomatsakis (changes requested):
This looks good! Took me a bit to get my head in the right space. I left comments, can you check that they are accurate and merge them in? After that, I'd say we can land this.

nikomatsakis (approved):

ShoyuVanilla:
Well, I was going to continue working on this but it seems that I'm too late, sorry. I had family affairs last week 😢 
I'll issue a new PR with fixes for quirks on this PR

Veykril:
No need to apologize you have no obligations so do not feel pressured to work on things asap

nikomatsakis:
@ShoyuVanilla yeah, no worries, I'm grateful you did the original PR! Very big impact. 

The main thing this PR needs is `cargo fmt`

tests/tracked_fn_multiple_args.rs
ShoyuVanilla: Yes. Though this PR is not directly related to that, but as it changes some codes in such hidden interned value, I was intending to add a regression test for it. Other case are pre-existing

src/zalsa.rs
ShoyuVanilla: I don't remember for sure - as it was weeks before 😅 -, but I think that it was because it had some issues due to creation order like using ingredients index before they are created.
I'll look into such order related things along with the one you commented [here](https://github.com/salsa-rs/salsa/pull/614/files/38ad6555ebe4374bbc69022e005f4136a1839ad9#r1879820347)

components/salsa-macro-rules/src/setup_tracked_fn.rs
ShoyuVanilla: I thought - and experimented with some cases - that the structs are created first to be passed to a tracked function but my understanding on salsa is very limited and I might be wrong. I'll investigate this more

davidbarsky:
@ShoyuVanilla if it's okay, i'll merge this PR: it'll make managing the stack of changes that Lukas and I have for https://github.com/salsa-rs/salsa/compare/master...davidbarsky:salsa:david-and-lukas/required-salsa-changes-for-ra a lot easier. Can you do the followups on a separate PR?

ShoyuVanilla:
> @ShoyuVanilla if it's okay, i'll merge this PR: it'll make managing the stack of changes that Lukas and I have for [master...davidbarsky:salsa:david-and-lukas/required-salsa-changes-for-ra](https://github.com/salsa-rs/salsa/compare/master...davidbarsky:salsa:david-and-lukas/required-salsa-changes-for-ra) a lot easier. Can you do the followups on a separate PR?

Yes, I'll issue a new follow-up PR (along with #599, maybe?)

davidbarsky:
> Yes, I'll issue a new follow-up PR (along with https://github.com/salsa-rs/salsa/issues/599, maybe?)

Up to you!

PR #615, opened 2024-11-17
Faster accumulators

MichaReiser:
This PR implements faster accumulators. 

Accumulators are relatively costly today because collecting the accumulated values requires iterating over the queries' entire dependency tree. 
This cost is negligible with a cold DB but adds up for very deep trees. 

For example, adding accumulators to Ruff resulted in a 30% perf regression https://github.com/astral-sh/ruff/pull/14116

This PR adds a new field to `AccumulatedMap` that tracks whether any dependency of the current query has any accumulated values. It checks that field when collecting the accumulated values, allowing it to skip entire subtrees. 

This approach should make accumulated values nearly "free" in the common cases where there are no or only very few diagnostics. 

## Test Plan

I added a benchmark and I verified that the perf regression in the linked Ruff PR is mostly gone after using this salsa version.

Running the benchmark locally gives me

```
     Running benches\accumulator.rs (target\release\deps\accumulator-51d27e7934278de6.exe)
accumulator             time:   [666.27 µs 676.29 µs 688.11 µs]
                        change: [-32.104% -30.905% -29.682%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  1 (1.00%) high mild
  11 (11.00%) high severe
```

where the baseline is main

### DIFF
--- a/src/accumulator/accumulated_map.rs
+++ b/src/accumulator/accumulated_map.rs
@@ -7,6 +7,10 @@ use super::{accumulated::Accumulated, Accumulator, AnyAccumulated};
 #[derive(Default, Debug)]
 pub struct AccumulatedMap {
     map: FxHashMap<IngredientIndex, Box<dyn AnyAccumulated>>,
+
+    /// [`InputAccumulatedValues::Empty`] if any input read during the query's execution
+    /// has any direct or indirect accumulated values.
+    inputs: InputAccumulatedValues,
 }
 
 impl AccumulatedMap {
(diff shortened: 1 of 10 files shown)


src/accumulator/accumulated_map.rs
```diff
@@ -17,6 +22,16 @@ impl AccumulatedMap {
             .accumulate(value);
     }
 
+    /// Marks the accumulator result that some nested query has some accumulated values.
+    pub fn add_dependency_with_accumulated_values(&mut self) {
```
MichaReiser, line 26: I don't like the naming at all but struggled to come up with something better.

src/accumulator/accumulated_map.rs
```diff
@@ -7,6 +7,11 @@ use super::{accumulated::Accumulated, Accumulator, AnyAccumulated};
 #[derive(Default, Debug)]
 pub struct AccumulatedMap {
     map: FxHashMap<IngredientIndex, Box<dyn AnyAccumulated>>,
+    /// `true` if any query called by this query has any accumulated values.
+    ///
+    /// Tracking whether any dependency has any accumulated values helps
+    /// skipping over entire subtrees when collecting accumulated values.
+    dependency_with_accumulated_values: bool,
```
davidbarsky, line 14: Small nit: consider making this an enum instead? It'd make the callsites a bit more legible.

davidbarsky (approved):
approving modulo one nit. I'll let you decide whether to merge or not.

src/accumulator/accumulated_map.rs
MichaReiser: This is a great suggestion. Thanks

PR #617, opened 2024-11-20
internal: pass `db_lt` to the `inner_fn`

davidbarsky:
This is necessary to support passing interned structs to queries in the query group imitation macro: https://github.com/davidbarsky/db-ext-macro/.

(I'll upstream that macro to Salsa, promise! The code quality is bad, however...)


MichaReiser (commented):
Could you add demonstrating its use case?

davidbarsky:
Pushed a fix for the unused lifetime lint.

> Could you add demonstrating its use case?

@MichaReiser Uh, probably. Without porting my macro over, it might be a little challenging, but I think I can add it as a desugared version.

davidbarsky:
Closing this, as this isn't needed anymore.

PR #618, opened 2024-11-25
feature: support interned structs without lifetimes

davidbarsky:
For background context, [see Zulip](https://salsa.zulipchat.com/#narrow/channel/333573-Contributing-to-Salsa/topic/Interning.20in.20New.20Salsa.2C.20Og-Salsa.20Style/near/484002306), but the long-and-short of it is that the `'db` lifetime on interned values made porting rust-analyzer to the new Salsa nearly impossible. However, I realized that I could simply get rid of the `'db` lifetime on interned structs, hence this PR.

I also ported over the `interned-struct` tests.

### DIFF
--- a/components/salsa-macro-rules/src/lib.rs
+++ b/components/salsa-macro-rules/src/lib.rs
@@ -19,6 +19,7 @@ mod maybe_default;
 mod setup_accumulator_impl;
 mod setup_input_struct;
 mod setup_interned_struct;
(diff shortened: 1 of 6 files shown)


MichaReiser:
I don't mind this but I would prefer to have all the backwards compat macros to be behind an off-by-default feature flag

davidbarsky:
Closed in favor of https://github.com/salsa-rs/salsa/pull/632.

PR #620, opened 2024-12-03
feature: allow overriding `interned_sans_lifetime`-annotated structs with custom IDs 

davidbarsky:
(This feature builds atop of https://github.com/salsa-rs/salsa/pull/618, but GitHub doesn't do `git range-diff` between PRs, so...)

This PR adds support for replacing the `salsa::Id` in `#[salsa::interned_sans_lifetime]`-annotated structs with a newtype that implements `salsa::plumbing::FromId` and `salsa::plumbing::AsId`. I expect that this would simplify migrating rust-analyzer, especially inside of the [`span` crate](https://github.com/rust-lang/rust-analyzer/tree/master/crates/span/src), as we use bit-packed `u32s`.

### DIFF
--- a/src/id.rs
+++ b/src/id.rs
@@ -29,8 +29,9 @@ impl Id {
     /// In general, you should not need to create salsa ids yourself,
     /// but it can be useful if you are using the type as a general
     /// purpose "identifier" internally.
+    #[doc(hidden)]
     #[track_caller]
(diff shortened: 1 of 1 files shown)


davidbarsky:
Closed in favor of https://github.com/salsa-rs/salsa/pull/632.

PR #624, opened 2024-12-13
Remove unnecessary `Option` from `ZalsaLocal::query_stack`

Veykril:
This removes a couple panic paths

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -1,4 +1,5 @@
 use std::{
+    mem,
     panic::panic_any,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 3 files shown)


src/runtime.rs
```diff
@@ -249,87 +250,87 @@ impl Runtime {
...
-            dg.for_each_cycle_participant(
-                from_id,
-                &mut from_stack,
-                database_key_index,
-                to_id,
-                |aqs| {
-                    aqs.iter_mut().for_each(|aq| {
-                        cycle_query.add_from(aq);
-                        v.push(aq.database_key_index);
+        let (me_recovered, others_recovered, cycle) = local_state.with_query_stack(|from_stack| {
```
Veykril, line 253: The diff here is just the closure wrapping causing an extra level of indentation

src/runtime.rs
```diff
@@ -203,18 +204,18 @@ impl Runtime {
...
-        );
-
-        local_state.restore_query_stack(stack);
+        let result = local_state.with_query_stack(|stack| {
+            let (new_stack, result) = DependencyGraph::block_on(
+                dg,
+                thread_id,
+                database_key,
+                other_id,
+                mem::take(stack),
```
Veykril, line 213: The `mem::take` is effectively what was `Option::take` before here

MichaReiser (approved):

PR #625, opened 2024-12-13
Fix CI

Veykril:


Veykril:
Ah the cause is `annotate_snippets@0.11.5`, `0.11.4` works fine

PR #626, opened 2024-12-13
Do not pass ownership of the QueryStack in `Runtime::block_on_or_unwind`

Veykril:
Builds on top of https://github.com/salsa-rs/salsa/pull/624, opened separately as this is this is separate concern that is more likely to be rejected.

This is more of a micro optimization as it reduces the size of `Edge` by a `usize` (and it removes one `mem::take` on a `Vec`)

Motivation for this change is to reduce the moving of the query stack `Vec`, that removes a bunch of potential memcopies (at the expense of some more unsafe)

### DIFF
--- a/src/runtime/dependency_graph.rs
+++ b/src/runtime/dependency_graph.rs
@@ -206,7 +206,8 @@ impl DependencyGraph {
         assert_ne!(from_id, to_id);
         debug_assert!(!self.edges.contains_key(&from_id));
         debug_assert!(!self.depends_on(to_id, from_id));
-        let (edge, guard) = edge::Edge::new(to_id, database_key, from_stack);
+        // SAFETY: The caller is responsible for ensuring that the `EdgeGuard` outlives the `Edge`.
+        let (edge, guard) = unsafe { edge::Edge::new(to_id, database_key, from_stack) };
         self.edges.insert(from_id, edge);
         self.query_dependents
             .entry(database_key)
@@ -246,7 +247,7 @@ impl DependencyGraph {
 }
 
 mod edge {
-    use std::{marker::PhantomData, ptr::NonNull, sync::Arc, thread::ThreadId};
+    use std::{marker::PhantomData, mem, sync::Arc, thread::ThreadId};
 
     use parking_lot::MutexGuard;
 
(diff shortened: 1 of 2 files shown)


MichaReiser:
@carljm is currently working on fix-point iteration. Would it make sense to wait with these changes to avoid introducing merge conflicts for him?

carljm:
I'll have to resolve some merge conflicts, but it looks fine. The cycle-handling code changed here totally disappears in the fixpoint-iteration branch, and the query-stack API changes shouldn't make any difference to that branch.

carljm:
(To be clear, I haven't carefully reviewed these changes for correctness, and won't have time to do that. But assuming they are reviewed and are good for current Salsa, I'm not worried about their impact on fixpoint iteration. And I don't think other changes should be delayed to wait for fixpoint iteration, since that branch is not ready yet. Resolving merge conflicts isn't a big deal.)

src/runtime/dependency_graph.rs
```diff
@@ -253,11 +237,100 @@ impl DependencyGraph {
...
+    use crate::{
+        runtime::{dependency_graph::DependencyGraph, ActiveQuery},
+        DatabaseKeyIndex,
+    };
+
+    #[derive(Debug)]
+    pub(super) struct Edge {
+        pub(super) blocked_on_id: ThreadId,
+        pub(super) blocked_on_key: DatabaseKeyIndex,
+        stack: SendNonNull<[ActiveQuery]>,
```
Veykril, line 262: Contemplating whether to use a raw pointer here, or just transmute the lifetimes and store a `&'static mut [ActiveQuery]` instead. The latter probably simplifies a bit

Veykril: Done that in 77c96c07f5db31b7047fcd2c6475b5697b55a4f3

Veykril:
As discussed, decision on this should either way until the fixpoint PR is finished.

Veykril:
https://github.com/salsa-rs/salsa/pull/603 renders this PR unnecessary

Veykril:
Actually not quite, but given this is included in #629 I will combine the changes there (looking at rebasing this on top of the fixpoint PR)

PR #628, opened 2024-12-13
Replace `Page` data Vec with boxed fixed array

Veykril:
No need to use a `Vec` here, we have a comptime length of the page so we unnecessarily store an extra length and capacity

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -4,10 +4,10 @@ use std::{
     mem::MaybeUninit,
     panic::RefUnwindSafe,
     ptr, slice,
+    sync::atomic::{AtomicUsize, Ordering},
 };
 
 use append_only_vec::AppendOnlyVec;
-use crossbeam::atomic::AtomicCell;
 use memo::MemoTable;
 use parking_lot::Mutex;
 use sync::SyncTable;
@@ -50,7 +50,7 @@ pub(crate) struct Page<T: Slot> {
     ingredient: IngredientIndex,
 
     /// Number of elements of `data` that are initialized.
-    allocated: AtomicCell<usize>,
+    allocated: AtomicUsize,
 
     /// The "allocation lock" is held when we allocate a new entry.
     ///
@@ -179,7 +179,7 @@ impl<T: Slot> Page<T> {
     }
 
     fn check_bounds(&self, slot: SlotIndex) {
-        let len = self.allocated.load();
+        let len = self.allocated.load(Ordering::Acquire);
         assert!(
             slot.0 < len,
             "out of bounds access `{slot:?}` (maximum slot `{len}`)"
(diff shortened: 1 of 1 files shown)


src/table.rs
```diff
@@ -169,15 +171,11 @@ impl Table {
 impl<T: Slot> Page<T> {
     #[allow(clippy::uninit_vec)]
     fn new(ingredient: IngredientIndex) -> Self {
-        let mut data = Vec::with_capacity(PAGE_LEN);
-        unsafe {
-            data.set_len(PAGE_LEN);
```
Veykril, line 174: Afaik this considered UB as we now have a Vec containing uninitialized values (`UnsafeCell` does not make this sound either, `MaybeUninit` would I believe)

src/table.rs
```diff
@@ -62,7 +64,7 @@ pub(crate) struct Page<T: Slot> {
 
     /// Vector with data. This is always created with the capacity/length of `PAGE_LEN`
     /// and uninitialized data. As we initialize new entries, we increment `allocated`.
-    data: Vec<UnsafeCell<T>>,
+    data: Box<[UnsafeCell<MaybeUninit<T>>; PAGE_LEN]>,
```
Veykril, line 67: We could drop the `Box` here entirely if we wanted as a `Page` is currently always allocated as a `Box<dyn>` anyways, but we'd need to be careful in that we never put a `Page` on the stack when we construct it then which is rather tricky in today's Rust

src/table.rs
```diff
@@ -252,14 +264,13 @@ impl<T: Slot> TablePage for Page<T> {
...
-        let mut data = std::mem::take(&mut self.data);
-        let len = self.allocated.load();
+        // Execute destructors for all initialized elements
+        let len = self.allocated.load(Ordering::Acquire);
+        // SAFETY: self.data is initialized for T's up to len
         unsafe {
-            data.set_len(len);
-            drop(std::mem::transmute::<Vec<UnsafeCell<T>>, Vec<T>>(data));
+            // FIXME: Should be ptr::from_raw_parts_mut but that is unstable
+            let to_drop = slice::from_raw_parts_mut(self.data.as_mut_ptr().cast::<T>(), len);
```
davidbarsky, line 272: what's the consequence of this being `slice` instead or `ptr`?

davidbarsky (commented):
lgtm, modulo clarification around the fixme in unsafe code

src/table.rs
Veykril: It creates an intermediate reference. It's not an issue in this case here but generally it ought to be avoided if not needed as references have more implications than raw pointers

davidbarsky (approved):

src/table.rs
davidbarsky: cool, merging!

PR #629, opened 2024-12-13
Do not pass ownership of the QueryStack in `Runtime::block_on_or_un…

Veykril:
`ActiveQuery` is 192 bytes and has a couple of collections within it (2 of which are not extract) that we can potentially re-use the allocations for. So pooling those instead of creating, pushing and popping the query from the stack might be beneficial to perf.

Builds on top of https://github.com/salsa-rs/salsa/pull/626

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -1,3 +1,5 @@
+use std::{mem, ops};
+
 use rustc_hash::FxHashMap;
 
 use super::zalsa_local::{EdgeKind, QueryEdges, QueryOrigin, QueryRevisions};
@@ -57,7 +59,7 @@ pub(crate) struct ActiveQuery {
 }
 
 impl ActiveQuery {
-    pub(super) fn new(database_key_index: DatabaseKeyIndex) -> Self {
+    pub(crate) fn new(database_key_index: DatabaseKeyIndex) -> Self {
         ActiveQuery {
             database_key_index,
             durability: Durability::MAX,
@@ -69,6 +71,28 @@ impl ActiveQuery {
             tracked_struct_ids: Default::default(),
             accumulated: Default::default(),
         }
+    }
+
+    fn reset(&mut self, new_database_key_index: DatabaseKeyIndex) {
+        let Self {
+            database_key_index,
+            durability,
+            changed_at,
+            input_outputs,
+            untracked_read,
+            cycle,
+            disambiguator_map,
+            // These two are cleared via `mem::take`` when popped off as revisions.
+            tracked_struct_ids: _,
+            accumulated: _,
+        } = self;
+        *database_key_index = new_database_key_index;
+        *durability = Durability::MAX;
+        *changed_at = Revision::start();
+        input_outputs.clear();
+        *untracked_read = false;
+        *cycle = None;
+        disambiguator_map.clear();
     }
 
     pub(super) fn add_read(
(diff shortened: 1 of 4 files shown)


MichaReiser:
Do you have any perf numbers from ra that confirm the improvement?

Veykril:
Not yet hence why I am fine with waiting on these PRs until salsa can be used for r-a as is so its easier for me to test

Veykril:
Huh that's odd, all I did was rebase on latest master and now we have a perf regression?

src/accumulator/accumulated_map.rs
```diff
@@ -50,6 +50,13 @@ impl AccumulatedMap {
             .unwrap()
             .extend_with_accumulated(output);
     }
+
+    pub(crate) fn take(&mut self) -> Self {
+        Self {
+            map: self.map.drain().collect(),
+            inputs: self.inputs,
+        }
+    }
```
MichaReiser, line 59: This looks wrong. We should use `std::mem::take` on `self.inputs`

Veykril: Oops, good catch, though this was a change I've done after seeing the rebase change

Veykril:
okay locally I see (compared to #626:
- `+2%` on `Mutating Inputs/new/InternedInput`
- `-10%` on `Mutating Inputs/amortized/InternedInput`
- `+19%` on `Mutating Inputs/new/Input`
- `-10%` on `Mutating Inputs/amortized/Input`
- `+9%` on `many_tracked_structs`

Veykril:
Looking at the call graph in the codspeed report the `SeqQueue` push seems to suddenly take a lot more time (due to a syscall!), so the regression here seems to unrelated (different allocation patterns seem to affect it calling out to a different malloc thing)

Veykril:
Looking at the profiling graph in codspeed this regressions to come from different allocation behavior (within unrelated code), so I don't think this is a direct regression?

MichaReiser:
Codspeed can be flaky at times but 22% suggest that there's something wrong. The `crossbeam_queue` call now takes significantly longer... I think we should look into why

Veykril:
It can be flaky but this report has been consistent across my 3 or so rebases earlier. My PR does change allocation patterns, so we might just be getting unlucky with the allocator here now. The `SegQueue` itself doesn't look special, it merely allocates in blocks of 31 items + pointer.

Veykril:
Okay after fixing the benchmarks, this regression seems a lot more workable (and reasonable regarding my changes)

Veykril:
Okay my latest drain changes are the culprit for that one. I do want to say that this PR will generally regress perf over our benches as none of them do multiple queries in succession (which is where this PR shines), they tend to do one or two which means this PR strictly does more allocation work than before.

Veykril:
The regression is a bit confusing to me, it looks like an allocation for the query stack happening, but the stack should already have enough capacity from the `setup` closure 😕

src/active_query.rs
```diff
@@ -116,30 +143,99 @@ impl ActiveQuery {
...
+            changed_at,
+            untracked_read,
+            accumulated_inputs,
+            ref mut cycle_heads,
+            ref mut input_outputs,
+            ref mut tracked_struct_ids,
+            ref mut accumulated,
+        } = self;
+
+        let edges = QueryEdges::new(input_outputs.drain(..));
```
MichaReiser, line 160: Should we use `mem::take` here too (and change `QueryEdges::new` to take a `Box<[QueryEdge]>`) to avoid the extra allocation?

Veykril: `input_outputs` is an `IndexSet` so that won't work

Veykril:
Alright figured out the cause of the regression, now its all green 🎉

MichaReiser (approved):

PR #631, opened 2024-12-15
Do not allocate input singleton data for non-singleton inputs

Veykril:
This does some type shenanigans (unfortunately can't gate on the previously used const as Rust's const story just isn't there yet), to swap the singleton field with `()` if it's not needed.

### DIFF
--- a/src/input/singleton.rs
+++ b/src/input/singleton.rs
@@ -1,0 +1,55 @@
+use crossbeam::atomic::AtomicCell;
+use parking_lot::Mutex;
+
+use crate::{id::FromId, Id};
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+pub trait SingletonChoice: sealed::Sealed + Default {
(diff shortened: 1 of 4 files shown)


src/input/singleton.rs
```diff
@@ -0,0 +1,55 @@
...
+impl Default for Singleton {
+    fn default() -> Self {
+        Self {
+            index: AtomicCell::new(None),
+            lock: Default::default(),
+        }
+    }
+}
+#[derive(Default)]
+pub struct NoSingleton;
```
davidbarsky, line 46: nit: maybe `NotSingleton` instead?

Veykril: Fine with either (apparently the concept of a `Multiton` exists as well, but that's not quite fitting either here)

Veykril:
I'll go ahead and merge this as I don't think there is anything controversial about this PR and its a minor cleanup (in my eyes)

PR #623, opened 2024-12-11
feature: access underlying `IngredientImpl` in interned structs

davidbarsky:
(Builds atop of https://github.com/salsa-rs/salsa/pull/620 and https://github.com/salsa-rs/salsa/pull/618).

This PR adds the ability to fetch interned data from the Salsa database by simply knowing the name of the interned struct and the corresponding ID. For example:

```rust
#[salsa::interned_sans_lifetime(id = salsa::Id)]
struct InternedString {
    data: String
}

let s1 = InternedString::new(db, "Hello, world!");

// later...

let interned_s1 = InternedString::ingredient(&db).fields(&db, s1).0;
```

PR #632, opened 2024-12-17
feature: add `#[salsa::interned_sans_lifetime]` macro

davidbarsky:
(This consolidates #618, #620, and #623 into a single PR, logical PR.)

Here's what this solves:
1. the `'db` lifetime on interned structs makes migrating some rust-analyzer's existing interned values *really* challenging.
2. rust-analyzer wants to fiddle with the underlying ID.
3. We'd like to be able to fetch the associated, interned data without having the wrapped, interned struct on hand.

We'll probably port rust-analyzer over the new, idiomatic API instead, but this makes the migration a lot easier!

### DIFF
--- a/components/salsa-macro-rules/src/lib.rs
+++ b/components/salsa-macro-rules/src/lib.rs
@@ -19,6 +19,7 @@ mod maybe_default;
 mod setup_accumulator_impl;
 mod setup_input_struct;
 mod setup_interned_struct;
+mod setup_interned_struct_sans_lifetime;
 mod setup_method_body;
 mod setup_tracked_fn;
 mod setup_tracked_struct;
--- a/components/salsa-macro-rules/src/setup_interned_struct_sans_lifetime.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct_sans_lifetime.rs
@@ -11,6 +11,10 @@ macro_rules! setup_interned_struct_sans_lifetime {
         // Name of the struct
         Struct: $Struct:ident,
 
+        // Name of the struct data. This is a parameter because `std::concat_idents`
+        // is unstable and taking an addition dependency is unnecessary.
+        StructData: $StructDataIdent:ident,
+
         // Name of the `'db` lifetime that the user gave
         db_lt: $db_lt:lifetime,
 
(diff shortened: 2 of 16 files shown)


MichaReiser:
We discussed that we want to discourage this pattern but it would also be nice if we can avoid duplicating the interned macro. Would it be possible to gate the functionality behind a feature? E.g. that the `interned` macro only supports a `no_lifetime` argument if `salsa_macros` is used with the `compat` feature.

Veykril:
Given the goal is for this to only ease the transition for r-a it should be fine (and honestly preferred in my eyes) to gate this behind a feature until r-a no longer needs it (at which point we can remove it again).

MichaReiser:
It's hard to tell for me what the difference is between the two macros but would it be possible to change `$db_lt` to either be `<'db'>` (so that it includes the `<..>` ` or an empty string if `no_lifetime` is used.

davidbarsky:
> It's hard to tell for me what the difference is between the two macros

I do too, but difftastic helped here. I don't think it's a particularly useful paste, but if you have it handy, `difft components/salsa-macro-rules/src/setup_interned_struct.rs components/salsa-macro-rules/src/setup_interned_struct_sans_lifetime.rs` will show the difference. The primary differences are:
1. The addition of a `StructData` parameter to the macro. I didn't want pull in `paste`, so I just made the ident in the proc macro.
2. The addition of the `$Id` parameter, to allow the ID overriding behavior we need in rust-analyzer.
3. Moving `salsa::plumbing::interned::Configuration` out of the `const _: () = {...} ` block in order to get the data through the underlying ID alone.
4. The removal of lifetimes and a few select additions of `'static`, as needed.

I honestly recommend cloning this PR and diffing via difftastic.

> would it be possible to change `$db_lt` to either be `<'db>` (so that it includes the `<..>`) or an empty string if `no_lifetime` is used? 

Not that I'm aware of, at least with `macro_rules!`-based macros. I don't think we can invoke `maybe_*`-style macros in the type position. From a "how can we unify these macros", I _think_ it'd require moving things away from `salsa-macro-rules` into a (mostly) pure `quote!`-based approach.

davidbarsky:
I fixed the rustdoc failures, but I can't reproduce the test failure w.r.t to cancellation on my Mac. That's a bit concerning, but I think I'll try and debug this tomorrow.

Veykril:
I question the need for the `id =` part actually, I'll raise the why in the rust-analyzer zulip but just so people know. I don't think r-a needs that part.

davidbarsky:
> I fixed the rustdoc failures, but I can't reproduce the test failure w.r.t to cancellation on my Mac. That's a bit concerning, but I think I'll try and debug this tomorrow.

Lukas found and fixed the issue in https://github.com/salsa-rs/salsa/pull/637/.

davidbarsky:
Closing in favor of https://github.com/salsa-rs/salsa/pull/661, which has substantially less duplication and removes support for [overriding IDs](https://github.com/salsa-rs/salsa/pull/632#issuecomment-2558538153).

PR #633, opened 2024-12-18
Implement self-referential/subset key interning 

Veykril:
This lacks a user facing API via the macro but I could not be bothered implementing that as it takes way too much effort with the decl macro quoting unfortunately (and I did not feel like cloning the entire macro for this either). This is enough for rust-analyzer's needs for the time being though as there is only one usage which we have handwritten for now.


nikomatsakis (approved):
Merging for now but obviously it'd be good to add a public API for this at some point, since iiuc you all will be relying on this in rust-analyzer?

PR #635, opened 2024-12-22
Replace unnecessary `Arc` by `Box`

Veykril:
`QueryEdges` were `Arc`ed before, but we only ever clone them for LRU evictions (and there the original one will drop right after eviction), this is a rare occurence. Additionally to save on constantly allocating an empty `Arc` in the case where no deps exist, we needed a `lazy_static`. Instead, we `Box` the slice now, dropping the need for ref-counting as well as the `lazy_static` as an empty slice never allocates in a `Box`. 

Another improvement we should make is use a `ThinBox` once that's a thing, we don't really make use of the length directly and it would shave off 4/8 bytes for Memo/QueryRevisions.

### DIFF
--- a/src/table/memo.rs
+++ b/src/table/memo.rs
@@ -194,11 +194,9 @@ impl MemoTable {
         unsafe { Self::from_dummy::<M>(arc_swap.swap(Self::to_dummy(memo))) };
     }
 
-    pub(crate) fn into_memos(
-        mut self,
-    ) -> impl Iterator<Item = (MemoIngredientIndex, Arc<dyn Memo>)> {
-        let memos = std::mem::take(self.memos.get_mut());
-        memos
+    pub(crate) fn into_memos(self) -> impl Iterator<Item = (MemoIngredientIndex, Arc<dyn Memo>)> {
+        self.memos
+            .into_inner()
             .into_iter()
             .zip(0..)
             .filter_map(|(mut memo, index)| memo.data.take().map(|d| (d, index)))
--- a/src/table/memo.rs
+++ b/src/table/memo.rs
@@ -163,6 +163,37 @@ impl MemoTable {
         unsafe { Some(Self::from_dummy(arc_swap.load_full())) }
     }
 
+    /// Calls `f` on the memo at `memo_ingredient_index` and replaces the memo with the result of `f`.
+    /// If the memo is not present, `f` is not called.
+    pub(crate) fn map_memo<M: Memo>(
+        &self,
+        memo_ingredient_index: MemoIngredientIndex,
+        f: impl FnOnce(Arc<M>) -> Arc<M>,
+    ) {
+        // If the memo slot is already occupied, it must already have the
+        // right type info etc, and we only need the read-lock.
+        let memos = self.memos.read();
+        let Some(MemoEntry {
+            data:
+                Some(MemoEntryData {
+                    type_id,
+                    to_dyn_fn: _,
+                    arc_swap,
+                }),
+        }) = memos.get(memo_ingredient_index.as_usize())
+        else {
+            return;
+        };
+        assert_eq!(
+            *type_id,
+            TypeId::of::<M>(),
+            "inconsistent type-id for `{memo_ingredient_index:?}`"
+        );
+        // SAFETY: type_id check asserted above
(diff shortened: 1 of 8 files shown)


Veykril:
```


---- parallel_map::execute_cancellation stdout ----
[tests/parallel/signal.rs:11:9] format!("signal({})", stage) = "signal(1)"
[tests/parallel/signal.rs:11:9] format!("signal({})", stage) = "signal(2)"
thread 'parallel_map::execute_cancellation' panicked at tests/parallel/parallel_map.rs:91:10:
called `Result::unwrap_err()` on an `Ok` value: [2, 3, 4, 5, 6, 7, 8, 9, 10, 11]

```
seems spurious to me as I can't reproduce it

src/table/memo.rs
```diff
@@ -163,11 +163,40 @@ impl MemoTable {
         unsafe { Some(Self::from_dummy(arc_swap.load_full())) }
     }
 
-    pub(crate) fn into_memos(
-        mut self,
-    ) -> impl Iterator<Item = (MemoIngredientIndex, Arc<dyn Memo>)> {
-        let memos = std::mem::take(self.memos.get_mut());
-        memos
+    /// Calls `f`` on the memo at `memo_ingredient_index` and replaces the memo with the result of f.
```
MichaReiser, line 166:
```suggestion
    /// Calls `f` on the memo at `memo_ingredient_index` and replaces the memo with the result of f.
```

MichaReiser (approved):

MichaReiser:
> ```
> 
> 
> ---- parallel_map::execute_cancellation stdout ----
> [tests/parallel/signal.rs:11:9] format!("signal({})", stage) = "signal(1)"
> [tests/parallel/signal.rs:11:9] format!("signal({})", stage) = "signal(2)"
> thread 'parallel_map::execute_cancellation' panicked at tests/parallel/parallel_map.rs:91:10:
> called `Result::unwrap_err()` on an `Ok` value: [2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
> ```
> 
> seems spurious to me as I can't reproduce it

I don't remember the test being spurious before...

Veykril:
Uuh, I cancelled the first CI run as I pushed twice and apparently that just skips the merge queue? 🤨

Veykril:
> > ```
> > 
> > 
> > ---- parallel_map::execute_cancellation stdout ----
> > [tests/parallel/signal.rs:11:9] format!("signal({})", stage) = "signal(1)"
> > [tests/parallel/signal.rs:11:9] format!("signal({})", stage) = "signal(2)"
> > thread 'parallel_map::execute_cancellation' panicked at tests/parallel/parallel_map.rs:91:10:
> > called `Result::unwrap_err()` on an `Ok` value: [2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > seems spurious to me as I can't reproduce it
> 
> I don't remember the test being spurious before... But @davidbarsky might know more

Not the first time it seems, it even failed on master before https://github.com/salsa-rs/salsa/actions/runs/12397086201/job/34606629011

Veykril:
I can reproduce it by making the main thread sleep for a bit after spawning the other thread it seems

davidbarsky:
> I can reproduce it by making the main thread sleep for a bit after spawning the other thread it seems

Ah, yeah, this is sufficient for me to reproduce:

```patch
diff --git a/tests/parallel/parallel_map.rs b/tests/parallel/parallel_map.rs
index 80e4ebeafe..278bbfedd8 100644
--- a/tests/parallel/parallel_map.rs
+++ b/tests/parallel/parallel_map.rs
@@ -79,6 +79,7 @@
         let db = db.clone();
         move || a1(&db, input)
     });
+    std::thread::sleep(std::time::Duration::from_millis(100));

     let counts = (2..=20).collect::<Vec<u32>>();
```

Veykril:
Ah I can see the problem I think, by the time the signal passes we will no longer do any cancellable work in the `a1` function, so we never end up cancelling (note how the test in `parallel_cancellation` differs by calling a query after the `wait_for` call)

Callofdump (commented):
WHERE?

PR #636, opened 2024-12-22
Shrink `QueryRevisions` size by 3 words

Veykril:


Veykril:
I wonder if https://github.com/salsa-rs/salsa/pull/629 would help here a bit by reducing the need to allocate additionally (or whether the cause of the perf regression is the extra indirection)

MichaReiser:
I haven't reviewed the code changes but any chance you undid the optimization introduced in https://github.com/salsa-rs/salsa/pull/615

Veykril:
Ah I believe I did by accident!

src/ingredient.rs
```diff
@@ -74,7 +74,10 @@ pub trait Ingredient: Any + std::fmt::Debug + Send + Sync {
         &'db self,
         db: &'db dyn Database,
         key_index: Id,
-    ) -> Option<&'db AccumulatedMap>;
+    ) -> (Option<&'db AccumulatedMap>, InputAccumulatedValues) {
+        _ = (db, key_index);
+        (None, InputAccumulatedValues::Any)
+    }
```
Veykril, line 80: One thing to note is that I have this trait method a default impl to reduce some code noise. I imagine the intent was to force the implementor to think about this function but it does make changing signatures really annoying as well.

src/active_query.rs
```diff
@@ -53,6 +55,10 @@ pub(crate) struct ActiveQuery {
     /// Stores the values accumulated to the given ingredient.
     /// The type of accumulated value is erased but known to the ingredient.
     pub(crate) accumulated: AccumulatedMap,
+
+    /// [`InputAccumulatedValues::Empty`] if any input read during the query's execution
+    /// has any direct or indirect accumulated values.
+    pub(super) accumulated_inputs: InputAccumulatedValues,
```
MichaReiser, line 61: I think the documentation is now outdated because you changed it to track whether the query itself or any of its dependencies have any accumulated values.

MichaReiser (approved):

src/accumulator/accumulated_map.rs
```diff
@@ -7,10 +9,6 @@ use super::{accumulated::Accumulated, Accumulator, AnyAccumulated};
 #[derive(Default, Debug)]
 pub struct AccumulatedMap {
     map: FxHashMap<IngredientIndex, Box<dyn AnyAccumulated>>,
```
MichaReiser, line 11: I guess a more local change would be to change `map` here to `Option<Box<FxHashMap>>`. It would have the advantage that all accumualted fields remain closer together.

Veykril: Generally not a bad idea but I think (will have to verify but I am fairly certain) that moving `InputAccumulatedValues` back into `AccumulatedMap` will increase the size of `QueryRevisions` as the field can't be batched together with other fields that are smaller than the alignment.

Veykril: I do think the field here should be option-boxed though instead, that should clean up the code interacting with `AccumulatedMap` at least

davidbarsky:
This was a pretty substantial improvement to rust-analyzer's memory benchmark (`analysis-stats`) when I rebased this change atop of [my forked branch of Salsa](https://github.com/salsa-rs/salsa/compare/master...davidbarsky:salsa:david-and-lukas/required-salsa-changes-for-ra): we went from 4053mb to 3968mb.

Veykril:
We could save another 24 bytes by option boxing `QueryRevisions::tracked_struct_ids` field, assuming a decent number of queries do not need it on average.

Veykril:
And another 8 bytes by having access to a `ThinBox` (I don't know if there is a stable way to createa `ThinBox<[T]>` type today)

PR #637, opened 2024-12-23
Fix parallel_map::execute_cancellation test

Veykril:
As discovered in https://github.com/salsa-rs/salsa/pull/635, the test here is racy. After resolving the `wait_for` it never actually executes cancellable work and so it might just finish computation without cancelling (the only place where it was able to cancel was when calling the `input.field(db)` call. This brings the test more in line with how `parallel_execute` does it.


davidbarsky (approved):
oh, that'll do it. thanks for noticing and fixing this!

PR #639, opened 2024-12-27
internal: add dhat to several `tracked_fn` tests

davidbarsky:
I'm investigating allocations in rust-analyzer. I'm pretty sure that the primary, remaining regression we have in rust-analyzer's memory usage can be attributed to #599 ([discussion on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Porting.20to.20Salsa.203.2E0/near/490944779)). I figured I'd establish some baselines, hence the changes and additions to these tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ ordered-float = "4.2.1"
 rustversion = "1.0"
 test-log = { version ="0.2.11", features = ["trace"] }
 trybuild = "1.0"
+dhat = "0.3"
 
 [[bench]]
 name = "compare"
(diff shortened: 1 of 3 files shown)


davidbarsky:
Hmm, on second though: I don't think this should be a test; it should be a benchmark, but criterion doesn't really allow modeling allocations (https://github.com/bheisler/criterion.rs/issues/97). For the purposes of figuring out the memory regression in rust-analyzer, I think the results these tests showed are more than sufficient. I'll close this PR.

PR #640, opened 2024-12-29
feature: Add support for dumping database contents

davidbarsky:
As discussed in [this HackMD](https://hackmd.io/3Mpv0fa7Rm6cn9-EH5ejkw), we want to dump the contents of a Salsa database in order to debug what rust-analyzer and Salsa are doing. The original Salsa had a [`DebugQueryTable`](https://github.com/rust-lang/rust-analyzer/blob/fa4a40bbe867ed54f5a7c905b591fd7d60ba35eb/crates/ra-salsa/src/debug.rs#L9-L29) that we'd use [like this](https://github.com/rust-lang/rust-analyzer/blob/0a706f7d2ac093985eae317781200689cfd48b78/crates/rust-analyzer/src/cli/analysis_stats.rs#L274-L286):

```rust
let mut total_file_size = Bytes::default();
for e in ide_db::base_db::ParseQuery.in_db(db).entries::<Vec<_>>() {
    total_file_size += syntax_len(db.parse(e.key).syntax_node())
}

let mut total_macro_file_size = Bytes::default();
for e in hir::db::ParseMacroExpansionQuery.in_db(db).entries::<Vec<_>>() {
    let val = db.parse_macro_expansion(e.key).value.0;
    total_macro_file_size += syntax_len(val.syntax_node())
}
eprintln!("source files: {total_file_size}, macro files: {total_macro_file_size}");
```

(`ParseQuery` and `ParseMacroExpansionQuery` [were generated](https://github.com/rust-lang/rust-analyzer/blob/59bc7b49d0ad319de8c477c63da552cbc8a05e4c/crates/ra-salsa/ra-salsa-macros/src/query_group.rs#L43-L44) from [`SourceDatabase::parse`](https://github.com/rust-lang/rust-analyzer/blob/59bc7b49d0ad319de8c477c63da552cbc8a05e4c/crates/base-db/src/lib.rs#L69C8-L69C13) and [`ExpandDatabase::parse_macro_expansion`](https://github.com/rust-lang/rust-analyzer/blob/59bc7b49d0ad319de8c477c63da552cbc8a05e4c/crates/hir-expand/src/db.rs#L69-L73), respectively.)

This PR takes a slightly different approach: instead of generating a `Query` struct corresponding to a tracked function or Salsa struct, this PR simply exposes the underlying pages on `salsa::Storage` via the `debug_input_entries`, `debug_input_entries`, and `debug_interned_entries` methods. This results in the following API:

```rust
let _ = InternedStruct::new(&db, "Hello, world!".to_string());

// test interned structs
let interned = db
    .storage()
    .debug_interned_entries::<InternedStruct>()
    .collect::<Vec<_>>();
    
assert_eq!(interned[0].data.0, "Hello, world!");
```

Few notes:
- I don't see any obvious way to make these methods generic over the underlying Salsa struct, but I'm not sure it's desirable in the first place.
- I'm not _entirely_ happy with the name or API, so suggestions to improve it would be welcome. I don't know if I want to expose this debug information on the Salsa structs themselves, but I'd be happy to hear arguments in favor.
- This approach requires exposing parts of `{interned, tracked, input}::Value` publicly, which _feels_ suboptimal, but we're all friends here, right?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,11 @@ salsa-macros = { path = "components/salsa-macros" }
 smallvec = "1"
 rayon = "1.10.0"
 
+[features]
+# FIXME: remove this as a default feature before 1.0.
+default = ["salsa_unstable"]
+salsa_unstable = []
+
 [dev-dependencies]
 annotate-snippets = "0.11.5"
 derive-new = "0.6.0"
@@ -33,7 +38,7 @@ eyre = "0.6.8"
 notify-debouncer-mini = "0.4.1"
 ordered-float = "4.2.1"
 rustversion = "1.0"
-test-log = { version ="0.2.11", features = ["trace"] }
+test-log = { version = "0.2.11", features = ["trace"] }
 trybuild = "1.0"
 
 [[bench]]
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -24,8 +24,8 @@ use super::Revision;
 pub trait Configuration: Sized + 'static {
     const DEBUG_NAME: &'static str;
 
-    /// The type of data being interned
-    type Data<'db>: InternedData;
+    /// The fields of the struct being interned.
+    type Fields<'db>: InternedData;
 
     /// The end user struct
     type Struct<'db>: Copy;
@@ -60,7 +60,7 @@ pub struct IngredientImpl<C: Configuration> {
     /// Maps from data to the existing interned id for that data.
     ///
     /// Deadlock requirement: We access `value_map` while holding lock on `key_map`, but not vice versa.
-    key_map: FxDashMap<C::Data<'static>, Id>,
+    key_map: FxDashMap<C::Fields<'static>, Id>,
 
     /// Stores the revision when this interned ingredient was last cleared.
     /// You can clear an interned table at any point, deleting all its entries,
@@ -74,9 +74,19 @@ pub struct Value<C>
 where
     C: Configuration,
 {
-    data: C::Data<'static>,
+    fields: C::Fields<'static>,
     memos: MemoTable,
     syncs: SyncTable,
+}
+
+impl<C> Value<C>
+where
+    C: Configuration,
+{
+    /// Fields of this interned struct.
+    pub fn fields(&self) -> &C::Fields<'static> {
+        &self.fields
+    }
 }
 
 impl<C: Configuration> Default for JarImpl<C> {
@@ -113,11 +123,11 @@ where
         }
     }
 
-    unsafe fn to_internal_data<'db>(&'db self, data: C::Data<'db>) -> C::Data<'static> {
+    unsafe fn to_internal_data<'db>(&'db self, data: C::Fields<'db>) -> C::Fields<'static> {
         unsafe { std::mem::transmute(data) }
     }
 
-    unsafe fn from_internal_data<'db>(data: &'db C::Data<'static>) -> &'db C::Data<'db> {
+    unsafe fn from_internal_data<'db>(data: &'db C::Fields<'static>) -> &'db C::Fields<'db> {
         unsafe { std::mem::transmute(data) }
     }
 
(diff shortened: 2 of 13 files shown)


src/storage.rs
```diff
@@ -58,6 +59,45 @@ impl<Db: Database> Default for Storage<Db> {
 }
 
 impl<Db: Database> Storage<Db> {
+    pub fn debug_input_entries<T>(&self) -> impl Iterator<Item = &input::Value<T>>
+    where
+        T: input::Configuration,
+    {
+        let zalza = self.zalsa_impl();
```
nikomatsakis, line 66:
```suggestion
        let zalsa = self.zalsa_impl();
```

src/storage.rs
```diff
@@ -58,6 +59,45 @@ impl<Db: Database> Default for Storage<Db> {
...
+        zalza
```
nikomatsakis, line 67:
```suggestion
        zalsa
```

src/tracked_struct.rs
```diff
@@ -216,7 +216,7 @@ where
 
     /// Fields of this tracked struct. They can change across revisions,
     /// but they do not change within a particular revision.
-    fields: C::Fields<'static>,
+    pub fields: C::Fields<'static>,
```
nikomatsakis, line 219: Why is this public? That sounds a bit worrisome.

src/interned.rs
```diff
@@ -73,7 +73,7 @@ pub struct Value<C>
 where
     C: Configuration,
 {
-    data: C::Data<'static>,
+    pub data: C::Data<'static>,
```
nikomatsakis, line 76: it feels like this should be renamed to `fields` for consistency; maybe `C::Fields`, even

nikomatsakis (changes requested):
This PR seems fine. I left some nits. I'm of mixed minds about exposing the fields directly. Also, returning a `&(A, B, C)` means that the data *must* store tuples directly, which perhaps we might want to change at some point.

I am thinking about returning a `(&A, &B, &C)` tuple... we'd have to do a bit more work to do that, but it would be very flexible if we wanted to refactor away from storing a tuple.

davidbarsky:
> This PR seems fine. I left some nits. I'm of mixed minds about exposing the fields directly. Also, returning a `&(A, B, C)` means that the data _must_ store tuples directly, which perhaps we might want to change at some point.

I'm also of mixed minds of exposing fields directly; it doesn't feel _great_.

> I am thinking about returning a `(&A, &B, &C)` tuple... we'd have to do a bit more work to do that, but it would be very flexible if we wanted to refactor away from storing a tuple.

I like the `(&A, &B, &C)` idea. I initially wrote a longer response proposing that we'd use a dedicated/standalone struct for this, but I realized that it's effectively a named variant of `(&A, &B, &C)` and raises questions about how and when said struct would be generated.

Anyways, for immediate next steps, I'll extend this to the outputs of tracked functions and fix some of nits. Longer term, I'm unsure if I'd like to revisit this before a 1.0 (moving to the `(&A, &B, &C)` approach) or feature-flagging this functionality behind an unstable feature flag. To expand on the former, I haven't looked into what work the `(&A, &B, &C)` approach entails, but I worry it requires writing some trait implementations for tuples of different lengths, but the flexibility to evolve is pretty darn compelling.

nikomatsakis:
@davidbarsky that sounds good to me. Maybe move it under a feature before we land?

davidbarsky:
@nikomatsakis Yeah. I'll move this to be under a Cargo feature instead of a `RUSTFLAG`, but I'll name it something like `unstable-debug` and say its exempt from semver. I also may the dumping a tracked function's outputs since I think the changes here might be sufficient for rust-analyzer's needs, but I don't know yet.

nikomatsakis:
@davidbarsky I propose you call it something like `semver-exempt-debug` and also that we rename [the `plumbing` module](https://github.com/salsa-rs/salsa/blob/master/src/lib.rs#L68) to be `semver-exempt`.

nikomatsakis:
or `semver-unstable`? idk, is there a widely used convention for this?

nikomatsakis:
in any case I'd prefer not to land this PR without the cargo feature, but we can if you like

nikomatsakis:
Agreed that renaming the plumbing module should be separate. I like the name semver-unstable.

nikomatsakis:
I'm indifferent about RUST FLAGS vs not as long as the name clearly signals the instability

nikomatsakis (approved):

Cargo.toml
```diff
@@ -24,6 +24,11 @@ salsa-macros = { path = "components/salsa-macros" }
 smallvec = "1"
 rayon = "1.10.0"
 
+[features]
+# FIXME: remove this as a default feature before 1.0.
+default = ["salsa_unstable"]
```
nikomatsakis, line 29: This can wait until a future PR, but I think we want to tweak the testing setup so that this feature is not included for (most) tests.

PR #641, opened 2024-12-31
Fix accumulated values for backdated queries

MichaReiser:
As reported on [Zulip](https://salsa.zulipchat.com/#narrow/channel/333573-Contributing-to-Salsa/topic/Accumulator.20performance)

The optimization introduced in https://github.com/salsa-rs/salsa/pull/621 doesn't account for queries that get backdated but now have accumulated values (or had accumulated values and now no more). 

This PR fixes this regression but I think it needs some cleanup before it can be landed. Specifically, I think the refactor in https://github.com/salsa-rs/salsa/pull/636 that moves the `AccumulatedInputValues` to the memo would be an improvement. I'm happy to follow up on this sometime when I've a little more time or someone else takes it over from here.

### DIFF
--- a/src/accumulator.rs
+++ b/src/accumulator.rs
@@ -12,7 +12,7 @@ use accumulated_map::AccumulatedMap;
 
 use crate::{
     cycle::CycleRecoveryStrategy,
-    ingredient::{fmt_index, Ingredient, Jar},
+    ingredient::{fmt_index, Ingredient, Jar, MaybeChangedAfter},
     plumbing::JarAux,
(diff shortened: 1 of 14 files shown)


Veykril:
I agree, the split done in https://github.com/salsa-rs/salsa/pull/636 would likely make this look a bit nicer in general (I undrafted that PR now, I am pretty happy with how it looks)

MichaReiser:
Closing in favor of https://github.com/salsa-rs/salsa/pull/662

PR #642, opened 2025-01-04
Improve safety comments on function/fetch

Veykril:


src/function/input_outputs.rs
Veykril, line 1: This file is dangling, it seems to have been split across `input.rs` and `accumulated.rs`

Veykril:
I'll be cheeky and save approve this given it doesn't change behavior, mainly to get rid of the clippy failure on nightly.

PR #643, opened 2025-01-04
Fix compare benchmarks not benching the right things

Veykril:


Veykril:
Not sure whats up with the regression for that one (the interned input benchs are a bit suspect anyways to me), but this should reduce noise within the mutation ones as we now reset the database freeing all the previous inputs properly

Veykril:
Though on a related note, I wonder if all our benches should actually have setup functions instead. It seems odd to me that we re-use databases between calls, feels like we arent benching what we want to bench in some cases because of that.

MichaReiser:
> Though on a related note, I wonder if all our benches should actually have setup functions instead. It seems odd to me that we re-use databases between calls, feels like we arent benching what we want to bench in some cases because of that.

That's at least what we do in Ruff and our benchmarks have proven to be very consistent.

benches/compare.rs
```diff
@@ -26,25 +28,29 @@ fn mutating_inputs(c: &mut Criterion) {
...
+                    (db, base_string, base_len, string, new_len)
+                },
+                |(db, base_string, base_len, string, new_len)| {
+                    let input = Input::new(db, base_string.clone());
+                    let actual_len = length(db, input);
+                    assert_eq!(base_len, actual_len);
+
+                    input.set_text(db).to(string.clone());
+                    let actual_len = length(db, input);
+                    assert_eq!(new_len, actual_len);
```
MichaReiser, line 50: This seems incorrect now. I think the idea of the benchmark is to mutate `input` by changing the text and measuring how long it takes to do the incremental computation. 

What this is measuring now is how long it takes to create a new input and changing that input.

MichaReiser (approved):

benches/compare.rs
Veykril: Well, that was the behavior before my change already. You are right that the name implies we want to only bench the mutation but it never did so. Will change that

Veykril:
I've changed the compare benchmarks to use appropriate setup functions now

MichaReiser:
I suspect that the regression now comes from the fact that we are a) allocating more and b) the cached `IngredientIndex` optimization no longer applies.

Veykril:
Yep I expect this to regress all of them now that we do them properly (not sharing state between runs)

benches/compare.rs
Veykril: [`99568ba` (#643)](https://github.com/salsa-rs/salsa/pull/643/commits/99568baff84b5b269accdce9b877a5a8644af485)

benches/compare.rs
```diff
@@ -56,34 +65,58 @@ fn inputs(c: &mut Criterion) {
...
-        let input = InternedInput::new(&db, "hello, world!".to_owned());
-        let _ = interned_length(&db, input);
-
-        b.iter(|| interned_length(&db, input));
+        b.iter_batched_ref(
+            || {
+                let db = salsa::DatabaseImpl::default();
+                // we can't pass this along otherwise, and the lifetime is generally informational
+                let input: InternedInput<'static> =
+                    unsafe { transmute(InternedInput::new(&db, "hello, world!".to_owned())) };
```
Veykril, line 85: This is a bit annoying, the alternative I suppose is to lift out the `db` binding, assign to that, then return a reference to that but I think I prefer this over that.

Veykril:
or well, mutating improving also makes sense given we now actually test the mutation only

PR #644, opened 2025-01-04
Remove unreachable unwraps via typing

Veykril:
Output edges can never contain a `None` key index so `mark_validated_output` and `remove_stale_output` unwrap these options. We can easily get around that with a bit more strict typing though.

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -1,16 +1,68 @@
+use core::fmt;
+
 use crate::{cycle::CycleRecoveryStrategy, zalsa::IngredientIndex, Database, Id};
 
 /// An integer that uniquely identifies a particular query instance within the
-/// database. Used to track dependencies between queries. Fully ordered and
+/// database. Used to track output dependencies between queries. Fully ordered and
 /// equatable but those orderings are arbitrary, and meant to be used only for
 /// inserting into maps and the like.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct DependencyIndex {
+pub struct OutputDependencyIndex {
+    ingredient_index: IngredientIndex,
+    key_index: Id,
+}
+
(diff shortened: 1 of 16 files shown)


src/key.rs
```diff
@@ -1,36 +1,35 @@
...
 /// An integer that uniquely identifies a particular query instance within the
-/// database. Used to track dependencies between queries. Fully ordered and
+/// database. Used to track output dependencies between queries. Fully ordered and
 /// equatable but those orderings are arbitrary, and meant to be used only for
 /// inserting into maps and the like.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct DependencyIndex {
-    pub(crate) ingredient_index: IngredientIndex,
-    pub(crate) key_index: Option<Id>,
+pub struct OutputDependencyIndex {
```
MichaReiser, line 10: Do we rely on the `Ord` implementation anywhere? I suspect not because `QueryEdge` doesn't implement `Ord`. Should we, therefore, remove it?

MichaReiser (approved):
Nice, I like how it also communicates the intent better (makes accidentally mixing inputs and outputs harder). 

My only open question is around the `Ord` definition. The old implementation allowed ordering indices with both inputs and outputs, now the ordering is per type first (unless we add a custom `Ord` to `QueryEdge`)

src/key.rs
Veykril: That is a good point to raise, I agree I'll remove them.

Veykril: I went a head and removed it from `DatabaseKeyIndex` as well (and hence `Cycle`), as it was only used in one place and there the ord call won't ever be reached (as its behind the ordering of an enumerator in a tuple), [`189fc09` (#644)](https://github.com/salsa-rs/salsa/pull/644/commits/189fc09055896f3a6306374532cdafc48a5990a5)

PR #646, opened 2025-01-05
Remove panicking paths from `maybe_changed_after`

Veykril:
Follow up on https://github.com/salsa-rs/salsa/pull/644

### DIFF
--- a/src/key.rs
+++ b/src/key.rs
@@ -87,9 +87,14 @@ impl InputDependencyIndex {
         db: &dyn Database,
         last_verified_at: crate::Revision,
     ) -> bool {
-        db.zalsa()
-            .lookup_ingredient(self.ingredient_index)
-            .maybe_changed_after(db, self.key_index, last_verified_at)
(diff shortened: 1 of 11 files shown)


src/key.rs
```diff
@@ -87,9 +87,14 @@ impl InputDependencyIndex {
         db: &dyn Database,
         last_verified_at: crate::Revision,
     ) -> bool {
-        db.zalsa()
-            .lookup_ingredient(self.ingredient_index)
-            .maybe_changed_after(db, self.key_index, last_verified_at)
+        match self.key_index {
+            // Only entries within a table can be queried for changes.
```
MichaReiser, line 91: I'd prefer changing the comment to match the `for_table` documentation so that it says that entries without a key remain valid until the entire table is reset (which is forever?). It makes it clear that `false` is the expected output for these cases and not just a "made up" fallback.

MichaReiser (approved):

PR #647, opened 2025-01-05
Do not re-hash already hashed keys disambiguator and struct identity maps

Veykril:
Additionally newtypes these maps.

We got no reason to hash what we already have hashed! (This does need hashbrown as std doesn't stably expose the required API yet)


MichaReiser:
Did you do any perf measurements on ra to verify that the extra complexity and memory usage is worth it?

Veykril:
Not yet, until r-a can use the master branch salsa, checking anything against r-a is kind of a pain as I'd need to work against david's fork I believe

src/tracked_struct.rs
```diff
@@ -151,10 +151,15 @@ where
...
 /// stored in the [`ActiveQuery`](`crate::active_query::ActiveQuery`)
 /// struct and later moved to the [`Memo`](`crate::function::memo::Memo`).
-#[derive(Debug, PartialEq, Eq, Hash, PartialOrd, Ord, Copy, Clone)]
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Copy, Clone)]
 pub(crate) struct Identity {
-    /// Hash of fields with id attribute
-    identity_hash: IdentityHash,
+    // conceptually, this contains an `IdentityHash`, but using the type directly will grow the size
+    // of if this struct by a word due to unusable padding, so we store the fields directly instead.
+    /// Index of the tracked struct ingredient.
```
davidbarsky, line 158: two nits:
- Phrasing
- Considering moving this comment up to `Identity`'s doc comment?

```suggestion
    /// Index of the tracked struct ingredient.
    ///
    /// Conceptually, this contains an `IdentityHash`, but using `IdentityHash` directly will grow the size
    /// of this struct struct by a `std::mem::size_of::<usize>()` due to unusable padding. To avoid this increase
    /// in size, `Identity` stores the fields directly instead.
```

src/tracked_struct.rs
```diff
@@ -181,6 +186,60 @@ pub struct IdentityHash {
     hash: u64,
 }
 
+#[derive(Default, Debug)]
+pub(crate) struct IdentityMap {
+    // we use a non-hasher hashmap here as our key contains its own hash (in a sense)
```
davidbarsky, line 191: do you mind elaborating on what that sense is, or linking towards `Identity`? sorry, the internals Salsa can be a bit subtle, and I'm reviewing by campground rules 😄

src/tracked_struct.rs
```diff
@@ -181,6 +186,60 @@ pub struct IdentityHash {
...
+}
+
+impl IdentityMap {
+    pub(crate) fn insert(&mut self, key: Identity, id: Id) -> Option<Id> {
+        use hashbrown::hash_map::RawEntryMut;
+
+        let eq_module_hash = |k: &Identity| {
+            k.ingredient_index == key.ingredient_index && k.disambiguator == key.disambiguator
+        };
+        let entry = self.map.raw_entry_mut().from_hash(key.hash, eq_module_hash);
```
davidbarsky, line 214: consider inlining `eq_module_hash`?

src/tracked_struct.rs
```diff
@@ -181,6 +186,60 @@ pub struct IdentityHash {
...
+                vacant.insert_with_hasher(key.hash, key, id, |k| k.hash);
+                None
+            }
+        }
+    }
+
+    pub(crate) fn get(&self, key: &Identity) -> Option<Id> {
+        let eq_module_hash = |k: &Identity| {
+            k.ingredient_index == key.ingredient_index && k.disambiguator == key.disambiguator
+        };
```
davidbarsky, line 227: consider inlining `eq_module_hash`?

src/tracked_struct.rs
```diff
@@ -231,8 +290,35 @@ where
...
+    // so we use the raw entry api instead to avoid the overhead of hashing unnecessarily
+    map: hashbrown::HashMap<IdentityHash, Disambiguator, ()>,
+}
+
+impl DisambiguatorMap {
+    pub(crate) fn disambiguate(&mut self, key: IdentityHash) -> Disambiguator {
+        use hashbrown::hash_map::RawEntryMut;
+
+        let eq_module_hash = |k: &IdentityHash| k.ingredient_index == key.ingredient_index;
+        let entry = self.map.raw_entry_mut().from_hash(key.hash, eq_module_hash);
```
davidbarsky, line 308: consider inlining `eq_module_hash`?

src/zalsa_local.rs
```diff
@@ -254,19 +254,19 @@ impl ZalsaLocal {
     #[track_caller]
     pub(crate) fn tracked_struct_id(&self, identity: &Identity) -> Option<Id> {
         self.with_query_stack(|stack| {
-            let top_query = stack.last().expect(
-                "cannot create a tracked struct disambiguator outside of a tracked function",
-            );
-            top_query.tracked_struct_ids.get(identity).copied()
+            let top_query = stack
+                .last()
+                .expect("cannot create a tracked struct id outside of a tracked function");
```
davidbarsky, line 259:
```suggestion
                .expect("cannot create a tracked struct ID outside of a tracked function");
```

src/zalsa_local.rs
```diff
@@ -254,19 +254,19 @@ impl ZalsaLocal {
...
 
     #[track_caller]
     pub(crate) fn store_tracked_struct_id(&self, identity: Identity, id: Id) {
         self.with_query_stack(|stack| {
-            let top_query = stack.last_mut().expect(
-                "cannot create a tracked struct disambiguator outside of a tracked function",
-            );
+            let top_query = stack
+                .last_mut()
+                .expect("cannot store a tracked struct id outside of a tracked function");
```
davidbarsky, line 269:
```suggestion
                .expect("cannot store a tracked struct ID outside of a tracked function");
```

davidbarsky (approved):
looks good to me, modulo documentation and style nits.

src/tracked_struct.rs
Veykril: I put it outside of the doc comment as this is just an implementation detail, it is not relevant to the `ingredient_index` field

src/tracked_struct.rs
Veykril: Oh, that was supposed to be `eq_modulo_hash`. I prefer keeping it outside as it documents what we check (also the formatting like this is more readable imo)

MichaReiser:
Backing out of this PR fixes the coarse grained dependency upgrade in Ruff. I've to take a look at this more closely tomorrow

Veykril:
That sounds surprising to me as this PR shouldn't have changed anything. I did find something odd about the coarse grained one though https://github.com/salsa-rs/salsa/pull/657#discussion_r1947962216

src/tracked_struct.rs
```diff
@@ -231,8 +291,35 @@ where
...
+    // we use a non-hasher hashmap here as our key contains its own hash (in a sense)
+    // so we use the raw entry api instead to avoid the overhead of hashing unnecessarily
+    map: hashbrown::HashMap<IdentityHash, Disambiguator, ()>,
+}
+
+impl DisambiguatorMap {
+    pub(crate) fn disambiguate(&mut self, key: IdentityHash) -> Disambiguator {
+        use hashbrown::hash_map::RawEntryMut;
+
+        let eq_modulo_hash = |k: &IdentityHash| k.ingredient_index == key.ingredient_index;
```
MichaReiser, line 308: It seems that we also need to include `k.hash == key.hash`....

ibraheemdev: To add a little context, hashbrown doesn't store a full hash of each key, and only compares partially. The equality function cannot rely on the hash being equal, it must uniquely identify an entry regardless (e.g. hashbrown could legally do a linear search for equality).

Veykril: Ah I see, well that would do it ... apologies

PR #651, opened 2025-01-07
feature: expose `Database::unwind_if_revision_cancelled`

davidbarsky:
This method is used all over rust-analyzer, most prominently in rust-analyzer's `parallel_prime_caches` method. We use this method in order to check for cancellation in an ad-hoc, map/reduce-shaped threadpool inside of `parallel_prime_caches`, and (re)introducing this method solves a hang there.

### DIFF
--- a/src/database.rs
+++ b/src/database.rs
@@ -54,6 +54,24 @@ pub trait Database: Send + AsDynDatabase + Any + ZalsaDatabase {
         )
     }
 
+    /// Starts unwinding the stack if the current revision is cancelled.
+    ///
+    /// This method can be called by query implementations that perform
+    /// potentially expensive computations, in order to speed up propagation of
+    /// cancellation.
+    ///
+    /// Cancellation will automatically be triggered by salsa on any query
+    /// invocation.
+    ///
+    /// This method should not be overridden by `Database` implementors. A
+    /// `salsa_event` is emitted when this method is called, so that should be
+    /// used instead.
+    fn unwind_if_revision_cancelled(&self) {
+        let db = self.as_dyn_database();
+        let zalsa_local = db.zalsa_local();
+        zalsa_local.unwind_if_revision_cancelled(db);
+    }
+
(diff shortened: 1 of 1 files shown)


Veykril:
Wait r-a hangs without this? As in, deadlocks? That sounds like a bug in salsa then. Also salsa should already be calling this for us, quoting the comment
>     /// Cancellation will automatically be triggered by salsa on any query
>    /// invocation.
>    ///

We should bring the function back either way but this sounds weird.

src/database.rs
```diff
@@ -54,6 +54,12 @@ pub trait Database: Send + AsDynDatabase + Any + ZalsaDatabase {
         )
     }
 
+    fn unwind_if_revision_cancelled(&self) {
```
Veykril, line 57: Would be good to lift the comment from `ZalsaLocal` on here

MichaReiser:
> Wait r-a hangs without this? As in, deadlocks? That sounds like a bug in salsa then. Also salsa should already be calling this for us, quoting the comment

I think what @davidbarsky means is that it's a useful function to call before an expensive computation (or during) to abort on a timely manner instead of computing something for a long time, just to throw it away later.

Veykril:
Ah okay I see, we block on threads/tasks returning (that will eat the cancellation unwind), sorry for the noise

davidbarsky:
> Ah okay I see, we block on threads/tasks returning (that will eat the cancellation unwind), sorry for the noise

Yes, what Lukas said. We should probably look into rewriting `parallel_prime_caches` to not swallow cancellation signals as we currently do and use Salsa 3's new parallelism functionality (especially if we can do progress reporting!)

src/database.rs
davidbarsky: done!

Veykril (approved):

davidbarsky:
I'll merge this, since Niko was okay with it yesterday.

PR #653, opened 2025-01-07
chore: fix completions in `#[salsa::tracked]` functions

davidbarsky:
This change allows rust-analyzer (and, I assume RustRover) to provide accurate completions inside of a `#[salsa::tracked]` function. Previously, I'd comment out the `#[salsa::tracked]` annotation when writing a tracked function, but this PR removes the need for that workaround.

As @Veykril put it, the _core_ issue is that rust-analyzer's `mbe` crate is not very error-resilient: we no longer pick the correct match arm of the `macro_rules` macro because the input doesn't match—`mbe` _should_ pick the closest match instead of giving up, but until `mbe` supports that, this PR ~~papers over a rust-analyzer weakness~~ 🌟  it makes the macro more IDE friendly 🌟 !

(Credit goes to Lukas for the tip, explanation, and joke!)

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -32,7 +32,7 @@ macro_rules! setup_tracked_fn {
         output_ty: $output_ty:ty,
 
         // Function body, may reference identifiers defined in `$input_pats` and the generics from `$generics`
-        inner_fn: $inner_fn:item,
+        inner_fn: {$($inner_fn:tt)*},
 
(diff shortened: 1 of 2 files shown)


Veykril (commented):
Clearly I should not be the one to approve this but if we do this we ought to give the same treatment to `setup_method_body`

davidbarsky:
Getting methods to work is a bit trickier: the "wrap the inner fn in a block" trick doesn't seem to be sufficient for methods, and I'm not enough of an expert on macros to figure that out. Can we make that a followup?

Issue #598, opened 2024-10-16
Coarse-grained tracked structs


nikomatsakis:
As described in [this hackmd](https://hackmd.io/@salsa/ryisALuT0), we want to change how tracked struct dependencies work. The basic idea is to switch the default: instead of labeling some fields as `#[id]` and having the rest be independently tracked, we will have fields be `#[id]` by default unless they are annotated with `#[salsa::tracked]`. This is a fairly straightforward change that can be done in the proc-macro.

## Example

Before:

```rust
#[salsa::tracked]
struct Function {
    #[id]
    name: String,
    body: String,
}
```

After:

```rust
#[salsa::tracked]
struct Function {
    name: String,

    #[salsa::tracked]
    body: String,
}
```

MichaReiser:
CC: @ibraheemdev

nikomatsakis:
# Mentoring instructions

Ok, I realized this is a bit more complicated than I thought. There are two parts required

1. Swap the default.
2. Change how we track reads of `#[id]` fields -- we don't currently distinguish them, but we should.

## swap the default

The decision about which fields are "id" fields is driven by this input to the macro-rules macro:

https://github.com/salsa-rs/salsa/blob/master/components/salsa-macro-rules/src/setup_tracked_struct.rs#L32-L33

which is provided here

https://github.com/salsa-rs/salsa/blob/c6c51a0ea0bb11ce5a449c4b9256c66c5e484fee/components/salsa-macros/src/tracked_struct.rs#L112

it should be fairly straightforward to modify the latter file to flip the defaults. (Note that the next section will also likely change this interface slightly.)

## change how we track reads of `#[id]` fields

This is the trickier bit. Currently we create a field ingredient for every field, regardless of whether it's part of the `#[id]` hash:

https://github.com/salsa-rs/salsa/blob/c6c51a0ea0bb11ce5a449c4b9256c66c5e484fee/src/tracked_struct.rs#L101-L115

the field accessor for every field is the same:

https://github.com/salsa-rs/salsa/blob/c6c51a0ea0bb11ce5a449c4b9256c66c5e484fee/components/salsa-macro-rules/src/setup_tracked_struct.rs#L198-L212

in particular it invokes the `field` method from the tracked struct on line 205. That function is defined here:

https://github.com/salsa-rs/salsa/blob/c6c51a0ea0bb11ce5a449c4b9256c66c5e484fee/src/tracked_struct.rs#L516-L524

[...]

PR #657, opened 2025-01-17
Coarse-grained tracked structs

ibraheemdev:
This PR makes tracked structs coarse-grained by default. To declare a field independently tracked, you now use the `#[tracked]` attribute. This is the opposite of the previous behavior, where `#[id]` effectively declared a field untracked. The `#[id]` attribute now has no effect. Importantly, reads of untracked fields do not require a read dependency to be declared.

There is some duplicated code in the macros as we need to generate separate accessors for tracked vs. untracked fields and I would like these to be static (as opposed to branching at runtime).

I believe the guide documentation also needs to be updated.

Resolves https://github.com/salsa-rs/salsa/issues/598.

### DIFF
--- a/examples/calc/ir.rs
+++ b/examples/calc/ir.rs
@@ -79,7 +79,6 @@ pub enum Op {
 pub struct Function<'db> {
     pub name: FunctionId<'db>,
 
-    #[tracked]
     name_span: Span<'db>,
 
     #[tracked]
--- a/examples/calc/ir.rs
+++ b/examples/calc/ir.rs
@@ -28,6 +28,7 @@ pub struct FunctionId<'db> {
 // ANCHOR: program
 #[salsa::tracked]
 pub struct Program<'db> {
+    #[tracked]
     #[return_ref]
     pub statements: Vec<Statement<'db>>,
 }
@@ -76,14 +77,16 @@ pub enum Op {
 // ANCHOR: functions
 #[salsa::tracked]
 pub struct Function<'db> {
-    #[id]
     pub name: FunctionId<'db>,
 
+    #[tracked]
     name_span: Span<'db>,
 
+    #[tracked]
     #[return_ref]
     pub args: Vec<VariableId<'db>>,
 
+    #[tracked]
     #[return_ref]
     pub body: Expression<'db>,
 }
@@ -91,7 +94,9 @@ pub struct Function<'db> {
 
 #[salsa::tracked]
 pub struct Span<'db> {
+    #[tracked]
     pub start: usize,
+    #[tracked]
     pub end: usize,
 }
 
--- a/src/tracked_struct.rs
+++ b/src/tracked_struct.rs
@@ -26,6 +26,7 @@ pub mod tracked_field;
 pub trait Configuration: Sized + 'static {
     const DEBUG_NAME: &'static str;
     const FIELD_DEBUG_NAMES: &'static [&'static str];
+    const TRACKED_FIELD_DEBUG_NAMES: &'static [&'static str];
 
     /// A (possibly empty) tuple of the fields for this struct.
     type Fields<'db>: Send + Sync;
@@ -107,9 +108,14 @@ impl<C: Configuration> Jar for JarImpl<C> {
     ) -> Vec<Box<dyn Ingredient>> {
         let struct_ingredient = <IngredientImpl<C>>::new(struct_index);
 
+        // If there are no tracked fields, we can skip creating any ingredients.
+        if C::TRACKED_FIELD_DEBUG_NAMES.is_empty() {
+            return vec![Box::new(struct_ingredient)];
+        }
+
         std::iter::once(Box::new(struct_ingredient) as _)
-            // Note that we create ingredients for untracked fields as well, in order to
-            // keep field indices relative to the entire struct.
+            // Otherwise, we have to create ingredients for untracked fields as well, in
+            // order to keep field indices relative to the entire struct.
             .chain((0..C::FIELD_DEBUG_NAMES.len()).map(|field_index| {
                 Box::new(<FieldIngredientImpl<C>>::new(struct_index, field_index)) as _
             }))
--- a/src/tracked_struct.rs
+++ b/src/tracked_struct.rs
@@ -26,7 +26,6 @@ pub mod tracked_field;
 pub trait Configuration: Sized + 'static {
     const DEBUG_NAME: &'static str;
     const FIELD_DEBUG_NAMES: &'static [&'static str];
-    const TRACKED_FIELD_DEBUG_NAMES: &'static [&'static str];
 
     /// A (possibly empty) tuple of the fields for this struct.
     type Fields<'db>: Send + Sync;
@@ -109,7 +108,9 @@ impl<C: Configuration> Jar for JarImpl<C> {
         let struct_ingredient = <IngredientImpl<C>>::new(struct_index);
 
         std::iter::once(Box::new(struct_ingredient) as _)
-            .chain((0..C::TRACKED_FIELD_DEBUG_NAMES.len()).map(|field_index| {
+            // Note that we create ingredients for untracked fields as well, in order to
+            // keep field indices relative to the entire struct.
+            .chain((0..C::FIELD_DEBUG_NAMES.len()).map(|field_index| {
                 Box::new(<FieldIngredientImpl<C>>::new(struct_index, field_index)) as _
             }))
             .collect()
--- a/src/tracked_struct.rs
+++ b/src/tracked_struct.rs
@@ -24,9 +24,14 @@ pub mod tracked_field;
 /// Implemented by the `#[salsa::tracked]` macro when applied
 /// to a struct.
 pub trait Configuration: Sized + 'static {
+    /// The debug name of the tracked struct.
     const DEBUG_NAME: &'static str;
+
+    /// The debug names of any fields.
     const FIELD_DEBUG_NAMES: &'static [&'static str];
-    const TRACKED_FIELD_DEBUG_NAMES: &'static [&'static str];
+
+    /// The absolute indices of any tracked fields.
+    const TRACKED_FIELD_INDICES: &'static [usize];
 
     /// A (possibly empty) tuple of the fields for this struct.
     type Fields<'db>: Send + Sync;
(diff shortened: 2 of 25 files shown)


examples/calc/ir.rs
```diff
@@ -76,22 +77,26 @@ pub enum Op {
 // ANCHOR: functions
 #[salsa::tracked]
 pub struct Function<'db> {
-    #[id]
     pub name: FunctionId<'db>,
 
+    #[tracked]
     name_span: Span<'db>,
```
ibraheemdev, line 83: I tried not adding this `#[tracked]` attribute because it seems unnecessary, but I actually got a test failure that I don't really understand:
```
---- type_check::fix_bad_variable_in_function stdout ----
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillExecute { database_key: parse_statements(Id(0)) } }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillExecute { database_key: type_check_program(Id(1400)) } }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillExecute { database_key: type_check_function(Id(1800)) } }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillExecute { database_key: type_check_function(Id(1801)) } }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillExecute { database_key: find_function(Id(1c00)) } }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillExecute { database_key: find_function(Id(1c01)) } }
Event: Event { thread_id: ThreadId(11), kind: DidSetCancellationFlag }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillExecute { database_key: parse_statements(Id(0)) } }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
Event: Event { thread_id: ThreadId(11), kind: WillCheckCancellation }
thread 'type_check::fix_bad_variable_in_function' panicked at /home/ibraheem/dev/sync/salsa/src/table.rs:281:9:
assertion `left == right` failed: page has hidden type `"salsa::table::Page<salsa::tracked_struct::Value<calc::ir::Function>>"` but `"salsa::table::Page<salsa::tracked_struct::Value<calc::ir::Program>>"` was expected
  left: TypeId(0xbeb3ca6f8846f92e6ae4fc7569b4bf5f)
 right: TypeId(0xbbdce2561688eb1bfc9fa5f5006a8342)
```

MichaReiser: I added some debug statements to `function::maybe_chagned_after` to log the dependencies and compared the logs between a run that succeeds and a run that fails. 

The most interesting difference is

**Successful**

```
[src/tracked_struct/tracked_field.rs:58:9] self = salsa::tracked_struct::tracked_field::FieldIngredientImpl<calc::ir::Program> {
    ingredient_index: IngredientIndex(
        13,
    ),
    field_index: 0,
}
Event: Event { thread_id: ThreadId(2), kind: WillCheckCancellation }
[src/function/maybe_changed_after.rs:154:9] &database_key_index = type_check_function(Id(1800))
[src/function/maybe_changed_after.rs:156:15] &old_memo.revisions.origin = Derived(
    QueryEdges {
        input_outputs: [
            Input(
                Function.args(Id(1000)),
            ),
            Input(
                Program(Id(1000)),
            ),
            Input(
                Span.start(Id(406)),
            ),
            Input(
                Span.end(Id(406)),
            ),
        ],
    },
)
```

**With Panic**

```
[src/tracked_struct/tracked_field.rs:58:9] self = salsa::tracked_struct::tracked_field::FieldIngredientImpl<calc::ir::Program> {
    ingredient_index: IngredientIndex(
        12,
    ),
    field_index: 0,
}
Event: Event { thread_id: ThreadId(2), kind: WillCheckCancellation }
[src/function/maybe_changed_after.rs:154:9] &database_key_index = type_check_function(Id(1800))
[src/function/maybe_changed_after.rs:156:15] &old_memo.revisions.origin = Derived(
    QueryEdges {
        input_outputs: [
            Input(
                Program(Id(1000)),
            ),
            Input(
                Program.statements(Id(1000)),
            ),
            Input(
                Span.start(Id(406)),
            ),
            Input(
                Span.end(Id(406)),
            ),
        ],
    },
)
```

The main difference is that the ingredient on which we run `maybe_changed_after` is different, AND so are the inputs. Looking at `type_check_function` then the inputs of the *panic* case make no sense. The query never calls `Program.statements` but it does call `Function.args`. I don't understand how it is happening but I suspect that we're incorrectly reusing a memo and turn a `Function` memo into a `Program` which would align with the panic

MichaReiser: Okay, I think I know what the issue is... 

You changed the tracked struct's `Configuration::create_ingredients` to only create ingredients for tracked fields. I think this change is correct. However, you call `Ingredient::Index::successor` in `tracked_field` with the `field_index` which is the absolute offset including all fields -- including untracked fields. What this means is that we now add incorrect dependencies as soon as we have a tracked struct where a `tracked` field follows an untracked field.

I hope that helps. I let you take a stab at correctly mapping the field indices. It would be great to have a unit test demonstrating the behavior and possibly an assertion somewhere that would help us catch this sooner.

MichaReiser (commented):
This looks great to me. I think the idea of `tracked` and `untracked` fields is easier to understand than `id` (I'm fairly certain we used `id` incorrectly in Red Knot). 

I think I have an understanding why the panic occurs and I added an inline comment with an explanation. I let you take a stab at fixing it.

examples/calc/ir.rs
ibraheemdev: Ah, that is a tricky. I tried a couple things here:
- Creating the ingredients with the absolute indices directly. Unfortunately this causes a panic, as some of the ingredient code relies on indices being create sequentially.
- Passing the correct relative index to `successor` in `tracked_field`. Unfortunately this also did not seem to work correctly and didn't result in the correct fields being tracked, but I'm not exactly sure why. The logging code also relies on the indices being absolute in order to print field names, so this was a little awkward.

Eventually I think the easiest solution is just creating ingredients for all fields. I added a failing test that is now passing with this fix (and also fails with the second idea I tried).

src/tracked_struct.rs
```diff
@@ -563,6 +566,25 @@ where
...
+        &'db self,
+        db: &'db dyn crate::Database,
+        s: C::Struct<'db>,
+        _field_index: usize,
+    ) -> &'db C::Fields<'db> {
+        let (zalsa, _) = db.zalsas();
+        let id = C::deref_struct(s);
+        let data = Self::data(zalsa.table(), id);
+
+        data.read_lock(zalsa.current_revision());
```
ibraheemdev, line 584: Is the `read_lock` necessary here? It seems not because `leak_fields` does not call it.

src/tracked_struct.rs
```diff
@@ -563,6 +565,25 @@ where
...
+    /// Note that this function returns the entire tuple of value fields.
+    /// The caller is responible for selecting the appropriate element.
+    pub fn untracked_field<'db>(
+        &'db self,
+        db: &'db dyn crate::Database,
+        s: C::Struct<'db>,
+        _field_index: usize,
+    ) -> &'db C::Fields<'db> {
+        let (zalsa, _) = db.zalsas();
+        let id = C::deref_struct(s);
```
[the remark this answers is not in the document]
ibraheemdev: I believe that is [here](https://github.com/salsa-rs/salsa/blob/e4d65a656fc68d0fb759b292ceae2aff2c785c5d/src/tracked_struct.rs#L299).

examples/calc/ir.rs
MichaReiser: Hmm, that's unfortunate but seems reasonable. Would it be possible to avoid creating any field ingredients if the structure only has tracked fields? Doing this optimization seems worthwhile, considering that all-fields-untracked is the new default

src/tracked_struct.rs
MichaReiser: Not a 100% but the comment on `Value::updated_at` and the inline comment inside `update` does suggest to me that the lock is required before accessing any field (handing out references to field) and it not being present in `leak_field` seems like a bug.

MichaReiser (approved):
thanks. This is great. I think we should be able to optimize the implementation to avoid creating any field-ingredients if all fields are tracked.

examples/calc/ir.rs
MichaReiser: I assume that keeping a mapping table from `field_index` to `relative_ingredient_index` (a `[Option<usize>]` where the value is `Some` for all tracked fields) is annoying to write the macro code for? If not, maybe that's worth a try. I otherwise suggest we keep it as is for now (but optimize for the no-tracked fields case)

MichaReiser:
Thanks. This looks good to me. I'll give @nikomatsakis a chance to review

nikomatsakis (commented):
This PR looks pretty good, but I'm a bit about why the indexing has to be absolute.

ibraheemdev:
I realized where I missed remapping when I tried relative indices. I think I can get it to work now.

nikomatsakis:
@ibraheemdev OK

nikomatsakis:
I think my preference would be to either create ingredients only for tracked fields or ALWAYS create ingredients no matter what. The current "create nothing or create extra ingredients" feels a bit confusing.

ibraheemdev:
I updated the implementation to only create ingredients for tracked fields. This should be good to go now.

src/tracked_struct.rs
```diff
@@ -668,7 +679,6 @@ where
         &'db self,
         db: &'db dyn crate::Database,
         s: C::Struct<'db>,
-        _field_index: usize,
```
MichaReiser: You removing the `_field_index` here makes me wonder if we should update `revisions` to only store the revisions for tracked fields because they seem wasted for untracked fields. Removing the bookkeeping for untracked fields could improve memory usage because it reduces the metadata stored for every tracked struct instance. It may also reduce the need to pass both relative and absolute indices to `tracked_field`. 

However, this might also be something that's fine to follow up in a separate PR

MichaReiser (approved):
This is great. I've one possible follow up improvement that we could make. Let me know if you want to tackle this as part of this PR or as a follow up. I otherwise think this is good to merge

MichaReiser:
I'll go ahead and merge it as is. We can follow up on the revision optimization in a separate PR

src/tracked_struct.rs
```diff
@@ -368,7 +381,7 @@ where
 
         let identity_hash = IdentityHash {
             ingredient_index: self.ingredient_index,
-            hash: crate::hash::hash(&C::id_fields(&fields)),
+            hash: crate::hash::hash(&C::untracked_fields(&fields)),
```
Veykril, line 384: Note that `IdentityHash`'s docs state

> This includes the ingredient index of that struct type plus the hash of its id fields.

So if I see this right we have flipped the fields that we hash here now right? before we hashed  the tracked fields, not hash the untracked fields which sounds wrong to me?

MichaReiser: Flipping here is correct. We should have updated the documentation. 

The identity of a tracked struct is now defined by all untracked fields. If any untracked field changes, then we just assume it's a different tracked struct. This "trick" ensures that we get away without tracking dependencies on individual fields because the id from rev `a` and rev `b` now no longer compare equal

ibraheemdev: Yeah "id fields" have effectively been renamed to "untracked fields".

PR #658, opened 2025-01-17
add release workflow

### LINKED ISSUE #489: Publish `salsa-2022` to crates.io
I've been using the `salsa-2022` crate in [my Language Server](https://github.com/Michael-F-Bryan/wit-lsp) implementation for a while now and really like it. The design with a `Db` trait, `#[salsa::tracked]`, and `crate::Jar` struct composes quite nicely and makes it easy to write clean code. 

Is there a roadmap for publishing `salsa-2022` to crates.io, either as the `salsa` crate or maybe under a temporary `salsa-2022` name?

g-plane:
This PR adds a GitHub Actions workflow for publishing crate to crates.io.

Once merged this PR, remember to :

1. Go to crates.io and generate a token.
2. Open repository settings on GitHub.
3. Select the "Actions" under the "Secrets and variables":
  image
4. Create a repository secret called `CARGO_REGISTRY_TOKEN` and paste the token into there.

To publish a new version to crates.io, just update the `version` field of Cargo.toml and create a corresponding Git tag.

Hope it helps. I'm looking forward to new versions of Salsa.

Close #489

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,16 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: cargo publish
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,16 +1,57 @@
-name: Release
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
 
 on:
   push:
-    tags:
-      - 'v*.*.*'
+    branches:
+      - master
 
 jobs:
-  release:
-    name: Release
+  # Release unpublished packages.
+  release-plz-release:
+    name: Release-plz release
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
-      - uses: actions/checkout@v4
-      - run: cargo publish
+      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


g-plane:
Done.

nikomatsakis:
Ooh, this is awesome, thanks!

nikomatsakis (changes requested):
Thanks for this. It won't work as is though because publishing salsa requires publishing both salsa and salsa-macros. Also, the two step process (edit version(s), issue git tag) is maybe a bit more complex than I'd like. 

My suggestion: let's create a [cargo xtask](https://github.com/matklad/cargo-xtask) like `cargo publish-salsa` and then modify it so that...

* `cargo publish-salsa` with no arguments dumps some help
* `cargo publish-salsa prepare` modifies `Cargo.toml` with new version and creates the git tag. It can also check that nothing is dirty in git. We should probably start maintaining release notes but that can wait.
* `cargo publish-salsa from-ci` actually does the work. Ideally it should also assert that the tag has been pushed to main so we avoid publishing things that don't exist in main. But that can be a FIXME.

g-plane:
Are they always published in same versions?

g-plane:
Should I update this PR and switch to release-plz?

MichaReiser:
@g-plane that would be great. I haven't checked if `release-plz` does everything we need but if does, then that'd be great

g-plane:
I've updated the workflow. You need to follow [release-plz's instructions](https://release-plz.dev/docs/github/quickstart) and add tokens to make it work.

MichaReiser:
Thanks. We'll have to wait for niko because only he has access to the repo's settings

nikomatsakis:
OK, the secrets are created.

nikomatsakis:
Let's....give this a try!

nikomatsakis (approved):

PR #659, opened 2025-01-21
Out of date comment in overview

pmarks:
*Like* an input, the new for a tracked struct only requires a &-reference to the database:

### DIFF
--- a/book/src/overview.md
+++ b/book/src/overview.md
@@ -163,8 +163,7 @@ struct Ast<'db> {
 }
 ```
 
-Just as with an input, new values are created by invoking `Ast::new`.
-Unlike with an input, the `new` for a tracked struct only requires a `&`-reference to the database:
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):

book/src/overview.md
```diff
@@ -163,8 +163,7 @@ struct Ast<'db> {
 }
 ```
 
-Just as with an input, new values are created by invoking `Ast::new`.
-Unlike with an input, the `new` for a tracked struct only requires a `&`-reference to the database:
+Just as with an input, new values are created by invoking `Ast::new`, and the `new` for a tracked struct only requires a `&`-reference to the database:
```
davidbarsky, line 166:
```suggestion
Just as with an input, new values are created by invoking `Ast::new`. The `new` function on a tracked struct only requires a `&`-reference to the database:
```

nikomatsakis (approved):
Thanks!

PR #660, opened 2025-01-23
Drop memos on separate thread

Veykril:
Stacked on top of https://github.com/salsa-rs/salsa/pull/684

Notably this replaces the `SeqQueue` with the standard libraries mpsc which is also implemented as a seqmented list, this simplified things a bit. We might want to consider moving to a different channel implementation for performance reasons, be it `crossbeam` or something else.


Veykril:
Whew thats not so fun

Veykril:
I removed the LRU changes from this PR that were unrelated to the eviction changes, slimming down things a bit

Veykril:
Rebased on top of https://github.com/salsa-rs/salsa/pull/673/files which was split out of this PR given its a soundness fix

MichaReiser:
It's somewhat interesting that this changes the performance so significantly (maybe just noise?)

Veykril:
The numbers seem to be consistent (judging across my rebases), so I think this might just be due to more favorable code layout due to the changes, given the only thing that really changed for the benches is one branch being removed in a hot path (none of them bench eviction/lru/dropping of the database)

Veykril:
Two more things I'd like to follow up with, setting the LRU capacity should require mutable database access. There is no reason I can see as to why one should be allowed to change the limit mid computation. That change allows to remove an atomic load. The second thing is we should be able to trigger eviction without bumping the revision. I won't include those changes here though as there is already enough in this PR.

MichaReiser:
> The second thing is we should be able to trigger eviction without bumping the revision. I won't include those changes here though as there is already enough in this PR.

Makes sense. I assume that would still require a `&mut db` for soundness?

Veykril:
Yes, otherwise there might be some computation still going on that could be borrowing from a memo.

src/table/memo.rs
```diff
@@ -22,6 +24,71 @@ pub(crate) trait Memo: Any + Send + Sync + Debug {
...
+            MemoDropAction::Drop(memo_drop) => drop(memo_drop),
+            // We were instructed to park the thread, this means we have dropped everything for the
+            // last revision.
+            MemoDropAction::Park(parker) => parker.park(),
+        });
+    })
+}
+
+pub fn memo_drop_channel() -> (MemoDropSender, MemoDropReceiver) {
+    let (tx, rx) = std::sync::mpsc::channel();
```
MichaReiser, line 61: Should we use a `sync_channel` instead to apply back pressure if the drop thread is getting overwhelmed?

What happens if a drop implementatio panics? I assume that leaves the entire database in a poisoned state because the  drop thread is now "broken"?

Veykril:
I didn't split the PR into two as the second PR would be stacked either way but I've come to the realization yesterday that the dropping change has a lot of extra implications that need to be discussed more so I'll split this PR in two later

src/table/memo.rs
Veykril: Good question regarding backpressure, unsure :)

> What happens if a drop implementatio panics? I assume that leaves the entire database in a poisoned state because the drop thread is now "broken"?

Bad things. I have actually forgotten about destructors panicking, right now that will tear down the dropper thread and cause us to panic when evicting. It should be simple enough to workaround though with a `catch_unwind`, the bigger question then though becomes if its fine for us to silently eat panic payloads in destructors ...

PR #661, opened 2025-01-23
feature: allow skipping the lifetime on `#[salsa::interned]` structs

davidbarsky:
This PR adds support for three features:

1. Writing `#[salsa::interned(no_lifetime)]` allows skipping the `'db` lifetime.
2. `#[salsa::interned]` structs now have public ingredients, which allows accessing the underlying, interned data through the `salsa::Id` alone.
3. Adds support for custom Salsa IDs. While rust-analyzer doesn't need this, their presence makes migration easier. I'd like to jettison this feature before the 1.0 release.
 
In my view, the use of the above features should be discouraged, but they are necessary for rust-analyzer to migrate to the new version of Salsa.
 
This PR supersedes https://github.com/salsa-rs/salsa/pull/632.

### DIFF
--- a/components/salsa-macro-rules/src/setup_interned_struct.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct.rs
@@ -23,6 +23,9 @@ macro_rules! setup_interned_struct {
         // optional db lifetime argument.
         db_lt_arg: $($db_lt_arg:lifetime)?,
 
+        // the salsa ID
+        id: $Id:path,
+
         // the lifetime used in the desugared interned struct.
         // if the `db_lt_arg`, is present, this is `db_lt_arg`, but otherwise,
         // it is `'static`.
@@ -69,7 +72,7 @@ macro_rules! setup_interned_struct {
         $(#[$attr])*
         #[derive(Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Hash)]
         $vis struct $Struct< $($db_lt_arg)? >(
-            salsa::Id,
+            $Id,
             std::marker::PhantomData < & $interior_lt salsa::plumbing::interned::Value <$StructWithStatic> >
         );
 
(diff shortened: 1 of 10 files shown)


components/salsa-macro-rules/src/setup_interned_struct.rs
```diff
@@ -11,9 +11,20 @@ macro_rules! setup_interned_struct {
         // Name of the struct
         Struct: $Struct:ident,
 
+        // Name of the struct data. This is a parameter because `std::concat_idents`
+        // is unstable and taking an addition dependency is unnecessary.
```
Wilfred, line 15:
```suggestion
        // is unstable and taking an additional dependency is unnecessary.
```

davidbarsky: thanks! fixed.

components/salsa-macro-rules/src/setup_interned_struct.rs
```diff
@@ -54,18 +68,32 @@ macro_rules! setup_interned_struct {
...
+                use salsa::plumbing::FromId;
+                $Struct(<$Id>::from_id(id), std::marker::PhantomData)
+            }
+            fn deref_struct(s: Self::Struct<'_>) -> salsa::Id {
+                use salsa::plumbing::AsId;
+                s.0.as_id()
+            }
+        }
+
+        type $StructDataIdent<$db_lt> = ($($field_ty,)*);
```
Veykril, line 90: Any reason this got moved up out of the `const _` scope?

tests/interned-structs.rs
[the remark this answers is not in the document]
davidbarsky: Done, fixed!

components/salsa-macro-rules/src/setup_interned_struct.rs
```diff
@@ -11,9 +11,26 @@ macro_rules! setup_interned_struct {
         // Name of the struct
         Struct: $Struct:ident,
 
+        // Name of the struct data. This is a parameter because `std::concat_idents`
+        // is unstable and taking an additional dependency is unnecessary.
+        StructData: $StructDataIdent:ident,
+
+        StructWithStatic: $StructWithStatic:ty,
```
nikomatsakis, line 18:
```suggestion
        // Name of the struct type with a `'static` argument (unless this type has no db lifetime,
        // in which case this is the same as `$Struct`)
        StructWithStatic: $StructWithStatic:ty,
```

nikomatsakis (approved):

PR #662, opened 2025-01-24
Fix accumulated values for backdated queries

Veykril:
Rebased version of https://github.com/salsa-rs/salsa/pull/641

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -1,4 +1,5 @@
 use crate::{
+    accumulator::accumulated_map::InputAccumulatedValues,
     ingredient::MaybeChangedAfter,
     key::DatabaseKeyIndex,
     zalsa::{Zalsa, ZalsaDatabase},
@@ -97,7 +98,10 @@ where
             return Some(if changed_at > revision {
                 MaybeChangedAfter::Yes
             } else {
-                MaybeChangedAfter::No(memo.revisions.accumulated_inputs.load())
+                MaybeChangedAfter::No(match &memo.revisions.accumulated {
+                    Some(_) => InputAccumulatedValues::Any,
+                    None => memo.revisions.accumulated_inputs.load(),
+                })
             });
         }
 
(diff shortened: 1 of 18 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -182,13 +202,19 @@ where
                 // valid, then some later input I1 might never have executed at all, so verifying
                 // it is still up to date is meaningless.
                 let last_verified_at = old_memo.verified_at.load();
+                let mut inputs = old_memo.revisions.accumulated_inputs.load();
```
Veykril, line 205: This had to change from defaulting to `Empty` to the memo's value due to https://github.com/salsa-rs/salsa/pull/636 as that changed the flag to track self and dependencies, instead of just dependencies. Notably this has a bigger implication though, it means a memo can never lose its `Any` accumulator status via verification, so if we the memo had a dependency that accumulated (which now no longer does) while not having accumulated itself we will still keep the flag as the memo having accumulated. (If am not making a mistake) I am not sure how bad this is, I feel like it shouldn't matter too much? It makes the optimization a bit less useful though I think this is a rather uncommon case to hit? @MichaReiser

MichaReiser: It makes the optimization less useful in an LSP case where it is very common that you have many files that had some diagnostics at some point. E.g you make edits to 20 files, each of those will have diagnostics (syntax) associated with them while making those edits but they'll ultimately converge. I rarely close my editor. That means that these "stale" accumulated states keep adding up. 

Why was it again that #636 had to change the tracking to track both self and its dependencies?

Veykril: > It makes the optimization less useful in an LSP case where it is very common to have many files that had some diagnostics at some point. For example, if you make edits to 20 files, each of those will have diagnostics (syntax) associated with it while making those edits, but they'll ultimately converge. I rarely close my editor. That means that these "stale" accumulated states keep adding up.

I would expect in the general case that if accumulations changed, that the general output of a given tracked function also changed though which will then reset the flag anyways. Your test here shows the case where that does not happen, but for that the accumulation producing run and the non accumulation producing run would need to produce the same memoized value to trigger this specific case here which does not sound like something that will happen too often imo.

> Why was it again that https://github.com/salsa-rs/salsa/pull/636 had to change the tracking to track both self and its dependencies?

Not sure, will have to dig into that whether that was accidentally or required.

MichaReiser: Good point. Although this can happen frequently with nested queries where the outer queries still produce the same result and it's only the inner queries that changed. E.g. some function local types changed, but the function return type remains unchanged.

Veykril: Restored the tracking behavior, it was accidental.

MichaReiser (approved):
Nice and thanks for rebasing

Issue #669, opened 2025-02-05
LRU drops memos before revision bump


Veykril:
This can result in dangling references to deallocated memos within a revision

PR #673, opened 2025-02-05
Fix LRU dropping memos too early

Veykril:
Fixes https://github.com/salsa-rs/salsa/issues/669

### DIFF
--- a/src/function/delete.rs
+++ b/src/function/delete.rs
@@ -18,8 +18,8 @@ impl<C: Configuration> Default for DeletedEntries<C> {
 }
 
 impl<C: Configuration> DeletedEntries<C> {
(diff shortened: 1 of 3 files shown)


MichaReiser (approved):

MichaReiser:
Huh, there's a LRU test failure now

Veykril:
Ahaha, well the test checks for how many entries are alive via global counter that is decreased on drop. But now we no longer drop them eagerly. So the test is actually wrong in that regard. (my other PR also had to change that)

Veykril:
Actually not quite, it was trying to trigger a new revision by creating an input but nowadays that actually doesn't trigger one (only setting an input does)

Issue #610, opened 2024-10-28
Returning `&` from a tracked function


nikomatsakis:
I noticed that code like 

```rust
#[salsa::input]
struct Input {
    #[return_ref]
    data: Vec<String>
}

#[salsa::tracked]
fn test<'db>(db: &'db dyn crate::Database, input: Input) -> &'db String {
    input.data(db).first().unwrap()
}
```

...will compile, but I don't think that's sound. We should be requiring the `Update` trait on the return value.

Filing this issue so I don't forget to think about it. =)

PR #674, opened 2025-02-05
Require `Update` trait for tracked function return values

MichaReiser:
Fixes https://github.com/salsa-rs/salsa/issues/610 by requiring the `Output` of tracked functions to implement `Update`

Let's see how annoying this will be...

### DIFF
--- a/src/update.rs
+++ b/src/update.rs
@@ -296,6 +296,10 @@ where
     unsafe fn maybe_update(old_pointer: *mut Self, new_arc: Self) -> bool {
         let old_arc: &mut Arc<T> = unsafe { &mut *old_pointer };
 
+        if Arc::ptr_eq(old_arc, &new_arc) {
+            return false;
+        }
+
         if let Some(inner) = Arc::get_mut(old_arc) {
             match Arc::try_unwrap(new_arc) {
                 Ok(new_inner) => T::maybe_update(inner, new_inner),
--- a/src/update.rs
+++ b/src/update.rs
@@ -2,6 +2,7 @@ use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet},
     hash::{BuildHasher, Hash},
     path::PathBuf,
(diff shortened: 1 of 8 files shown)


src/update.rs
```diff
@@ -288,6 +289,29 @@ where
     }
 }
 
+unsafe impl<T> Update for Arc<T>
```
MichaReiser, line 292: I don't feel too confident about the `Update` implementations here. I never fully understood how `Update` is supposed to work...

Veykril: Agreed, the docs on the trait are not that clear unfortunately, Niko clearly tried making them clear but its still a bit difficult to understand. From what I understand (by reading the existing implementations) we likely want to "peel" away one layer of types and delegate the `maybe_update` down?

src/update.rs
```diff
@@ -288,6 +289,29 @@ where
...
+where
+    T: Update,
+{
+    unsafe fn maybe_update(old_pointer: *mut Self, new_arc: Self) -> bool {
```
Veykril, line 296: I wonder if it would make sense to do a ptr equality check at the start and return `false` if they are equal. That probably is a very unlikely scenario (usually you will compute an entirely new arc allocation I believe?).

Veykril (approved):
The impls do look correct to me from what I underestood about this.

src/update.rs
MichaReiser: Yeah, that's the pattern I tried to follow. I'm not sure if it's worth doing this entire ceremony for `Arc` by trying to update them in place but it seemed reasonable to do?

MichaReiser:
I'm sure this will get sort of painful. We'll probably need to add many implementations (e.g. we use `SmallVec`.

Veykril:
A shame we can't use the `Fallback` trick with trait bounds then we wouldn't be needing the `Update` derive in the majority of cases ...

PR #676, opened 2025-02-06
feature: expose database contents as inherent method on Ingredients

davidbarsky:
This PR redesigns the functionality introduced in #640 so that the dump-all-database contents methods are on the Salsa struct ingredient instead of the database's `Storage`. I think this is an improvement for two reasons:
1. This design is closer to the original [`DebugQueryTable`](https://docs.rs/salsa/0.16.1/salsa/debug/trait.DebugQueryTable.html) trait.
2. At least within rust-analyzer, it's not feasible to get access to the underlying storage from methods that only have a `&dyn SomeDatabase`. However, it _is_ feasible to access the ingredient

## Example

Here's an example for interned structs (input and tracked structs have an identical API): 

```rust
#[salsa::interned]
struct InternedStruct<'db> {
    name: String,
}

let _ = InternedStruct::new(&db, "Hello, world!".to_string());

let interned = InternedStruct::ingredient(&db)
    .entries(&db)
    .collect::<Vec<_>>();
```

Note that the `entries` method is defined on the `{interned, input, tracked}::IngredientImpl`. The more optimal API would be to define `entries` on the Salsa struct itself, alongside `ingredient`. However, this is substantially more annoying to feature-flag through macros.

## Notes

While I don't have a good idea how to access a tracked function's memos, I also don't really know how much it makes sense to do so outside of some migrations.

### DIFF
--- a/src/input.rs
+++ b/src/input.rs
@@ -184,6 +184,20 @@ impl<C: Configuration> IngredientImpl<C> {
         &value.fields
     }
 
+    #[cfg(feature = "salsa_unstable")]
+    /// Returns all data corresponding to the input struct.
+    pub fn entries<'db>(
+        &'db self,
+        db: &'db dyn crate::Database,
+    ) -> impl Iterator<Item = &'db Value<C>> {
+        db.zalsa()
+            .table()
+            .pages
+            .iter()
+            .filter_map(|page| page.cast_type::<crate::table::Page<Value<C>>>())
+            .flat_map(|page| page.slots())
+    }
+
     /// Peek at the field values without recording any read dependency.
     /// Used for debug printouts.
     pub fn leak_fields<'db>(&'db self, db: &'db dyn Database, id: C::Struct) -> &'db C::Fields {
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -253,6 +253,20 @@ where
         self.data(db, C::deref_struct(s))
     }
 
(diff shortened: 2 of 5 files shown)


MichaReiser:
Exposing the methods on the ingredient does make sense to me. 

> While I don't have a good idea how to access a tracked function's memos, I also don't really know how much it makes sense to do so outside of some migrations.

Not saying we have to do this as part of this PR but having the same functionality for queries seems useful because most queries (e.g. the source text or parsed AST) in Ruff aren't tracked structs or inputs but getting a count of alive values would be a useful debug statistics.

Veykril (approved):

davidbarsky:
> Exposing the methods on the ingredient does make sense to me. 

How would you feel about exposing this on the Salsa struct itself once this stops exposing the underlying storage mechanism?

> > While I don't have a good idea how to access a tracked function's memos, I also don't really know how much it makes sense to do so outside of some migrations.
> 
> 
> 
> Not saying we have to do this as part of this PR but having the same functionality for queries seems useful because most queries (e.g. the source text or parsed AST) in Ruff aren't tracked structs or inputs but getting a count of alive values would be a useful debug statistics. 

Ah, I naively assumed that most queries in Ruff were Salsa structs. In that case, we should add this functionality to tracked functions, but I'd rather do that in a followup.

Issue #578, opened 2024-09-26
Idea: salsa::supertype and upcasting on `Id`


nikomatsakis:
I've noticed that I often have an enum whose variants are all different salsa structs:

```rust
enum Item<'db> {
    Struct(Struct<'db>),
    Function(Function<'db>),
    ...
}
```

I was thinking about how this enum will use 2 words and not 1 and I realized that, given that each variant is just storing an `Id`, and that we can go from an `Id` to an `IngredientIndex` (it's stored on the page), we could have a proc-macro that takes such an enum and returns a special struct that just stores a single `Id`. It could then be converted into an enum readily enough. You also downcast it.

I'm imagining the following:

First, add two methods to `salsa::Id`, `fn is<S>(self, db: &dyn Database) -> bool` and `fn downcast<S>(self, db: &dyn Database) -> Option<S>`, where `S: SalsaStructInDb<'_>`. These methods can lookup the id in `db.table`, identify the ingredient index on the from the `Page`, compare that to the ingredient index for `S`, and then do the appropriate thing. (We might also want to tweak the `FromId` or other traits in a similar-ish way.)

Next, create a proc macro named `supertype` (bikeshed) which expects an enum with one lifetime argument and where each variant has a single argument:

```rust
#[salsa::supertype]
enum Foo<'db> {
    A(Bar<'db>),
    ...
    Z(Baz<'db>),
}
```

we generate the following:

```rust
// A struct to store the id itself
#[derive(Copy, Clone, ...)]
struct Foo<'db> { id: Id }

// Original enum, but with a tweaked name (should be configurable)
enum FooVariants<'db> {
    A(Bar<'db>),
    ...
    Z(Baz<'db>),
}

// From impls to "upcast" into the struct; these just take the `id`
impl<'db> From<Bar<'db>> for Foo<'db> { ... }

// From impls to "upcast" into the enum; these pick the appropriate variant
impl<'db> From<Bar<'db>> for FooVariants<'db> { ... }

// From impls from the variants struct
impl<'db> From<FooVariants<'db>> for Foo<'db> { ... }

impl Foo<'db> {
    // A method to convert from the struct into the variants 
    fn variants(self, db: &dyn salsa::Database) -> FooVariants<'db> {
        if let Some(v) = self.id.downcast::<Foo<'db>>(db) {
            return v.into();
        }

        // ...

        if let Some(v) = self.id.downcast::<Bar<'db>>(db) {
            return v.into();
        }

        unreachable!()
[...]

PR #677, opened 2025-02-06
feature: add `#[salsa::supertype]`

davidbarsky:
This PR implements https://github.com/salsa-rs/salsa/issues/578 and is (currently) stacked atop of https://github.com/salsa-rs/salsa/pull/676. All authorship/credit for this work goes to @ChayimFriedman2 for the implementation, I largely renamed some stuff.

Two notes on this PR:
1. There's no corresponding `macro_rules!` macro for `#[salsa::supertype]`. We can add it, need be.
2. I've commented out some hand-written `PartialEq` implementations because they caused a bunch of test failures in rust-analyzer. Those test failures stemmed from rust-analyzer's native, non-Salsa powered interner, which persists even after we changed databases in tests! For more details, see what [Chayim wrote on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Porting.20to.20Salsa.203.2E0/near/494538002). Instead of commenting things out, I'll make these options on the attribute macro.

### DIFF
--- a/src/function/fetch.rs
+++ b/src/function/fetch.rs
@@ -1,4 +1,5 @@
 use super::{memo::Memo, Configuration, IngredientImpl};
+use crate::zalsa::MemoIngredientIndex;
 use crate::{
     accumulator::accumulated_map::InputAccumulatedValues, runtime::StampedValue,
     zalsa::ZalsaDatabase, AsDynDatabase as _, Id,
@@ -40,17 +41,26 @@ where
         db: &'db C::DbView,
         id: Id,
     ) -> &'db Memo<C::Output<'db>> {
+        let memo_ingredient_index = self.memo_ingredient_index(db.zalsa(), id);
         loop {
-            if let Some(memo) = self.fetch_hot(db, id).or_else(|| self.fetch_cold(db, id)) {
+            if let Some(memo) = self
+                .fetch_hot(db, id, memo_ingredient_index)
+                .or_else(|| self.fetch_cold(db, id, memo_ingredient_index))
+            {
                 return memo;
             }
         }
     }
 
     #[inline]
-    fn fetch_hot<'db>(&'db self, db: &'db C::DbView, id: Id) -> Option<&'db Memo<C::Output<'db>>> {
+    fn fetch_hot<'db>(
+        &'db self,
+        db: &'db C::DbView,
+        id: Id,
+        memo_ingredient_index: MemoIngredientIndex,
+    ) -> Option<&'db Memo<C::Output<'db>>> {
         let zalsa = db.zalsa();
-        let memo_guard = self.get_memo_from_table_for(zalsa, id);
+        let memo_guard = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
         if let Some(memo) = &memo_guard {
             if memo.value.is_some()
                 && self.shallow_verify_memo(db, zalsa, self.database_key_index(id), memo)
@@ -63,7 +73,12 @@ where
         None
     }
 
-    fn fetch_cold<'db>(&'db self, db: &'db C::DbView, id: Id) -> Option<&'db Memo<C::Output<'db>>> {
+    fn fetch_cold<'db>(
+        &'db self,
+        db: &'db C::DbView,
+        id: Id,
+        memo_ingredient_index: MemoIngredientIndex,
+    ) -> Option<&'db Memo<C::Output<'db>>> {
         let (zalsa, zalsa_local) = db.zalsas();
         let database_key_index = self.database_key_index(id);
 
@@ -72,7 +87,7 @@ where
             db.as_dyn_database(),
             zalsa_local,
             database_key_index,
-            self.memo_ingredient_index,
+            memo_ingredient_index,
         )?;
 
         // Push the query on the stack.
@@ -80,7 +95,7 @@ where
 
         // Now that we've claimed the item, check again to see if there's a "hot" value.
         let zalsa = db.zalsa();
-        let opt_old_memo = self.get_memo_from_table_for(zalsa, id);
+        let opt_old_memo = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
         if let Some(old_memo) = &opt_old_memo {
             if old_memo.value.is_some() && self.deep_verify_memo(db, old_memo, &active_query) {
                 // Unsafety invariant: memo is present in memo_map and we have verified that it is
--- a/src/ingredient.rs
+++ b/src/ingredient.rs
@@ -27,7 +27,7 @@ pub trait Jar: Any {
     where
         Self: Sized,
     {
-        IngredientIndices::uninitialized()
+        IngredientIndices::empty()
     }
 
     /// Create the ingredients given the index of the first one.
--- a/src/ingredient.rs
+++ b/src/ingredient.rs
@@ -6,8 +6,9 @@ use std::{
 use crate::{
     accumulator::accumulated_map::{AccumulatedMap, InputAccumulatedValues},
     cycle::CycleRecoveryStrategy,
+    plumbing::IngredientIndices,
     table::Table,
-    zalsa::{IngredientIndex, MemoIngredientIndex},
+    zalsa::{IngredientIndex, Zalsa},
     zalsa_local::QueryOrigin,
     Database, DatabaseKeyIndex, Id,
 };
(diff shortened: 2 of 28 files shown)


Veykril:
Of note is that this implements a different design of https://github.com/salsa-rs/salsa/issues/578, this design here having the benefit of retaining the enum being an enum (where as the issue would've transformed it into just another id that you'd need to access variants via methods).

Veykril:
> I've commented out some hand-written PartialEq implementations because they caused a bunch of test failures in rust-analyzer. Those test failures stemmed from rust-analyzer's native, non-Salsa powered interner, which persists even after we changed databases in tests! For more details, see what [Chayim wrote on Zulip](https://rust-lang.zulipchat.com/#narrow/channel/185405-t-compiler.2Frust-analyzer/topic/Porting.20to.20Salsa.203.2E0/near/494538002). Instead of commenting things out, I'll make these options on the attribute macro.

Hmm that sounds problematic, I don't see how adding an option to the macro fixes this? Are you saying we'd want to opt out of those impls in rust-analyzer until we replaced the type interner with salsa (by using the new trait solver)?

davidbarsky:
> Are you saying we'd want to opt out of those impls in rust-analyzer until we replaced the type interner with salsa (by using the new trait solver)?

Yes. In the meantime, Jack has also asked us to avoid making changes to the interner, as his changes depend on it being in place until Salsa's interned structs can natively handle/represent a `TyKind`.

src/salsa_struct.rs
```diff
@@ -1,5 +1,52 @@
...
+    /// It could ask each variant "is this value yours?" and then invoke [`FromId`][crate::id::FromId]
+    /// with the correct variant, but that will duplicate the work: now `InnerEnum` will have to do
+    /// the same thing for its variants.
+    ///
+    /// Instead, we keep track of the [`TypeId`] of the ID struct, and ask each variant to "cast" it. If
+    /// it succeeds, we return that value; if not, we go to the next variant.
+    ///
+    /// Why `TypeId` and not `IngredientIndex`? Because it's cheaper and easier. The `TypeId` is readily
+    /// available at compile time, while the `IngredientIndex` requires a runtime lookup.
+    fn cast(id: Id, type_id: TypeId) -> Option<Self>;
```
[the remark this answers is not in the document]
ChayimFriedman2: Yes. I thought this won't be a problem, because the use-case seems dubious.

Veykril:
> > Are you saying we'd want to opt out of those impls in rust-analyzer until we replaced the type interner with salsa (by using the new trait solver)?
> 
> Yes. In the meantime, Jack has also asked us to avoid making changes to the interner, as his changes depend on it being in place until Salsa's interned structs can natively handle/represent a `TyKind`.

Understandable but won't this have effects on the incrementality of rust-analyzer?

src/salsa_struct.rs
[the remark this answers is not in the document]
Veykril: Yes, I don't think there is any point in having a differing behavior here, just wanted to clarify. So you set analogy makes sense.

Veykril:
Propopsing a different store scheme https://github.com/davidbarsky/salsa/pull/1 that wastes a tiny bit of memory for faster `IngredientIndex` to `MemoIngredientIndex` mapping

src/ingredient.rs
```diff
@@ -16,40 +17,33 @@ use super::Revision;
 /// A "jar" is a group of ingredients that are added atomically.
 /// Each type implementing jar can be added to the database at most once.
 pub trait Jar: Any {
+    /// This creates the ingredient dependencies of this jar. We need to split this from `create_ingredients()`
+    /// because while `create_ingredients()` is called, a lock on the ingredient map is held (to guarantee
+    /// atomicity), so other ingredients could not be created.
+    ///
+    /// Only tracked fns use this.
+    fn create_dependencies(_zalsa: &Zalsa) -> IngredientIndices
```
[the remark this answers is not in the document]
davidbarsky: wait, what'd you call this instead?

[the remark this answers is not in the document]
Veykril: Well they are dependencies, but a different kind, unsure about a better name though

davidbarsky:
Okay, I think this is ready for review. I've addressed Lukas's comments, but I've decided to punt on the derive/equality stuff  until later—I'd have to rewrite this derive as an attribute macro and I don't really want to do that right now. I've opened https://github.com/salsa-rs/salsa/issues/694 to track this.

MichaReiser:
Do we understand why this decreases performance by as much as 10%?

davidbarsky:
> So maybe this is something that has come up or has been discussed before but do we understand why this decreases performance by as much as 10%? I'm a bit surprised by this because none of the existing benchmarks use this feature, right?

I'm a bit surprised as well. The most recent runs ([16 hours ago](https://codspeed.io/salsa-rs/salsa/runs/compare/67b3aa36f68edeee7a82cbbe..67b3e9b7026d2efabccfe3f5), as of me writing this comment) shows that 10% regression on `amortized[InternedInput]`, whereas older runs [from 11 days ago](https://codspeed.io/salsa-rs/salsa/runs/compare/67b3aa36f68edeee7a82cbbe..67a67ae90371e701f20a6172) show a far more minor 1% decrease in performance on the same benchmark. However, the changes from *12* days ago show a [6% decrease](https://codspeed.io/salsa-rs/salsa/runs/compare/67b3aa36f68edeee7a82cbbe..67a53fa791813e464c045247) in performance.

This is a somewhat long-winded of saying "no, not really", as it wasn't really on my radar until you pointed it out. I don't think those regression are pure noise, especially since we've added `src/memo_ingredient_indices.rs` and introduced a bit of indirection with the `FromIdWithDb`, where the latter now requires callers to pass in a database alongside the ID.

davidbarsky:
Frustratingly, I can't seem to get *all* changes the that I've force-pushed https://github.com/salsa-rs/salsa/pull/684, but I recall rebasing last week over https://github.com/salsa-rs/salsa/pull/673 and https://github.com/salsa-rs/salsa/pull/684. I wonder if the changes introduced  by those soundness fixes interacted suboptimally with the changes in _this_ PR: I remember thinking a bunch as to how to resolve those conflicts appropriately.

davidbarsky:
If I were to guess—and again, this is a guess that I haven't really dug into properly—the changes to `reset_for_new_revision` in https://github.com/salsa-rs/salsa/pull/684/ and PR resulted in more substantial regression than I otherwise expected.

MichaReiser:
> If I were to guess—and again, this is a guess that I haven't really dug into properly—the changes to `reset_for_new_revision` in #684 and PR resulted in more substantial regression than I otherwise expected.

You mean that specific PR or that your PR is now slower because it doesn't play well with the changes made in #684?

davidbarsky:
> > If I were to guess—and again, this is a guess that I haven't really dug into properly—the changes to `reset_for_new_revision` in #684 and PR resulted in more substantial regression than I otherwise expected.
> 
> 
> 
> You mean that specific PR or that your PR is now slower because it doesn't play well with the changes made in #684?

The latter. Alone, the changes seem fine, but together, we get a ~10% regression. I don't think we can or should revert #684, I just think the current state is fixable bad luck.

src/function/fetch.rs
```diff
@@ -40,17 +41,26 @@ where
         db: &'db C::DbView,
         id: Id,
     ) -> &'db Memo<C::Output<'db>> {
+        let memo_ingredient_index = self.memo_ingredient_index(db.zalsa(), id);
```
ChayimFriedman2, line 44: If I had to guess, the perf regressions are from here, because of calling `zalsa()` on `dyn Database` twice (we call it in `fetch_hot()` too), and from `Zalsa::ingredient_index()`, because we call the `dyn Page` to retrieve the ingredient.

If I'm right, it is possible to fix them, by giving the zalsa we use here to `fetch_hot()`, and batching the page calls, by delaying `memo_ingredient_index()`, and instead adding a `Page::ingredient_index_and_memos()` method that combines `Page::memos()` and `Page::ingredient_index()`, and call it from `Zalsa::memo_table_for()`.

davidbarsky: @ChayimFriedman2 If I understand you correctly, you're saying that there are two things that we can do:

1. Batch up calls to `zalsa` in `src/function/fetch.rs`. I did this locally and this helped marginally in my local benchmarks, but it could also just be noise.
2. Combine calls to `ingredient_index ` until the _latest_ possible point. I'm having difficulty understanding _this_ part.

On the second point, it seems like the only case where this would be useful is in `src/function.rs`' `IngredientImpl::reset_for_new_revision`:

```rust
fn reset_for_new_revision(&mut self, table: &mut Table) {
    self.lru.for_each_evicted(|evict| {
        let ingredient_index = table.ingredient_index(evict);
        self.evict_value_from_memo_for(table.memos_mut(evict), ingredient_index);
    });
    std::mem::take(&mut self.deleted_entries);
}
```

davidbarsky: Oh, I see what you mean now. For whatever reason, I was searching for calls to `ingredient_index` instead of `get_memo_from_table_for`. That makes far more sense.

Veykril:
regressions are down to 5% worst now, I assume you merely rebased?

davidbarsky:
> regressions are down to 5% worst now, I assume you merely rebased?

Yeah, I just rebased, which notably captured the boxcar changes.

Veykril:
Opened a PR to your branch to add a small benchmark that exerts the supertype stuff https://github.com/davidbarsky/salsa/pull/3

src/function/fetch.rs
Veykril: We actually unnecessarily call zalsa here for non supertype indices (its unused in that path) so we can be lazy here. I'll fix that up as a follow up to this PR though. Imo we should rebase this now and then merge it to make progress.

davidbarsky: Sounds good. That would get things closer to an alpha.1 release and gets us closer to unblocking @carljm from landing his fixpoint iteration changes.

@MichaReiser Are you okay with us landing this and fixing the regressions after (within, say, a day)?

Veykril: I rebased this PR in https://github.com/salsa-rs/salsa/pull/727 and added the proposed change, no regressions anymore

PR #671, opened 2025-02-05
Impl `HashEqLike<T>` for `&T`

Veykril:
Fixes https://github.com/salsa-rs/salsa/issues/670

PR #678, opened 2025-02-07
Implement `HashEqLike<&T>` for `T`

ibraheemdev:
It looks like the fix in https://github.com/salsa-rs/salsa/pull/671 was the wrong way around, I should have confirmed that it fixed the regression.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -376,11 +376,13 @@ pub trait HashEqLike<O> {
 pub trait Lookup<O> {
     fn into_owned(self) -> O;
 }
+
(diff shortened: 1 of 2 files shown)


MichaReiser (approved):
Thanks. Would you mind adding a test to https://github.com/MichaReiser/salsa/blob/205b70e79734d4df721c23901508376b502308aa/tests/interned-structs.rs#L95-L106 that demonstrates the new impl. A test would help us prevent future regressions like this (and validate that we know implement the trait for the right types ;)

PR #679, opened 2025-02-08
Replace all uses of `AtomicCell` with appropriate atomic primitives

Veykril:
I believe Ibrahim talked about this and it seemed reasonable to me to do. This just carries over the orderings that atomic cell uses for now (with one exception). We also drop the crossbeam dependency now, replacing it with `crossbeam-queue` and `crossbeam-channel`.

### DIFF
--- a/src/accumulator/accumulated_map.rs
+++ b/src/accumulator/accumulated_map.rs
@@ -58,7 +58,7 @@ impl Clone for AccumulatedMap {
 ///
 /// Knowning whether any input has accumulated values makes aggregating the accumulated values
 /// cheaper because we can skip over entire subtrees.
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 pub enum InputAccumulatedValues {
     /// The query nor any of its inputs have any accumulated values.
     #[default]
@@ -119,5 +119,22 @@ impl AtomicInputAccumulatedValues {
         } else {
             InputAccumulatedValues::Empty
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn atomic_input_accumulated_values() {
+        let val = AtomicInputAccumulatedValues::new(InputAccumulatedValues::Empty);
+        assert_eq!(val.load(), InputAccumulatedValues::Empty);
+        val.store(InputAccumulatedValues::Any);
+        assert_eq!(val.load(), InputAccumulatedValues::Any);
+        let val = AtomicInputAccumulatedValues::new(InputAccumulatedValues::Any);
+        assert_eq!(val.load(), InputAccumulatedValues::Any);
+        val.store(InputAccumulatedValues::Empty);
+        assert_eq!(val.load(), InputAccumulatedValues::Empty);
     }
 }
--- a/src/accumulator/accumulated_map.rs
+++ b/src/accumulator/accumulated_map.rs
@@ -1,4 +1,7 @@
-use std::ops;
+use std::{
+    ops,
+    sync::atomic::{AtomicBool, Ordering},
+};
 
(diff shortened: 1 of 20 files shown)


MichaReiser (commented):
@ibraheemdev would you be interested in reviewing this PR? You are way more knowledged when it comes to concurrency than I.

src/input/singleton.rs
```diff
@@ -1,52 +1,56 @@
...
         let id = cb();
-        self.index.store(Some(id));
-        drop(_guard);
+        if self
+            .index
+            .compare_exchange(0, id.as_u32() + 1, Ordering::SeqCst, Ordering::SeqCst)
+            .is_err()
+        {
+            panic!("singleton struct may not be duplicated");
+        }
```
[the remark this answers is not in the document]
Veykril: There is no need to, if we end up calling this twice we want to panic so there is no need to retry

src/accumulator/accumulated_map.rs
```diff
@@ -91,3 +94,30 @@ impl ops::BitOrAssign for InputAccumulatedValues {
         *self = *self | rhs;
     }
 }
+
+#[derive(Debug, Default)]
+pub struct AtomicInputAccumulatedValues(AtomicBool);
```
davidbarsky, line 99: so, this is a slightly nit-picking comment, but do you mind writing a unit test or for this one? the reason i ask is that my eyes kinda glaze over the contents and having some usage examples by the way of unit tests would be nice.

src/revision.rs
```diff
@@ -55,10 +67,62 @@ impl AtomicRevision {
...
+        }
     }
 
     pub(crate) fn store(&self, r: Revision) {
         self.data.store(r.as_usize(), Ordering::SeqCst);
     }
 }
+
+#[derive(Debug)]
+pub(crate) struct OptionalAtomicRevision {
```
davidbarsky, line 83: same note about unit testing this one as well, fwiw.

carljm:
> This does just carry over the SeqCst ordering for now as to not change behavior

Curious about this, [AtomicCell docs](https://docs.rs/crossbeam/latest/crossbeam/atomic/struct.AtomicCell.html) say it uses Acquire/Release ordering, not SeqCst?

Veykril:
Huh it does, where did I pick up the `SeqCst` from then 😕

Veykril:
Okay fixed the orderings, thanks for double checking

src/input/singleton.rs
```diff
@@ -1,52 +1,56 @@
...
+    fn with_scope(&self, cb: impl FnOnce() -> Id) -> Id {
+        if self.index.load(Ordering::Acquire) != 0 {
             panic!("singleton struct may not be duplicated");
         }
         let id = cb();
-        self.index.store(Some(id));
-        drop(_guard);
+        if self
+            .index
+            .compare_exchange(0, id.as_u32() + 1, Ordering::AcqRel, Ordering::Acquire)
```
ibraheemdev, line 26: Why do we have to add 1 here? `id` is a `NonZeroU32` right?

ibraheemdev (approved):

src/input/singleton.rs
ibraheemdev: Oh I see that `as_u32` subtracts 1, that feels a little weird.

src/input/singleton.rs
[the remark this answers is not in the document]
ibraheemdev: We could also use a `OnceLock` here, but this seems fine as well.

src/input/singleton.rs
Veykril: The `Id` from/to `u32` conversion methods increment/decrement by 1 so we undo that here such that we can keep using `0` as `None`. We could also not do that and use `u32::MAX` as `None`, I have no strong preference here

MichaReiser (approved):

PR #680, opened 2025-02-08
Use *fallback* trick for tracked-fn `Update` constraint

MichaReiser:
Use the Fallback trick similar to what the `derive(Update)` does to constrain the tracked-fn's return type to `Update`. 

This removes the need to implement `Update` for types with a `'static` lifetime.

I also added a few `Update` implementations that are needed for Ruff. The not good news is, Ruff now panics with some assertions and I haven't figured out the root cause yet. 

Edit: Okay, the panics go away if I back out of the coarse-grained dependency changes... So something goes wrong there, uff.

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -55,6 +55,8 @@ macro_rules! setup_tracked_fn {
         // True if we `return_ref` flag was given to the function
         return_ref: $return_ref:tt,
 
+        maybe_update_fn: {$($maybe_update_fn:tt)*},
+
         // Annoyingly macro-rules hygiene does not extend to items defined in the macro.
         // We have the procedural macro generate names for those items that are
         // not used elsewhere in the user's code.
@@ -150,14 +152,8 @@ macro_rules! setup_tracked_fn {
             ///
             /// # Safety
             /// The same safety rules as for `Update` apply.
-            unsafe fn _implements_update<'db>(old_pointer: *mut $output_ty, new_value: $output_ty) -> bool {
-                unsafe {
-                    use $zalsa::UpdateFallback;
-                    $zalsa::UpdateDispatch::<$output_ty>::maybe_update(
-                        old_pointer, new_value
-                    )
-                }
-            }
+            $($maybe_update_fn)*
+
 
             impl $zalsa::function::Configuration for $Configuration {
                 const DEBUG_NAME: &'static str = stringify!($fn_name);
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -145,6 +145,20 @@ macro_rules! setup_tracked_fn {
                 }
             }
 
+            /// This method isn't used anywhere. It only exitst to enforce the `Self::Output: Update` constraint
+            /// for types that aren't `'static`.
+            ///
+            /// # Safety
+            /// The same safety rules as for `Update` apply.
+            unsafe fn _implements_update<'db>(old_pointer: *mut $output_ty, new_value: $output_ty) -> bool {
+                unsafe {
+                    use $zalsa::UpdateFallback;
+                    $zalsa::UpdateDispatch::<$output_ty>::maybe_update(
+                        old_pointer, new_value
+                    )
+                }
+            }
+
             impl $zalsa::function::Configuration for $Configuration {
                 const DEBUG_NAME: &'static str = stringify!($fn_name);
 
--- a/components/salsa-macros/src/update.rs
+++ b/components/salsa-macros/src/update.rs
@@ -1,4 +1,5 @@
 use proc_macro2::{Literal, TokenStream};
+use syn::spanned::Spanned;
 use synstructure::BindStyle;
 
 use crate::hygiene::Hygiene;
@@ -34,7 +35,7 @@ pub(crate) fn update_derive(input: syn::DeriveInput) -> syn::Result<TokenStream>
                 .bindings()
                 .iter()
                 .fold(quote!(), |tokens, binding| quote!(#tokens #binding,));
-            let make_new_value = quote! {
+            let make_new_value = quote_spanned! {variant.ast().ident.span()=>
                 let #new_value = if let #variant_pat = #new_value {
                     (#make_tuple)
                 } else {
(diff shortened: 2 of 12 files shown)


components/salsa-macros/src/update.rs
```diff
@@ -52,14 +52,18 @@ pub(crate) fn update_derive(input: syn::DeriveInput) -> syn::Result<TokenStream>
...
 
+                    let update_field = quote_spanned! {
+                        binding.binding.span() =>
+                        unsafe {
+                            salsa::plumbing::UpdateDispatch::<#field_ty>::maybe_update(
+                                #binding,
+                                #new_value.#field_index,
+                            )
+                        }
+                    };
```
MichaReiser, line 63: I hoped that this would give us better compiler errors that point to the relevant field but it seems that using spanned here doesn't help

tests/compile-fail/tracked_fn_return_ref.stderr
```diff
@@ -6,37 +6,24 @@ warning: unused import: `salsa::Update`
...
+error: lifetime may not live long enough
+  --> tests/compile-fail/tracked_fn_return_ref.rs:15:1
    |
-16 | fn tracked_fn_return_ref<'db>(db: &'db dyn Db, input: MyInput) -> &'db str {
-   |                                                                   ^^^^^^^^ the trait `Update` is not implemented for `&'db str`
+15 | #[salsa::tracked]
+   | ^^^^^^^^^^^^^^^^^
+   | |
+   | lifetime `'db` defined here
+   | requires that `'db` must outlive `'static`
```
MichaReiser, line 16: The errors are, unfortunately, not very helpful

src/update.rs
```diff
@@ -259,7 +283,7 @@ macro_rules! maybe_update_map {
 
 unsafe impl<K, V, S> Update for HashMap<K, V, S>
 where
-    K: Update + Eq + Hash,
+    K: Eq + Hash,
```
Veykril, line 286: I'd expect this bound to be necessary as otherwise we could store non update keys in the map? That is `HashMap<&'db (), ()>`

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -158,6 +158,15 @@ macro_rules! setup_tracked_fn {
...
                 const CYCLE_STRATEGY: $zalsa::CycleRecoveryStrategy = $zalsa::CycleRecoveryStrategy::$cycle_recovery_strategy;
 
+                unsafe fn maybe_update<'db>(old_pointer: *mut Self::Output<'db>, new_value: Self::Output<'db>) -> bool {
+                    unsafe {
+                        use $zalsa::UpdateFallback;
+                        salsa::plumbing::UpdateDispatch::<Self::Output<'db>>::maybe_update(
+                            old_pointer, new_value
+                        )
+                    }
+                }
```
Veykril, line 168: Given this is not called and just used as an comptime assertion, there isn't really a reason for us to make this a trait method is there? We can just emit it as a private inherent function here

src/update.rs
MichaReiser: Oh right. It's sort of annoying because we can't use the "Fallback" trick for containers. Or I just don't see how

Veykril: The fallback should apply right? I guess the issue is that we won't do a fine grained update on the elements within but instead on the thing as a whole?

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -158,6 +158,15 @@ macro_rules! setup_tracked_fn {
 
                 const CYCLE_STRATEGY: $zalsa::CycleRecoveryStrategy = $zalsa::CycleRecoveryStrategy::$cycle_recovery_strategy;
 
+                unsafe fn maybe_update<'db>(old_pointer: *mut Self::Output<'db>, new_value: Self::Output<'db>) -> bool {
```
Veykril, line 161:
```suggestion
                unsafe fn maybe_update<$db_lt>(old_pointer: *mut Self::Output<$db_lt>, new_value: Self::Output<$db_lt>) -> bool {
```

This slightly improves things, but not much

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -158,6 +158,15 @@ macro_rules! setup_tracked_fn {
...
+                    unsafe {
+                        use $zalsa::UpdateFallback;
+                        salsa::plumbing::UpdateDispatch::<Self::Output<'db>>::maybe_update(
```
Veykril, line 164:
```suggestion
                        $zalsa::UpdateDispatch::<Self::Output<'db>>::maybe_update(
```

Veykril, line 164: Looking at hte actual diagnostic being emitted (it differs slightly due to having more spans to introspect when testing outside of an expansion I think)
```
error[E0521]: borrowed data escapes outside of function
  --> tests\tracked_fn_no_eq.rs:55:9
   |
52 | fn assert_update_on_return_type<'db>(old_pointer: *mut &'db u32, new_value: &'db u32) -> ! {
   |                                 ---  ----------- `old_pointer` is a reference that is only valid in the function body
   |                                 |
   |                                 lifetime `'db` defined here
...
55 |         salsa::plumbing::UpdateDispatch::<&u32>::maybe_update(old_pointer, new_value);
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |         |
   |         `old_pointer` escapes the function body here
   |         argument requires that `'db` must outlive `'static`
   |
   = note: requirement occurs because of a mutable pointer to `&u32`
   = note: mutable pointers are invariant over their type parameter
   = help: see <https://doc.rust-lang.org/nomicon/subtyping.html> for more information about variance
```

The triggering span seems to be the `maybe_update` identifier (and/or the full path span), so to improve the diagnostics here we'd need to pass that path / identifier from the proc-macro to the decl macro with an adjusted span

src/update.rs
MichaReiser: Right, it does apply if both key and value are static. It doesn't apply if the `value` requires `Update` but the `Key` otherwise wouldn't

Veykril: The bounds don't apply the fallback trick, right...

components/salsa-macro-rules/src/setup_tracked_fn.rs
MichaReiser: There are two things that are "bad" about the diagnostic:

* The error isn't very actionable: I have to know that the solution here is to implement the `Update` trait
* The span point into some non existing code. Ideally, we'd use the span of the return type (possibly?)

MichaReiser: It at least points users in the right direction -> add `Partialeq` or implement `Update`

Veykril: I wonder if https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute is useful here? (unsure given we go through the fallback trick)

MichaReiser: Yeah, that might work. Let me create an issue for this

Veykril (approved):

PR #682, opened 2025-02-09
Fix Disambiguator- and IdentityMap hashing

Veykril:
As was raised by @ibraheemdev in https://github.com/salsa-rs/salsa/pull/647, the hashmap may consider slots/buckets that do not uniquely correspond to a single hash so the hash itself is required for equality in our case.

Added a test that exerts the disambiguator mapping as we have no test that actually hits it, though even without the changes here that test won't fail. I am not quite sure what a setup for that would be.

### DIFF
--- a/tests/tracked_struct_disambiguates.rs
+++ b/tests/tracked_struct_disambiguates.rs
@@ -1,0 +1,107 @@
+//! Test that disambiguation works, that is when we have a revision where we track multiple structs
+//! that have the same hash, we can still differentiate between them.
+#![allow(warnings)]
+
+use std::hash::Hash;
+
+use rayon::iter::Either;
+use salsa::Setter;
+
+#[salsa::input]
+struct MyInput {
+    field: u32,
+}
+
+#[salsa::input]
+struct MyInputs {
+    field: Vec<MyInput>,
+}
+
+#[salsa::tracked]
+struct TrackedStruct<'db> {
+    field: DumbHashable,
+}
+
+#[salsa::tracked]
+struct TrackedStruct2<'db> {
+    field: DumbHashable,
+}
+
+#[derive(Debug, Clone)]
+pub struct DumbHashable {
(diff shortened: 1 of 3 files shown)


tests/tracked_struct_disambiguates.rs
```diff
@@ -0,0 +1,107 @@
...
+    let mut db = salsa::DatabaseImpl::new();
+    let inputs = MyInputs::new(
+        &db,
+        (0..1024)
+            .into_iter()
+            .map(|i| MyInput::new(&db, i))
+            .collect(),
+    );
+    let trackeds = batch(&db, inputs);
+    for (id, tracked) in trackeds.into_iter().enumerate() {
```
MichaReiser, line 94: I don't think this test exercises the right code path. We need at lesst two revisions, were both construct roughly the same tracked stucts because the identity map isn't used in a single revision (and we then just end up disambiguating too often)

It might be easier to write a unit test for each map

MichaReiser (commented):
Thanks. Hashbrowns raw entry api is somewhat easy to hold wrong and the documentation is very sparse.

tests/tracked_struct_disambiguates.rs
Veykril: I was thinking of writing a unit test, but then we'd still not actually test that disambiguating works as expected (as no test actual hits the disambiguator). I guess the test here + a unit test should cover things though.

Veykril: Added 2 unit tests

Veykril:
> Thanks. Hashbrowns raw entry api is somewhat easy to hold wrong and the documentation is very sparse.

It is rather sparse but this was entirely my mistake, the hash is part of the identity/eq of our struct here so it doesn't make sense to leave it out when comparing

MichaReiser (approved):
Thanks

PR #684, opened 2025-02-12
LRU eviction at revision bump

Veykril:
This change makes us LRU evict once a new revision starts, instead of at the very moment we reach the limit while computing tracked functions.


src/table/memo.rs
```diff
@@ -189,12 +204,24 @@ impl MemoTable {
             TypeId::of::<M>(),
             "inconsistent type-id for `{memo_ingredient_index:?}`"
         );
+        // arc-swap does not expose accessing the interior mutably at all unfortunately
+        // https://github.com/vorner/arc-swap/issues/131
+        // so we are required to allocate a nwe arc within `f` instead of being able
```
MichaReiser, line 209:
```suggestion
        // so we are required to allocate a new arc within `f` instead of being able
```

MichaReiser (approved):

PR #685, opened 2025-02-12
Allow triggering LRU eviction without increasing the current revision

Veykril:
Best reviewed by commit as it contains a couple cleanup commits

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -136,6 +136,14 @@ macro_rules! setup_tracked_fn {
                     })
                 }
 
+                pub fn fn_ingredient_mut(db: &mut dyn $Db) -> &mut $zalsa::function::IngredientImpl<Self> {
+                    let zalsa_mut = db.zalsa_mut();
+                    zalsa_mut.reset_cancellation_flag();
+                    let index = zalsa_mut.add_or_lookup_jar_by_type(&$Configuration);
+                    let (ingredient, _) = zalsa_mut.lookup_ingredient_mut(index);
+                    ingredient.assert_type_mut::<$zalsa::function::IngredientImpl<Self>>()
+                }
+
                 $zalsa::macro_if! { $needs_interner =>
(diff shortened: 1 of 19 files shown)


components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -284,9 +291,15 @@ macro_rules! setup_tracked_fn {
                 }
 
                 $zalsa::macro_if! { if0 $lru { } else {
+                    /// Sets the lru capacity
+                    ///
+                    /// **WARNING:** Just like an ordinary write, this method triggers
+                    /// cancellation. If you invoke it while a snapshot exists, it
+                    /// will block until that snapshot is dropped -- if that snapshot
+                    /// is owned by the current thread, this could trigger deadlock.
```
Veykril, line 299: I am a bit iffy about how our cancellation triggering functions all need this comment. I wonder if it would make sense to introduce a `CancellationToken` type (not to be confused with async cancellation tokens...) that one would need to supply to acknowledge the deadlock conditions as such an API forces the caller to think about this, where as the current status quo is having to have read the docs of the function.

Veykril: Right, that would infest all the input setters I believe hmm

src/zalsa.rs
```diff
@@ -45,6 +45,8 @@ pub unsafe trait ZalsaDatabase: Any {
     ///
     /// **WARNING:** Triggers cancellation to other database handles.
     /// This can lead to deadlock!
+    ///
+    /// Note that this needs to be paired with a call to `Zalsa::new_revision` or `Zalsa::clear_cancellation_flag`.
```
nikomatsakis: This comment I think is on the wrong commit?

nikomatsakis (approved):
Seems ok-- I don't entirely understand the point of this, but I trust you

Veykril:
In r-a we basically want to trigger LRU if no cancellation have occurred for some while as the LRU won't trigger if no inputs change anymore, this gives us the flexibility to do that

PR #688, opened 2025-02-12
internal: switch to `boxcar` from `append-only-vec`

davidbarsky:
Boxcar now uses [Loom](https://github.com/ibraheemdev/boxcar/pull/25), and I personally trust Loom/Shuttle-tested libraries more than those that aren't despite their apparent simplicity.

(cc: @ibraheemdev)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ dashmap = { version = "6", features = ["raw-api"] }
 hashlink = "0.9"
 hashbrown = "0.14.3"
 indexmap = "2"
-append-only-vec = "0.1.5"
+boxcar = "0.2.9"
 tracing = "0.1"
 parking_lot = "0.12"
 rustc-hash = "2"
--- a/src/input.rs
+++ b/src/input.rs
@@ -194,7 +194,7 @@ impl<C: Configuration> IngredientImpl<C> {
             .table()
             .pages
             .iter()
-            .filter_map(|page| page.cast_type::<crate::table::Page<Value<C>>>())
+            .filter_map(|(_, page)| page.cast_type::<crate::table::Page<Value<C>>>())
             .flat_map(|page| page.slots())
     }
 
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -263,7 +263,7 @@ where
             .table()
             .pages
             .iter()
-            .filter_map(|page| page.cast_type::<crate::table::Page<Value<C>>>())
+            .filter_map(|(_, page)| page.cast_type::<crate::table::Page<Value<C>>>())
             .flat_map(|page| page.slots())
     }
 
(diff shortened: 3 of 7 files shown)


MichaReiser:
The perf difference is interesting

ibraheemdev:
I don't see any relevant changes in the CodSpeed diffs, but I assume the difference is because of https://github.com/salsa-rs/salsa/pull/555#issuecomment-2367327893. If all write paths for the ingredients vector go through a lock, I can add a `push_single_writer` method (or a new SPMC split type) to `boxcar` to avoid the extra unnecessary synchronization if you think it's worth doing, but it sounds like that isn't a huge bottleneck anyways.

Veykril (commented):
Can we use the type in a qualified manner, that is as `boxcar::Vec` instead of importing it? Or give the import an alias? I find it confusing to overwrite a prelude type like this

davidbarsky:
> I don't see any relevant changes in the CodSpeed diffs, but I assume the difference is because of https://github.com/salsa-rs/salsa/pull/555#issuecomment-2367327893. If all write paths for the ingredients vector go through a lock, I can add a push_single_writer method (or a new SPMC split type) to boxcar to avoid the extra unnecessary synchronization if you think it's worth doing, but it sounds like that isn't a huge bottleneck anyways.

Ibraheem is far more of an expert on this than I am, but I'm guessing that that the benchmark benefits substantially from a spinlock, but those benefits evaporate under real, day-to-day conditions. At the very least, I'm _extremely_ skeptical of spinlocks outside a few, very narrow circumstances.

> Can we use the type in a qualified manner, that is as `boxcar::Vec` instead of importing it? Or give the import an alias? I find it confusing to overwrite a prelude type like this

Yeah, I'll refactor usages to be fully-qualified.

davidbarsky:
Updated as per review. I'm inclined to merge this given that most of the regression(s) could be explained away by "spinlocks look good in benchmarks".

ibraheemdev:
We see a similar 4% regression on red-knot incremental performance with this change. That said, I am also inclined to merge because I think it's likely that we can avoid holding the lock during writes to boxcar (I have an idea to make this possible), which would mean we aren't single-writer anymore and the spinlocks could start to matter. I am also fine to hold off until that change happens and we have measurable results, but boxcar's implementation is more robust for future changes, and if the Loom support is important we could go either way.

davidbarsky:
I think Loom support/tests are sufficiently important from a correctness perspective that I’d prefer a Loom-tested library over a non-Loom tested library.

I’m also extremely interested in seeing your locking changes: I imagine it’d be useful regardless of whether I land this PR or not.

Veykril (approved):
I'm also in favor of this change despite the small regression

davidbarsky:
Alright, I've rebased and I'll land this change.

PR #690, opened 2025-02-14
Experiment with single-writer `boxcar::Vec`

ibraheemdev:
Testing to see if we can avoid the performance hit in https://github.com/salsa-rs/salsa/pull/688 by using a single-writer optimized `push` method (which I added to boxcar in a branch). It looks like we can only do this for the `ingredient_vec` in `Zalsa`, not the page table or viewcaster vector.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ dashmap = { version = "6", features = ["raw-api"] }
 hashlink = "0.9"
 hashbrown = "0.14.3"
 indexmap = "2"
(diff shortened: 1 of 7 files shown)


ibraheemdev:
Looks like this doesn't really help much. I would guess where `append-only-vec` is winning is avoiding per-entry initialization flags because it serializes insertion order using the spinlock, which gives it better cache-locality. That would require a completely new type in `boxcar` for the single-writer case, not just a new method.

PR #691, opened 2025-02-15
Emit better diagnostic for invalid tracked function return types

Veykril:
I also slimmed down the generated code for this given it shouldn't be called in the first place, so it being a no-op doesn't matter


MichaReiser (approved):
Nice

PR #693, opened 2025-02-17
Don't `as_view` cast unless needed

Veykril:
`maybe_changed_after` only needs a casted view if shallow verification fails. This does mean we might cast multiple times if we fail to claim the sync table. Therefore this change optimizes for the hot path, but pessimizes for the cold path.

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -4,7 +4,7 @@ use crate::{
     key::DatabaseKeyIndex,
     zalsa::{Zalsa, ZalsaDatabase},
     zalsa_local::{ActiveQueryGuard, QueryEdge, QueryOrigin},
-    AsDynDatabase as _, Id, Revision,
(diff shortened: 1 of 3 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -30,14 +30,15 @@ where
...
-                if self.shallow_verify_memo(db, zalsa, database_key_index, memo) {
+                if self.shallow_verify_memo(db.as_dyn_database(), zalsa, database_key_index, memo) {
                     return if memo.revisions.changed_at > revision {
                         MaybeChangedAfter::Yes
                     } else {
                         MaybeChangedAfter::No(memo.revisions.accumulated_inputs.load())
                     };
                 }
                 drop(memo_guard); // release the arc-swap guard before cold path
+                let db = db.as_view::<C::DbView>();
```
Veykril, line 41: The cast has moved here

Veykril:
Looks like improvements and regressions alike, so probably not worth it.

PR #696, opened 2025-02-19
Update boxcar

ibraheemdev:
It looks like the reason for the regression in https://github.com/salsa-rs/salsa/pull/688 might have been simpler than I thought and come down to some poor codegen when allocating new boxcar buckets (see https://github.com/ibraheemdev/boxcar/pull/28). Hoping CodSpeed also picks up these changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ description = "A generic framework for on-demand, incrementalized computation (e
 salsa-macro-rules = { version = "0.19.0", path = "components/salsa-macro-rules" }
 salsa-macros = { version = "0.19.0", path = "components/salsa-macros", optional = true }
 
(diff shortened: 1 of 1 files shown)


ibraheemdev:
The CodSpeed regression is really interesting.. I see a very clear improvement locally.

davidbarsky:
Maybe one way to handle the noise from codespeed is to look at instruction counts before/after the change?

ibraheemdev:
Looks like this is a win across the board now. I suspect that's due to (https://github.com/salsa-rs/salsa/pull/725). I'm not sure what allocator CodSpeed was using previously, but the change should be a clear improvement.

Veykril (approved):

PR #699, opened 2025-02-20
fix: fix missing version in Cargo manifest

g-plane:
https://github.com/salsa-rs/salsa/actions/runs/13416786546/job/37479444250 is failed due to lack of `version` of `salsa-macros` in Cargo manifest. This PR fixes it.


MichaReiser (approved):
Thanks for double checking if the release worfklow works and following up with a fix right away :)

PR #701, opened 2025-02-20
Update `hashbrown` (0.15) and `hashlink` (0.10)

waywardmonkeys:
This leaves only `dashmap` pulling in the 0.14 series of `hashbrown`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,8 @@ arc-swap = "1"
 compact_str = { version = "0.8", optional = true }
 crossbeam-queue = "0.3.11"
 dashmap = { version = "6", features = ["raw-api"] }
-hashlink = "0.9"
-hashbrown = "0.14.3"
+hashbrown = "0.15"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -13,8 +13,8 @@ arc-swap = "1"
 compact_str = { version = "0.8", optional = true }
 crossbeam-queue = "0.3.11"
 dashmap = { version = "6", features = ["raw-api"] }
-hashlink = "0.9"
-hashbrown = "0.14.3"
+hashbrown = "0.15.2"
```
MichaReiser, line 16: I think we only specified the patch version before because we used some `0.14.3` or newer features. It should now be okay to only specify major and minor.
```suggestion
hashbrown = "0.15"
```

waywardmonkeys: Applied and edited the commit message.

MichaReiser:
Thanks!

PR #703, opened 2025-02-20
Remove dev dependency `derive-new`

waywardmonkeys:
This was only used for 4 structs in the `calc` example and the expanded code isn't that large.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ salsa_unstable = []
 
 [dev-dependencies]
 annotate-snippets = "0.11.5"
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
I'd seen that there was a newer release of `derive-new`, but figured it might be just as well to remove it and expand the code.

MichaReiser (approved):
IMO, this has the nice side effect that the code becomes more approachable which seems important, considering that it is an example on how to use salsa (we shouldn't overload users with other unfamiliar language constructs).

PR #705, opened 2025-02-21
docs: update release steps

g-plane:
Original guide doesn't reflect the switching to release-plz, and this PR updates it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,5 @@ but we have a [zulip chat](https://salsa.zulipchat.com/) as well.
 
 ## Contributing
 
-To create a release and publish to crates.io, follow the steps:
-
-1. Update the `version` field in Cargo.toml.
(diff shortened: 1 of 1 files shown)


MichaReiser:
To me it seems that release-plz now immediately creates a PR with all changes since the last release and bumps the version number automatically. All that seems necessary is to merge that PR?

g-plane:
I think no. Should I remove this section in readme?

MichaReiser:
I'd find instruction useful on how to do a release because it's currently unclear to me what needs doing. I just noticed the automatically created release PR, that makes me wonder if manually bumping the version is still necessary

g-plane:
Maybe no need to do it manually, but I didn't try release-plz before.

MichaReiser (approved):
Let's merge this and we can iterate on the content as we gain more experience with release-plz

nikomatsakis (approved):

PR #707, opened 2025-02-21
Remove some `ZalsaDatabase::zalsa` calls

Veykril:
Instead pass around `&Zalsa` to callers more to reduce dynamic dispatch, in most of these cases the functions are only called once so the compiler should have enough knowledge to make the extra argument passing virtually free


MichaReiser (approved):
Wow, this is way more than I expected. 

Not sure if this will require another frustrating rebase for @carljm

Veykril:
This can wait on fixpoint to land

carljm:
Nah just go ahead and land it. That's a great perf win, the rebase doesn't look bad

PR #710, opened 2025-02-21
refactor(perf): Remove extra page indirection in `Table`

Veykril:
By manually erasing the concrete slot types and dealing with the vtable ourselves

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -1,9 +1,11 @@
 use std::{
+    alloc::Layout,
     any::{Any, TypeId},
     cell::UnsafeCell,
-    mem::MaybeUninit,
-    panic::RefUnwindSafe,
-    ptr, slice,
+    marker::PhantomData,
+    mem::{self, MaybeUninit},
+    ptr::{self, NonNull},
+    slice,
     sync::atomic::{AtomicUsize, Ordering},
 };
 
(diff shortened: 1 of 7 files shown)


Veykril:
Results look promising

Veykril:
We could likely give `Accumulated` a similar treatment to remove the `Box<dyn Accumulated>`, though that might be a bit trickier since we need to take care of the layout instability of `Vec` used within somehow

src/table.rs
```diff
@@ -230,161 +286,112 @@ impl Table {
...
 
     pub(crate) fn allocate<V>(&self, page: PageIndex, value: V) -> Result<Id, V>
     where
         V: FnOnce(Id) -> T,
     {
-        let guard = self.allocation_lock.lock();
-        let index = self.allocated.load(Ordering::Acquire);
-        if index == PAGE_LEN {
+        let _guard = self.0.allocation_lock.lock();
+        let index = self.0.allocated.load(Ordering::Acquire);
```
Veykril, line 307: Given we only ever modify `allocated` in this function while the `allocation_lock` is held, could we use `Relaxed` ordering for the load here ? I believe so as the `lock` call should already incur a happens before relationship for the participating threads right?

Veykril:
I like the new lower limit, it makes my PRs pop out more now :^)

Veykril:
Unsure why this is regressing in one benchmark after rebasing to latest now. Looks like noise to me as there isn't anything special happening in that regressed codepath

PR #712, opened 2025-02-22
update MSRV to 1.80 and test it in CI

carljm:
We specified an MSRV of 1.76 in `Cargo.toml`, but this was a lie, because it wasn't tested anywhere.

This updates our MSRV to 1.80, the oldest Rust that current main branch actually compiles on, and adds a CI run on 1.80 so that it's tested.

This requires also updating the `rustversion` macro on the `compile_fail` test so that it runs only on stable Rust later than 1.84; that test snapshots compiler output and fails on any Rust version earlier than 1.84.

(Over in #603 I asked about upping our MSRV to 1.83, but the first step is to actually have an MSRV that isn't a lie.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,7 @@ jobs:
     strategy:
       matrix:
         rust:
+          - 1.80.0
           - stable
           - beta
         experimental:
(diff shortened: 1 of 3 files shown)


MichaReiser (approved):

Veykril (approved):

Veykril:
To add context as far as rust-analyzer is concerned regarding MSRV, our policy there is that we have no minimum. (we have a maximum of second latest released stable)

Veykril:
Looks like codspeed doesn't like merge groups https://github.com/CodSpeedHQ/action/issues/126. Kicked it out from the required jobs for merging now

PR #713, opened 2025-02-22
Enforce `unsafe_op_in_unsafe_fn`

Veykril:
We'll need to do this for the 2024 edition upgrade. The motivation for this change thouhgh is so I can quickly check whether the merge queue works now (and so I'll self approve this)

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -163,7 +163,7 @@ where
         &'this self,
         memo: &memo::Memo<C::Output<'this>>,
     ) -> &'this memo::Memo<C::Output<'this>> {
-        std::mem::transmute(memo)
+        unsafe { std::mem::transmute(memo) }
     }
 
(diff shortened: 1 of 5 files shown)


MichaReiser:
> We'll need to do this for the 2024 edition upgrade.

Is my understanding correct that an 2024 edition upgrade would require bumping the MSRV to 1.85? Bumping the MSRV is technically a breaking change for Ruff because users can build Ruff from source

Veykril:
Yes it would require 1.85 (or later). I didn't mean to imply we should do that ASAP (though I think Niko would love 1.86 for trait upcasting).

Veykril:
Okay I think the merge queue works now (its blocking on the CI check introduced in https://github.com/salsa-rs/salsa/pull/712)

PR #716, opened 2025-02-23
Remove unnecessary `Arc` and `Clone` impl from `Views`

Veykril:


Veykril:
Somethings weird with the CI benching

src/views.rs
```diff
@@ -22,7 +19,7 @@ use std::{
 /// ```
 pub struct Views {
     source_type_id: TypeId,
-    view_casters: Arc<boxcar::Vec<DynViewCaster>>,
+    view_casters: Box<boxcar::Vec<DynViewCaster>>,
```
Veykril, line 22: I seem to be missing something here, replacing the `Arc`, either with nothing or with a `Box` regresses performance even though `view_casters` is never cloned?

PR #717, opened 2025-02-23
Use `Mutex<Vec>` for `ingredients_requiring_reset` instead of `boxcar`

Veykril:
We only insert into it when holding the `jar_map` lock, while reading only happens when we have `&mut` access anyways


Veykril:
Any ideas on how to maybe get rid of allocator noise like this? 
image

MichaReiser:
> Any ideas on how to maybe get rid of allocator noise like this? image

Not with the standard allocator. But we could switch to jemalloc and disable the decay:

https://github.com/astral-sh/ruff/blob/1ab975b1423e296732597420b0c7d22c568fe800/crates/ruff_benchmark/benches/linter.rs#L21-L49

PR #718, opened 2025-02-23
Don't clone strings in benchmarks

Veykril:
This results in frequent allocator calling which generally hurts reliability of the benchmarks. Saw `malloc` popup as a 10% regression in https://github.com/salsa-rs/salsa/pull/717

### DIFF
--- a/benches/compare.rs
+++ b/benches/compare.rs
@@ -8,6 +8,7 @@ use salsa::Setter;
 
 #[salsa::input]
 pub struct Input {
+    #[return_ref]
     pub text: String,
 }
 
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):
Hmm, I should have brought this up earlier. 

I consider salsa's default of cloning the return type a huge perf footgun. It's so easy to forget a `return_ref`. We've run into this more than once on the ruff side. Should we change the default to always `return_ref`?

Veykril:
> Hmm, I should have brought this up earlier.
> 
> I consider salsa's default of cloning the return type a huge perf footgun. It's so easy to forget a `return_ref`. We've run into this more than once on the ruff side. Should we change the default to always `return_ref`?

Heh, I was considering raising the same question actually after seeing this. It seems to me one would almost always want to return a reference (except for `Copy` types). So I'd personally be in favor of flipping the default.

PR #721, opened 2025-02-23
Move `unwind_if_revision_cancelled` from `ZalsaLocal` to `Zalsa`

Veykril:
This allows reducing some dynamic dispatching calls


Veykril:
4% regression reported looks like noise as those codepaths aren't affected by this change

davidbarsky:
Something about the `amortized[InternedInput]` benchmark appears to be *extremely* noisy. Not sure what its deal is.

davidbarsky (approved):

PR #725, opened 2025-02-24
Use jemalloc in benches with decay disabled

Veykril:


Veykril:
I guess a faster allocator alone will probably make allocator noise less visible

PR #729, opened 2025-02-25
Emit compiler errors for unknown field attributes

### LINKED ISSUE #728: Unknown annotations in `salsa::tracked` structs do not lead to an error
Tracked struct fields that are annotated with unknown attributes are just silently ignored. For example, when changing `ParsedFile` in `examples/lazy-input/main.rs` to this …
```rs
#[salsa::tracked]
struct ParsedFile<'db> {
    value: u32,
    #[returns_ref]  // oops, should be 'return_ref'!
    links: Vec<ParsedFile<'db>>,
}
```
… everything still compiles without any errors, but `parsed_file.links(db)` returns a clone of the vector now.

It looks to me like the developer experience could possibly be improved here.

sharkdp:
Emit an error if an unknown field attribute is used on a `salsa::tracked` struct. For example, when changing the attribute in `ParsedFile` inside the `examples/lazy-input/main.rs` example from `return_ref` to `returns_ref`, the compiler emits:

    error: unrecognized field attribute
       --> examples/lazy-input/main.rs:164:5
        |
    164 |     #[returns_ref]
        |     ^^^^^^^^^^^^^^

I'm opening this as a draft for two reasons:

* It's unclear to me if this is actually desired behavior. Could non-salsa attributes be mixed with salsa attributes?
* The error message above seems fine, but there are actually a lot more downstream compiler errors if that change is made to the example. Maybe there is a way to get rid of those downstream errors?

closes #728

### DIFF
--- a/components/salsa-macros/src/salsa_struct.rs
+++ b/components/salsa-macros/src/salsa_struct.rs
@@ -380,10 +380,19 @@ impl<'s> SalsaField<'s> {
 
         // Scan the attributes and look for the salsa attributes:
         for attr in &field.attrs {
+            let mut found = false;
             for (fa, func) in FIELD_OPTION_ATTRIBUTES {
                 if attr.path().is_ident(fa) {
                     func(attr, &mut result);
+                    found = true;
                 }
+            }
+
+            if !found {
+                return Err(syn::Error::new_spanned(
+                    attr,
+                    "unrecognized field attribute",
+                ));
             }
         }
 


MichaReiser:
I'm not sure if possible but it would be nice if we could carry along unknown attributes. I would have found that useful in the past, e.g. to suppress a clippy warning (allow a reference to some deprecated code).

It would also be great to have a test for this. You can add a compilation test by adding a new file in `tests/compile-fail`.

sharkdp:
I'm unlikely to finish this any time soon, so I'm going to close this. @MichaReiser I completely agree with your suggestions here: a proper fix should attempt to preserve unknown attributes, and it should come with a compile-fail test.

PR #735, opened 2025-02-26
Add scratch space for `diff_outputs`

Veykril:
Likely wont have an affect on the benches as we do not keep zalsa local state for them


Veykril:
It does prove that allocations are still the root cause of noise in the benchmarks though 😬

Veykril:
Needs more involved benchmarks to tell

Issue #456, opened 2023-07-08
Can't use miette::Report in accumulator due to Clone requirement


DCNick3:
It's quite convenient to use `miette` to provide diagnostic type.

Unfortunately, it's not possible to use the type-erased error type of `miette::Report` due to it not implementing clone =(

```rust
// ERROR: the trait bound `miette::Report: std::clone::Clone` is not satisfied [E0277]
// Note: required by a bound in `salsa_2022::accumulator::AccumulatorIngredient`
#[salsa::accumulator]
pub struct Diagnostics(miette::Report);
```

Would it be possible to use something like `#[return_ref]` as in `#[salsa::tracked]` to not require it?

(Possible workarounds: using a concrete type like `miette::MietteDiagnostic` or wrapping the `miette::Report` in `Arc`)

DropDemBits:
I believe the main reason why `Clone` is required is because of this function:
https://github.com/salsa-rs/salsa/blob/d4a94fbf07bb837f3d9d0a4caa5db4d5db29243f/components/salsa-2022/src/accumulator.rs#L76-L98

Wonder if would be possible to add some indirection to the accumulator data?

PR #739, opened 2025-02-28
Drop clone requirement for accumulated values

Veykril:
I don't see a reason as to why we require it, in fact, why we even clone the elements out of it in the first place. The map is stored in the memo, so we can just borrow from it.

Fixes https://github.com/salsa-rs/salsa/issues/456


MichaReiser (approved):
Oh nice. That certainly makes accumulators more appealing

Issue #455, opened 2023-06-27
assert failed components/salsa-2022/src/function/specify.rs:130:48


Chronostasys:
After I added a cycle recover logic to my language server, it triggers assertion failed.

error log:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `DatabaseKeyIndex { ingredient_index: IngredientIndex(32), key_index: Id { value: 17 } }`,
 right: `DatabaseKeyIndex { ingredient_index: IngredientIndex(49), key_index: Id { value: 17 } }`', ~/.cargo/git/checkouts/salsa-61760caba2b17ca5/d4a94fb/components/salsa-2022/src/function/specify.rs:130:48
```

xffxff:
@Chronostasys Would it be possible for you to share a minimum, reproducible code with us? This would help us to better understand the issue

Chronostasys:
> @Chronostasys Would it be possible for you to share a minimum, reproducible code with us? This would help us to better understand the issue

Not yet. I've tried it yesterday, as the real-world case is really complecated, it may take more time for me to find out how to reproduce it on a sample project.

Chronostasys:
After moving the cycle handler to another participant function in the cycle, the panic position changed:

```
thread 'main' panicked at 'expected a query assigned by `compile_dry_file(11)`, not `Program_emit(11)`'
```

and during recover, the participants be like:

```
["compile_dry_file(11)", "Program_emit(11)", "compile_dry_file(12)", "Program_emit(12)", "compile_dry_file(15)", "Program_emit(15)"]
```

Note that in my case, panic recovery logic is triggered three times in a single revision.

Chronostasys:
> @Chronostasys Would it be possible for you to share a minimum, reproducible code with us? This would help us to better understand the issue

@xffxff 
https://github.com/Chronostasys/salsa-panic-example

xffxff:
A simpler example to reproduce this issue can be found at https://gist.github.com/xffxff/47dbd41a078c8e705aa35d7a9d6f1303. Here's a brief version:

```rust
#[salsa::tracked(jar = Jar)]
fn memoized(db: &dyn Db, input: MyInput) -> u32 {
    memoized_a(db, MyTracked::new(db, input.field(db)))
}

#[salsa::tracked(jar = Jar)]
fn memoized_a(db: &dyn Db, tracked: MyTracked) -> u32 {
    MyTracked::new(db, 0);
    return memoized_b(db, tracked);
[...]

PR #740, opened 2025-02-28
Add regression test for salsa-rs#455 

Veykril:
See https://github.com/salsa-rs/salsa/issues/455, this still fails today but with a different panic. Given recovery is involved this might be fixed by the fixpoint changes

### DIFF
--- a/tests/regression_455.rs
+++ b/tests/regression_455.rs
@@ -1,0 +1,40 @@
+use salsa::{Database, Setter};
+
+#[salsa::tracked]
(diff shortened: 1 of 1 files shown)


Veykril:
@carljm might be good to check if this is fixed in your PR (feel free to pick the test over to your branch)

carljm:
This test behaves as expected under fixpoint iteration cycle handling in https://github.com/salsa-rs/salsa/pull/603. I went ahead and committed a slightly-modified version (that doesn't iterate forever) in that branch. I think this can be closed.

PR #741, opened 2025-02-28
Do not alias fields of tracked_struct Values when updating

Veykril:


src/tracked_struct.rs
```diff
@@ -809,64 +800,54 @@ where
...
-impl<C> Slot for Value<C>
+impl<C> Slot for ValueWithLock<C>
 where
     C: Configuration,
 {
+    // FIXME: `&self` may alias here before the lock is taken?
     unsafe fn memos(&self, current_revision: Revision) -> &crate::table::memo::MemoTable {
         // Acquiring the read lock here with the current revision
         // ensures that there is no danger of a race
         // when deleting a tracked struct.
```
Veykril, line 835: This is a bit odd, there might already be a write lock taken due to delete racing in which case our `&self` may alias with mutable references to its contents. If this scenario ever happens we will panic anyways though as taking a readlock while a write lock is held (and vice versa) is an invalid state.

Nevertheless this is technically UB so we might need to change the `Slot` interface

src/tracked_struct.rs
```diff
@@ -524,55 +525,57 @@ where
...
             panic!(
                 "failed to acquire write lock, id `{id:?}` must have been leaked across threads"
             );
         }
-
-        // UNSAFE: Marking as mut requires exclusive access for the duration of
+        // SAFETY: Marking as mut requires exclusive access for the duration of
         // the `mut`. We have now *claimed* this data by swapping in `None`,
         // any attempt to read concurrently will panic.
-        let data = unsafe { &mut *data_raw };
```
Veykril, line 537: This was technically invalid before as it would alias the `updated_at` field

Veykril:
I believe (if I understand the code correctly) this fixes a theoretical unsoundness when a user leaks a tracked struct value across threads. In that scenario we might try to acquire the read lock and write lock at the same time which could cause aliasing (before we inevitably panic).

The perf regression likely comes from taking the read lock in `tracked_field` now. I believe we are required to take the lock there as otherwise the access could alias (given the above scenario).

Veykril:
I'll see if we can craft a test that actually hits that lock setup resulting in a panic (to check if miri might be able to diagnose the theoretical issue)

src/tracked_struct.rs
```diff
@@ -488,8 +491,9 @@ where
         // In that case we should not modify or touch it because there may be
         // `&`-references to its contents floating around.
         //
-        // Observing `Some(current_revision)` can happen in two scenarios: leaks (tsk tsk)
-        // but also the scenario embodied by the test test `test_run_5_then_20` in `specify_tracked_fn_in_rev_1_but_not_2.rs`:
+        // Observing `Some(current_revision)` can happen in two scenarios:
+        // - leaks, see tests\preverify-struct-with-leaked-data.rs and tests\preverify-struct-with-leaked-data-2.rs
+        // - or the following scenario (FIXME verify this? There is no test that covers this behavior):
```
Veykril, line 496: I am a bit curious about this second scenario. The test that used to exist here (its gone now) does not actually work anymore (I brought it back to check). So something within salsa's behavior must've changed here, @nikomatsakis any idea? Is this scenario working out differently nowadays, given you removed the test in one of your PRs? The comment was added in https://github.com/salsa-rs/salsa/commit/188f759555bed374435467f2df42a035dc834643#diff-545b847b1fec894075887ca68e52e020a61bc8f8e3d3ec6fe1bcbc27bb46b5d5 and the test was removed in https://github.com/nikomatsakis/salsa/commit/3d2b2d3a65288bfea50eba46cb63767d56fdf9a7

src/tracked_struct/tracked_field.rs
```diff
@@ -62,7 +62,7 @@ where
         revision: crate::Revision,
     ) -> MaybeChangedAfter {
         let zalsa = db.zalsa();
-        let data = <super::IngredientImpl<C>>::data(zalsa.table(), input);
+        let data = <super::IngredientImpl<C>>::data(zalsa.table(), input, zalsa.current_revision());
```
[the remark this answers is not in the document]
nikomatsakis: Thinking about it more, I *think* -- it is correct to get the read-lock *but* if everyone is using it correctly, the lock should have already been acquired, because if the field may have changed, the creating query should have re-executed.

src/tracked_struct.rs
```diff
@@ -524,56 +532,74 @@ where
...
+        let updated = unsafe { C::update_fields(current_revision, revisions, old_fields, fields) };
+
+        // SAFETY: The mutable accesses to the following metadata fields will not alias
+        // as attempting to read any of these would first verify whether we are up to date, either
+        // blocking on our update or implying we would not have ended up in this method here in the
+        // first place.
+        // FIXME: why can these not alias
+        // - `durability`:  with `tracked_field` / `untracked_field`
+        // - `revisions`:  with `tracked_field::maybe_changed_after`
+        // - `created_at`: with `untracked_field` / `maybe_changed_after`
```
Veykril, line 567: These I still need to check as to why we can say they are not going to end up aliasing (as Niko was talking about). I believe that is correct for all of them except for `revisions`, as that one is used to check if a field is outdated after all. Correct me if I am wrong but we first check that field to tell if we need to update the struct no? In which case we are in fact aliasing on that field

MichaReiser:
I think I'm running into the aliasing bug with Red Knot's WASM based playground. 

We see instances where it panics with 

```
access to field whilst the value is being initialized
```

I rebased your PR and I'm no longer seeing the same panic (or it at least makes it much less likely to reproduce 😆)

I forgot the specifics of what's left. Should I take this PR from you?

Veykril:
Just whatever is left regarding the review comments I've put up I think, though I am not sure that what you observe would be fixed by this PR (unless we are in fact running into UB!). Feel free to take it over, I'm busy this week with a rust-analyzer refactor

src/tracked_struct.rs
```diff
@@ -742,9 +756,10 @@ where
         revision: Revision,
     ) -> MaybeChangedAfter {
         let zalsa = db.zalsa();
-        let data = Self::data(zalsa.table(), input);
+        let data = Self::data_raw(zalsa.table(), input);
+        let created_at = unsafe { (*data).created_at };
```
MichaReiser, line 760: I wonder if `created_at` should be an `AtomicRevision` because we mutate it in place during `update`

MichaReiser:
@ibraheemdev you're probably the better reviewer for this than I.

@Veykril it seems that this PR uses some unstable language features

Veykril:
That would, iirc raw borrows were stabilized in 1.81

src/tracked_struct.rs
```diff
@@ -557,34 +557,30 @@ where
...
 
-        // UNSAFE: Marking as mut requires exclusive access for the duration of
-        // the `mut`. We have now *claimed* this data by swapping in `None`,
-        // any attempt to read concurrently will panic.
-        let data = unsafe { &mut *data_raw };
+        // SAFETY: We have now *claimed* mutable access to the `value` field by swapping in `None`,
+        // any attempt to read concurrently will panic so it is safe to take exclusive references.
+        let old_fields = unsafe { &raw mut (*data_raw).value.fields }.cast::<C::Fields<'_>>();
+
+        // SAFETY: FIXME: why can this not alias with tracked_field::maybe_changed_after?
```
ibraheemdev: I think it can, it should be a `&[AtomicRevision]` instead.

Veykril:
Go ahead

src/tracked_struct.rs
```diff
@@ -864,19 +874,18 @@ where
...
-        data.read_lock(zalsa.current_revision());
+        let data = Self::data_raw(zalsa.table(), id);
 
-        let field_changed_at = data.revisions[relative_tracked_index];
+        let field_changed_at = unsafe { (&(*data).revisions)[relative_tracked_index].load() };
 
         zalsa_local.report_tracked_read_simple(
             DatabaseKeyIndex::new(field_ingredient_index, id),
-            data.durability,
+            unsafe { (*data).durability },
```
ibraheemdev, line 888: Is it okay to report the read and load the revision/durability before acquiring the read lock?

src/tracked_struct.rs
```diff
@@ -648,17 +652,15 @@ where
         // that is still live.
 
         {
-            // SAFETY: Guaranteed by caller.
-            let data = unsafe { &*data_raw };
-
-            let last_updated_at = data.updated_at.load();
+            // SAFETY: `updated_at` is never exclusively borrowed, so borrowing it is sound
+            let last_updated_at = unsafe { (*data_raw).updated_at.load() };
```
ibraheemdev, line 656: The way the fields are read here (and in a few other places) still creates a temporary `&ValueWithMetadata`. You'd need a raw borrow to avoid that aliasing (or put an `UnsafeCell` around the relevant fields).

Veykril: Really? That is news to me, I thought a field projection like this here will only create a reference to the field?

src/tracked_struct.rs
Veykril: Ah good catch, that must've gotten lost in the rebase, this should take the lock

src/tracked_struct.rs
ibraheemdev: Sorry I think you're right, this should be fine.

src/tracked_struct.rs
```diff
@@ -724,24 +724,43 @@ where
...
+    fn data_raw(table: &Table, id: Id) -> *mut ValueWithMetadata<C> {
+        table.get_raw(id)
     }
 
-    fn data_raw(table: &Table, id: Id) -> *mut Value<C> {
-        table.get_raw(id)
+    /// # Safety
+    ///
+    /// `data` must be a valid pointer to a `ValueWithMetadata<C>`
+    unsafe fn fields(
```
ibraheemdev: Might be helpful to name this `lock_fields` for clarity.

ibraheemdev (approved):

Veykril:
We stored it before too though

PR #742, opened 2025-03-01
refactor: Clean up `par_map` a bit

Veykril:


Veykril:
@davidbarsky bad news, the current impl of `par_map` is unsound, we are smuggling the database reference across threads before calling clone on it.

davidbarsky:
> @davidbarsky bad news, the current impl of `par_map` is unsound, we are smuggling the database reference across threads before calling clone on it.

Crap, I'm hoping this is fixable.

Veykril:
It is, we just need to eagerly `fork_db`, sending a `Box<dyn Database>` is fine, but `&dyn Database` is not given its not `Send`

Veykril:
Fixed the issue and added some more tests for the other new APIs

davidbarsky (approved):

Veykril:
We could

PR #743, opened 2025-03-02
Things

Veykril:


Veykril:
Oof https://github.com/salsa-rs/salsa/pull/743/commits/85a3ff257e7e6b79c0d5fc22d7f188afb0d4cd66 makes quite a difference, the event system is certainly not free as it requires us to thread a dyn database handle to a bunch of places just for the event reporting.

I wonder if it would be better to have an API that lets you install an event hook into `Zalsa`.

PR #747, opened 2025-03-06
`Update` derive field overwrite support

Veykril:
I believe this should allow reducing the need for coherence dependencies like https://github.com/salsa-rs/salsa/blob/ceb9b083b3c0f6a1634e5a0b75b7bb5c7ca7b33f/Cargo.toml#L29-L30
The main annoyances is that this technically makes the derive unsafe (which is not expressible today)


Veykril:
Actually, @MichaReiser I believe this could be a viable alternative for the `Update` requirement issue. You'd define a helper function `fn(_: *mut Map<CompactStr, ...>, Map<CompactStr, ...>) -> bool` and then just reference that in the places where you need it here. That avoids the coherence problems with third-party crate types while also not requiring a wrapper newtype that implements `Update` for the given foreign type.

MichaReiser:
Yeah, it does reduce the urgency for having `Update` impls in salsa but I think it's still less ergonomic because it now requires a change at every use.

Veykril:
Either way I believe this to be a nice addition to the derive

MichaReiser:
> Either way I believe this to be a nice addition to the derive. Improved the diagnostics a bit so this is ready now

I agree with this. It definitely raises the bar for which Crate Salsa should provide `Update` implementations. But it comes at an ergonomics cost, which is why we'd probably want to keep providing `Update` implementations for some popular crates.

Veykril:
Rebased this again now, any concerns? I am not trying to advocate with this to remove some orphan rule avoidance dependencies for now, the feature itself seems generally useful to me.

davidbarsky:
Yeah, I'll review in a second. Sorry!

components/salsa-macros/src/update.rs
```diff
@@ -35,7 +53,7 @@ pub(crate) fn update_derive(input: syn::DeriveInput) -> syn::Result<TokenStream>
                 .bindings()
                 .iter()
                 .fold(quote!(), |tokens, binding| quote!(#tokens #binding,));
-            let make_new_value = quote_spanned! {variant.ast().ident.span()=>
+            let make_new_value = quote! {
```
davidbarsky, line 56: why remove the span? confuses ra?

Veykril: Yes, attaching all your tokens to spans randomly has less than desireable effects. The tokens in this quote! are all meaningless so there is no need to give them a specific input span

PR #748, opened 2025-03-07
feat: Implement #599

Veykril:
Currently looking into https://github.com/salsa-rs/salsa/issues/599, this change makes things a bit simpler to overview removing some back and forth in code paths

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -43,6 +43,7 @@ pub trait Configuration: Any {
 
     /// The input to the function
     type Input<'db>: Send + Sync;
+    type MapKey<'db>: Send + Sync + std::hash::Hash + Eq + std::fmt::Debug + Any;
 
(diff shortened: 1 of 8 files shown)


Veykril:
So implementing the memo table part is actually trivial, the real issue is the dependency tracking as that goes through the type erased ingredient trait and now we have potentially more keys than just an `ID` ...

Veykril:
I just realized, all of this is probably unnecessary! Instead of interning the additional arguments, can't we just synthesize a new tracked struct with each arg being a field? That would trivial fix things from what I can tell?

MichaReiser:
That won't work because you can't create tracked structs from outside a query. It would limit multi argument queries to only be callable from other queries. 

It's also not clear to me how that's different from interning?

PR #749, opened 2025-03-07
internal: use `portable-atomic` in `IngredientCache` to compile on `powerpc-unknown-linux-gnu`

davidbarsky:
rust-analyzer compiles on `powerpc-unknown-linux-gnu`. Since #687, Salsa does not.

However, since the Rust Project considers `powerpc-unknown-linux-gnu` to be a [Tier 2 platform with host tool support](https://doc.rust-lang.org/nightly/rustc/target-tier-policy.html#tier-2-with-host-tools), we really should avoid breaking folks on that platform!

For context, see [this discussion](https://salsa.zulipchat.com/#narrow/channel/145099-Using-Salsa/topic/Use.20of.20AtomicU64.20in.20Salsa/with/503583305) on Zulip.


Veykril (approved):

davidbarsky:
It looks like portable-atomic effectively treats this the item a lock on 32 bit platforms; we can always vendor.

PR #750, opened 2025-03-08
Have salsa not depend on salsa-macros

Veykril:
That way salsa can build without having to build the proc-macro ecosystem, improving build times for dependants of salsa


src/database.rs
```diff
@@ -89,6 +88,21 @@ pub trait Database: Send + AsDynDatabase + Any + ZalsaDatabase {
...
+
+    #[doc(hidden)]
+    #[inline(always)]
+    unsafe fn downcast(db: &dyn Database) -> &dyn Database
+    where
+        Self: Sized,
+    {
+        // No-op
+        db
+    }
```
Veykril, line 105: Interestingly enough, this change also means that one does not actually need to use `#[salsa::db]` on trait impls for the baseline `Database`

src/database.rs
```diff
@@ -89,6 +88,21 @@ pub trait Database: Send + AsDynDatabase + Any + ZalsaDatabase {
     {
         crate::attach::attach(self, || op(self))
     }
+
+    #[doc(hidden)]
+    fn zalsa_register_downcaster(&self) {
+        // The no-op downcaster is special cased in view caster construction.
+    }
```
Veykril, line 95: The perf changes come from this being a no-op now, so they are kind of fake as you only get that benefit if you don't use a custom database trait

davidbarsky: should we update the benchmarks to use a custom
database? I assume any production use of Salsa would need a custom database to do the lazy input pattern to load, e.g., files.

davidbarsky: Lukas said: 

> unrelated to [this] PR I'd say, but I don't think that's relevant for the benchmarks either. The only difference is a no-op vs upcast (a dyn trait pointer reconstruction)

davidbarsky (approved):

MichaReiser:
I'm very busy this week and don't know when I'll get around to review this or other salsa PRs. Feel free to go ahead without my review.

Veykril:
👍  We can always roll this back if needed

PR #754, opened 2025-03-12
tests(parallel): remove `#[cfg_attr(miri, ignore)]`

davidbarsky:
I wasn't able to reproduce the issue again locally, so... trying this out.


Veykril:
The failures are due to strict provenance -> https://github.com/crossbeam-rs/crossbeam/pull/1184. Given crossbeam has a fairly low MSRV this is unlikely to get fixed soon.

PR #755, opened 2025-03-13
bug [salsa-macros]: Improve debug name of tracked methods

MichaReiser:
The debug name of every tracked method always was `inner_fn_name`. This PR improves the function name to be `<fn_name>_`. This is not perfect but much more useful than the same name for every tracked method.

An alternative is to use `std::any::type_name` and we could then try to detect if the type name is from a `TrackedTrait` but that seems complicated.


carljm (approved):
Awesome, thank you!

PR #756, opened 2025-03-15
feat: Drop `Debug` requirements and flip implementation defaults

Veykril:
The default caused a pretty severe hang for us in rust-analyzer due to a debug implementation formatting the entire dependency tree for each dependency recursively by accident so this proposes flipping the default.

This also drops a bunch of debug bounds internal which are unnecessary (without that  you are required to somehow implement `Debug` for most salsa structs).

### DIFF
--- a/benches/compare.rs
+++ b/benches/compare.rs
@@ -26,7 +26,7 @@ pub struct InternedInput<'db> {
     pub text: String,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, salsa::Supertype)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, salsa::Supertype)]
 enum SupertypeInput<'db> {
     InternedInput(InternedInput<'db>),
     Input(Input),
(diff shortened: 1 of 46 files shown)


MichaReiser (commented):
This makes sense to me. 

Would it be possible to inspect the structs derive attribute and only generate the Debug impl if someone derived Debug (and then remove it). Or should we add our own DebugSalsa derive macro?

Veykril:
> Would it be possible to inspect the structs derive attribute and only generate the Debug impl if someone derived Debug (and then remove it). Or should we add our own DebugSalsa derive macro?

I believe that should be doable but:
1. We can only do this syntactically, so it would fail if the user was to do something odd like qualifying the `Debug` name etc.
2. More importantly, it would make it impossible to derive debug to get a debug impl that merely prints the ID.

I think a custom salsa debug derive might be the better option for this. It definitely sounds like the more proper way to do this (though I'd do that as a follow up).

MichaReiser (approved):

PR #759, opened 2025-03-15
perf: Reduce unnecessary conditional work in `deep_verify_memo`

Veykril:


src/function/maybe_changed_after.rs
```diff
@@ -228,8 +219,24 @@ where
...
+    pub(super) fn validate_may_be_provisional(
+        &self,
+        db: &C::DbView,
+        zalsa: &Zalsa,
+        database_key_index: DatabaseKeyIndex,
+        memo: &Memo<C::Output<'_>>,
+    ) -> bool {
+        // Wouldn't it be nice if rust had an implication operator ...
+        // may_be_provisional -> validate_provisional
+        !memo.may_be_provisional() || self.validate_provisional(db, zalsa, database_key_index, memo)
```
carljm, line 234: Would it be just as good to not add this additional wrapper, and instead just check `memo.may_be_provisional` inside `validate_provisional`? This is the only call site of `validate_provisional`.

carljm (approved):
Nice, thank you!

src/function/maybe_changed_after.rs
Veykril: This seems to regress perf, presumably due to inlining changes

PR #760, opened 2025-03-15
perf: Use a `Vec` for `CycleHeads`

Veykril:
The number of cycle participants is generally low so a hashset likely has more overhead compared to a simple vec for such small numbers of elements.

### DIFF
--- a/src/cycle.rs
+++ b/src/cycle.rs
@@ -90,7 +90,8 @@ pub enum CycleRecoveryStrategy {
 /// the cycle head(s) (can be plural in case of nested cycles) representing the cycles it is part
 /// of. This struct tracks these cycle heads.
 #[derive(Clone, Debug, Default)]
-pub(crate) struct CycleHeads(Option<Box<Vec<DatabaseKeyIndex>>>);
+#[allow(clippy::box_collection)]
+pub struct CycleHeads(Option<Box<Vec<DatabaseKeyIndex>>>);
 
 impl CycleHeads {
     pub(crate) fn is_empty(&self) -> bool {
@@ -114,6 +115,17 @@ impl CycleHeads {
             self.0.take();
         }
         true
+    }
+
+    #[inline]
+    pub(crate) fn insert_into(self, cycle_heads: &mut Vec<DatabaseKeyIndex>) {
+        if let Some(heads) = self.0 {
+            for head in *heads {
+                if !cycle_heads.contains(&head) {
+                    cycle_heads.push(head);
+                }
+            }
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


Veykril:
[b4929f1](https://github.com/salsa-rs/salsa/pull/760/commits/b4929f102c96510226cd45722109172e43249c30) has quite the impact

src/function/maybe_changed_after.rs
```diff
@@ -384,7 +383,12 @@ where
             //    from cycle heads. We will handle our own memo (and the rest of our cycle) on a
             //    future iteration; first the outer cycle head needs to verify itself.
 
-            let in_heads = cycle_heads.remove(&database_key_index);
+            let found = cycle_heads
+                .iter()
+                .position(|&head| head == database_key_index);
+            let in_heads = found
+                .inspect(|&head| _ = cycle_heads.swap_remove(head))
+                .is_some();
```
carljm, line 391:
```suggestion
            let in_heads = cycle_heads
                .iter()
                .position(|&head| head == database_key_index)
                .inspect(|&head| _ = cycle_heads.swap_remove(head))
                .is_some();
```

src/cycle.rs
```diff
@@ -121,36 +133,39 @@ impl std::iter::Extend<DatabaseKeyIndex> for CycleHeads {
...
     fn into_iter(self) -> Self::IntoIter {
         self.0.map(|heads| *heads).unwrap_or_default().into_iter()
     }
 }
 
-// This type can be removed once MSRV is 1.83+ and we have Default for hashset iterators.
-pub(crate) struct CycleHeadsIter<'a>(
-    Option<std::collections::hash_set::Iter<'a, DatabaseKeyIndex>>,
-);
+pub struct CycleHeadsIter<'a>(std::slice::Iter<'a, DatabaseKeyIndex>);
```
carljm, line 158: I don't know if you consider this an improvement, but we can eliminate `CycleHeadsIter` and implement `IntoIterator` for `&'a CycleHeads` like so:

```rs
impl<'a> std::iter::IntoIterator for &'a CycleHeads {
    type Item = DatabaseKeyIndex;
    type IntoIter = std::iter::Copied<std::slice::Iter<'a, DatabaseKeyIndex>>;

    fn into_iter(self) -> Self::IntoIter {
        self.0
            .as_ref()
            .map(|heads| heads.iter())
            .unwrap_or_default()
            .copied()
    }
}
```

carljm (approved):
Looks great, thank you!

src/cycle.rs
Veykril: This causes `new[InternedInput]` to regress due to a `report_tracked_read` becoming more expensive, so something optimizes differently there 😕 Will drop our `Extend` impl as we can optimize around knowing the concrete iterator anyways.

PR #766, opened 2025-03-20
Relax atomic loads and stores in `Table`

Veykril:


Veykril:
@davidbarsky here is your reminder for this for your loom work

ibraheemdev:
> `just shuttle` passes

[I don't think shuttle actually models weak atomics](https://docs.rs/shuttle/latest/shuttle/sync/atomic/index.html#warning-about-relaxed-behaviors).

ibraheemdev (commented):
I'm not sure this is correct if something like `slots_of` is used concurrently?

ibraheemdev:
`allocated` is the flag for a slot being initialized and establishes happens-before, e.g. `if i < allocated.load(Acquire) { *slot.get(i) }`. This only matters if you are reading a slot that was initialized by a different thread, which I believe can only happen with the `slots_of` function. Otherwise, most slots are tied to a specific ingredient `Id` which must be allocated by the current-thread. However, even apart from `slots_of`, this invariant is not clear in the code so I would be wary of relaxing the load.

PR #768, opened 2025-03-20
refactor: Shrink `QueryRevisions` by 3 usize by boxing `IdentityMap`

Veykril:
Not all queries actually create tracked structs and so this may be empty in those cases


src/zalsa_local.rs
```diff
@@ -309,7 +309,7 @@ pub(crate) struct QueryRevisions {
     ///   previous revision. To handle this, `diff_outputs` compares
     ///   the structs from the old/new revision and retains
     ///   only entries that appeared in the new revision.
-    pub(super) tracked_struct_ids: IdentityMap,
+    pub(super) tracked_struct_ids: Option<Box<IdentityMap>>,
```
davidbarsky, line 312:
```suggestion
    ///
    /// Since not all queries produce a tracked struct, wrapping 
    /// `IdentityMap` in an `Option<Box<T>>` reduces the size of
    /// `QueryRevisions` 3 words (24 bytes on a 64-bit platform).
    pub(super) tracked_struct_ids: Option<Box<IdentityMap>>,
```

davidbarsky (approved):
lgtm, modulo doc comment nit

MichaReiser:
Do you see any improvement of peak memory consumption in r-a with this change? I'm just curious on how we make the decisions on perf vs memory consumption (both are very important to us and trading perf for memory consumption can certainly be a worth trade-off)

Veykril:
I have not yet, I agree for this one specifically we should as it has performance implications. We'll need to move r-a to the fixpoint changes, then we can test this PR out on our stats afterwards

Veykril:
image
I am not sure what happened here, I am fairly certain I did not press the button there 🤨

PR #773, opened 2025-03-22
refactor: Keep edge condvar on stack instead of allocating it in an `Arc`

Veykril:


Veykril:
Okay I think this does confirm that the benches are really sensitive to allocation changes. This should not regress performance whatsoever.

MichaReiser (approved):
I'm a bit concerned about the amount of unsafe code in salsa. But maybe that's just me because I'm not used to writing `unsafe` code. Avoiding an allocation in `add_edge` certainly is nice. But it's not entirely clear when we think that using `unsafe` is justified

Issue #697, opened 2025-02-19
Enforce `clippy::undocumented_unsafe_blocks`


Veykril:
We have a ton of unsafe in the codebase, I believe we should take care that we document those appropriately

PR #776, opened 2025-03-23
chore: Document most safety blocks

detrumi:
Makes a start on #697, but doesn't forbid `clippy::undocumented_unsafe_blocks` yet.
Some files with many undocumented blocks allow the lint on file level for now to reduce noise.

### DIFF
--- a/src/function/fetch.rs
+++ b/src/function/fetch.rs
@@ -18,13 +18,13 @@ where
         zalsa.unwind_if_revision_cancelled(db);
 
         let memo = self.refresh_memo(db, id);
-        // SAFETY: We just refreshed the memo so it is guaranteed to contain a value now.
         let StampedValue {
             value,
             durability,
             changed_at,
         } = memo
             .revisions
+            // SAFETY: We just refreshed the memo so it is guaranteed to contain a value now.
             .stamped_value(unsafe { memo.value.as_ref().unwrap_unchecked() });
 
         self.lru.record_use(id);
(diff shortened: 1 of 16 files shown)


src/function/fetch.rs
```diff
@@ -18,13 +18,13 @@ where
...
         let memo = self.refresh_memo(db, id);
-        // SAFETY: We just refreshed the memo so it is guaranteed to contain a value now.
         let StampedValue {
             value,
             durability,
             changed_at,
         } = memo
             .revisions
+            // SAFETY: We just refreshed the memo so it is guaranteed to contain a value now.
             .stamped_value(unsafe { memo.value.as_ref().unwrap_unchecked() });
```
Veykril, line 28:
```suggestion
        let memo = self.refresh_memo(db, id);
        // SAFETY: We just refreshed the memo so it is guaranteed to contain a value now.
        let stamped_value = memo
            .revisions
            .stamped_value(unsafe { memo.value.as_ref().unwrap_unchecked() });
        let StampedValue {
            value,
            durability,
            changed_at,
        } = stamped_value;
```
That way we keep the comment close to the relevant part without interrupting the flow

src/zalsa.rs
```diff
@@ -393,6 +393,7 @@ where
     phantom: PhantomData<fn() -> I>,
 }
 
+#[allow(clippy::undocumented_unsafe_blocks)] // TODO(#697) document safety
 unsafe impl<I> Sync for IngredientCache<I> where I: Ingredient + Sync {}
```
Veykril, line 397: is this impl even necessary? Looking at the definition it should hold by auto trait rules already if I am not mistaken

src/function/fetch.rs
detrumi: That would still trigger the lint, but pulling out the argument instead does the trick.

src/zalsa.rs
detrumi: You're right, I'll remove it.

src/function/fetch.rs
Veykril: Oh does clippy specifically check by following line? I see. Well, your change is better either way!

Veykril (approved):
Thanks!

PR #778, opened 2025-03-25
fix: Use `changed_at` revision when updating fields

Veykril:
Setting the field's revision to `current_revision` when updating is not technically wrong, but its pessimizing. If all previous reads only observed a change of up to `changed_at`, then it's fine for us to do the same as nothing we derive from has read from `current_revision`.

### DIFF
--- a/src/tracked_struct.rs
+++ b/src/tracked_struct.rs
@@ -553,7 +553,7 @@ where
         // to meet its safety invariant.
         unsafe {
             if C::update_fields(
-                current_revision,
+                current_deps.changed_at,
                 &mut data.revisions,
                 self.to_self_ptr(std::ptr::addr_of_mut!(data.fields)),
                 fields,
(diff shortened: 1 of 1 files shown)


Veykril:
Looks like one of my parallel tests are flaky (note I don't think there is a bug in salsa, but a race in our testing signal infra)

MichaReiser (approved):

src/tracked_struct.rs
```diff
@@ -568,7 +568,7 @@ where
             }
         }
         if current_deps.durability < data.durability {
-            data.revisions = C::new_revisions(current_revision);
+            data.revisions = C::new_revisions(current_deps.changed_at);
             data.created_at = current_revision;
```
MichaReiser, line 572: It might be a bit surprising that `crated_at` can now be larger than `changed_at`

Veykril: That is already the case when we allocate the struct

PR #779, opened 2025-03-26
chore: Normalize imports style

Veykril:
Effectively reformatted everything with
```toml
imports_granularity = "Module"
group_imports = "StdExternalCrate"
reorder_imports = true
```

### DIFF
--- a/benches/dataflow.rs
+++ b/benches/dataflow.rs
@@ -2,10 +2,11 @@
 //!
 //! This benchmark simulates a (very simplified) version of a real dataflow analysis using fixpoint
 //! iteration.
-use codspeed_criterion_compat::{criterion_group, criterion_main, BatchSize, Criterion};
-use salsa::{CycleRecoveryAction, Database as Db, Setter};
 use std::collections::BTreeSet;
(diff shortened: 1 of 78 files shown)


MichaReiser:
How do we plan on preserving this format style in the future or are we okay if the import style denormalizes over time?

Veykril:
Great question, the main annoyance is that the rustfmt configs are still unstable 🙃 rust-analyzer will attempt to keep the ordering and style consistent (unsure if RustRover has a similar thing going on). I think it is fine if the style degrades over time, we can always do another one of these PRs every now and then, but right now imports are fairly chaotic.

Veykril:
Also I picked `imports_granularity = "Module"` given it reduces git conflicts between PRs to some extent. The option here is up to debate.

davidbarsky:
For "I hate resolving conflicts" and "trivial `git blame` reasons, I'd personally vote for `imports_granularity = "Item"`, but I know the format is ~extremely ugly~ controversial 😅

MichaReiser (approved):

PR #780, opened 2025-03-26
refactor: Discard unnecessary atomic load

Veykril:


src/function/maybe_changed_after.rs
```diff
@@ -205,12 +205,7 @@ where
         );
         if last_changed <= verified_at {
             // No input of the suitable durability has changed since last verified.
-            memo.mark_as_verified(
-                db,
-                revision_now,
-                database_key_index,
-                memo.revisions.accumulated_inputs.load(),
```
Veykril, line 212: This one specifically, we load the value here just to set it to the same one

src/accumulator/accumulated_map.rs
```diff
@@ -74,6 +74,14 @@ impl InputAccumulatedValues {
     pub const fn is_empty(self) -> bool {
         matches!(self, Self::Empty)
     }
+
+    pub fn or_shortcircuiting(self, other: impl FnOnce() -> Self) -> Self {
```
MichaReiser, line 78: Would `or` be enough? Similar to `Result::or` and `Option::or` that both "short circuit" if `self` is `Ok` or `Some`

src/function/maybe_changed_after.rs
```diff
@@ -412,10 +403,7 @@ where
                             continue 'cycle;
                         }
                     }
-                    break 'cycle VerifyResult::Unchanged(
-                        InputAccumulatedValues::Empty,
-                        CycleHeads::from(cycle_heads),
-                    );
+                    break 'cycle VerifyResult::Unchanged(inputs, CycleHeads::from(cycle_heads));
```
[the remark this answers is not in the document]
MichaReiser: CC: @carljm

Veykril:
> CC: @carljm

Unsure why I can't reply to that but right, forgot to call that change out. It looks like a bug to me but please double check

MichaReiser:
Sorry for the ping but @carljm did you see @Veykril comment?

src/function/maybe_changed_after.rs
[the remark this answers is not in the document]
carljm: Oh, sorry for missing this ping, apparently twice! I think this is change is right. It would probably be ideal to write a test that fails without this fix, though. If this fix is needed, I don't think the test should even require a cycle, as this is also the path for non-cycles. It would probably be a test with accumulated values across multiple revisions, with some inputs changing but some sub-graph (that has accumulated values in it) not changing?

src/zalsa_local.rs
```diff
@@ -158,18 +156,18 @@ impl ZalsaLocal {
...
+            input, revisions.durability, revisions.changed_at
         );
+        // We don't access the accumulator anyways
+        let revisions = AssertUnwindSafe(revisions);
         self.with_query_stack(|stack| {
             if let Some(top_query) = stack.last_mut() {
-                top_query.add_read(input, durability, changed_at, accumulated, cycle_heads);
+                top_query.add_read(input, { revisions }.0, cycle_heads);
             }
         })
```
MichaReiser, line 172: I think I prefer the old method signature in that case to avoid this very *easy to get outdated* comment about accumulators. 

Not changing the signature (or only remuving accumulated) also has the benefit that it makes the diff a bit clearer.

src/function/memo.rs
```diff
@@ -217,12 +216,12 @@ impl<V> Memo<V> {
 
     /// Mark memo as having been verified in the `revision_now`, which should
     /// be the current revision.
+    #[inline]
```
MichaReiser, line 219: It would be great to have a comment here mentioning that it's the caller's responsibility to update `accumulated` if their accumulated values have changed since  the query ran last

MichaReiser (approved):

src/zalsa_local.rs
Veykril: Went for that and `#[inline]`'d the function given ti has only one use site

src/function/maybe_changed_after.rs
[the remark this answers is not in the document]
Veykril: Hmm, I gave this a quick try with building a test but couldn't really figure out one that hits this. I don't really know the accumulation stuff well enough here

PR #781, opened 2025-03-28
fix: `#[doc(hidden)]` `plumbing` module

Veykril:

PR #786, opened 2025-04-02
allow reuse of cached provisional memos within the same cycle iteration

carljm:
Currently, we don't allow provisional memos from cycle iteration to be reused as a memoized result. This is important in case another thread "peeks" in while we are still iterating a cycle (it shouldn't see a provisional result), and also so that iteration N+1 of a cycle doesn't reuse results from iteration N; it needs to recompute instead.

But this also means that if the same query is called many times within a single iteration of a cycle, we will re-execute it many times. For example, if query A calls query B, and query B calls query C 1000 times, and query C calls query A, each time we iterate this cycle we will execute query C 1000 times. This can lead to massive blow-up in execution times.

To fix this, we need to record in `cycle_heads` of a provisional memo the iteration-count for each cycle in which this memo was recorded. When validating whether a provisional memo can be used by the caller, we must allow it to be used only if we are still within the same iteration of those cycles, and not otherwise.

This PR fixes an execution-time blow-up on red-knot that was bad enough that it looked like a hang.


carljm:
The last 3 commits on this PR are all viable working fixes, which should be evaluated against each other on maintainability and performance.

* The 3rd-from-last commit (17c8eb879466bdef3f77d5fbdf7b814a16ca2626) walks the active query stack once per cycle head in a provisional memo, to check if each cycle head is present in the active query stack.
* The 2nd-from-last commit (2d7948612be3f411c7787829ff2db74e3a575f0c) only walks the active query stack once, but iterates the provisional memo's cycle heads once per entry in the active query stack.
* The last commit (4ea3d850b53a449dda7af7ceb4cc565d3a64001c) instead tracks the currently-iterating cycles in a new `active_cycles` field on `ZalsaLocal`, and just checks whether that field contains all cycle heads in the memo.

All 3 commits are showing a small regression on a couple Salsa benchmarks. Oddly, the benchmarks showing regression are not the one with cycles, but this fix should have little or no impact when there are not cycles. This suggests the regression might be from compilation, e.g. an inlining change? 17c8eb879466bdef3f77d5fbdf7b814a16ca2626 reported the smallest regression.

I've tried all 3 of these commits in release builds of red-knot, on a module with a cycle that previously appeared to hang (due to runaway re-execution) before this fix. The performance differences observed between the commits are small (3-4%) and not consistent: sometimes 17c8eb879466bdef3f77d5fbdf7b814a16ca2626 is faster, sometimes 2d7948612be3f411c7787829ff2db74e3a575f0c is faster. I've never observed 4ea3d850b53a449dda7af7ceb4cc565d3a64001c reported as faster.

Regarding maintainability, I think the first two commits (that walk the active query stack) are better; correctly maintaining `ZalsaLocal::active_cycles` is a bit finicky and took me a while to get right. Walking the active query stack is a more reliable and easy-to-maintain way to know which cycle heads are currently executing.

Based on this, I'm inclined to go back to either 17c8eb879466bdef3f77d5fbdf7b814a16ca2626 or 2d7948612be3f411c7787829ff2db74e3a575f0c and abandon 4ea3d850b53a449dda7af7ceb4cc565d3a64001c.

I welcome reviewer feedback here (and/or other suggestions for how to reclaim some perf.)

Veykril:
> Oddly, the benchmarks showing regression are not the one with cycles, but this fix should have little or no impact when there are not cycles. This suggests the regression might be from compilation, e.g. an inlining change? 
 
 I believe we are actually seeing allocator noise. You are changing the size of the `ActiveQuery` which is heap allocated so this hits the allocator differently (and regressions in those benches are also mostly in allocating paths). I've noticed this in a couple PRs that affect allocation size and/or order.

src/function/maybe_changed_after.rs
```diff
@@ -260,6 +263,33 @@ where
...
+        tracing::debug!(
+            "{database_key_index:?}: validate_same_iteration(memo = {memo:#?})",
+            memo = memo.tracing_debug()
+        );
+        for cycle_head in &memo.revisions.cycle_heads {
+            if !db.zalsa_local().with_query_stack(|stack| {
+                stack.iter().any(|entry| {
+                    entry.database_key_index == cycle_head.database_key_index
+                        && entry.iteration_count == cycle_head.iteration_count
+                })
```
Veykril, line 284: I wonder, wouldn't it be slightly better to iterate from the back here? I imagine our cycle should tend to be closer to the top of the stack than the bottom

Veykril (approved):
I agree with your choice of which strategy to prefer here 👍

src/cycle.rs
```diff
@@ -86,12 +86,19 @@ pub enum CycleRecoveryStrategy {
...
+    pub iteration_count: u32,
+}
+
+/// Any provisional value generated by any query in a cycle will track the cycle head(s) (can be
+/// plural in case of nested cycles) representing the cycles it is part of, and the current
+/// iteration count for each cycle head. This struct tracks these cycle heads.
 #[derive(Clone, Debug, Default)]
 #[allow(clippy::box_collection)]
-pub struct CycleHeads(Option<Box<Vec<DatabaseKeyIndex>>>);
+pub struct CycleHeads(Option<Box<Vec<CycleHead>>>);
```
Veykril, line 101: Unrelated to this PR (but something I just noticed we could investigate in a separate PR), we could check if using `ThinVec` here might be better than a `Box<Vec<_>>` (I think we have that dep nowadays)

src/function/fetch.rs
```diff
@@ -83,7 +83,8 @@ where
         if let Some(memo) = memo_guard {
             let database_key_index = self.database_key_index(id);
             if memo.value.is_some()
-                && self.validate_may_be_provisional(db, zalsa, database_key_index, memo)
+                && (self.validate_may_be_provisional(db, zalsa, database_key_index, memo)
+                    || self.validate_same_iteration(db, database_key_index, memo))
```
MichaReiser, line 87: I suspect that the performance regression comes from `fetch_hot` now requiring at least two function calls if the `database_key_index` isn't provisional. Can you try marking `self.validate_same_iteration` as `#[cold]` so as to signal the compiler that this branch is unlikely.

src/function/maybe_changed_after.rs
```diff
@@ -260,6 +263,33 @@ where
...
+    pub(super) fn validate_same_iteration(
+        &self,
+        db: &C::DbView,
+        database_key_index: DatabaseKeyIndex,
+        memo: &Memo<C::Output<'_>>,
+    ) -> bool {
+        tracing::debug!(
+            "{database_key_index:?}: validate_same_iteration(memo = {memo:#?})",
+            memo = memo.tracing_debug()
+        );
```
MichaReiser, line 278: I wonder how expensive the `tracing` calls are in a function as hot as `fetch_hot`. It might be worth reducing them to `trace` (They would be removed from the release bundle in red knot)

src/active_query.rs
```diff
@@ -60,6 +60,9 @@ pub(crate) struct ActiveQuery {
 
     /// Provisional cycle results that this query depends on.
     cycle_heads: CycleHeads,
+
+    /// If this query is a cycle head, iteration count of that cycle.
+    pub(crate) iteration_count: u32,
```
MichaReiser, line 65: Should we make this private and provide an accessor instead? It would make it clear that no-one is writing the field outside of `ActiveQuery`.

MichaReiser (approved):

carljm:
The Salsa benchmarks seem very noisy here and not useful. I just pushed two small changes (iterating active query stack in reverse, and making `ActiveQuery::iteration_count` private with accessor) which should not impact performance of any benchmark other than `converge_diverge` (the one with cycles, which already showed no impact), and now the perf regression on those other two input benchmarks is gone. I think this supports @Veykril theory that it's allocator noise and not really meaningful.

src/function/maybe_changed_after.rs
carljm: I did this anyway, "compiled out" seems best, and these are very in-the-weeds tracing outputs, I think having them at "trace" level is reasonable.

Veykril:
Also note that 1.86 released earlier today. So until you rebase onto latest main ( #781 ) you will likely bench between different rust versions.

carljm:
Rebased on latest main.

carljm:
Currently both Salsa and red-knot benchmarks are suggesting around a 1% regression here. I'm going to go ahead and merge; we can explore optimizations separately.

Issue #785, opened 2025-04-02
tracked-structs: access to field whilst the value is being initialized


MichaReiser:
We see more spurious panics in the `read_lock` function of `TrackedStructs`:

```
panicked at /home/runner/.cargo/git/checkouts/salsa-e6f3bb7c2a062968/d758691/src/tracked_struct.rs:824:21:
access to field whilst the value is being initialized
```

I haven't been able to create a minimal reproduction yet. We see it both in our tests (where we heavily relly on incrementality for faster tests) and in the playground (IDE like environment)

This might be solved by https://github.com/salsa-rs/salsa/pull/741

PR #788, opened 2025-04-09
fix: Dereferencing freed memos when verifying provisional memos

MichaReiser:
Fixes https://github.com/salsa-rs/salsa/issues/785

The root cause of #785 is that Salsa tries to load the memo of a cycle head that is stored on a tracked struct that has been freed. 

The main problem is that we tried to verify `validate_provisional` even in cases where the query is from a past revision, in which case we can't trust any inputs to still exist (and be valid). 

The fix in this PR is to change the order in which `shallow_verify` and `validate_provisional` are called. This requires extracting the *mutating* bits out of `shallow_verify` because we only want to call those if the struct isn't provisional (at least, that's what we did before and I tried to preserve this behavior).

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -60,10 +60,16 @@ where
 
             // Check if we have a verified version: this is the hot path.
             let memo_guard = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
-            if let Some(memo) = memo_guard {
-                if self.validate_may_be_provisional(db, zalsa, database_key_index, memo)
-                    && self.shallow_verify_memo(db, zalsa, database_key_index, memo)
-                {
+            let Some(memo) = memo_guard else {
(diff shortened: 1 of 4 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -433,14 +456,17 @@ where
                         );
 
                         if in_heads {
-                            // Iterate our dependency graph again, starting from the top. We clear the
-                            // cycle heads here because we are starting a fresh traversal. (It might be
-                            // logically clearer to create a new HashSet each time, but clearing the
-                            // existing one is more efficient.)
-                            cycle_heads.clear();
+                            if is_provisional {
+                                old_memo
```
MichaReiser, line 460: I don't think this is technically required because a later read will set the `finalize` flag but doing it here allows short-circuting in later reads (e.g. adding an `assert` here that `verified_final` is `true` panics.

PR #789, opened 2025-04-09
[fix] Use `validate_maybe_provisional` instead of `validate_provisional`

MichaReiser:
I accidentally changed the `validate_maybe_provisional` calls to `validate_provisional` in https://github.com/salsa-rs/salsa/pull/788. This is bad for performance because it prevents Salsa from short-circuiting if the memo is already finalized.


MichaReiser:
I confirmed on the Red Knot repo that this fixes the 3% incremental perf regression

PR #790, opened 2025-04-10
add WillIterateCycle event

carljm:
I would like to write tests asserting about cycles iterated and log cycles iterated as a metric. Add a `WillIterateCycle` event and fire it whenever a cycle head is about to iterate the cycle again. Update some cycle tests accordingly.

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -59,6 +59,11 @@ pub enum EventKind {
         database_key: DatabaseKeyIndex,
     },
 
+    WillIterateCycle {
+        /// The database-key for the cycle head. Implements `Debug`.
+        database_key: DatabaseKeyIndex,
+    },
+
     /// Indicates that `unwind_if_cancelled` was called and salsa will check if
     /// the current revision has been cancelled.
     WillCheckCancellation,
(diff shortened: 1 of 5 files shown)


src/event.rs
```diff
@@ -59,6 +59,11 @@ pub enum EventKind {
         database_key: DatabaseKeyIndex,
     },
 
+    WillIterateCycle {
+        /// The database-key for the cycle head. Implements `Debug`.
+        database_key: DatabaseKeyIndex,
```
MichaReiser, line 64: Should we also include the iteration count? I think I'd find this useful to see in logs.

MichaReiser (approved):
I think that's not just useful for third-party tests but also for our own. We don't have good infrastructure to set up tracing in tests, which is why I mainly use the db events to debug test errors (with occasional println sprinkled in every now and there)

MichaReiser (approved):
I think that's not just useful for third-party tests but also for our own. We don't have good infrastructure to set up tracing in tests, which is why I mainly use the db events to debug test errors (with occasional println sprinkled in every now and there). In fact, I had to add multiple println expressions to verify that my test used the cycle head that I intended.

src/event.rs
carljm: Yes, good idea

PR #791, opened 2025-04-10
[refactor] Reuse the same stack for all cycles heads in `validate_same_iteration`

MichaReiser:
Retrieve the `QueryStack` once (including `zalsa_local`) instead of once for each cycle head

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -293,18 +293,21 @@ where
             "{database_key_index:?}: validate_same_iteration(memo = {memo:#?})",
             memo = memo.tracing_debug()
         );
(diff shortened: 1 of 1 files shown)


MichaReiser:
Wow, that's unexpected. Maybe this is because we now avoid claiming the stack if there are no cycle heads?

carljm (approved):
Nice! That's odd because I thought I had tried this and seen no difference. But I didn't have the short-circuit if empty; maybe that is the difference? It doesn't quite make sense to me, because we should only enter this function for provisional memos, which mean they have cycle heads.

But anyway, the code looks fine, and certainly can't argue with the benchmark results!

davidbarsky:
Oh wow, this basically removes the (small, worthwhile!) performance penalty that #603 introduced.

PR #792, opened 2025-04-10
[refactor] Simplify `fetch_hot`

MichaReiser:
The basic promise of `fetch_hot` is to do the minimal amount of work for the common case and fall back to `fetch_cold` 
which repeats some of the checks for the rare cases where `fetch_hot` wasn't successful (because we failed to claim a guard, or a deep update is required).

This PR simplifies `fetch_hot` by moving the `same_iteration` check for cycle handling into `fetch_cold` (which has the advantage that `maybe_changed_after` also uses it). 
Moving the check wasn't as trivial as I hoped because it's important that we don't push the new query until after verifying if the memo is a provisional value from the same iteration. 

## Performance 
Codspeed reports a 2-3% perf improvement for most benchmarks.

There's a 1% perf improvement on Red Knot's incremental benchmarks https://codspeed.io/astral-sh/ruff/branches/micha%2Ftest-salsa-perf

### DIFF
--- a/src/function/fetch.rs
+++ b/src/function/fetch.rs
@@ -171,7 +171,7 @@ where
             ClaimResult::Claimed(guard) => guard,
         };
 
-        let mut active_query = LazyActiveQuery::new(database_key_index, db.zalsa_local());
+        let mut active_query = LazyActiveQueryGuard::new(database_key_index);
 
         // Now that we've claimed the item, check again to see if there's a "hot" value.
         let opt_old_memo = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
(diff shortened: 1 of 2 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -452,12 +456,10 @@ where
                             inputs,
                         );
 
-                        if is_provisional {
```
carljm, line 455: I guess it's cheaper to unconditionally update this, even when it's not necessary?

carljm (approved):
Looks good, thank you!

src/function/maybe_changed_after.rs
MichaReiser: I guess I could revert this now. I first split `deep_verify_memo` into two functions so that I didn't had access to `is_provisional` anymore. I'll revert to reduce the size of the change.

src/function/fetch.rs
```diff
@@ -192,8 +189,38 @@ where
...
+    pub(super) fn get(&mut self) -> &ActiveQueryGuard<'me> {
+        self.guard
+            .get_or_insert_with(|| self.zalsa_local.push_query(self.database_key_index, 0))
+    }
+
+    pub(super) fn into_inner(self) -> ActiveQueryGuard<'me> {
+        self.guard
+            .unwrap_or_else(|| self.zalsa_local.push_query(self.database_key_index, 0))
+    }
+}
```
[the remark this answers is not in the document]
MichaReiser: That's a good point and shouldn't be too annoying.

PR #793, opened 2025-04-11
[refactor] More `fetch_hot` simplification

MichaReiser:
Move the `validate_maybe_provisional` out of `fetch_hot` and only test if the memo is provisional (in which case we shouldn't mark the memo as updated). 
Rely on `fetch_cold` to lazily validate the memo. 

Codspeed thinks this is slightly better in some benchmarks (but the results change between runs). There's no meaningful change in Red Knot's benchmarks. I think it simplifies the logic a tiny bit. But I'm also okay not landing this change.

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -67,7 +67,7 @@ where
 
             if let Some(shallow_update) = self.shallow_verify_memo(zalsa, database_key_index, memo)
             {
-                if self.validate_may_be_provisional(db, zalsa, database_key_index, memo) {
+                if !memo.may_be_provisional() {
                     self.update_shallow(db, zalsa, database_key_index, memo, shallow_update);
 
                     return if memo.revisions.changed_at > revision {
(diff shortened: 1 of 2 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -252,7 +252,9 @@ where
     ) -> bool {
         // Wouldn't it be nice if rust had an implication operator ...
         // may_be_provisional -> validate_provisional
-        !memo.may_be_provisional() || self.validate_provisional(db, zalsa, database_key_index, memo)
+        !memo.may_be_provisional()
```
carljm, line 255: The doc comment for this method (above and unchanged in the PR, so I can't inline-comment on it directly) needs to be updated to mention the same-iteration caveat.

carljm (commented):
I agree this is nicer, since it means only `validate_may_be_provisional` has to know/care about `validate_same_iteration`

carljm (approved):

Issue #704, opened 2025-02-20
Panic with predictable value if there is no initial fn in cycle


nikomatsakis:
When there is a cycle with no initial value we currently panic with a string, but we should have a more recognizable value.

_Originally posted by @nikomatsakis in https://github.com/salsa-rs/salsa/pull/603#discussion_r1964188695_

Veykril:
Note that if we do this, we should not be panicking but unwinding with a such a value as to not print the now rather confusing `Box<dyn Any>` backtrace.

PR #796, opened 2025-04-16
Print query stack when encountering unexpected cycle

Veykril:
Example output
```
dependency graph cycle when validating impl_self_ty_shim(Id(4000)), set cycle_fn/cycle_initial to fixpoint iterate.
Query stack:
[
    infer_shim(Id(4400)),
    trait_solve_shim(Id(6800)),
    trait_impls_in_deps_shim(Id(1c00)),
    trait_impls_in_crate_shim(Id(1c00)),
    impl_trait_shim(Id(4000)),
    impl_trait_with_diagnostics_shim(Id(4000)),
    impl_self_ty_shim(Id(4000)),
    impl_self_ty_with_diagnostics_shim(Id(4000)),
]
```

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -258,10 +258,25 @@ impl ActiveQuery {
     }
 }
 
-#[derive(Debug, Default)]
+#[derive(Default)]
 pub(crate) struct QueryStack {
     stack: Vec<ActiveQuery>,
     len: usize,
+}
+
+impl std::fmt::Debug for QueryStack {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if f.alternate() {
+            f.debug_list()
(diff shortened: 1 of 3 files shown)


carljm (approved):
I think this is very useful, thank you!

Noting the potential overlap with https://github.com/salsa-rs/salsa/issues/704 -- @nikomatsakis felt that we should not panic with a string.

Personally I would find panicking with a non-string to be a major usability regression by default, because by default that shows up as just `Box<dyn Any>` in the panic message, unless you add extra code to your project to catch panics, pull out the value, and display it in a more useful way. And the use-case served by panicking with a non-string value (that you do want to catch panics and handle them in some way) is not a use case I've encountered so far.

But if we switch to panicking with a non-string value, presumably it should also include the query stack in some form.

carljm:
(Test failure here looks like the flaky failure from parallel_scope, which was already removed in another PR.)

PR #799, opened 2025-04-17
chore: Adjust `#[inline]` annotations

Veykril:
A lot of attached calls/definitions seem to hover around the default inlining limit making the codspeed reports difficult to interpret at times, generally these should benefit from being inlined either way as the "attached" code paths are quite hot.

Aside from those, this PR touches up on a lot of other call sites for instructing inlining behavior, as just changing the attached calls has a negative impact overall on the remainder of the inlining behaviors. These additional explicit annotations bring things back in line.

### DIFF
--- a/src/function/fetch.rs
+++ b/src/function/fetch.rs
@@ -3,7 +3,7 @@ use crate::function::{Configuration, IngredientImpl, VerifyResult};
 use crate::table::sync::ClaimResult;
 use crate::zalsa::{MemoIngredientIndex, Zalsa, ZalsaDatabase};
 use crate::zalsa_local::{ActiveQueryGuard, QueryRevisions, ZalsaLocal};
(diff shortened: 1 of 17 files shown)


src/function/fetch.rs
```diff
@@ -49,13 +49,7 @@ where
...
                 // thread completing fixpoint iteration of the cycle, and then we can re-query for
                 // our no-longer-provisional memo.
-                if !(memo.may_be_provisional()
-                    && memo.provisional_retry(
-                        db.as_dyn_database(),
-                        zalsa,
-                        self.database_key_index(id),
-                    ))
-                {
+                if !memo.provisional_retry(db, zalsa, self.database_key_index(id)) {
```
Veykril, line 52: `provisional_retry` itself already calls `may_be_provisional()`

Veykril:
I'll self-approve this, as this doesn't change behavior and hopefully makes CI benches more reliable.

PR #806, opened 2025-04-21
perf: Don't push an unnecessary active query for `deep_verify_memo`

Veykril:
Turns out the only thing that makes use of the pushed query within the scope of the push and drop of the guard in `deep_verify_memo` is `function::Ingredient::maybe_changed_after` which itself already pushes the query if it needs to. So all of this is actually unnecessary.


carljm (approved):
Looks reasonable, and tests are passing. Thank you!

PR #813, opened 2025-04-24
Use `DatabaseKey` for interned events

MichaReiser:
Similar to other events, use `DatabaseKey` for interned events.
This has the advantage that logging the event reveals the id *and the struct name*

## Test plan

See updated snapshot tests


MichaReiser:
I very much question the performance improvements 😆

Veykril (approved):

PR #816, opened 2025-04-24
shrink_to_fit `IdentityMap` before storing it

Veykril:
Notable we do not do this for cycles, given there the memo is likely temporary anyways

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -48,7 +48,7 @@ pub(crate) struct ActiveQuery {
 
     /// Map from tracked struct keys (which include the hash + disambiguator) to their
     /// final id.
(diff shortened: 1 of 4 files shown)


src/function/execute.rs
```diff
@@ -105,6 +105,7 @@ where
             // outputs and update the tracked struct IDs for seeding the next revision.
             self.diff_outputs(zalsa, database_key_index, old_memo, &mut revisions);
         }
+        revisions.tracked_struct_ids.shrink_to_fit();
```
MichaReiser, line 108: Should we move this call to `Memo::new`? It would then also apply for memos participating in cycles

PR #817, opened 2025-04-24
Access to tracked-struct that was freed during fixpoint

MichaReiser:
This adds a test for a fix point related bug. I'll write up more tomorrow. I don't know how to fix this...


MichaReiser:
Unclear why this would improve performance unless it's tracing

Veykril:
Formatting can have effects on codegen (that includes tracing, println etc) as that captures the address of locals. Looking at the flamegraph this perf is mainly due to inline differences here, see https://github.com/salsa-rs/salsa/pull/818

tests/cycle_output.rs
```diff
@@ -136,7 +136,7 @@ fn revalidate_no_changes() {
 
     db.assert_logs(expect![[r#"
         [
-            "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(401)) })",
+            "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(403)) })",
```
MichaReiser, line 139: The `id` change here is because Salsa reused the IDs between iterations, even though the `Output` had different values. Now, Salsa keeps all intermediate constructed `Output` structs around.

tests/cycle_output.rs
```diff
@@ -163,16 +163,15 @@ fn revalidate_with_change_after_output_read() {
...
             "salsa_event(WillExecute { database_key: query_d(Id(800)) })",
             "salsa_event(WillExecute { database_key: query_a(Id(0)) })",
-            "salsa_event(WillExecute { database_key: read_value(Id(400)) })",
+            "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(400)) })",
```
MichaReiser, line 169: Because salsa now keeps the intermediate tracked struct around, it can now also reuse the results on those queries. That means, this fix also improves incremental caching in cycles across revisions.

MichaReiser: Okay, I was a bit too optimistic. I think this optimization could lead to a memory leak :( I'll write a more detailed comment on the relevant line

src/function/execute.rs
```diff
@@ -54,6 +54,17 @@ where
             // previous execution as the starting point for the new one.
             if let Some(old_memo) = opt_old_memo {
                 active_query.seed_tracked_struct_ids(&old_memo.revisions.tracked_struct_ids);
+
+                // Copy over all outputs from a previous iteration.
+                // This is necessary to ensure that tracked struct created during the previous iteration
+                // (and are owned by the query) alive even if the query in this iteratoin no longer creates them.
+                // The query not re-creating the tracked struct doesn't guarantee that there
+                // aren't any other queries depending on it.
+                if old_memo.verified_at.load() == revision_now && old_memo.may_be_provisional() {
```
MichaReiser, line 63: My first version didn't had the check that `old_memo` is from the same revision. This had the nice side-effect of allowing us to remove the `provisional` from `remove_stale_output` and it unlocked reusing mid-cycle query results on tracked structs.

One such example is the `revalidate_with_change_after_output_read` test case where `query_a` creates a new `Output` in each iteration and calls `read_value(db, output)`. The intermediate `read_value` query results can't be reused across revisions if the check `old_memo.verified_at.load() == revision_now` exists because their `Output'`s only get created in later iterations of `query_a` and Salsa frees all `Output`s after the first iteration that haven't been created yet. 

We could remove the `old_memo.verified_at.load() == revision_now` restriction here but my concern is that this might lead to ever-growing output lists for intermediate queries that never get finalized. This shouldn't happen for most queries because they're likely to create the same outputs but this isn't guaranteed. 


What's not entirely clear to me is that we can't remove the `provisional` special handling in `remove_stale_output` if we have the check here. I believe it is because we end up reading some tracked structs when validating dependend tracked struct queries. This is unfortunate. It would have been nice if we could remove that special handling all together

MichaReiser:
Hmm, I can't reproduce the test failures locally.

tests/cycle_tracked_own_input.rs
```diff
@@ -0,0 +1,130 @@
...
+        node.name(db),
+        node.type_params(db).map(|tp| infer_type_param(db, tp)),
+    ))
+}
+
+#[salsa::tracked]
+fn infer_type_param<'db>(db: &'db dyn salsa::Database, node: TypeParamNode) -> TypeParam<'db> {
+    if let Some(constraint) = node.constraint(db) {
+        // Reuse the type param from the class if any.
+        // The example is a bit silly, because it's a reduction of what we have in Red Knot
```
carljm, line 66: Future Salsa maintainer: what is a "Red Knot"? 😆

carljm (approved):
Thank you for tracking this down!! The fix looks reasonable to me, and this looks ready to land to me, once we solve out the test failures.

src/function/execute.rs
```diff
@@ -54,6 +54,15 @@ where
             // previous execution as the starting point for the new one.
             if let Some(old_memo) = opt_old_memo {
                 active_query.seed_tracked_struct_ids(&old_memo.revisions.tracked_struct_ids);
+
+                // Copy over all outputs from a previous iteration.
+                // This is necessary to ensure that tracked struct created during the previous iteration
+                // (and are owned by the query) alive even if the query in this iteration no longer creates them.
```
Veykril, line 60:
```suggestion
                // (and are owned by the query) are alive even if the query in this iteration no longer creates them.
```

MichaReiser:
Hmm, I can't reproduce the test failure on my arch x86 system :(

src/function/execute.rs
```diff
@@ -54,6 +54,15 @@ where
...
+                // The query not re-creating the tracked struct doesn't guarantee that there
+                // aren't any other queries depending on it.
+                if old_memo.may_be_provisional() && old_memo.verified_at.load() == revision_now {
+                    active_query.seed_outputs(old_memo.revisions.origin.outputs());
```
Veykril, line 64: I am not sure about the test failure but given that the order of inputs and outputs for dependencies is fairly important, it might be more correct to seed the outputs *after* having executed the query such that the seeding does not influence the executions dependency order? Maybe that is causing issues

MichaReiser: Let me try this. I'd be surprised if it's the case because `input_outputs` isn't used for ID seeding. But who knows, maybe it changes `maybe_changed_after` execution order (although it shouldn't?)

Veykril: Same, but figured it might be worth a try. Seems to not have been the case though

MichaReiser:
What's also notable is that the identity map entries are ordered differently because local and CI

**local**
```
     tracked_struct_ids: IdentityMap {
        map: {
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 0,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(400),
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 12583873379513078615,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(401),
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 858131991118035463,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(403),
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 6721002685315557039,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(402),
        },
    },
```

**CI**

```
    tracked_struct_ids: IdentityMap {
        map: {
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 0,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(400),
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 18026959284135166402,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(403),
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 12157901119326311915,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(401),
            Identity {
                ingredient_index: IngredientIndex(
                    4,
                ),
                hash: 5869058164875963351,
                disambiguator: Disambiguator(
                    0,
                ),
            }: Id(402),
        },
```

MichaReiser:
Uff, found it. The problem was that discarding the tracked struct depended on the internal hash set ordering which means that the order of tracaked structs in the free list were different.

src/function/execute.rs
MichaReiser: I'll keep this regardless because I think it's nice to keep the input/output order as close as possible to the actual execution order.

Issue #683, opened 2025-02-09
Cryptic *missing `Update` implementation* compiler errors


MichaReiser:
A missing `Update` implementation for a salsa struct field or a query return type can read rather cryptic and it's not clear what a user should do (e.g. implement `Update` or remove the `&'db` reference. 

> I wonder if [doc.rust-lang.org/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-diagnosticon_unimplemented-attribute) is useful here? (unsure given we go through the fallback trick)
> @Veykril

detrumi:
Can you give a code example where this happens?

Veykril:
I believe we can consider this fixed by https://github.com/salsa-rs/salsa/pull/691

Veykril:
https://github.com/salsa-rs/salsa/pull/691 added a compile fail test for the return type part at least

PR #820, opened 2025-04-28
Add a compile-fail test for a `'static` `!Update` struct

Veykril:
Closes https://github.com/salsa-rs/salsa/issues/683

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -46,8 +46,8 @@ macro_rules! setup_tracked_fn {
         // If true, this is specifiable.
         is_specifiable: $is_specifiable:tt,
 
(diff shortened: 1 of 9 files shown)


MichaReiser (approved):
Nice

Can you verify that this change also improves the error message when using a `NotUpdate` type as a tracked/interned-struct field because that would be required to entirely solve https://github.com/salsa-rs/salsa/issues/683

Veykril:
Ah right, I was somehow only focusing on the return type case where the proposed attribute didn't make a difference. Let me check the field case.

Veykril:
Added a test for a tracked field. Its not the best message, but I think its informative enough? Improving this message is really difficult with the macro_rules quoting setup as getting the spans right would require moving a lot of it back into the proc-macro.

MichaReiser (commented):
That's much better! I think that should do it for now

PR #821, opened 2025-04-28
cleanup: Remove unnecessary `Array` abstraction

Veykril:


MichaReiser (approved):

MichaReiser:
I wonder what broke codspeed. Something is definitely off now (or salsa becomes 5% faster with every PR :P)

PR #823, opened 2025-04-28
Attempt to fix codspeed

Veykril:
I removed this last week but I think that made codspeed no longer update its baseline


MichaReiser (approved):
Ohh, that could be it!

PR #826, opened 2025-04-28
fix: correct debug output for tracked fields

### LINKED ISSUE #814: debug name of tracked struct fields is off
```
#[salsa::tracked(debug)]
pub struct Definition<'db> {
    /// The file in which the definition occurs.
    pub(crate) file: File,

    /// The scope in which the definition occurs.
    pub(crate) file_scope: FileScopeId,

    /// The symbol defined.
    pub(crate) symbol: ScopedSymbolId,

    /// WARNING: Only access this field when doing type inference for the same
    /// file as where `Definition` is defined to avoid cross-file query dependencies.
    #[no_eq]
    #[return_ref]
    #[tracked]
    pub(crate) kind: DefinitionKind<'db>,

    /// This is a dedicated field to avoid accessing `kind` to compute this value.
    pub(crate) is_reexported: bool,

    count: countme::Count<Definition<'static>>,
}
```

`report_tracked_read` incorrectly reports a read on `definition.file` when accessing the `kind` field. I think it is because `TRACKED_FIELD_INDICES` stores relative indices but `DEBUG_NAMES` lists all field names.

I don't think we need all `DEBUG_NAMES`. A fix could be to only add the tracked fields to `DEBUG_NAMES`.

gvozdvmozgu:
close #814

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_struct.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_struct.rs
@@ -109,8 +109,8 @@ macro_rules! setup_tracked_struct {
             impl $zalsa_struct::Configuration for $Configuration {
                 const DEBUG_NAME: &'static str = stringify!($Struct);
 
-                const FIELD_DEBUG_NAMES: &'static [&'static str] = &[
-                    $(stringify!($field_id),)*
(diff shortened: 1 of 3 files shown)


MichaReiser (approved):
Nice, thank you

PR #827, opened 2025-04-29
feat: Implement a query stack `Backtrace` analog

Veykril:
Example output:
```
query stacktrace:
   0: query_e(Id(3)) -> (R1, Durability::LOW)
             at tests/backtrace.rs:30
   1: query_cycle(Id(3)) -> (R1, Durability::HIGH, iteration = 0)
             at tests/backtrace.rs:43
             cycle heads: query_cycle(Id(3)) -> 0
   2: query_f(Id(3)) -> (R1, Durability::HIGH)
             at tests/backtrace.rs:38
```

### DIFF
--- a/tests/backtrace.rs
+++ b/tests/backtrace.rs
@@ -1,0 +1,127 @@
+use expect_test::expect;
+use salsa::{Backtrace, Database, DatabaseImpl};
+use test_log::test;
+
+#[salsa::input(debug)]
+struct Thing {
+    detailed: bool,
+}
+
+#[salsa::tracked]
+fn query_a(db: &dyn Database, thing: Thing) -> String {
+    query_b(db, thing)
+}
+
+#[salsa::tracked]
+fn query_b(db: &dyn Database, thing: Thing) -> String {
+    query_c(db, thing)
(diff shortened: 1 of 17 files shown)


tests/backtrace.rs
```diff
@@ -0,0 +1,127 @@
...
+                     at tests/backtrace.rs:25
+           4: query_e(Id(0))
+                     at tests/backtrace.rs:30
+    "#]]
+    .assert_eq(&backtrace);
+
+    let backtrace = query_a(&db, Thing::new(&db, true)).replace("\\", "/");
+    expect![[r#"
+        query stacktrace:
+           0: query_a(Id(1)) -> (R1, Durability::HIGH)
```
MichaReiser, line 89: Wow, this is so useful that I sort of want this for non-tests too :)

Veykril: This is pretty nice actually

MichaReiser: I find it more useful than most tracing we have in place today.

MichaReiser (approved):
This is great!

PR #833, opened 2025-04-30
refactor: Clean up `function::execute`

Veykril:
it's gotten a bit tough to reason about imo due to the different execution modes we have now

### DIFF
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -1,11 +1,11 @@
-use std::sync::atomic::Ordering;
+use std::sync::atomic::{AtomicBool, Ordering};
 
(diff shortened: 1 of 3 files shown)


src/function/execute.rs
```diff
@@ -39,199 +38,244 @@ where
...
+        self.insert_memo(
+            zalsa,
+            id,
+            Memo::new(Some(new_value), zalsa.current_revision(), revisions),
+            memo_ingredient_index,
+        )
+    }
+
+    #[inline]
+    fn execute_maybe_cycle<'db>(
```
MichaReiser, line 127:
```suggestion
    fn execute_maybe_iterate<'db>(
```

MichaReiser (approved):

PR #835, opened 2025-04-30
bug: Avoid panic in `Backtrace::catpure` if `query_stack` is already mutably borrowed

MichaReiser:
This PR fixes a panic in `Backtrace::catpure` if `local_state.query_stack` is already mutably borrowed by
returning `None`. 

This was possible because some of our cycle queries used `with_query_stack` when panicking to include the query stack
and `with_query_stack` always aquired a mutable borrow. I split `with_query_stack` into a reference, mutable reference, and `try` variant
which should avoid this from triggering in most places


Veykril:
That does make this less useful for panic handlers (noticed r-a always aborts now when we panic thanks to the double panic haha)

Issue #834, opened 2025-04-30
assertion failed: old_memo.revisions.changed_at <= revisions.changed_at


Veykril:
https://github.com/rust-lang/rust-analyzer/issues/19715
> **rust-analyzer version**: 0.3.2441-standalone ([d8887c0](https://github.com/rust-lang/rust-analyzer/commit/d8887c0758bbd2d5f752d5bd405d4491e90e7ed6) 2025-04-26) [c:\Users\~\.vscode\extensions\rust-lang.rust-analyzer-0.3.2441-win32-x64\server\rust-analyzer.exe]
> 
> **rustc version**: rustc 1.86.0 (05f9846f8 2025-03-31)
> 
> **editor or extension**: Visual Studio Code Version: 1.99.3 (user setup) Commit: 17baf841131aa23349f217ca7c570c76ee87b957 Date: 2025-04-15T23:18:46.076Z Electron: 34.3.2 ElectronBuildId: 11161073 Chromium: 132.0.6834.210 Node.js: 20.18.3 V8: 13.2.152.41-electron.0 OS: Windows_NT x64 10.0.26100 Extension rust-analyzer 0.3.2441
> 
> **relevant settings**: -
> 
> **repository link (if public, optional)**: [respectZ/oxide_eval](https://github.com/respectZ/oxide_eval/)
> 
> **code snippet to reproduce**: -
> 
> Log:
> 
> [rust-analyzer Language Server.log](https://github.com/user-attachments/files/19959339/rust-analyzer.Language.Server.log)
> 
> I'm not sure what's causing this, but this specific log caught my attention:
> 
> ```
> thread 'Worker' panicked at C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\salsa-0.20.0\src\function\backdate.rs:31:17:
> assertion failed: old_memo.revisions.changed_at <= revisions.changed_at
> ```
> 
> Did I mess something up? Or this is salsa issue? If I opened my `lib.rs` it's just loading till panicked. It's also taking up huge amount of memory.
> 
> ![Image](https://github.com/user-attachments/assets/1a1ccf38-c02f-4dc8-a6cf-ab7c1d97943d)
> 
> Any leads / helps appreciated, thanks!

MichaReiser:
Relevant salsa thread 

https://salsa.zulipchat.com/#narrow/channel/333573-Contributing-to-Salsa/topic/Another.20fixpoint.20bug/with/515074372

PR #836, opened 2025-04-30
fix: change detection for fixpoint queries

MichaReiser:
This PR fixes two bugs in `maybe_changed_after` for cyclic queries

## Fix number 1

Salsa doesn't invalidate a cycle of `query_a` -> `query_b` if `query_b` only depends on the input in the first iteration. 

Fixing this requires that we carry over inputs to later iterations. This is a very similar fix to what I did for tracked structs. 

This fixes https://github.com/salsa-rs/salsa/issues/834.

## Fix number 2

We saw many *access to interned struct that wasn't interned in this revision* panics. This was caused by `maybe_changed_after` re-running a query in the attempt to backdate it even when the query participates in a cycle (in which case salsa returned *Fixpoint initial* on which the query that Salsa tries to backdate now depends on). This is problematic the assumption before calling `execute` is that all its inputs are *valid* but this isn't the case when returning *fixpoint initial*, Salsa then intentionally defers validating all the input/outputs. 

The fix here is to only perform the backdating if not in a cycle. If incide a cycle, then the backdating must happen for the cycle head (which then calls `maybe_changed_after` of inner queries again)

## Performance

[The performance results on ty](https://codspeed.io/astral-sh/ruff/branches/micha%2Fperf-fixpoint-fix) (former Red Knot) are neutral. The performance should mainly regress if cycles have many different inputs/outputs between iterations because they're now carried over


MichaReiser:
It's not entirely clear what behavior we want for accumulated values. E.g. `a` depends on `b` but only in its first iteration and `b` emits an accumulated value. Is that value now returned when retrieving the accumulated values of `a`?

If no, then we'll need a way to distinguish between dependencies from the current and previous iterations.

MichaReiser:
I think the reason why the fix is unsafe is because it's a possibility that the default value is wrong (a leftover from changing the signature from `int | None` to `None`. A user has to explicitly express their intention.

tests/cycle_output.rs
```diff
@@ -170,29 +173,29 @@ fn revalidate_with_change_after_output_read() {
     db.assert_logs(expect![[r#"
         [
             "salsa_event(DidSetCancellationFlag)",
-            "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(403)) })",
+            "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(400)) })",
+            "salsa_event(DidReinternValue { key: query_d::interned_arguments(Id(800)), revision: R2 })",
+            "salsa_event(WillExecute { database_key: query_b(Id(0)) })",
```
MichaReiser, line 178: This test captures fix 2. Salsa tries to see if `query_b` changed (cycle head). `query_b` depends on `query_a` and `query_a` depends on `query_b` and `query_d`. 

Salsa returns a provisional value for `query_b`.

**Before**

`maybe_changed_after`:

* `query_b`: Unchanged
	* `query_a`: 
  	* Calls `maybe_changed_after` for `query_b`: Returns `Unchanged` (but fixpoint initial)
  	* Calls `maybe_changed_after` for `query_d`: Detects that one input changed. Tries backdating by executing the query (which isn't safe because it may depend on a struct interned by `query_b` which was never fully verified

**Now**

`query_d` now returns `Changed` without trying to backdate, same for `query_a` because both see that they're part of a cycle. Salsa then executes `query_b`, which may get backdated.

src/function/maybe_changed_after.rs
```diff
@@ -369,20 +411,22 @@ where
...
                 // where the value is specified
                 // in rev 1 but not in rev 2.
-                VerifyResult::Changed
+                VerifyResult::changed()
+            }
+            QueryOrigin::FixpointInitial if old_memo.may_be_provisional() => {
+                VerifyResult::changed()
             }
-            QueryOrigin::FixpointInitial if old_memo.may_be_provisional() => VerifyResult::Changed,
             QueryOrigin::FixpointInitial => VerifyResult::unchanged(),
```
MichaReiser, line 419: @carljm I'm not 100% sure about the two fixpoint arms here.

tests/cycle_output.rs
carljm: And I guess since you reverted the earlier (not really right) change I'd made to avoid having output-validation fail in this case, this test will also now fail without the other fix in this PR (and not only due to the log assertions.)

tests/cycle_maybe_changed_after.rs
```diff
@@ -0,0 +1,213 @@
...
+    } else {
+        value
+    }
+}
+
+fn query_a_initial(_db: &dyn Database, _input: Input) -> u32 {
+    0
+}
+
+fn query_a_recover(
```
carljm, line 42: Maybe these shouldn't be named `query_a_*` since they are used by both `query_a` and `query_d`? Looks a bit weird above.

src/function/maybe_changed_after.rs
carljm: I'm not either :) I don't have a memory of why this is this way -- maybe it was needed in some much-earlier version of fixpoint iteration?

Locally it seems that replacing both arms with this doesn't fail any tests, and it feels more clearly correct (but I'd want to also check it for perf regression on ty):

```
QueryOrigin:FixpointInitial => VerifyResult::changed(),
```

carljm (approved):
So good!!

This all makes sense to me, fantastic work.

src/function/maybe_changed_after.rs
MichaReiser: This makes me wonder if we even need this extra variant or if we could just use `Derived` everywhere because we also have `revisions.maybe_changed_after` to identify provisional values.

PR #837, opened 2025-04-30
better debug name for interned query arguments

MichaReiser:
The debug name `Configuration` isn't very helpful when debugging salsa. Use a name constructed from the query name.

I'm open to suggestions on how to avoid collisions with actual query names. e.g. we could prefix the name with two underscores or similar.

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -139,7 +139,7 @@ macro_rules! setup_tracked_fn {
                             file: file!(),
                             line: line!(),
                         };
-                        const DEBUG_NAME: &'static str = "Configuration";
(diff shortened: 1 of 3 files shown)


components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -139,7 +139,7 @@ macro_rules! setup_tracked_fn {
                             file: file!(),
                             line: line!(),
                         };
-                        const DEBUG_NAME: &'static str = "Configuration";
+                        const DEBUG_NAME: &'static str = concat!(stringify!($fn_name), "_arguments");
```
davidbarsky, line 142: maybe something like?

```suggestion
                        const DEBUG_NAME: &'static str = concat!(stringify!($fn_name), "_interned_args");
```

(In general, I'm a +1 on this change.)

davidbarsky (approved):

PR #839, opened 2025-04-30
Simple LRU garbage collection for interned values

ibraheemdev:
Basic implementation of https://github.com/salsa-rs/salsa/issues/597.

The garbage collection strategy used here is very simple.
- Slots that have not been accessed in the last 3 *active revisions* are always reused, and nothing else. Active revisions are those in which interned values were actually read to avoid counting empty revisions.
- Slots are never swept proactively, though this would be simple to add as an option.
- We also cannot collect low durability values, which would require `&mut db`. This could be supported with the above.

This is by no means a perfect strategy, but we can always tune it as needed once the infrastructure is in place. We may also want to expose the `N` active revisions constant as a configuration option in the `#[interned]` macro (it seems somewhat related to durability).

The LRU list is sharded by hash, very similar to dashmap. There seems to be no performance difference from the previous implementation in multi-threaded `ty` benchmarks (though there was a significant regression without sharding). There is also no significant performance difference in single-threaded benchmarks thanks to https://github.com/salsa-rs/salsa/pull/864, which lets us avoid additional read dependencies.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -151,6 +151,19 @@ struct ValueShared {
     durability: Durability,
 }
 
+impl ValueShared {
+    /// Returns `true` if this value slot can be reused when interning, and should be added to the LRU.
+    fn is_reusable(&self) -> bool {
+        // Collecting higher durability values requires invalidating the revision for their
+        // durability (see `Database::synthetic_write`, which requires a mutable reference to
+        // the database) to avoid short-circuiting calls to `maybe_changed_after`. This is
+        // necessary because `maybe_changed_after` for interned values is not "pure"; it updates
+        // the `last_interned_at` field before validating a given value to ensure that it is not
+        // reused after read in the current revision.
+        self.durability == Durability::LOW
+    }
+}
+
 impl<C> Value<C>
 where
     C: Configuration,
@@ -322,15 +335,17 @@ where
                     })
                 });
 
-                // Move the value to the front of the LRU list.
-                //
-                // SAFETY: We hold the lock for the shard containing the value, and `value` was
-                // previously interned, so is in the list.
-                unsafe { shard.lru.cursor_mut_from_ptr(value).remove() };
+                if value_shared.is_reusable() {
+                    // Move the value to the front of the LRU list.
+                    //
+                    // SAFETY: We hold the lock for the shard containing the value, and `value` is
+                    // a reusable value that was previously interned, so is in the list.
+                    unsafe { shard.lru.cursor_mut_from_ptr(value).remove() };
 
-                // SAFETY: The value pointer is valid for the lifetime of the database
-                // and never accessed mutably directly.
-                unsafe { shard.lru.push_front(UnsafeRef::from_raw(value)) };
+                    // SAFETY: The value pointer is valid for the lifetime of the database
+                    // and never accessed mutably directly.
+                    unsafe { shard.lru.push_front(UnsafeRef::from_raw(value)) };
+                }
             }
 
             // Record the maximum durability across all queries that intern this value.
@@ -379,12 +394,6 @@ where
             // will not find any stale slots.
             if !self.revision_queue.is_stale(value_shared.last_interned_at) {
                 break;
-            }
-
-            // We can't collect higher durability values until we have a mutable reference to the database.
-            if value_shared.durability != Durability::LOW {
-                cursor.move_prev();
-                continue;
             }
 
             // We should never reuse a value that was accessed in the current revision.
@@ -484,11 +493,13 @@ where
             // correct type.
             unsafe { self.clear_memos(zalsa, &mut memo_table, new_id) };
 
-            // Move the value to the front of the LRU list.
-            //
-            // SAFETY: The value pointer is valid for the lifetime of the database.
-            // and never accessed mutably directly.
-            shard.lru.push_front(unsafe { UnsafeRef::from_raw(value) });
+            if value_shared.is_reusable() {
+                // Move the value to the front of the LRU list.
+                //
+                // SAFETY: The value pointer is valid for the lifetime of the database.
+                // and never accessed mutably directly.
+                shard.lru.push_front(unsafe { UnsafeRef::from_raw(value) });
+            }
 
             return new_id;
         }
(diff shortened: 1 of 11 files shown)


MichaReiser:
> I'm seeing a lot of panics in red-knot, mostly Data was not interned in the latest revision for its durability. I'm not sure exactly why this is happening, but I suspect it's related to durabilities.

Interesting. I suspect this is new but I'm investigating a similar issue related to interned in cycles

MichaReiser:
That makes me wonder if your PR might help to come up with simpler repros for the fixpoint issues that we're seeing. I should give this a try tomorrow.

MichaReiser:
I quickly tried the branch in Red Knot but I also got new panics that we didn't see before. That makes me suspect that some revision tracking might be off in this PR?

ibraheemdev:
I think the issue is unrelated to durabilities but about interned query arguments. If a query takes an interned struct as an argument then an older memo may be incorrectly verified after the slot has been reused, but the IDs still match. I think we have to check that the memo and first_interned_at revisions match to avoid this, but I'm not exactly sure where that would be.

MichaReiser:
I think I understand. What you're saying is that `query(interned)` may be checked for changes in a later revision where `interned` has been collected.

I think will need something similar to what we have in tracked structs where salsa informs other ingredients that the memo was collected so that they can free resources as well if necessary:

https://github.com/MichaReiser/salsa/blob/953dc34ce5cfd7fca6ad040241c3cbebc11830e7/src/tracked_struct.rs#L597-L655

The second part in tracked struct that protects against this is that we separately track `created_at` and `updated_at`. `maybe_changed_after` can then compare `created_at` to see if this struct was recrated since `rev_before`.

ibraheemdev:
I updated the PR to add read dependencies and clear memos on reuse (I believe we will need to defer reclamation of the memos), but there are a still some ty tests failing that I'm unsure about.

ibraheemdev:
Hmm I'm not sure we can remove the read dependency on creation with generational IDs, we still need `last_interned_at` to be updated for any dependent queries. Maybe it would be cheaper to update that field when a value is read?

ibraheemdev:
I updated the PR description with the new changes, this should be ready for review.

src/interned.rs
```diff
@@ -43,60 +45,116 @@ pub struct JarImpl<C: Configuration> {
...
 where
     C: Configuration,
 {
-    fields: C::Fields<'static>,
-    memos: MemoTable,
+    /// The index of the shard containing this value.
+    shard: u16,
+
+    /// An intrusive linked list for LRU.
+    link: LinkedListLink,
```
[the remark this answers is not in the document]
ibraheemdev: Yes but the per-ingredient list is intrusive instead of being allocated separately.

src/interned.rs
```diff
@@ -400,29 +732,37 @@ where
...
-            // The slot was reused.
+
+        // SAFETY: `value.shard` is guaranteed to be in-bounds for `self.shards`.
+        let _shard = unsafe { self.shards.get_unchecked(value.shard as usize) }.lock();
+
+        // SAFETY: We hold the lock for the shard containing the value.
+        let value_shared = unsafe { &mut *value.shared.get() };
+
+        // The slot was reused.
+        if value_shared.first_interned_at > revision {
```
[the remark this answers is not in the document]
ibraheemdev: I think you're right, we don't actually need this anymore.

src/interned.rs
```diff
@@ -189,185 +277,429 @@ where
...
+                // and never accessed mutably directly.
+                unsafe { shard.lru.push_front(UnsafeRef::from_raw(value)) };
             }
+
+            // Record the maximum durability across all queries that intern this value.
+            if let Some((_, stamp)) = zalsa_local.active_query() {
+                value_shared.durability = std::cmp::max(value_shared.durability, stamp.durability);
+            }
+
+            // Record a dependency on the value.
```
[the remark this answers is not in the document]
ibraheemdev: I added a comment on the `id` field of `ValueShared`.

src/interned.rs
```diff
@@ -396,33 +737,41 @@ where
...
-            // The slot was reused.
+
+        // SAFETY: `value.shard` is guaranteed to be in-bounds for `self.shards`.
+        let _shard = unsafe { self.shards.get_unchecked(value.shard as usize) }.lock();
+
+        // SAFETY: We hold the lock for the shard containing the value.
+        let value_shared = unsafe { &mut *value.shared.get() };
+
+        // The slot was reused.
+        if value_shared.id.generation() > input.generation() {
```
MichaReiser, line 773: Do we need this check, given that the `id` changes when the generation changes?

(which in turn means that no existing query can depend on the now "new" interned value)

src/interned.rs
```diff
@@ -189,185 +284,431 @@ where
...
+            // soundly, but we also do not want to collect values that have been read recently.
+            //
+            // Note that the list is sorted by LRU, so if the tail of the list is not stale, we
+            // will not find any stale slots.
+            if !self.revision_queue.is_stale(value_shared.last_interned_at) {
+                break;
+            }
 
-                id
+            // We can't collect higher durability values until we have a mutable reference to the database.
```
Veykril: Why is that? (I did not read up on the discussions on the PR) Would be good to add that to the comment?

src/interned.rs
```diff
@@ -452,12 +801,86 @@ where
...
-        &self.memos
+        // SAFETY: The fact that we have a reference to the `Value` means it must
+        // have been interned, and thus validated, in the current revision.
+        unsafe { &*self.memos.get() }
     }
 
     #[inline(always)]
     fn memos_mut(&mut self) -> &mut MemoTable {
-        &mut self.memos
+        // SAFETY: We have `&mut self`.
```
Veykril: There is no unsafe operation here?

src/interned.rs
ibraheemdev: I think this check is still needed, because it allows queries that depend on the *old* value to see that the slot was reused. There is a test in ty that fails if I remove it.

src/interned.rs
ibraheemdev: Added a comment.

src/interned.rs
```diff
@@ -43,60 +45,136 @@ pub struct JarImpl<C: Configuration> {
...
+    /// The revision the value was most-recently interned in.
+    last_interned_at: Revision,
 
     /// The minimum durability of all inputs consumed by the creator
-    /// query prior to creating this tracked struct. If any of those
+    /// query prior to creating this interned struct. If any of those
     /// inputs changes, then the creator query may create this struct
     /// with different values.
-    durability: AtomicU8,
+    durability: Durability,
```
Veykril, line 151: Note that this field wastes 8 bytes while only occupying a single byte due to padding and alignment of `ValueShared`

PR #842, opened 2025-05-06
Add loom support

ibraheemdev:
Doesn't add any tests yet, but should still be good to merge to avoid having to rebase. I think the one thing missing here is a polyfill for crossbeam-queue. I also added a workflow to make sure everything compiles under cfg(loom).

cc @davidbarsky

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -59,6 +59,8 @@ jobs:
         run: cargo test --workspace --all-features --doc
       - name: Check (without default features)
         run: cargo check --workspace --no-default-features
+      - name: Check (loom)
+        run: RUSTFLAGs="--cfg loom" cargo check --workspace
 
   miri:
     name: Miri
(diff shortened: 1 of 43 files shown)


davidbarsky:
Nice! Thanks for getting it over the line.

davidbarsky (approved):

.github/workflows/test.yml
```diff
@@ -59,6 +59,8 @@ jobs:
         run: cargo test --workspace --all-features --doc
       - name: Check (without default features)
         run: cargo check --workspace --no-default-features
+      - name: Check (loom)
+        run: RUSTFLAGs="--cfg loom" cargo check --workspace
```
MichaReiser, line 63: Could you add a `just` script for loom similar to miri (and test)

MichaReiser (approved):
Nice, thank you

Issue #831, opened 2025-04-30
assertion failed: existing.iteration_count == head.iteration_count


Veykril:
https://github.com/rust-lang/rust-analyzer/issues/19717
> **rust-analyzer version**: rust-analyzer version: `0.4.2442-standalone`
> 
> **rustc version**: `rustc 1.86.0 (05f9846f8 2025-03-31)`
> 
> **editor or extension**: VSCode
> 
> **repository link (if public, optional)**: [ruffle-rs/ruffle](https://github.com/ruffle-rs/ruffle/)
> 
> One way to reproduce is to try to hover over the `ui` parameter of `show_avm2_button` in `core/src/debug_ui/display_object.rs`
> ## Backtrace
> 
> ```
> thread 'Worker' panicked at /github/home/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/salsa-0.20.0/src/cycle.rs:164:17:
> assertion failed: existing.iteration_count == head.iteration_count
> stack backtrace:
>    0: rust_begin_unwind
>    1: core::panicking::panic_fmt
>    2: core::panicking::panic
>    3: salsa::active_query::ActiveQuery::add_read
>    4: salsa::function::fetch::<impl salsa::function::IngredientImpl<C>>::fetch
>    5: <DB as hir_ty::db::HirDatabase>::variances_of::variances_of_shim
>    6: hir_ty::variance::Context::add_constraints_from_ty
>    7: hir_ty::variance::Context::add_constraints_from_ty
>    8: hir_ty::variance::Context::solve::{{closure}}
>    9: hir_ty::variance::variances_of
>   10: salsa::function::execute::<impl salsa::function::IngredientImpl<C>>::execute
>   11: salsa::function::fetch::<impl salsa::function::IngredientImpl<C>>::fetch_cold
>   12: salsa::function::fetch::<impl salsa::function::IngredientImpl<C>>::fetch
>   13: <DB as hir_ty::db::HirDatabase>::variances_of::variances_of_shim
>   14: hir_ty::variance::Context::add_constraints_from_ty
>   15: hir_ty::variance::Context::add_constraints_from_ty
>   16: hir_ty::variance::Context::solve::{{closure}}
>   17: hir_ty::variance::variances_of
>   18: salsa::function::execute::<impl salsa::function::IngredientImpl<C>>::execute
>   19: salsa::function::fetch::<impl salsa::function::IngredientImpl<C>>::fetch_cold
[...]

PR #843, opened 2025-05-06
fix: incorrect caching for queries participating in fixpoint

MichaReiser:
This is a follow up to https://github.com/salsa-rs/salsa/pull/836 and may fix the hang in https://github.com/salsa-rs/salsa/issues/831

I'm very sorry for not adding any tests but I'm very short on time right now. I'll create an issue if we go ahead with
merging this change to follow up with tests


MichaReiser:
The perf change is interesting, considering that we're invalidating more often but maybe it's because we return `Unchanged` for initial

carljm (approved):
Thank you!

carljm:
@MichaReiser note I rebased this on latest main branch to unblock @dhruvmanila (needed the loom gating fix)

Veykril:
I just checked, this does not fix the hang for us unfortunately.

MichaReiser:
> I just checked, this does not fix the hang for us unfortunately.

Hmm, that's sad

MichaReiser:
I created an issue to explore renabling backdating https://github.com/salsa-rs/salsa/issues/847

PR #845, opened 2025-05-06
Add a basic loom test

ibraheemdev:
Fixes a couple issues when running under loom and adds a basic concurrency test. Unfortunately, this test is still very slow (it ran for >45 minutes and >20M iterations before I killed it on my machine), so I didn't add it to CI yet. It will probably only be practical for us to run very fine-grained unit tests under loom, or switch to something like shuttle for larger tests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,8 @@ salsa-macros = { version = "0.21.1", path = "components/salsa-macros", optional 
 boxcar = "0.2.11"
 crossbeam-queue = "0.3.11"
 dashmap = { version = "6", features = ["raw-api"] }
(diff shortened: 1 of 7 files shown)


MichaReiser:
It might be a flake but any idea why the benchmarks are regressing

ibraheemdev:
It seems like a flake. It's possible that the compiler isn't inlining the `assume_init_ref` calls but that seems unlikely.

MichaReiser (approved):
Thanks

I don't understand much of loom but this looks reasonable. 

The fact that not even the most simple tests complete in reasonable time makes me question the value of loom and whether it is worth the complexity (and indirection) it introduces. 

What do you think?

PR #849, opened 2025-05-07
Move salsa event system into `Zalsa`

Veykril:
The existence of `Database::salsa_event` is incredibly annoying as a lot of internal code takes a `&dyn Database` solely to do event reporting where as usually a `&Zalsa` handle would otherwise suffice. This becomes even more annoying when mutable handles are involved as it requires some odd code structuring to avoid mutability xor shared errors, it also makes optimizations more difficult for the compiler due to dynamic dispatches it can't look through.

So this moves the trait method away to a separate callback that can be registered, and if not registered, can be skipped entirely.

### DIFF
--- a/components/salsa-macro-rules/src/setup_accumulator_impl.rs
+++ b/components/salsa-macro-rules/src/setup_accumulator_impl.rs
@@ -23,13 +23,12 @@ macro_rules! setup_accumulator_impl {
 
             // Suppress the lint against `cfg(loom)`.
             #[allow(unexpected_cfgs)]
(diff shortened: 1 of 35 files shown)


MichaReiser (approved):
I didn't do a line by line review but this looks great.

MichaReiser:
Huh, I'm pretty sure I wrote one more comment but it seems it didn't make it. 

Roughly, it would be nice if `Storage::new` could remain argument free and we would isntead have a `with_event_logger` mutator method. I think the challenge with this is that `zalsa` is wrapped in an `Arc`. Would it be possible to use our normal cancellation to get a reference to storage and swap the evnet handler?

Veykril:
Ye that should be doable

PR #850, opened 2025-05-07
Fix memo table growth condition

ibraheemdev:
The `additional_len` calculation assumes that `memos.len() <= memo_ingredient_index`, which is not true in light of concurrent inserts. We're seeing spurious overflow panics because of this. It looks like this was changed in https://github.com/salsa-rs/salsa/pull/803.


carljm (approved):
Thank you!!

PR #851, opened 2025-05-08
Do not re-verify already verified memoized value in cycle verification 

Veykril:
Stacked on top of https://github.com/salsa-rs/salsa/pull/852

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -19,9 +19,6 @@ pub enum VerifyResult {
     ///
     /// The inner value tracks whether the memo or any of its dependencies have an
     /// accumulated value.
-    ///
-    /// Don't mark memos verified until we've iterated the full cycle to ensure no inputs changed
-    /// when encountering this variant.
     Unchanged(InputAccumulatedValues),
 }
 
@@ -36,10 +33,6 @@ impl VerifyResult {
 
     pub(crate) fn unchanged() -> Self {
         Self::Unchanged(InputAccumulatedValues::Empty)
-    }
-
-    pub(crate) const fn is_unchanged(&self) -> bool {
-        matches!(self, Self::Unchanged(_))
     }
 }
 
@@ -146,11 +139,11 @@ where
         // Check if the inputs are still valid. We can just compare `changed_at`.
         let deep_verify =
             self.deep_verify_memo(db, zalsa, old_memo, database_key_index, cycle_heads);
-        if deep_verify.is_unchanged() {
+        if let VerifyResult::Unchanged(accumulated_inputs) = deep_verify {
             return Some(if old_memo.revisions.changed_at > revision {
                 VerifyResult::Changed
             } else {
-                VerifyResult::Unchanged(old_memo.revisions.accumulated_inputs.load())
+                VerifyResult::Unchanged(accumulated_inputs)
             });
         }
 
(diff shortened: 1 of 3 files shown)


Veykril:
Seems like this has a slight overall green perf implication

src/function/maybe_changed_after.rs
```diff
@@ -191,26 +157,23 @@ where
         // the cycle head returned *fixpoint initial* without validating its dependencies.
         // `in_cycle` tracks if the enclosing query is in a cycle. `deep_verify.cycle_heads` tracks
         // if **this query** encountered a cycle (which means there's some provisional value somewhere floating around).
-        if old_memo.value.is_some() && !in_cycle && deep_verify.cycle_heads().is_empty() {
+        if old_memo.value.is_some() && cycle_heads.is_empty() {
```
Veykril, line 160: `!in_cycle` is now implied by `cycle_heads` being re-used

src/function/maybe_changed_after.rs
```diff
@@ -408,57 +372,49 @@ where
...
             }
             // Return `Unchanged` similar to the initial value that we insert
             // when we hit the cycle. Any dependencies accessed when creating the fixpoint initial
             // are tracked by the outer query. Nothing should have changed assuming that the
             // fixpoint initial function is deterministic.
-            QueryOrigin::FixpointInitial => VerifyResult::Unchanged(
-                InputAccumulatedValues::Empty,
-                CycleHeads::initial(database_key_index),
-            ),
+            QueryOrigin::FixpointInitial => VerifyResult::unchanged(),
```
Veykril, line 381: I am not 100% sure but I believe we do not need to push `database_key_index` into the cycle heads here, we only need to extend it with the dependency cycle heads, but `FixpointInitial` has no relevant dependencies (unless the initial kicks off a cycle itself which iirc is not supported scenario).

src/function/maybe_changed_after.rs
```diff
@@ -546,7 +510,7 @@ pub(super) enum ShallowUpdate {
 
     /// The revision for the memo's durability hasn't changed. It can be marked as verified
     /// in this revision.
-    HigherDurability(Revision),
```
Veykril, line 549: Removed as `Revision` always tracks `zalsa.current_revision()`

src/function/maybe_changed_after.rs
MichaReiser: We need to push the cycle head here. We've seen hangs without it. My reasoning behind this is that it otherwise behaves differently from when we push an initial value when we encounter the cycle the first time. 

In the end, this still signifies a cycle and we can only backdate the cycle head (and not the current query that only participates in the cycle)

src/function/maybe_changed_after.rs
```diff
@@ -511,14 +467,18 @@ where
...
-                            zalsa,
-                            zalsa.current_revision(),
-                            database_key_index,
-                        );
+                    let was_in_heads = cycle_heads.remove(&database_key_index);
+                    let heads_non_empty = !cycle_heads.is_empty();
+                    if heads_non_empty {
+                        // case 2 / 4
+                        break 'cycle VerifyResult::Unchanged(inputs);
+                    } else if !first_iteration {
```
MichaReiser, line 475: Is the `elif !first_iteration` a semantic change?

tests/cycle.rs
```diff
@@ -882,7 +882,6 @@ fn cycle_unchanged() {
         [
             "salsa_event(DidValidateMemoizedValue { database_key: min_iterate(Id(1)) })",
             "salsa_event(DidValidateMemoizedValue { database_key: min_panic(Id(2)) })",
-            "salsa_event(DidValidateMemoizedValue { database_key: min_iterate(Id(1)) })",
```
MichaReiser, line 885: Can you explain why we see behavior changes here?

MichaReiser (commented):
I like the `&mut CycleHead` refactor. I think it helps clarify some things

Veykril:
Split out the first commit into https://github.com/salsa-rs/salsa/pull/852 to leave the question of interest here

MichaReiser (approved):
This looks correct to and I verified that this doesn't introduce any new hangs but I'd prefer if Carl could take a look at this as well

tests/cycle.rs
carljm: This is removing a redundant verification (`min_iterate(1)` was already verified above), which is the point of the change as I understand it

carljm (approved):
Very nice! This looks great to me.

MichaReiser:
Unrelated to your PR but one thing I and carl discussed is that we should verify that its guarantee that we exit the loop after N iterations and don't loop forever (hang).

MichaReiser (commented):
Did this fix the r-a hang?

Veykril:
No it did not

PR #862, opened 2025-05-15
Short-circuit `block-on` if same thread

MichaReiser:
Early return from `block_on` if both the current and other thread are the same to skip the need to traverse the dependency tree (which requires a lock)

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -175,6 +175,11 @@ impl Runtime {
         let dg = self.dependency_graph.lock();
         let thread_id = thread::current().id();
 
+        // Cycle in the same thread.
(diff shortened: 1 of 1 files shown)


src/runtime.rs
```diff
@@ -175,6 +175,11 @@ impl Runtime {
         let dg = self.dependency_graph.lock();
         let thread_id = thread::current().id();
 
+        // Cycle in the same thread.
+        if thread_id == other_id {
```
Veykril, line 179: Do we even need to acquire the graph lock here?

MichaReiser: I don't think so

Veykril (approved):

PR #864, opened 2025-05-16
Use generational identifiers for tracked structs

ibraheemdev:
Pack a generation into input IDs.

The generation of a tracked struct is incremented after it is reused, allowing us to avoid read dependencies. Generational IDs were originally meant for https://github.com/salsa-rs/salsa/pull/839, as adding the necessary read dependency on interned structs that may be reused introduced a large (~50%) regression to ty's incremental performance.

This increases the size of `Id` from a `u32` to a `u64`. However, if the generation is restricted to `u16`, and ingredient indices are restricted to `u16`, this does not increase the size of `DatabaseKeyIndex`, so the memory usage effect is limited (~5% increase to ty's peak memory usage). However, this should allow us to implement garbage collection for interned values without significant performance concerns, so memory usage over time should benefit.

If the generation exceeds `u16::MAX`, we can fallback to adding read dependencies on tracked structs. An alternative would be to leak the slot, which would also allow us to remove the `created_at` field on tracked structs and may alleviate the memory usage concerns. This might be more feasible if the generation stole a few more bits from ingredient indices (as the number of ingredients is effectively static for a given salsa program).

This has a small (~4%) performance improvement on ty's benchmarks.

### DIFF
--- a/src/id.rs
+++ b/src/id.rs
@@ -39,7 +39,7 @@ impl Id {
     #[doc(hidden)]
     #[track_caller]
     #[inline]
-    pub const unsafe fn from_data(v: u32) -> Self {
+    pub const unsafe fn from_index(v: u32) -> Self {
         debug_assert!(v < Self::MAX_U32);
 
         Id {
@@ -66,13 +66,24 @@ impl Id {
         }
     }
 
+    /// Returns a new `Id` with same index, but the generation incremented by one.
+    ///
+    /// Returns `None` if the generation would overflow, i.e. the current generation
+    /// is `u32::MAX`.
+    #[inline]
+    pub fn next_generation(self) -> Option<Id> {
+        self.generation()
+            .checked_add(1)
+            .map(|generation| self.with_generation(generation))
+    }
+
     /// Mark the `Id` with a generation.
     ///
     /// This `Id` will refer to the same page and slot in the database,
     /// but will differ from other identifiers of the slot based on the
     /// provided generation.
     #[inline]
-    pub fn with_generation(self, generation: u32) -> Id {
+    pub const fn with_generation(self, generation: u32) -> Id {
         let mut value = self.value.get();
 
         value &= 0xFFFFFFFF;
@@ -84,9 +95,9 @@ impl Id {
         }
     }
 
-    /// Return the data portion of this `Id`.
+    /// Return the index portion of this `Id`.
     #[inline]
-    pub const fn data(self) -> u32 {
+    pub const fn index(self) -> u32 {
         // Truncate the high-order bits.
         (self.value.get() as u32) - 1
     }
@@ -108,9 +119,9 @@ impl Id {
 impl Debug for Id {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         if self.generation() == 0 {
-            write!(f, "Id({:x})", self.data())
+            write!(f, "Id({:x})", self.index())
         } else {
-            write!(f, "Id({:x}g{:x})", self.data(), self.generation())
+            write!(f, "Id({:x}g{:x})", self.index(), self.generation())
         }
     }
 }
--- a/src/id.rs
+++ b/src/id.rs
@@ -6,10 +6,8 @@ use crate::zalsa::Zalsa;
 
 /// The `Id` of a salsa struct in the database [`Table`](`crate::table::Table`).
 ///
-/// The high-order bits of an `Id` store a 16-bit generation counter
-/// as well as an optional 16-bit ingredient index.
-///
-/// The low-order bits pack a [`PageIndex`](`crate::table::PageIndex`) and
+/// The high-order bits of an `Id` store a 32-bit generation counter, while
+/// the low-order bits pack a [`PageIndex`](`crate::table::PageIndex`) and
 /// [`SlotIndex`](`crate::table::SlotIndex`) within the page.
 ///
 /// The low-order bits of `Id` are a `u32` ranging from `0..Id::MAX_U32`.
@@ -27,9 +25,6 @@ pub struct Id {
 impl Id {
     pub const MAX_U32: u32 = u32::MAX - 0xFF;
     pub const MAX_USIZE: usize = Self::MAX_U32 as usize;
-
-    const INGREDIENT_MASK: u64 = 0x0000FFFFFFFFFFFF;
-    const GENERATION_MASK: u64 = 0xFFFF0000FFFFFFFF;
 
     /// Create a `salsa::Id` from a u32 value, without a generation. This
     /// value should be less than [`Self::MAX_U32`].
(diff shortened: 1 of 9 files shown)


MichaReiser:
> This has a small (~4%) performance improvement on ty's benchmarks.

This is huge! I'm leaning towards removing the `created_at` field and leaking memos. It's still an improvement over what we have today and we can explore the right solution once we hit that limit (which could as well be that we increase the id size further)

tests/cycle_output.rs
```diff
@@ -152,6 +152,9 @@ fn revalidate_no_changes() {
             "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(400)) })",
             "salsa_event(DidReinternValue { key: query_d::interned_arguments(Id(800)), revision: R2 })",
             "salsa_event(DidValidateMemoizedValue { database_key: query_d(Id(800)) })",
+            "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(401)) })",
```
MichaReiser, line 155: Huh, this is interesting

src/tracked_struct.rs
```diff
@@ -560,22 +554,27 @@ where
...
-                    data.created_at = current_revision;
+                    // Consider this a new tracked-struct when any non-tracked field got updated.
+                    // This should be rare and only ever happen if there's a hash collision.
+                    //
+                    // Note that we hold the lock and have exclusive access to the tracked struct data,
+                    // so there should be no live instances of IDs from the previous generation. We clear
+                    // the memos and return a new ID here as if we have allocated a new slot.
+
+                    // SAFETY: We already set `updated_at` to `None` to acquire the lock.
+                    self.delete_entity(zalsa, id, true);
```
ibraheemdev, line 565: I'm a little unsure about this but it seems correct?

MichaReiser:
Is there any perf difference on ty? (When running on codspeed)

MichaReiser:
I plan to review this tomorrow.

I'm okay with the regression given that it enables interned GC without a 50% incremental perf and memory regression. If you haven't done so already, maybe take an hour or two to see if you can spot the source of the regression in a local recorded profile (take the benchmark that regresses the most)

This is a substantial change where I'd like to get at least a thumbs up from r-a too, given that the performance is now regressing on micro benchmarks. Cc @Veykril

ibraheemdev:
> I'm okay with the regression given that it enables interned GC without a 50% incremental perf and memory regression. If you haven't done so already, maybe take an hour or two to see if you can spot the source of the regression in a local recorded profile (take the benchmark that regresses the most)

I don't think there is a specific source, I think the regression is directly related to the size of `DatabaseKeyIndex` increasing (e.g. worse cache utilization, and it specifically no longer fits into a standard register). It makes sense that the microbenchmarks are regressing, I'm not sure there's much we can do about it.

src/tracked_struct.rs
```diff
@@ -396,12 +387,18 @@ where
...
                 tracing::trace!("Reuse tracked struct {id:?}", id = index);
-                // The struct already exists in the intern map.
                 zalsa_local.add_output(index);
-                self.update(zalsa, current_revision, id, &current_deps, fields);
-                FromId::from_id(id)
+
+                let updated_id = self.update(zalsa, current_revision, id, &current_deps, fields);
+                if id != updated_id {
+                    // Overwrite the previous ID if we are reusing the slot with new fields.
+                    zalsa_local.store_tracked_struct_id(identity, updated_id);
```
ibraheemdev, line 398: This feels a little wrong, but I think it is correct. If the struct is recreated with the same fields, we want to return the latest generation for the tracked structs to align with the memos. If it is recreated with new fields, we have to create a new generation and invalidate the previous version.

src/id.rs
```diff
@@ -38,23 +39,79 @@ impl Id {
...
+
+        Id {
+            // SAFETY: The niche of `value` is in the lower bits, which we did not touch.
+            value: unsafe { NonZeroU64::new_unchecked(value) },
+        }
+    }
+
+    /// Return the data portion of this `Id`.
+    #[inline]
+    pub const fn data(self) -> u32 {
```
MichaReiser, line 89: Bikeshedding: Maybe rename this to `index`. An id consists of a generation and its index into the database. This seems clearer to me than the generic data. The `index` terminology would also match with what you use in `DatabaseKeyIndex`

src/key.rs
```diff
@@ -31,7 +33,8 @@ impl DatabaseKeyIndex {
     }
 
     pub fn key_index(self) -> Id {
```
MichaReiser, line 35: Can we make this `const`?

src/id.rs
```diff
@@ -38,23 +39,79 @@ impl Id {
     #[doc(hidden)]
     #[track_caller]
     #[inline]
-    pub const unsafe fn from_u32(v: u32) -> Self {
+    pub const unsafe fn from_data(v: u32) -> Self {
```
MichaReiser, line 42: bikeshedding: The name doesn't make it clear that this is a new id without a generation. Maybe `new_without_generation` or `new_with_index`.

src/tracked_struct.rs
```diff
@@ -435,7 +431,11 @@ where
         };
 
         if let Some(id) = self.free_list.pop() {
-            Self::data_raw(zalsa.table(), id).with_mut(|data_raw| {
+            // Increment the ID generation before reusing it, as if we have allocated a new
+            // slot in the table.
+            let id = id.with_generation(id.generation() + 1);
```
MichaReiser, line 436: Nit: I think it would be nice to have a `id.increment_generation` or `new_generation`. We should also handle the case where generation overflows.

src/tracked_struct.rs
```diff
@@ -635,20 +638,26 @@ where
...
             fn drop(&mut self) {
-                // SAFETY: We have verified that no more references to these memos exist and so we are good
+                // SAFETY: We have `&mut MemoTable`, so no more references to these memos exist and we are good
                 // to drop them.
                 unsafe { self.0.drop() };
             }
         }
         let mut table_guard = TableDropGuard(table);
-        // SAFETY: We have verified that no more references to these memos exist and so we are good
+        // SAFETY: We have `&mut MemoTable`, so no more references to these memos exist and we are good
```
MichaReiser, line 660: I think the `self.free_list.push` shouldn't be in `clear_memos` because we don't want to push the memo into the free list when updating a tracked struct

tests/cycle_output.rs
MichaReiser: Do you know why this test changed?

src/tracked_struct.rs
```diff
@@ -560,22 +560,25 @@ where
...
-                    data.created_at = current_revision;
+                    // Consider this a new tracked-struct when any non-tracked field got updated.
+                    // This should be rare and only ever happen if there's a hash collision.
+                    //
+                    // Note that we hold the lock and have exclusive access to the tracked struct data,
+                    // so there should be no live instances of IDs from the previous generation. We clear
+                    // the memos and return a new ID here as if we have allocated a new slot.
+                    let mut table = data.take_memo_table();
+                    self.clear_memos(zalsa, &mut table, id);
+                    id = id.with_generation(id.generation() + 1);
```
MichaReiser, line 571:
```suggestion
                    id = id.next_generation();
```

MichaReiser (approved):
This is awesome. I've a few smaller comments that should be addressed before landing (assuming r-a is on board with this change)

Veykril:
I would like to cut another release before merging this if possible. I expect this to have a noticable memory impact for rust-analyzer, having this be part of a separate release cycle would make that easier to check.

src/id.rs
ibraheemdev: Done.

src/key.rs
ibraheemdev: Done (though `const` doesn't actually have any effect unless the method is actually used in a `const`-context).

src/tracked_struct.rs
```diff
@@ -635,20 +638,26 @@ where
             unsafe { (*data).assume_init_mut() }.take_memo_table()
         });
 
+        self.clear_memos(zalsa, &mut memo_table, id);
+    }
+
+    /// Clears the given memo table.
+    pub(crate) fn clear_memos(&self, zalsa: &Zalsa, memo_table: &mut MemoTable, id: Id) {
         // SAFETY: We use the correct types table.
```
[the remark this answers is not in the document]
ibraheemdev: Done.

src/tracked_struct.rs
ibraheemdev: Good catch, thanks.

tests/cycle_output.rs
ibraheemdev: I'm not sure. Maybe the memos are being validated instead of the function being re-executing, if the function previously had a read dependency on a tracked struct (and that dependency no longer exists)?

ibraheemdev:
When rebased onto this branch, https://github.com/salsa-rs/salsa/pull/839 is actually a ~5% improvement to ty's incremental performance. It also allows us to remove the `first_interned_at` field from interned values, so there should be some more memory saving.

MichaReiser:
@Veykril what's your timeline on the next release?

Veykril:
I'm fine with cutting a release now if you want to land this to try it out in ruff/ty asap

MichaReiser:
I'll land this once all feedback is addressed.

Issue #866, opened 2025-05-17
Ways to work around the orphan rule


Veykril:
For the integration of hte new trait solver in rust-analyzer, we now have queries that return types from the trait solver. These types mention the `'db` lifetime and as such do not fulfill the `Dispatcher` `Update` / `PartialEq + 'static` bounds. Newtyping is not an option here, it would hamper the usability of things too much. I am wondering if it would make sense for a helper attribute that allows returning a non `Update` types while making the tracked attribute unsafe

MichaReiser:
Didn't you open a PR a while ago that allowed overriding `Update` for a field (by calling a function)?

Veykril:
Yes but that is to customize behavior for the derive. We want to avoid having to new type things though. I have different idea in mind for the tracked function case here that I'll look into the next week.

PR #867, opened 2025-05-19
Implement an `!Update` bound escape hatch for tracked fn

Veykril:
Closes https://github.com/salsa-rs/salsa/issues/866


Veykril:
Ideally we wouldn't require `Update` in the first place, instead having an auto trait that is implemented for anything but references. Unfortunately that's not a thing in stable Rust.

PR #870, opened 2025-05-20
Replace `ThreadId` usages with `ZalsaLocalid`

MichaReiser:
Experiment if this helps with performance. I may also need this to enforce a strict ordering between different threads that all run the same fixpoint iteration (`ThreadId::as_u64` is nightly)

### DIFF
--- a/src/runtime/dependency_graph.rs
+++ b/src/runtime/dependency_graph.rs
@@ -5,9 +5,9 @@ use smallvec::SmallVec;
 
 use crate::key::DatabaseKeyIndex;
 use crate::loom::sync::MutexGuard;
-use crate::loom::thread::ThreadId;
 use crate::runtime::dependency_graph::edge::EdgeCondvar;
 use crate::runtime::WaitResult;
+use crate::zalsa_local::ZalsaLocalId;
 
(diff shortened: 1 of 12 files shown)


src/runtime/dependency_graph.rs
```diff
@@ -5,33 +5,33 @@ use smallvec::SmallVec;
...
 #[derive(Debug, Default)]
 pub(super) struct DependencyGraph {
     /// A `(K -> V)` pair in this map indicates that the runtime
     /// `K` is blocked on some query executing in the runtime `V`.
     /// This encodes a graph that must be acyclic (or else deadlock
     /// will result).
-    edges: FxHashMap<ThreadId, edge::Edge>,
+    edges: FxHashMap<ZalsaLocalId, edge::Edge>,
 
     /// Encodes the `ThreadId` that are blocked waiting for the result
```
Veykril, line 20:
```suggestion
    /// Encodes the `ZalsaLocalId` that are blocked waiting for the result
```

Veykril (approved):
Sounds reasonable, given we can't share `ZalsaLocal`'s across threads (due to it being `!Sync`)

PR #871, opened 2025-05-21
Fix hang in nested fixpoint iteration

MichaReiser:
This fixes a hang in multithreaded fixpoint iteration where the `Memo::provisional_retry` retried over and over again because all its cycle heads other than itself were finalized at this point. 

This also fixes the panic from the `kopf` repro listed in https://github.com/salsa-rs/salsa/issues/831

### DIFF
--- a/src/function/memo.rs
+++ b/src/function/memo.rs
@@ -150,13 +150,10 @@ impl<V> Memo<V> {
             cycle_heads: &CycleHeads,
         ) -> bool {
             let mut retry = false;
+            let mut hit_cycle = false;
 
             for head in cycle_heads {
                 let head_index = head.database_key_index;
-
-                if head_index == database_key_index {
-                    continue;
-                }
 
                 let ingredient = zalsa.lookup_ingredient(head_index.ingredient_index());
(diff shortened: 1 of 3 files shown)


src/function/memo.rs
```diff
@@ -189,7 +188,10 @@ impl<V> Memo<V> {
                     // We hit a cycle blocking on the cycle head; this means it's in
                     // our own active query stack and we are responsible to resolve the
                     // cycle, so go ahead and return the provisional memo.
-                    return false;
+                    tracing::debug!(
+                        "Waiting for {head_index:?} results in a cycle, return {database_key_index:?} once all other cycle heads completed to allow the outer cycle to make progress."
+                    );
+                    hit_cycle = true;
```
MichaReiser, line 194: I changed this to early-return in https://github.com/salsa-rs/salsa/pull/861 but I think that was incorrect. Not early-returning fixes the ty panic for `kopf`. I might spend some more time to come up with a test case for it but I might also just leave it at it (I already spent another week on this :s)

MichaReiser:
This does not fix all occurences of the cycle iteration panic, I can still reproduce it on pandas-stubs

carljm (approved):

PR #874, opened 2025-05-22
Flexible memo dropping behavior

Veykril:
This allows users to opt-into receiving a channel receiver on storage creation that can be used to decide when and where to drop destructors of stale memo results.

Notably this does not yet handle LRU evictions, so those are still synchronous. Handling that is more complicated or will have to incur allocations so I am not yet sure how to best approach this. We could maybe consider a double buffer (triple buffer) approach for the deletion queues per ingredient instead and if those are exhausted fall back to blocking in case of too frequent revision bumps. That is something I will consider in a separate follow up PR though, for now this is ready for a review.

Note that the perf changes are noise due to allocation order differences.


Veykril:
Regressions are kind of expected given we no longer keep the boxcar allocation from before. Should be fixable if we make the behavior swappable (also kind of required so that we don't have to implement a loom variant for the parker stuff)

src/table/memo.rs
```diff
@@ -22,6 +24,126 @@ pub(crate) struct MemoTable {
 pub trait Memo: Any + Send + Sync {
     /// Returns the `origin` of this memo
     fn origin(&self) -> &QueryOrigin;
+    fn clear_value(&mut self);
+}
+
+/// An untyped memo that can only be dropped.
+#[derive(Debug)]
+pub struct MemoDrop(NonNull<DummyMemo>, unsafe fn(NonNull<DummyMemo>));
```
davidbarsky, line 32: Three things:
1. I didn't know it was legal to write `unsafe fn(NonNull<DummyMemo>)`. I'm guessing that if you pass it a safe closure, this will fail to compile?
2. Can you make these named fields instead of tuple structs?
3. I *really* think this should be tested with Shuttle. if you're okay with me pushing to your branch, i can do that?

Veykril:
Okay, new approach. The user is now in charge of things. We now have two dropping modes, synchronous (what we currently have) and channel based, where the user will receive a receiver on storage construction that they can poll to drop things as they like.

src/table/memo.rs
Veykril: > I didn't know it was legal to write unsafe fn(NonNull<DummyMemo>). I'm guessing that if you pass it a safe closure, this will fail to compile?

No, safe functions/closures can be coerced to unsafe ones just fine. After all calling a safe function in an unsafe context isn't problematic. Only the other direction won't work for obvious reasons.

The other two questions won't be relevant anymore once im finished writing out the new approach

Veykril:
So, delaying the LRU clears is actually a bit more difficult and as it turns out, my previous approach was even unsound. We need to remove the values immediately, as otherwise we might be racing with computations of the next revision. That is a bit unfortunate though, as the values are not necessarily allocated and so we can't just take the pointer as an opaque droppable thing. That leaves two options, allocate LRU'd values that are to be dropped elsewhere individually, or use a (paged) arena allocator and move the value into that. Neither approach seems nice to me ...

src/storage.rs
```diff
@@ -123,6 +137,24 @@ impl<Db: Database> Storage<Db> {
...
+    ) -> (Self, DropChannelReceiver) {
+        let (handle, receiver) = StorageHandle::new_with_drop_channel(event_callback, capacity);
+        (
+            Self {
+                handle,
+                zalsa_local: ZalsaLocal::new(),
+            },
+            receiver,
+        )
+    }
```
MichaReiser, line 143: Could we add this to the `StorageBuilder` instead?

davidbarsky (approved):
LGTM, but I think I'd prefer someone from Astral to also approve this—conflict of interest and all.

MichaReiser:
We now see this issue in ty :) Worst case scenarios are a 9-12s hang where 64 threads block on a single thread dropping old parsed modules (for an operation that takes 60ms to compute)

MichaReiser:
A somewhat naive alternative would be to use rayon to spawn multiple threads. Certainly not ideal.

MichaReiser:
Another alternative is to bound how many items are dropped per revision.

MichaReiser:
I misunderstood this PR. I assumed it's for LRU/eviction policy. But it's mainly about replacing memos. 

Can't we special case this case and drop the memo value in place? We already hold the lock.

Veykril:
If you mean eagerly dropping memos on eviction, no we cannot as we do not know whether another query is using the memo still.

MichaReiser:
> If you mean eagerly dropping memos on eviction, no we cannot as we do not know whether another query is using the memo still.

Can you say more about when another query would still use that memo after execute completed? Specifically, I think we can drop the value (not the memo) in `execute` after inserting the new memo (we may need to reshuffle some instructions in `fetch_hot`), because we hold the query's lock.

MichaReiser:
What I have in mind is something like this https://github.com/salsa-rs/salsa/compare/master...MichaReiser:salsa:drop-replaced-memo-values?expand=1 (this is all codex generated and some safety claims don't hold without making the enclosing function unsafe too)

I'm not saying it's worth pursuing, especially if we need a different solution for LRU anyway. But I do like the idea of eagerly eviction to reduce peak memory usage and distributing work load across multiple threads.

Veykril:
Ah I misunderstood you, I thought you were talking about LRU/eviction 😅 Yea I think eager dropping on re-execution should be fine.

MichaReiser:
I considering adding this to my next 2-months goals (a solution to memo eviction, it might not be exactly this, and it might also be different solution dependent on whether it's an LRU'd vlaue or a replaced memo). 

Are there any rust-analyzer issues that I could put into my goal document to motivate why I should spend time on this (other than that ty also needs this)? I did find https://github.com/rust-lang/rust-analyzer/issues/20220 but, given that it's test only, it's not the most convincing issue ;)

Veykril:
I don't think we have specifics aside from wanting to drop some (though ideally all) memos in the background so cancellation stays fast. LRU is the important one actually as it turns out now (which iirc this PR doesnt even do), because dropping our sytnax trees takes the longest.

MichaReiser:
Too bad. You also don't have any metrics or tests that I could benchmark against?

>  LRU is the important one actually as it turns out now (which iirc this PR doesnt even do), because dropping our sytnax trees takes the longest.

Yes, this is also the main issue in ty (and part of the reason why we don't want to enable LRU for more queries, at least once the SIEVE PR lands)

PR #879, opened 2025-05-25
Don't report stale outputs if there is newer generation in new_outputs

### LINKED ISSUE #878: Crash with BadHash input and multiple changes
```rust 
#[salsa::input]
    struct TmpInput {
        pub id: u64,
    }

    #[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone)]
    struct BadHash {
        field: u64,
    }

    impl From<u64> for BadHash {
        fn from(value: u64) -> Self {
            Self { field: value }
        }
    }

    impl std::hash::Hash for BadHash {
        fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
            state.write_i16(0);
        }
    }

    #[salsa::tracked(debug)]
    struct Tmp<'db> {
        id: BadHash,
    }

    #[salsa::tracked]
    fn tmp_getter<'db>(db: &'db dyn base_db::BaseDb, input: TmpInput) -> Tmp<'db> {
        Tmp::new(db, input.id(db).into())
    }

    #[test]
    fn generations_test() {
        tracing_subscriber::fmt().init();
        let mut db = TestDatabase::default();
        let input = TmpInput::new(&db, 9);
        dbg!(tmp_getter(&db, input).0);
        input.set_id(&mut db).to(10);
        dbg!(tmp_getter(&db, input).0);
        input.set_id(&mut db).to(9);
        dbg!(tmp_getter(&db, input).0);
    }
```

logs:
```
---- tests::eq_ord_test stdout ----
[test.rs:1077:9] tmp_getter(&db, input).0 = Id(800)
[test.rs:1079:9] tmp_getter(&db, input).0 = Id(800g1)

thread 'resolver::tests::eq_ord_test' panicked at C:\Users\user\.cargo\git\checkouts\salsa-e6f3bb7c2a062968\0414d89\src\tracked_struct.rs:652:53:
cannot delete read-locked id `Id(800)`; value leaked across threads or user functions not deterministic
stack backtrace:
...
```

puuuuh:
Fixes #878

### DIFF
--- a/src/function/diff_outputs.rs
+++ b/src/function/diff_outputs.rs
@@ -3,7 +3,7 @@ use crate::function::{Configuration, IngredientImpl};
 use crate::hash::FxIndexSet;
 use crate::zalsa::Zalsa;
 use crate::zalsa_local::QueryRevisions;
-use crate::{DatabaseKeyIndex, Event, EventKind};
+use crate::{DatabaseKeyIndex, Event, EventKind, Id};
 
 impl<C> IngredientImpl<C>
 where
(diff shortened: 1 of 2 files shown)


src/function/diff_outputs.rs
```diff
@@ -25,7 +25,12 @@ where
         revisions: &mut QueryRevisions,
     ) {
         // Iterate over the outputs of the `old_memo` and put them into a hashset
```
MichaReiser, line 27: I would find it useful if we could extend the comment so that it explains why we ignore the generation part.

PR #882, opened 2025-05-27
fix: multithreaded nested fixpoint iteration

MichaReiser:
This PR fixes two bugs related to nested fixpoint iteration that can lead threads to see incorrect results or panics due to https://github.com/salsa-rs/salsa/issues/831

The main issue with our fixpoint iteration was that it assumed that a cycle would always be entered from the same query. This is given in single-threaded environments but cycles (especially if they have nested cyles) are likely entered from different queries.

The scenario is:

* thread a: has a nested loop, it iterates the outermost loop
* thread b: shares an inner cycle with thread a, but uses a different entry query than thread a.

The tests push this to the extreme with 4 threads and multiple queries. Shuttle revealed multiple bugs in this scenario.

Closes https://github.com/salsa-rs/salsa/issues/831
Closes https://github.com/salsa-rs/salsa/issues/893

Huge shoutout to @ibraheemdev for adding shuttle support. Narrowing this down without a shuttle would have been extremely hard, if not impossible. Shuttle also gives me a lot of confidence that the fixes from this PR cover most if not all cases.

## Issue 1:

`cycle_head_kind` only returned `Provisional` if the memo's cycle heads contained itself (it's a provisional in its own cycle). The problem with this is that (even in non multi-threaded cases) we don't re-run queries that participate in an outer cycle: Specifically, in the added test:

* `query_d` depends on `query_c`'s initial value
* `query_e` depends on `query_c`'s initial value
* `query_c` is an inner cycle

Both queries need to be re-run if `query_c` finished (as an inner cycle) and the outer cycle starts a second iteration, which includes starting a fresh round on `query_c`. Previously, `cycle_head_kind` returned `NotProvisional` because the provisional value for `query_c` no longer contained itself in `cycle_heads`; it only contains the cycle head of the outer query. Because of that, `validate_provisional` considered `query_d` and `query_e` as final after `query_c` completed the first time (it set `verified_final` to `true`). 

The fix is that any memo where `verified_final` is false should be considered provisional, and `provisional_retry` should retry (which might finalize the memos), and `validate_provisional` should return `false` (and fallback to `validate_same_iteration` or re-execute). 

I suspect that this is from before we had `validate_same_iteration`. This obviously leads to a significant performance regression (about 10%) because we now (correctly) run inner cycle queries more often.

## Issue 2

The second issue is due to a similar problem. Thread 'a' releases the lock on the inner cycle once it reaches a final value. It then iterates the outer cycle. Thread `b` can now acquire the lock on the inner cycle. The result is that we iterate different cyles of the same outer cycle at the same time. This is problematic because the different threads can override each other's provisional values with fixpoint initial values whenever they hit a cycle. Thread b re-executes the query because the cycle heads aren't finalized (the outer cycle is still iterating), and `validate_same_iteration` will fail because thread b has a different query stack. This can lead to https://github.com/salsa-rs/salsa/issues/831. 

The fix in this PR is that salsa blocks on all cycle heads before re-executing a provisional memo from the same revision.

## Issue 3

`validate_same_iteration` didn't work if a cycle runs on multiple threads because it only ever looked at the current query stack. This can result in hangs if the same query is re-executed repeatedly. The fix here is to look up the current memo and compare against its current iteration if requiring that memo results in a cycle with another thread, because this indicates that the two threads work on the same cycle.

## Test plan

* Added unit tests. I ran the tests with 10k iteration and a depth of 100 without seeing any panics or incorrect results
* I verified that the r-a example form https://github.com/salsa-rs/salsa/issues/831 now passes
* ~~I verified that this fixes all known multithreaded mypy primer hangs~~.


MichaReiser:
Interesting, I can't reproduce this shuttle panic locally

tests/parallel/cycle_nested_deep.rs
MichaReiser, line 1: The test looks so innocent!

carljm (commented):
Need to head out for a bit but submitting a partial review with some comments. Will come back to this later.

src/function/fetch.rs
```diff
@@ -200,9 +202,40 @@ where
...
+        // By waiting, we ensure that thread 1 completes a (based on a provisional value for `b`) and `b`
+        // becomes the new outer cycle, which thread 2 drives to completion.
+        if let Some(old_memo) = opt_old_memo {
+            if old_memo.value.is_some()
+                && old_memo.may_be_provisional()
+                && old_memo.verified_at.load() == zalsa.current_revision()
+            {
+                // TODO(micha) 29th May 2025: It would be nice if `block_on_heads` would release
+                // and re-acquire the lock if it blocks on any other thread. That would allow any other thread
+                // to claim this query (e.g. thread a in the example above) and make progress on it.
```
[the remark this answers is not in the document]
carljm: > I would have expected that thread a completes its cycle using the fixpoint initial value for `b`, which then adds `b` as cycle head to query `a` (which then waits in `provisional_retry`)

Yeah, that sounds right.

> The part that I don't fully understand is when it's safe to hold which lock?

Yeah, I'm not sure either (in terms of what actually causes broken behavior). But it does seem that the fewer locks we hold here, the better (triggers fewer extra cycle heads).

carljm (approved):

src/function/memo.rs
```diff
@@ -141,62 +144,110 @@ impl<V> Memo<V> {
...
+
+    /// Blocks on all cycle heads (recursively) that this memo depends on.
+    ///
+    /// Returns `true` if awaiting the cycle heads resulted in a cycle.
+    #[inline(always)]
+    pub(super) fn block_on_heads(&self, zalsa: &Zalsa, zalsa_local: &ZalsaLocal) -> bool {
+        // IMPORTANT: If you make changes to this function, make sure to run `cycle_nested_deep` with
+        // shuttle with at least 10k iterations.
+
+        // The most common case is that the entire cycle is running in the same thread.
```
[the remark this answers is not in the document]
MichaReiser: I agree that it seems problematic but I haven't found an alternative yet and my thinking was that, ideally one thread drives the entire cycle in the end. But maybe that's impossible if each thread already made too much progress because we can't unwind?

MichaReiser:
I'll need to clean this up a bit but this now fixes all known issues for ty. 

The third issue was that `validate_same_iteration` only worked if all cycle participants ran on the same thread.

src/function.rs
```diff
@@ -246,28 +250,44 @@ where
...
+            .map(|memo| memo.revisions.iteration())
+    }
+
+    fn cycle_heads<'db>(&self, zalsa: &'db Zalsa, input: Id) -> &'db CycleHeads {
+        self.get_memo_from_table_for(zalsa, input, self.memo_ingredient_index(zalsa, input))
+            .map(|memo| memo.cycle_heads())
+            .unwrap_or(empty_cycle_heads())
+    }
+
     /// Attempts to claim `key_index`, returning `false` if a cycle occurs.
```
carljm, line 285: Need doc comment update here

src/function.rs
```diff
@@ -246,28 +250,44 @@ where
...
+        iteration: Option<IterationCount>,
+    ) -> CycleHeadKind {
         let is_provisional = self
             .get_memo_from_table_for(zalsa, input, self.memo_ingredient_index(zalsa, input))
             .is_some_and(|memo| {
-                memo.cycle_heads()
-                    .into_iter()
-                    .any(|head| head.database_key_index == self.database_key_index(input))
+                Some(memo.revisions.iteration()) != iteration
+                    || !memo.revisions.verified_final.load(Ordering::Relaxed)
```
carljm, line 263: If the memo is verified final, but the iteration doesn't match, we still want to report it as provisional? This seems surprising to me, that we would ever want to report a verified-final memo as provisional.

src/function/fetch.rs
```diff
@@ -109,15 +113,30 @@ where
...
+        let claim_guard = match self.sync_table.try_claim(zalsa, id) {
+            ClaimResult::BlockedOn(blocked_on) => {
+                blocked_on.wait_for(zalsa);
+
+                let memo = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
+
+                if let Some(memo) = memo {
+                    // This isn't strictly necessary, but if this is a provisional memo for an inner cycle,
+                    // await all outer cycle heads to give the thread driving it a chance to complete
+                    // (we don't want multiple threads compute for the queries participating in the same cycle).
```
carljm, line 132:
```suggestion
                    // (we don't want multiple threads competing for the queries participating in the same cycle).
```

carljm (approved):
Took a quick look now, have to go -- can do a deeper review later if that would be useful. (I realize you said you plan to clean this up still, some of my comments probably duplicate what you already planned.)

src/function.rs
MichaReiser: I think what we should do here is to change the return type to include the iteration for a final memo. I believe I added a comment at the call site why it's important (we may have a memo that is provisional and depends on a non-final iteration memo because the final iteration didn't depend on that memo. We should not finalize this memo)

MichaReiser:
Hmm, the `cycle_panic` tests are now running forever?

MichaReiser:
There definitely seems to be some new issue with that test. Not sure what's happening. I can't repro locally (that was to be expected)

MichaReiser:
Todo: verify if claim result increased

src/function/memo.rs
```diff
@@ -141,62 +144,110 @@ impl<V> Memo<V> {
...
+                            tracing::trace!(
+                            "Dependent cycle head {head:?} has been released (there's a new memo)"
+                        );
+
+                            // Recursively wait for all cycle heads that this head depends on.
+                            // This is normally not necessary, because cycle heads are transitively added
+                            // as query dependencies (they aggregate). The exception to this are queries
+                            // that depend on a fixpoint initial value. They only depend on the fixpoint initial
+                            // value but not on its dependencies because they aren't known yet. They're only known
+                            // once the cycle completes but the cycle heads of the queries don't get updated.
```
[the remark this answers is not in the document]
MichaReiser: There are two cases: 
1. A later iteration adds new cycle heads because some new condition is met, making the function call more queries
2. Fixpoint initial values only depend on itself. They don't track their transitive dependencies until the first iteration completed. This is sort of the same as 1.

PR #883, opened 2025-05-29
Capture execution backtrace when throwing `UnexpectedCycle`

Veykril:
cc https://github.com/salsa-rs/salsa/pull/856


Veykril:
Note with this if you have a cycle to debug in a test @ChayimFriedman2 you can at least wrap your test temporarily in a `UnexpectedCycle::catch` the stuff

carljm:
I'm going to go ahead and merge this; I think it's strictly an improvement over the status quo. We can still separately discuss possibly going back to simple panic.

PR #885, opened 2025-05-29
Pack `QueryOrigin` memory layout

ibraheemdev:
Packs `QueryOrigin` into a `(u64, u32, u8)` union layout to take advantage of the padding bytes in `QueryRevisions`. This cuts down the size of memos by two words.

### DIFF
--- a/src/function/inputs.rs
+++ b/src/function/inputs.rs
@@ -1,15 +1,15 @@
 use crate::function::{Configuration, IngredientImpl};
 use crate::zalsa::Zalsa;
-use crate::zalsa_local::QueryOrigin;
+use crate::zalsa_local::QueryOriginRef;
 use crate::Id;
 
 impl<C> IngredientImpl<C>
 where
     C: Configuration,
 {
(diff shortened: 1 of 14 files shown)


src/function/inputs.rs
```diff
@@ -1,15 +1,15 @@
...
 
 impl<C> IngredientImpl<C>
 where
     C: Configuration,
 {
-    pub(super) fn origin(&self, zalsa: &Zalsa, key: Id) -> Option<QueryOrigin> {
+    pub(super) fn origin<'db>(&self, zalsa: &'db Zalsa, key: Id) -> Option<QueryOriginRef<'db>> {
         let memo_ingredient_index = self.memo_ingredient_index(zalsa, key);
         self.get_memo_from_table_for(zalsa, key, memo_ingredient_index)
-            .map(|m| m.revisions.origin.clone())
```
ibraheemdev, line 13: I believe we were cloning here unnecessarily. Removing this is likely the reason for the accumulator benchmark improvement.

PR #886, opened 2025-05-29
Pack `QueryEdge` memory layout

ibraheemdev:
If we reserve a bit in `IngredientIndex` we can avoid an extra discriminator byte for each `QueryEdge` in the `input_outputs` list, which is quite meaningful.

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -91,7 +91,7 @@ impl ActiveQuery {
     ) {
         self.durability = self.durability.min(durability);
         self.changed_at = self.changed_at.max(changed_at);
-        self.input_outputs.insert(QueryEdge::Input(input));
+        self.input_outputs.insert(QueryEdge::input(input));
         self.accumulated_inputs = self.accumulated_inputs.or_else(|| match has_accumulated {
             true => InputAccumulatedValues::Any,
             false => accumulated_inputs.load(),
@@ -107,7 +107,7 @@ impl ActiveQuery {
     ) {
         self.durability = self.durability.min(durability);
         self.changed_at = self.changed_at.max(revision);
-        self.input_outputs.insert(QueryEdge::Input(input));
+        self.input_outputs.insert(QueryEdge::input(input));
     }
 
     pub(super) fn add_untracked_read(&mut self, changed_at: Revision) {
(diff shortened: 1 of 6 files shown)


src/active_query.rs
```diff
@@ -91,7 +91,7 @@ impl ActiveQuery {
     ) {
         self.durability = self.durability.min(durability);
         self.changed_at = self.changed_at.max(changed_at);
-        self.input_outputs.insert(QueryEdge::Input(input));
+        self.input_outputs.insert(QueryEdge::input(input));
```
MichaReiser, line 94: The other approach that we could have considered (I don't think order matters between inputs and outputs) is to store the inputs and outputs (which most queries don't have) in separate boxed slices (and outputs in a thinvec).

src/zalsa.rs
```diff
@@ -77,9 +77,14 @@ static NONCE: NonceGenerator<StorageNonce> = NonceGenerator::new();
 pub struct IngredientIndex(u32);
```
MichaReiser, line 77: Could we use a `NonZeroU32` instead of a tag? or does that not work because `QueryEdge` is wrapping more than just `IngredientIndex`.

MichaReiser (approved):
Neat!

src/zalsa.rs
Veykril: I think the main annoyance here comes with the fact that we would want a `NonMax` variant, and so we'd have to do the -1 / +1 dance on use / construction then. I wonder if we can just put a `NonZeroU32` here and allocate an unused dummy ingredient into slot 0 to avoid having to do that?

ibraheemdev: The issue is that we need to reserve a whole bit for the tag, not just a niche (so this cuts the range of indices in half).

src/zalsa_local.rs
```diff
@@ -637,8 +637,51 @@ impl std::fmt::Debug for QueryOrigin {
     }
 }
 
+/// An input or output query edge.
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+pub struct QueryEdge {
```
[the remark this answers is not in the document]
ibraheemdev: Added a comment.

src/active_query.rs
ibraheemdev: That might also slightly improve iteration over inputs or outputs, but the extra allocation might be costly.

MichaReiser (approved):

PR #887, opened 2025-05-29
Unchecked page access

ibraheemdev:
Interested to see the performance results...

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -167,7 +167,7 @@ impl Table {
     #[inline]
     pub fn ingredient_index(&self, id: Id) -> IngredientIndex {
         let (page_idx, _) = split_id(id);
-        self.pages[page_idx.0].ingredient
+        unsafe { self.pages.get_unchecked(page_idx.0).ingredient }
     }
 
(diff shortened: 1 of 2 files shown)


Veykril:
I assume this is not doable for the same reason we cannot ever omit the type checks, someone could smuggle an index from one database to another.

ibraheemdev:
Maybe an (unsound) feature flag with the invariant that your application only has a single database? Or maybe we could even have a global that panics if you try to create multiple databases?

Veykril:
That does sound like a possible escape, that should also allow dropping the nonce check in that situation I think? Testing that seems a bit iffy though

ibraheemdev:
> I assume this is not doable for the same reason we cannot ever omit the type checks, someone could smuggle an index from one database to another. The type checks are a lot more expensive, I'd love if we could get rid of them

Actually, how is this possible given the nonce check?

ibraheemdev:
Opened https://github.com/salsa-rs/salsa/pull/937. I think you're right that we can't do this for slots at least, because those may actually vary across databases.

PR #888, opened 2025-05-29
Lazily allocate extra memo state

ibraheemdev:
Similar to https://github.com/salsa-rs/salsa/pull/768, except also lazily allocating the `CycleHeads`. I did some profiling in ty, and on a cold run ~1% of memos have non-empty `CycleHeads` and ~0.5% create tracked structs, out of >2M memos on large projects. Lazily allocating saves a meaningful amount of memory for us (4 words on memos without extras) and doesn't affect performance for ty (if anything it seems to improve performance very slightly, but that may just be noise).

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -1,4 +1,3 @@
-use std::ops::Not;
 use std::{fmt, mem, ops};
 
 use crate::accumulator::accumulated_map::{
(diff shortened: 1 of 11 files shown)


src/active_query.rs
```diff
@@ -203,18 +203,27 @@ impl ActiveQuery {
...
             .then(|| Box::new(mem::take(accumulated)));
-        let tracked_struct_ids = mem::take(tracked_struct_ids);
+
+        let (verified_final, extra) = if tracked_struct_ids.is_empty() && cycle_heads.is_empty() {
+            (true, None)
+        } else {
+            let extra = QueryRevisionsExtra {
+                tracked_struct_ids: mem::take(tracked_struct_ids),
+                cycle_heads: mem::take(cycle_heads),
+            };
```
MichaReiser, line 213: Nit: I think it would be nice if `QueryRevisionsExtra` could abstract the fact that it is an `Option<Box<Data>>` internally to keep the code here simpler (`QueryRevisions::new(final, tracked_struct_ids, cycle_heads)`)

src/zalsa_local.rs
```diff
@@ -328,6 +328,30 @@ pub(crate) struct QueryRevisions {
     /// How was this query computed?
     pub(crate) origin: QueryOrigin,
 
+    pub(super) accumulated: Option<Box<AccumulatedMap>>,
```
MichaReiser, line 331: We could probably move `AccumulatedMap` too

I assume you didn't move `accumualted_inputs` because it fits into what otherwise would be padding (similar to verified_final)

MichaReiser (commented):
This is great. I wonder if it would simplify the code if `QueryRevisionsExtra` would abstract the `Option<Box<Data>>` internally. Curious to hear your perspective.

I think we could also move `AccumulatedValues`. Let me know if you want to do this in the same or a separate PR

src/zalsa_local.rs
Veykril: I'd say lets move accumulated map in there as well, given we already have it lazily allocated (and neither you folks or rust-analyzer seem to use it much at the moment). In fact I still don't see a future where r-a wants to use it tbh 😅

PR #889, opened 2025-05-30
Add calc example to Cargo.toml so it's easy to run

MatrixFrog:


Cargo.toml
```diff
@@ -89,6 +89,10 @@ harness = false
 [workspace]
 members = ["components/salsa-macro-rules", "components/salsa-macros"]
 
+[[example]]
+name = "calc"
+crate-type = ["bin"]
```
MichaReiser, line 94: Can you tell me more why this change is necessary? I can run `cargo run --example calc` on `master`.

MatrixFrog: oh cool, sorry nevermind then!

PR #892, opened 2025-05-31
Store tracked struct ids as ThinVec on Revisions

MichaReiser:
We don't need a map. All we use it for is to seed new ids. This helps to reduce the `QueryRevisionsExtra` size (which will make up for the regression that I introduce in https://github.com/salsa-rs/salsa/pull/882)

### DIFF
--- a/tests/parallel/cycle_a_t1_b_t2_fallback.rs
+++ b/tests/parallel/cycle_a_t1_b_t2_fallback.rs
@@ -11,8 +11,7 @@
 //!  |                    |
 //!  +--------------------+
 //! ```
-use crate::sync::thread;
-use crate::{Knobs, KnobsDatabase};
+use crate::KnobsDatabase;
 
 const FALLBACK_A: u32 = 0b01;
 const FALLBACK_B: u32 = 0b10;


Veykril (commented):
Can't this be a `Box<[T]>` even? I don't think we ever push into it?

src/tracked_struct.rs
```diff
@@ -220,12 +220,18 @@ impl Clone for IdentityMap {
...
     }
-    fn clone_from(&mut self, source: &Self) {
-        self.table.clone_from(&source.table);
-    }
 }
 
 impl IdentityMap {
+    pub(crate) fn clone_from_slice(&mut self, source: &[(Identity, Id)]) {
+        self.table.clear();
+        self.table.reserve(source.len(), |(k, _)| k.hash);
```
MichaReiser, line 228: I'm not sure if there's a more elegant way of doing this

MichaReiser:
> Can't this be a `Box<[T]>` even? I don't think we ever push into it?

Is it possible to delete from a box slice? We use `retain`

MichaReiser:
I guess, we could use a `ThinVec`

ibraheemdev:
This looks good to me, I was trying to refactor the code to avoid the `retain` call and instead filter the tracked struct IDs when we convert the map to a list. That way we could use `Box<[_]>` instead, but I'm not sure if that's possible.

src/function/diff_outputs.rs
```diff
@@ -46,16 +46,14 @@ where
             ));
         }
 
-        if old_outputs.is_empty() {
+        let Some(tracked_struct_ids) = revisions.tracked_struct_ids_mut() else {
             return;
```
ibraheemdev: Is this correct? The old check was on the old revision, not the new one. We still need to remove stale outputs.

src/tracked_struct.rs
ibraheemdev: `self.table.extend(source)` should work.

src/function/diff_outputs.rs
MichaReiser: Oh no, this is a merge error. We should keep the `old_outputs` check to avoid calling `retain` if `old_outputs` is empty

MichaReiser: Actually no, this check is redudant today. It's already on line 36

MichaReiser:
> This looks good to me, I was trying to refactor the code to avoid the `retain` call and instead filter the tracked struct IDs when we convert the map to a list. That way we could use `Box<[_]>` instead, but I'm not sure if that's possible.

That would be nice! It might still be worth it from a perf perspective to avoid the linear traversal

PR #896, opened 2025-06-03
Sam/compile err non update ref

samscott89:
I had a return type like `Result<Tracked<'db>, Error>` and the error message was complaining about `'db` not being `'static`. It took me a while to figure out that the `'db` wasn't the problem, it was that my `Error` type doesn't implement `Update`.

Since tracking down where the `'static` bound comes it's fairly obvious why, but for future people this might provide a slightly easier debugging experience.

### DIFF
--- a/components/salsa-macros/src/tracked_fn.rs
+++ b/components/salsa-macros/src/tracked_fn.rs
@@ -91,6 +91,12 @@ impl Macro {
             ty
         });
         let output_ty = self.output_ty(&db_lt, &item)?;
+        let mut has_lifetime_visitor = HasLifetimeVisitor {
+            has_lifetime: false,
+        };
+        syn::visit::visit_type(&mut has_lifetime_visitor, &output_ty);
+        let has_lifetime = has_lifetime_visitor.has_lifetime;
+
         let (cycle_recovery_fn, cycle_recovery_initial, cycle_recovery_strategy) =
             self.cycle_recovery()?;
         let is_specifiable = self.args.specify.is_some();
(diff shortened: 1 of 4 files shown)


src/update.rs
```diff
@@ -494,6 +494,7 @@ fallback_impl! { compact_str::CompactString, }
 
 macro_rules! tuple_impl {
     ($($t:ident),*; $($u:ident),*) => {
+        #[diagnostic::do_not_recommend]
```
samscott89: This also seemed like a nice quality of life improvement. The error message just lists all the tuple types otherwise.

samscott89: Ah, not supported in 1.80 though

samscott89: Removed https://github.com/salsa-rs/salsa/pull/896/commits/d812631d157a94f1f22d307a256d5bd575832694

Veykril: You can bring this back, we bumped the MSRV now (needs a rebase then)

components/salsa-macros/src/tracked_fn.rs
```diff
@@ -307,6 +325,19 @@ impl syn::visit_mut::VisitMut for ToDbLifetimeVisitor {
...
+}
+
+impl<'ast> syn::visit::Visit<'ast> for HasLifetimeVisitor {
+    fn visit_lifetime(&mut self, l: &'ast syn::Lifetime) {
+        // We don't consider `'static` to be a lifetime in this context.
+        if l.ident != "static" {
+            self.has_lifetime = true;
+        }
+    }
+}
```
Veykril, line 339: This unfortunately breaks down with macro calls in the type, `-> foo!()` might expand to something introducing an elided lifetime. Likewise a user could use a `Foo<'_>` but without specifying the `<'_>` which warns post 2018 edition, but not error.

src/update.rs
samscott89: ah nice, will do then

components/salsa-macros/src/tracked_fn.rs
samscott89: Both of those seem like reasonable gaps to me, since it will still fallback to the current behaviour. Although I could see the argument that it might create even more confusing behaviour.

What do you think the best path forward is? I have no problems with closing this PR if it doesn't seem worth it.

PR #897, opened 2025-06-03
make UnexpectedCycle backtrace fields public

carljm:
In ty, we want to format these back-traces differently (as diagnostics), not just via `UnexpectedCycle::fmt`, so we need these fields to be public.

### DIFF
--- a/src/cycle.rs
+++ b/src/cycle.rs
@@ -63,8 +63,8 @@ use crate::sync::OnceLock;
 pub const MAX_ITERATIONS: IterationCount = IterationCount(200);
 
 pub struct UnexpectedCycle {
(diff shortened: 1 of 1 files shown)


src/cycle.rs
```diff
@@ -63,8 +63,8 @@ use crate::sync::OnceLock;
 pub const MAX_ITERATIONS: IterationCount = IterationCount(200);
 
 pub struct UnexpectedCycle {
```
Veykril, line 65: Not that it matters right now given we publish new versions a lot still, but for future proofing it might be good to `#[non_exhaustive]` this struct then

carljm:
Closing this since we've gone with #898 instead

Issue #891, opened 2025-05-30
No cycle error in logs


puuuuh:
After https://github.com/salsa-rs/salsa/commit/4327d6bb943650a7cca2dc51fb0592dc1c1839d7 there is no error message when cycle occurs, just silent panic instead

PR #898, opened 2025-06-04
panic with string message again for cycle panics

carljm:
Switch back to regular panic with string message, instead of `resume_unwind` with custom payload, for cycle panics.

This avoids having the default behavior be silent panic with no information, and allows the use of a panic hook to collect additional information, if needed.

It's not clear what the use case is for catching cycle panics specifically and treating them differently from other panics; they are not more recoverable.

Fixes https://github.com/salsa-rs/salsa/issues/891

### DIFF
--- a/src/function/fetch.rs
+++ b/src/function/fetch.rs
@@ -1,4 +1,4 @@
-use crate::cycle::{CycleHeads, CycleRecoveryStrategy, IterationCount, UnexpectedCycle};
+use crate::cycle::{CycleHeads, CycleRecoveryStrategy, IterationCount};
 use crate::function::memo::Memo;
 use crate::function::sync::ClaimResult;
 use crate::function::{Configuration, IngredientImpl, VerifyResult};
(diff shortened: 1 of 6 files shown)


src/function/fetch.rs
```diff
@@ -161,7 +161,13 @@ where
                 }
                 // no provisional value; create/insert/return initial provisional value
                 return match C::CYCLE_STRATEGY {
-                    CycleRecoveryStrategy::Panic => UnexpectedCycle::throw(),
+                    CycleRecoveryStrategy::Panic => zalsa_local.with_query_stack(|stack| {
+                        panic!(
+                            "dependency graph cycle when querying {database_key_index:#?}, \
+                            set cycle_fn/cycle_initial to fixpoint iterate.\n\
+                            Query stack:\n{stack:#?}",
```
MichaReiser, line 168: Nit: I'm a bit undecided if we should print the query stack here. I think it's a useful default behavior, but it can be redundant when using a custom panic hook handler.

MichaReiser (approved):

carljm:
> do think we want to roll back #883 then as well fwiw. The user can capture the normal backtrace themselves then if desired

Agreed, but doesn't this PR already do that by removing the `UnexpectedCycle` struct entirely?

src/function/fetch.rs
carljm: I had the same question, and I know it ends up being redundant output for us in ty... but I think for Salsa we should optimize for the default experience, not the assumption that someone has installed a custom panic hook. And this is clearly valuable debugging information.

Someone with a custom panic hook can also eliminate this from the printed message, if they care enough to do so.

PR #901, opened 2025-06-04
Assert size for interned `Value`

Veykril:


MichaReiser:
CC: @ibraheemdev

ibraheemdev (approved):

src/interned.rs
```diff
@@ -732,8 +735,16 @@ where
     pub fn entries<'db>(
         &'db self,
         db: &'db dyn crate::Database,
-    ) -> impl Iterator<Item = &'db Value<C>> {
-        db.zalsa().table().slots_of::<Value<C>>()
+    ) -> impl Iterator<Item = &'db ValueForConfiguration<C>> {
+        db.zalsa().table().slots_of::<ValueForConfiguration<C>>()
```
ibraheemdev, line 739: I've been trying to add some debug dumps for memory usage information and was running into some weird segfaults, I think this is the root cause. `entries` is now incorrect because the type-id assertions is only based on `C::Fields`, not `C`. This leads to us returning slots for other interned structs that have identical fields, which shows up specifically for constant query functions `fn query(&dyn Db)`. To prevent this we need `Value` to have the `Configuration` generic, do you think we could revert this change? For the size assertion we could create a dummy configuration type.

Veykril: Oh yes absolutely

Issue #895, opened 2025-06-03
Include field rustdocs in generated code


dcreager:
If you add rustdocs to a tracked or interned struct:

```rust
/// A description of the struct itself
#[salsa::tracked]
struct Yay {
    /// An exquisitely useful description of this field
    field: u32,
}
```

then the comment for the struct itself is carried over to the generated code, but the comments for each field are not. It would be nice if those were carried over to the accessor functions that are generated for the field.

puuuuh:
Looks like duplicate of https://github.com/salsa-rs/salsa/issues/728 I'll try to implement this

PR #905, opened 2025-06-08
Preserve attributes on interned/tracked structs

puuuuh:
Fixes https://github.com/salsa-rs/salsa/issues/728 and https://github.com/salsa-rs/salsa/issues/895

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_struct.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_struct.rs
@@ -72,6 +72,10 @@ macro_rules! setup_tracked_struct {
         // (see e.g. @return_mode below).
         untracked_options: [$($untracked_option:tt),*],
 
+        // Attrs for each field.
+        tracked_field_attrs: [$([$(#[$tracked_field_attr:meta]),*]),*],
+        untracked_field_attrs: [$([$(#[$untracked_field_attr:meta]),*]),*],
+
         // Number of tracked fields.
         num_tracked_fields: $N:literal,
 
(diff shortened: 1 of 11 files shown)


MichaReiser:
Nice, For #728, is this already "fixed" for regular salsa queries or input structs?

components/salsa-macro-rules/src/setup_tracked_struct.rs
```diff
@@ -256,6 +260,7 @@ macro_rules! setup_tracked_struct {
                 }
 
                 $(
+                    $(#[$tracked_field_attr])*
```
MichaReiser, line 263: Should we also set the attributes on the setter and builder methods. E.g. it would be useful to have the documentation everywhere.

src/table.rs
```diff
@@ -377,7 +377,7 @@ impl Page {
     }
 
     #[inline]
-    fn assert_type<T: Slot>(&self) -> PageView<T> {
+    fn assert_type<T: Slot>(&self) -> PageView<'_, T> {
```
MichaReiser, line 380: Is this a change to make clippy happy or what's the reason for it?

puuuuh: Yes, nightly clippy complains about it

puuuuh:
> Nice, For #728, is this already "fixed" for regular salsa queries or input structs?

Completely forgot about inputs. For queries we already pass all attributes to generated code

MichaReiser:
> Completely forgot about inputs. For queries we already pass all attributes to generated code
Do you plan on doing this as part of this PR or as a follow up (or not at all :))?

components/salsa-macro-rules/src/setup_tracked_struct.rs
puuuuh: Idk how we can distinguish between doc comments for setter and getter then. Also it seems more natural for me to pass attributes from "field" to getter with the same name, but not to setter

MichaReiser: I'd be fine having the same doc comment but we can also defer this for now

puuuuh: I'm completely fine with getter-only doc comments. If anyone really want attributes for setters, it wont be hard to implement, so lets leave it as is for now

MichaReiser:
Thank you

Issue #904, opened 2025-06-08
Crash after interned reuse


puuuuh:
```rust
    #[test]
    fn repro() {
        #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone)]
        struct BadHash(String);

        impl std::hash::Hash for BadHash {
            fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
                0.hash(state);
            }
        }

        #[salsa::input]
        struct Input {
            field: String
        }

        #[salsa::interned]
        struct Interned {
            field: String
        }

        #[salsa::tracked]
        struct Output<'db> {
            pub id: String,

            #[tracked]
            field: Interned<'db>
        }


        #[salsa::tracked]
        fn outer_query<'db>(db: &'db dyn Database, input: Input) -> Vec<Output<'db>> {
            let res: Vec<Output<'_>> = inner_query(db, input)
                .into_iter()
                .map(|i| Output::new(db, input.field(db), i))
                .collect();
            sleep(Duration::from_millis(500));
            db.unwind_if_revision_cancelled();
            for r in &res {
                r.id(db);
                r.field(db);
            }
            res
        }

        #[salsa::tracked]
        fn inner_query<'db>(db: &'db dyn Database, input: Input) -> Vec<Interned<'db>> {
            let res: Vec<Interned<'db>> = (0..10).map(|i| {
                Interned::new(db, format!("{}_{i}", input.field(db)))
            }).collect();
            res
        }


        tracing_subscriber::fmt()
            .with_max_level(Level::ERROR)
            .with_ansi(false)
            .with_writer(std::io::stderr)
            .init();
        let mut threads = vec![];
        let mut db = salsa::DatabaseImpl::default();
        let i1 = Input::new(&db, "i1".to_owned());
        let i2 = Input::new(&db, "i2".to_owned());
        let i3 = Input::new(&db, "i3".to_owned());
        let d = db.clone();
        threads.push(spawn(move || {
            outer_query(&d, i1);
        }));
        let d = db.clone();
        threads.push(spawn(move || {
            outer_query(&d, i2);
        }));
        sleep(Duration::from_millis(1000));
        db.synthetic_write(Durability::HIGH);
        let d = db.clone();
        threads.push(spawn(move || {
            outer_query(&d, i2);
        }));
        sleep(Duration::from_millis(250));
        db.synthetic_write(Durability::HIGH);
        let d = db.clone();
        threads.push(spawn(move || {
[...]

PR #908, opened 2025-06-10
Use latest revision for dependencies on interned values

ibraheemdev:
Resolved https://github.com/salsa-rs/salsa/issues/904.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -357,12 +357,13 @@ where
 
             // Record a dependency on the value.
             //
-            // Note that we can use `Revision::start()` here as the ID with the given generation
-            // is unique to this use of the interned slot.
+            // See `intern_id_cold` for why we need to use `current_revision` here. Note that just
+            // because this value was previously interned does not mean it was previously interned
+            // by *our query*, so the same considerations apply.
             zalsa_local.report_tracked_read_simple(
                 index,
                 value_shared.durability,
-                Revision::start(),
+                current_revision,
             );
 
             return value_shared.id;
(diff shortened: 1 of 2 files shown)


src/interned.rs
```diff
@@ -357,12 +357,13 @@ where
...
-            // Note that we can use `Revision::start()` here as the ID with the given generation
-            // is unique to this use of the interned slot.
+            // See `intern_id_cold` for why we need to use `current_revision` here. Note that just
+            // because this value was previously interned does not mean it was previously interned
+            // by *our query*, so the same considerations apply.
             zalsa_local.report_tracked_read_simple(
                 index,
                 value_shared.durability,
-                Revision::start(),
+                current_revision,
```
MichaReiser, line 366: I think this change technically means that, for any query creating or depending on a query creating interned values, this path can no longer be taken.

https://github.com/MichaReiser/salsa/blob/8d3455f0e6451c0dd6f88412db2aa9c49f41bc3f/src/function/maybe_changed_after.rs#L169-L176

Ideally, we'd use `first_interned_at` because that's since when this query hasn't changed. But we no longer track this for interned values. We could use `last_interned_at` (before it gets updated) but that seems a bit hacky because it only does the "backdating" for the query first validating the interned value. 

Maybe something to discuss during todays meeting: Whether we want to bring back `first_interned_at` or if there's another way to avoid using `current_revision` here.

ibraheemdev: I wasn't sure if `first_interned_at` was correct even for this case, but I agree that it is, because if the slot was recreated in a new revision because of garbage collection that must have happened *after* the initial creation, which is sufficient to update the memo.

PR #911, opened 2025-06-11
Add an option to tune interned garbage collection

ibraheemdev:
Adds the ability to tune and disable interned value garbage collection via an attribute:

```rust
#[salsa::interned(revisions = <1 | 2 | ... | usize::MAX>)]
struct Interned<'db>;
```

`revisions` is the minimum number of active revisions that must pass before a stale value is considered for garbage collection. Note that an *active revision* is one in which an instance of the interned struct has been read, so consecutive writes that create new revisions in sequence do not invalidate all interned values. Currently this constant is hard-coded to 3 internally for no good reason. Setting it to `usize::MAX` will disable garbage collection altogether (cc @Veykril).

I'm torn on whether to hard-code `usize::MAX` as the immortal value or to make that more explicit, e.g. `salsa::interned(durability = "immortal")` or `salsa::interned(immortal)`.

### DIFF
--- a/components/salsa-macro-rules/src/setup_interned_struct.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct.rs
@@ -28,6 +28,9 @@ macro_rules! setup_interned_struct {
         // the salsa ID
         id: $Id:path,
 
+        // The minimum number of revisions to keep the value interned.
+        revisions: $($revisions:expr)?,
+
         // the lifetime used in the desugared interned struct.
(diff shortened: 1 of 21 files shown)


Veykril (approved):
Generally looks fine to me, I do agree on the naming being a bit odd but I can't think of something better right now either.

One future neat thing would be to drop the `LinkedListLink` field for immortal interneds, but that as usual requires some trait gymnastics with assoc types to make that swappable.

PR #912, opened 2025-06-11
Fix race in `MemoTableTypes`

ibraheemdev:
Noticed this while reading the code. `boxcar::Vec::count` does not guarantee that there are `N` contiguous entries initialized, which the current code is assuming. It seems fine to just spin here because, from what I understand, `MemoTableTypes` is only setup the first time a query is called? I also don't think we've ever seen this bug in practice.

### DIFF
--- a/src/table/memo.rs
+++ b/src/table/memo.rs
@@ -5,6 +5,7 @@ use std::{
     ptr::{self, NonNull},
 };
 
+use portable_atomic::hint::spin_loop;
 use thin_vec::ThinVec;
 
 use crate::sync::atomic::{AtomicPtr, Ordering};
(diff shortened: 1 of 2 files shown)


src/table/memo.rs
```diff
@@ -120,22 +122,34 @@ impl MemoTableTypes {
         memo_type: &MemoEntryType,
     ) {
         let memo_ingredient_index = memo_ingredient_index.as_usize();
-        while memo_ingredient_index >= self.types.count() {
-            self.types.push(MemoEntryType {
-                data: OnceLock::new(),
-            });
```
ibraheemdev, line 126: In the case that the latest entry has been claimed but is still being initialized, this loop would previously spin waiting for the entry to be initialized while rapidly creating new values, so the new spin loop should be strictly better.

Veykril (approved):

PR #915, opened 2025-06-14
Revert "Assert size for interned Value" & Mark `Slot` trait as unsafe

Veykril:
This reverts commit 793dc41.
cc https://github.com/salsa-rs/salsa/pull/901/files#r2146090617


ibraheemdev (approved):
Thanks for the quick fix. If you want to keep the size assertion we could create a dummy configuration struct for the value.

Veykril:
Nah its fine I think

Issue #918, opened 2025-06-18
10x slowndown when using multiple dbs


MichaReiser:
I'm about to set up new benchmarks for ty and noticed that, after the first run, there's a 10x slowdown when using multithreading. I've been able to narrow it down to our `IngredientCache::get_or_create_index` optimization where the first DB can take a fast path and all other DBs have to go through a slow path which requires a lock. It seems the slow path is really slow! 

https://github.com/MichaReiser/salsa/blob/87a730fccf07add73aabc6f2c842d92fd2bfe466/src/zalsa.rs#L476-L509

I'm not sure what better approach we could take here but I was only able to figure this out because I knew about this salsa optimization and, even then, it still took me a fair amount of time to narrow it down.

@ibraheemdev any chance you have an idea on how we could make this cheaper if multiple dbs exist?

MichaReiser:
Not sure if using a `RwLock` instead of a regular `Mutex` would help here. Or we decide to use a separate lock for protecting the ingredient vec for when we hit the slow-slow path where we need to add an ingredient.

MichaReiser:
We could probably consider adding a secondary LRU cache to `IngredientCache` that stores the indices for up to N databases. This could either be a simple `Vec` or `DashMap`. The number of active databases should be pretty low

PR #919, opened 2025-06-18
Shard jar map

ibraheemdev:
Replaces the jar map with a `DashMap`. This depends on https://github.com/ibraheemdev/boxcar/pull/35, which allows us to claim multiple sequential ingredient indices without holding a global lock.

This should also help with https://github.com/salsa-rs/salsa/issues/918, because the slow path for accessing an uncached ingredient now goes through a sharded `DashMap` instead of a single `Mutex`, but I don't think we have benchmarks that use multiple databases yet (I'll add some). The ideal fix for that issue might also be to extend the `IngredientCache`, but this should at least help. We could also look into using [papaya](https://github.com/ibraheemdev/papaya) here (or more extreme, `ArcSwap`?), because the jar map is almost purely read-heavy after the table is initially filled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,10 +12,13 @@ description = "A generic framework for on-demand, incrementalized computation (e
 salsa-macro-rules = { version = "0.22.0", path = "components/salsa-macro-rules" }
 salsa-macros = { version = "0.22.0", path = "components/salsa-macros", optional = true }
 
-boxcar = "0.2.13"
+boxcar = { git = "https://github.com/ibraheemdev/boxcar", rev = "574c893f0a6d8b2cde17dd9933fd9e3d74ea8e0f" }
 crossbeam-queue = "0.3.11"
 crossbeam-utils = "0.8.21"
+dashmap = { version = "6", features = ["raw-api"] }
 hashbrown = "0.15"
+# The version of hashbrown used by dashmap.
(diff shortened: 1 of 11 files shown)


ibraheemdev:
The benchmarks are probably misleading here because they're single threaded.

MichaReiser:
> This should also help with https://github.com/salsa-rs/salsa/issues/918, because the slow path for accessing an uncached ingredient now goes through a sharded DashMap instead of a single Mutex, but I don't think we have benchmarks that use multiple databases yet (I'll add some). The ideal fix for that issue might also be to extend the IngredientCache, but this should at least help. We could also look into using [papaya](https://github.com/ibraheemdev/papaya?rgh-link-date=2025-06-18T22%3A25%3A57Z) here (or more extreme, ArcSwap?), because the jar map is almost purely read-heavy after the table is initially filled.

Did you try changing the ty benchmark by removing the thread pool override in the setup here:

https://github.com/astral-sh/ruff/blob/e352a50b74329268589cdc18eafab123832559ac/crates/ruff_benchmark/benches/ty_walltime.rs#L240-L250

This should be a very good benchmark to show the impact of this change

MichaReiser (approved):
Excellent! It would be nice to reduce this even further (e.g. could we use a lock free map that's optimized for reads?) but this is a great first step.

PR #922, opened 2025-06-20
Do manual trait casting

Veykril:
This removes occurences of `dyn Database` internally where only the zalsas are needed for (hopefully) a perf increase.

Lastly it renames downcasting to upcasting, because thats what we are doing, `dyn Database` --downcast-> `Concrete` --upcast-->`dyn CustomDatabaseTrait`

### DIFF
--- a/src/database.rs
+++ b/src/database.rs
@@ -1,13 +1,39 @@
-use std::any::Any;
 use std::borrow::Cow;
+use std::ptr::NonNull;
 
-use crate::views::DatabaseDownCaster;
+use crate::views::DatabaseUpCaster;
 use crate::zalsa::{IngredientIndex, ZalsaDatabase};
 use crate::{Durability, Revision};
 
+#[derive(Copy, Clone)]
+pub struct RawDatabasePointer<'db> {
+    pub(crate) ptr: NonNull<()>,
+    _marker: std::marker::PhantomData<&'db dyn Database>,
+}
+
+impl<'db, Db: Database + ?Sized> From<&'db Db> for RawDatabasePointer<'db> {
+    #[inline]
+    fn from(db: &'db Db) -> Self {
+        RawDatabasePointer {
+            ptr: NonNull::from(db).cast(),
+            _marker: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<'db, Db: Database + ?Sized> From<&'db mut Db> for RawDatabasePointer<'db> {
+    #[inline]
+    fn from(db: &'db mut Db) -> Self {
+        RawDatabasePointer {
+            ptr: NonNull::from(db).cast(),
+            _marker: std::marker::PhantomData,
+        }
+    }
+}
+
 /// The trait implemented by all Salsa databases.
 /// You can create your own subtraits of this trait using the `#[salsa::db]`(`crate::db`) procedural macro.
-pub trait Database: Send + AsDynDatabase + Any + ZalsaDatabase {
+pub trait Database: Send + ZalsaDatabase + AsDynDatabase {
     /// Enforces current LRU limits, evicting entries if necessary.
     ///
     /// **WARNING:** Just like an ordinary write, this method triggers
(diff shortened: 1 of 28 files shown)


Veykril:
Benchmarks look fairly noisy again to me

MichaReiser:
Not sure, 5% is a bit more than usual.

Veykril:
Okay the regression came from a dumb mistake

src/database.rs
```diff
@@ -84,59 +110,40 @@ pub trait Database: Send + AsDynDatabase + Any + ZalsaDatabase {
...
         // No-op
-        db
+        self
     }
 }
 
 /// Upcast to a `dyn Database`.
 ///
-/// Only required because upcasts not yet stabilized (*grr*).
+/// Only required because upcasting does not work for unsized generic parameters.
```
Veykril, line 131: I am quite annoyed by this but that's a limitation of Rust, a `&T` of `T: ?Sized + SomeTrait` cannot be coerced to a `&dyn SomeTrait` (or a supertrait) due to `T` possibly being an unrelated unsized type like `str`

src/function.rs
```diff
@@ -124,13 +124,14 @@ pub struct IngredientImpl<C: Configuration> {
...
     lru: lru::Lru,
 
-    /// A downcaster from `dyn Database` to `C::DbView`.
+    /// An upcaster to `C::DbView`.
     ///
     /// # Safety
     ///
     /// The supplied database must be be the same as the database used to construct the [`Views`]
+    /// instances that this upcaster was derived from.
     /// instances that this downcaster was derived from.
```
ibraheemdev:
```suggestion
    /// instances that this upcaster was derived from.
```

src/key.rs
```diff
@@ -36,16 +36,27 @@ impl DatabaseKeyIndex {
...
         &self,
-        db: &dyn Database,
+        db: crate::database::RawDatabasePointer<'_>,
         zalsa: &Zalsa,
         last_verified_at: crate::Revision,
         cycle_heads: &mut CycleHeads,
     ) -> VerifyResult {
         // SAFETY: The `db` belongs to the ingredient
         unsafe {
+            // heer, `db` has to be either the correct type already, or a subtype (as far as trait
```
ibraheemdev:
```suggestion
            // here, `db` has to be either the correct type already, or a subtype (as far as trait
```

src/key.rs
```diff
@@ -36,16 +36,27 @@ impl DatabaseKeyIndex {
...
+            // hierarchy is concerned)
             zalsa
                 .lookup_ingredient(self.ingredient_index())
-                .maybe_changed_after(db, self.key_index(), last_verified_at, cycle_heads)
+                .maybe_changed_after(
+                    zalsa,
+                    // lets say we do turn this into an opaque pair of data and vtable pointer
+                    // then we also need an upcast function, from our dbview to the ingredients
+                    // dbview
```
ibraheemdev: I'm not sure I understand this comment.

src/views.rs
```diff
@@ -13,116 +18,145 @@ pub struct Views {
...
 }
 
-type ErasedDatabaseDownCasterSig = unsafe fn(&dyn Database) -> *const ();
-type DatabaseDownCasterSig<DbView> = unsafe fn(&dyn Database) -> &DbView;
+type ErasedDatabaseUpCasterSig = unsafe fn(RawDatabasePointer<'_>) -> NonNull<()>;
+type DatabaseUpCasterSigRaw<DbView> =
+    for<'db> unsafe fn(RawDatabasePointer<'db>) -> NonNull<DbView>;
+type DatabaseUpCasterSig<DbView> = for<'db> unsafe fn(RawDatabasePointer<'db>) -> &'db DbView;
+type DatabaseUpCasterSigMut<DbView> =
+    for<'db> unsafe fn(RawDatabasePointer<'db>) -> &'db mut DbView;
```
ibraheemdev: Can we remove these two aliases in favor of `DatabaseUpCasterSigRaw`, and dereference the returned `NonNull<DbView>` instead of implicitly transmuting it? I found all of these type aliases a little confusing.

src/views.rs
```diff
@@ -13,116 +18,145 @@ pub struct Views {
...
+            // SAFETY: The type-erased function pointer is guaranteed to be valid for `DbView`
+            return unsafe { &*(caster as *const ViewCaster as *const DatabaseUpCaster<DbView>) };
         }
 
-        self.view_casters.push(ViewCaster::new::<DbView>(func));
-        DatabaseDownCaster(self.source_type_id, func)
+        // SAFETY: We are type erasing the function pointer for storage, and we will unerase it
+        // before we call it.
+        let caster = ViewCaster::new::<DbView>(unsafe {
+            mem::transmute::<fn(&Concrete) -> &DbView, DatabaseUpCasterSigRaw<DbView>>(func)
```
ibraheemdev: It's questionable to use the `&Concrete -> &DbView` function for the mutable downcaster. We should probably change the signature to use raw pointers.

src/database.rs
```diff
@@ -1,13 +1,39 @@
...
 use std::borrow::Cow;
+use std::ptr::NonNull;
 
-use crate::views::DatabaseDownCaster;
+use crate::views::DatabaseUpCaster;
 use crate::zalsa::{IngredientIndex, ZalsaDatabase};
 use crate::{Durability, Revision};
 
+#[derive(Copy, Clone)]
+pub struct RawDatabasePointer<'db> {
```
ibraheemdev: Should we just call this `RawDatabase`? The name is a little long.

src/database.rs
ibraheemdev: Can we add the `Sized` requirement?

src/function.rs
```diff
@@ -154,6 +155,7 @@ where
     C: Configuration,
 {
     pub fn new(
+        _zalsa: &Zalsa,
```
ibraheemdev: Was this left on purpose?

src/database.rs
Veykril: No, specifically this is used solely for `attach` which needs to take a dyn database of some form.

src/key.rs
Veykril: Oh, these were old notes for myself that I forgot to remove

src/function.rs
Veykril: No, this was a rebasing change (your inventory changes changed things here), forgot to remove the param again

PR #925, opened 2025-06-24
Add option to track heap memory usage of memos

ibraheemdev:
`#[salsa::tracked(heap_size)]` will require the return type of a tracked function to implement the [`get_size2::GetSize`](https://docs.rs/get-size2/latest/get_size2/trait.GetSize.html) trait, and include the heap size in the new memory usage APIs. I'm not sure `get-size2` is the best crate to use, but it seems relatively comprehensive, and integrating it into ty was relatively simple with its derive macros. Note that it's not actually a dependency of salsa, it's just called by the generated proc-macro code.

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -55,6 +55,9 @@ macro_rules! setup_tracked_fn {
         // If true, the input needs an interner (because it has >1 argument).
         needs_interner: $needs_interner:tt,
 
+        // The function used to implement `C::heap_size`.
+        heap_size_fn: $($heap_size_fn:path)?,
+
         // LRU capacity (a literal, maybe 0)
         lru: $lru:tt,
 
@@ -195,6 +198,12 @@ macro_rules! setup_tracked_fn {
                 const CYCLE_STRATEGY: $zalsa::CycleRecoveryStrategy = $zalsa::CycleRecoveryStrategy::$cycle_recovery_strategy;
 
                 $($values_equal)+
+
+                $(
+                    fn heap_size(value: &Self::Output<'_>) -> usize {
+                        $heap_size_fn(value)
+                    }
+                )?
 
                 fn execute<$db_lt>($db: &$db_lt Self::DbView, ($($input_id),*): ($($interned_input_ty),*)) -> Self::Output<$db_lt> {
                     $($assert_return_type_is_update)*
--- a/components/salsa-macros/src/input.rs
+++ b/components/salsa-macros/src/input.rs
@@ -64,6 +64,8 @@ impl crate::options::AllowedOptions for InputStruct {
     const ID: bool = false;
 
     const REVISIONS: bool = false;
+
+    const HEAP_SIZE: bool = false;
 }
 
 impl SalsaStructAllowedOptions for InputStruct {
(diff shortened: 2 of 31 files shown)


components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -196,6 +199,12 @@ macro_rules! setup_tracked_fn {
 
                 $($values_equal)+
 
+                $zalsa::macro_if! { $heap_size =>
+                    fn allocation_size(value: &Self::Output<'_>) -> usize {
+                        ::get_size2::GetSize::get_heap_size(value)
```
MichaReiser: Did you explore an approach where the `heap_size` argument takes a function expression so that Salsa doesn't depent on a specific crate?E.g something like


```rust
#[salsa::tracked(heap_size = my_tracked_size)]
struct MyStruct { ... }
```

where `my_tracked_size` must be a function with the signature `&Value -> usize`.

I guess the main downside of this is that it may not work with generic types and I suspect it isn't possible to use `::get_size2::GetSize::get_heap_size` as function directly? 

I guess, similar to how getsize allows you to override `get_size` for foreign types:  https://docs.rs/get-size2/latest/get_size2/index.html#using-a-helper-function

ibraheemdev: That's a lot nicer. `heap_size = GetSize::get_heap_size` should work fine.

Cargo.toml
```diff
@@ -63,6 +63,7 @@ expect-test = "1.5.0"
 rustversion = "1.0"
 test-log = { version = "0.2.11", features = ["trace"] }
 trybuild = "1.0"
+get-size2 = { version = "0.4.1", features = ["derive"] }
```
MichaReiser: Can we remove the dependency and use a manual `size` implementation in tests instead?

components/salsa-macros/src/input.rs
```diff
@@ -64,6 +64,8 @@ impl crate::options::AllowedOptions for InputStruct {
     const ID: bool = false;
 
     const REVISIONS: bool = false;
+
+    const HEAP_SIZE: bool = false;
```
MichaReiser, line 68: Do you plan on adding `HEAP_SIZE` for salsa-structs (input, tracked, interned) too (in a follow up PR?)

src/function/memo.rs
```diff
@@ -312,19 +305,23 @@ impl<V> Memo<V> {
     }
 }
 
-impl<V: Send + Sync + Any> crate::table::memo::Memo for Memo<V> {
+impl<C: Configuration> crate::table::memo::Memo for Memo<'static, C>
```
MichaReiser, line 308: Given that memo now has access to the `Configuration`. Doesn't that mean that we could get the debug name too?

ibraheemdev: Yes, great catch. I updated the PR.

components/salsa-macros/src/input.rs
ibraheemdev: Yeah, as we discussed, this is a little more difficult because it requires a `db` parameter. I'll do this in a future PR.

src/function/memo.rs
MichaReiser: We can also do this in a separate PR. It was more a thought that I had. As you preferred

PR #927, opened 2025-06-26
Emit self ty for query debug name of assoc function queries

Veykril:


tests/tracked_assoc_fn.rs
```diff
@@ -80,6 +80,6 @@ fn debug_name() {
     assert_eq!(output.field(&db), 88);
     db.assert_logs(expect![[r#"
         [
-            "salsa_event(WillExecute { database_key: tracked_trait_fn_(Id(0)) })",
+            "salsa_event(WillExecute { database_key: MyOutput < 'db >::tracked_trait_fn_(Id(0)) })",
```
Veykril, line 83: The formatting here is a bit unfortunate, but we the only other options are to either hand format these (I'd rather not) instead of relying on stringify! behavior or pulling in something like syn pretty (forgot the exact name). I think this is fine (and either way an improvement over the status quo)

components/salsa-macros/src/options.rs
```diff
@@ -433,3 +455,82 @@ impl<A: AllowedOptions> syn::parse::Parse for Options<A> {
         Ok(options)
     }
 }
+impl<A: AllowedOptions> quote::ToTokens for Options<A> {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
```
MichaReiser, line 459: I haven't read the diff to closely but what's this used for (should it replace some existing code?)

components/salsa-macros/src/tracked_impl.rs
```diff
@@ -83,11 +83,20 @@ impl Macro {
...
-        let salsa_tracked_attr = fn_item.attrs.remove(tracked_attr_index);
-        let args: FnArgs = match &salsa_tracked_attr.meta {
+        let mut salsa_tracked_attr = fn_item.attrs.remove(tracked_attr_index);
+        let mut args: FnArgs = match &salsa_tracked_attr.meta {
             syn::Meta::Path(..) => Default::default(),
             _ => salsa_tracked_attr.parse_args()?,
         };
+        if args.self_ty.is_none() {
+            // If the user did not specify a self_ty, we use the impl's self_ty
+            args.self_ty = Some(self_ty.clone());
```
MichaReiser, line 93: What's the reason or use case where a user needs to provide a self type? Can you add a test for it (or remove the attribute)

tests/tracked_assoc_fn.rs
MichaReiser: You mean https://github.com/dtolnay/prettyplease

I don't think it's worth the dependency, given that this is for debugging only.

components/salsa-macros/src/tracked_impl.rs
Veykril: The user isn't really supposed to use this, this is solely needed for tracked impls to pass through their self type to the assoc tracked fn so that it can adjust the debug name appropriately

components/salsa-macros/src/options.rs
Veykril: This is used to inject arguments into the `#[tracked]` invocation supplied by the user, see https://github.com/salsa-rs/salsa/pull/927/files#diff-7091e579afd9cfe8ee883486f5951094096ebf6419e4e5fc5d1d9be926a858fdR86-R99

MichaReiser: Okay, this goes beyond what I understand about macros 😅

Veykril:
Will cut a new release after this merges so we can update r-a (and see if things break for us given the generation changes)

Issue #581, opened 2024-09-28
Proposal: "late" fields on tracked structs


nikomatsakis:
## TL;DR

Permit fields to be tagged as "late"

```rust
#[salsa::tracked]
struct Variable {
    name: Id<'db>,
    #[late] ty: Ty<'db>,
}
```

When you create a tracked struct, you specify only the early fields:

```rust
let v = Variable::new(name);
```

but you can later specify the late fields, so long as it is within the same tracked function (query):

```rust
v.set_ty(db, ty);
```

Panics occur if:

* there is an attempt to access `v.ty(db)` but it has not yet been set
* you reach the end of the tracked function without setting `ty`

## Background

Salsa currently supports a rather obscure feature called "specified" functions that lets you emulate this:

```rust
#[salsa::tracked(specify)]
fn ty<'db>(db: &'db dyn Database, v: Variable<'db>) -> Ty<'db> {
    panic!("ty not specified");
}
```

This lets you specify the value of `ty` for some `v` so long as `v` was created in the current query (and `ty` has not yet been called on `s`):

```rust
ty::specify(db, v, value)
```

Supporting the late pattern directly would

* let us remove specifiable functions, simplifying salsa in my opinion;
* be mildly more efficient, since we could create space in the fields struct for the field instead of storing the value in the memo table
* let us check at the end of the query that all late fields were assigned

PR #933, opened 2025-07-06
WIP: late fields POC

puuuuh:
Basic impl of #581


puuuuh:
Also probably its better to store durability in LateField, so we can change field durability after set is called

PR #934, opened 2025-07-09
Use `inventory` for static ingredient registration

ibraheemdev:
Salsa currently registers ingredients dynamically the first time they are accessed, but persistent caching will require that all serializable ingredients are registered up-front. It's possible that we may switch to a static ingredient registration API, but as a first step, we can use `inventory` to ensure all ingredients have stable indices. This should enable persistent caching and also give us the same performance benefits of static ingredient registration without a large breaking API change.

Inventory has pretty good platform support, I believe it includes all the platforms we care about. There is [a caveat on WebAssembly](https://docs.rs/inventory/latest/inventory/#webassembly-and-constructors), but I don't think it's a dealbreaker.

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -112,8 +112,8 @@ macro_rules! setup_tracked_fn {
                         std::marker::PhantomData<fn() -> &$db_lt ()>,
                     );
 
-                    static $INTERN_CACHE: $zalsa::GlobalIngredientCache<$zalsa::interned::IngredientImpl<$Configuration>> =
-                        $zalsa::GlobalIngredientCache::new();
+                    static $INTERN_CACHE: $zalsa::IngredientCache<$zalsa::interned::IngredientImpl<$Configuration>> =
+                        $zalsa::IngredientCache::new();
 
                     impl $zalsa::SalsaStructInDb for $InternedData<'_> {
                         type MemoIngredientMap = $zalsa::MemoIngredientSingletonIndex;
@@ -165,22 +165,9 @@ macro_rules! setup_tracked_fn {
             impl $Configuration {
                 fn fn_ingredient(db: &dyn $Db) -> &$zalsa::function::IngredientImpl<$Configuration> {
                     let zalsa = db.zalsa();
-                    $FN_CACHE.get_or_create(zalsa, || {
-                        let index = zalsa.lookup_jar_by_type::<$fn_name>();
-
-                        let ingredient = zalsa.lookup_ingredient(index)
-                            .assert_type::<$zalsa::function::IngredientImpl<$Configuration>>();
-
-                        // If the ingredient has already been initialized, we know that the downcaster
-                        // has also been registered. This is a fast-path for multi-database use cases
-                        // that bypass the ingredient cache and will always execute this closure.
-                        if ingredient.is_initialized() {
-                            return (index, ingredient);
-                        }
-
-                        let view = <dyn $Db as $Db>::zalsa_register_downcaster(db);
-                        (index, ingredient.init(view))
-                    })
+                    $FN_CACHE
+                        .get_or_create(zalsa, || zalsa.lookup_jar_by_type::<$fn_name>())
+                        .get_or_init(|| <dyn $Db as $Db>::zalsa_register_downcaster(db))
                 }
 
                 pub fn fn_ingredient_mut(db: &mut dyn $Db) -> &mut $zalsa::function::IngredientImpl<Self> {
@@ -189,7 +176,7 @@ macro_rules! setup_tracked_fn {
                     let index = zalsa_mut.lookup_jar_by_type::<$fn_name>();
                     let (ingredient, _) = zalsa_mut.lookup_ingredient_mut(index);
                     let ingredient = ingredient.assert_type_mut::<$zalsa::function::IngredientImpl<Self>>();
-                    ingredient.init(view);
+                    ingredient.get_or_init(|| view);
                     ingredient
                 }
 
@@ -199,8 +186,7 @@ macro_rules! setup_tracked_fn {
                     ) -> &$zalsa::interned::IngredientImpl<$Configuration> {
                         let zalsa = db.zalsa();
                         $INTERN_CACHE.get_or_create(zalsa, || {
-                            let index = zalsa.lookup_jar_by_type::<$fn_name>().successor(0);
-                            (index, zalsa.lookup_ingredient(index).assert_type())
+                            zalsa.lookup_jar_by_type::<$fn_name>().successor(0)
                         })
                     }
                 }
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -106,8 +106,8 @@ macro_rules! setup_tracked_fn {
                         std::marker::PhantomData<fn() -> &$db_lt ()>,
                     );
 
-                    static $INTERN_CACHE: $zalsa::IngredientCache<$zalsa::interned::IngredientImpl<$Configuration>> =
-                        $zalsa::IngredientCache::new();
+                    static $INTERN_CACHE: $zalsa::GlobalIngredientCache<$zalsa::interned::IngredientImpl<$Configuration>> =
+                        $zalsa::GlobalIngredientCache::new();
 
                     impl $zalsa::SalsaStructInDb for $InternedData<'_> {
                         type MemoIngredientMap = $zalsa::MemoIngredientSingletonIndex;
@@ -193,9 +193,7 @@ macro_rules! setup_tracked_fn {
                     ) -> &$zalsa::interned::IngredientImpl<$Configuration> {
                         let zalsa = db.zalsa();
                         $INTERN_CACHE.get_or_create(zalsa, || {
-                            let index = zalsa.lookup_jar_by_type::<$Configuration>().successor(0);
-                            let ingredient = zalsa.lookup_ingredient(index).assert_type();
-                            (index, ingredient)
+                            zalsa.lookup_jar_by_type::<$Configuration>().successor(0)
                         })
                     }
                 }
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -91,8 +91,14 @@ macro_rules! setup_tracked_fn {
 
             struct $Configuration;
 
-            $zalsa::submit! {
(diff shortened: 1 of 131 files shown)


components/salsa-macros/src/db.rs
```diff
@@ -135,10 +135,11 @@ impl DbMacro {
         };
 
         input.items.push(parse_quote! {
+            #[cold]
+            #[inline(never)]
```
ibraheemdev, line 139: This was previously marked as `inline(always)`, but it's only called in the slow path for both single and multi-database use cases.

src/zalsa.rs
```diff
@@ -502,68 +500,63 @@ where
...
+        let nonce = zalsa.nonce().into_u32().get() as u64;
+        let packed = (nonce << u32::BITS) | (index.as_u32() as u64);
+        debug_assert_ne!(packed, IngredientCache::<I>::UNINITIALIZED);
+
+        // Discard the result, whether we won over the cache or not doesn't matter.
+        _ = self.cached_data.compare_exchange(
+            IngredientCache::<I>::UNINITIALIZED,
+            packed,
+            Ordering::Release,
+            Ordering::Relaxed,
```
ibraheemdev: I relaxed the failure ordering here because we don't synchronize on failure. FWIW this `compare_exchange` could even just be a `store` because it encapsulates all the cached data, but it doesn't really matter.

ibraheemdev:
I believe the benchmark results are somewhat misleading here because they use multiple databases. The fast path for accessing an ingredient is affected less (it doesn't include the `HashMap` lookup, only the `Vec`).

ibraheemdev:
Hmm looks like Miri doesn't support this yet https://github.com/rust-lang/miri/issues/450. That might be the dealbreaker I was anticipating.

Darksonn:
There are also many projects out there that ban life-before-main including inventory for various reasons including linker issues. For example, if salsa used inventory, we would never be able to import it into the Android codebase.

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -184,8 +193,7 @@ macro_rules! setup_tracked_fn {
                     ) -> &$zalsa::interned::IngredientImpl<$Configuration> {
                         let zalsa = db.zalsa();
                         $INTERN_CACHE.get_or_create(zalsa, || {
-                            <dyn $Db as $Db>::zalsa_register_downcaster(db);
```
ibraheemdev, line 187: I'm not exactly sure why this was here before, I don't think it's necessary. It's clearer now because we register the downcaster on the function ingredient directly.

src/database_impl.rs
```diff
@@ -7,7 +7,7 @@ use crate::{Database, Storage};
 /// require any custom user data.
 #[derive(Clone)]
 pub struct DatabaseImpl {
-    storage: Storage<Self>,
+    pub storage: Storage<Self>,
```
MichaReiser: What's the reason for making `storage` `pub?`

src/zalsa.rs
```diff
@@ -139,23 +140,20 @@ pub struct Zalsa {
...
 
     /// A map from the `IngredientIndex` to the `TypeId` of its ID struct.
     ///
     /// Notably this is not the reverse mapping of `jar_map`.
-    ingredient_to_id_struct_type_id_map: RwLock<FxHashMap<IngredientIndex, TypeId>>,
+    ingredient_to_id_struct_type_id_map: FxHashMap<IngredientIndex, TypeId>,
 
     /// Vector of ingredients.
     ///
     /// Immutable unless the mutex on `ingredients_map` is held.
```
MichaReiser:
```suggestion
```

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -156,26 +166,31 @@ macro_rules! setup_tracked_fn {
...
-                        if let Some(index) = jar_entry.get() {
-                            return index;
+                        if ingredient.is_initialized() {
+                            return (index, ingredient);
                         }
 
-                        <dyn $Db as $Db>::zalsa_register_downcaster(db);
-                        jar_entry.get_or_create()
+                        let view = <dyn $Db as $Db>::zalsa_register_downcaster(db);
+                        (index, ingredient.init(view))
```
MichaReiser: I understand that `FN_CACHE` uses `IngredientCache` over `GlobalIngredientCache` because it needs to run the downcaster registration for each `db`. 

Looking at `zalsa_register_downcaster`: The fast path is as simple as a `OnceLock` read. Is it worth keeping the `FN_CACHE` (which favors the first Db) or is the overhead so small that we could just run it always (and use `GlobalIngredientCache` instead). 

My other thinking is: Could we register the views statically during jar the jar registration? E.g. by adding an associated type to `HasJar` which is `$Db`?

src/ingredient_cache.rs
```diff
@@ -0,0 +1,187 @@
...
+/// Note that ingredients are statically registered with `inventory`, so their
+/// indices should be stable across any databases.
+///
+/// If ingredient initialization is database dependent, e.g. for registering
+/// view casters, [`IngredientCache`] should be used instead.
+pub struct GlobalIngredientCache<I>
+where
+    I: Ingredient,
+{
+    ingredient_index: AtomicU32,
```
MichaReiser: Could this be a `OnceLock` instead?

tests/manual_registration.rs
```diff
@@ -0,0 +1,49 @@
...
+fn test() {
+    let db = salsa::DatabaseImpl {
+        storage: salsa::Storage::builder()
+            .ingredient::<ingredients::track>()
+            .ingredient::<ingredients::intern>()
+            .ingredient::<ingredients::MyInput>()
+            .ingredient::<ingredients::MyTracked<'_>>()
+            .ingredient::<ingredients::MyInterned<'_>>()
+            .build(),
+    };
```
MichaReiser: Can we add a test where two databases are used but where the ingredients registered by one database are only a subset of the other database (and they are registered in different ordering).

src/zalsa.rs
```diff
@@ -177,19 +175,38 @@ impl RefUnwindSafe for Zalsa {}
...
+        let mut jars = inventory::iter::<ErasedJar>()
+            .copied()
+            .chain(jars)
+            .collect::<Vec<_>>();
+
+        #[cfg(not(feature = "inventory"))]
+        let mut jars = jars;
+
+        // Ensure structs are initialized before tracked functions.
+        jars.sort_by_key(|jar| jar.kind);
```
MichaReiser, line 204:
```suggestion
        let mut jars = if cfg!(feature = "inventory") {
        	inventory::iter::<ErasedJar>()
            .copied()
            .chain(jars)
            .collect::<Vec<_>>()
        } else {
        	jars
      	};

        // Ensure structs are initialized before tracked functions.
        jars.sort_by_key(|jar| jar.kind);
```

src/ingredient_cache.rs
```diff
@@ -0,0 +1,187 @@
...
+    fn get_or_create_index_slow(
+        &self,
+        load_index: impl Fn() -> IngredientIndex,
+    ) -> IngredientIndex {
+        let ingredient_index = load_index();
+
+        // It doesn't matter if we overwrite any stores, as `create_index` should
+        // always return the same index.
+        self.ingredient_index
+            .store(ingredient_index.as_u32(), Ordering::Release);
```
MichaReiser: I'm not sure this is guaranteed when using manual registration because different databases can have different registered ingredients.

components/salsa-macro-rules/src/setup_tracked_fn.rs
ibraheemdev: > Looking at zalsa_register_downcaster: The fast path is as simple as a OnceLock read. Is it worth keeping the FN_CACHE (which favors the first Db) or is the overhead so small that we could just run it always (and use GlobalIngredientCache instead).

Hmm where are you seeing a `OnceLock`? [`zalsa_register_downcaster` calls `Views::add`](https://github.com/salsa-rs/salsa/blob/master/components/salsa-macros/src/db.rs#L141) which does a linear search through a `boxcar::Vec` to find the viewcaster.

> My other thinking is: Could we register the views statically during jar the jar registration? E.g. by adding an associated type to `HasJar` which is `$Db`?

I don't think so, the downcaster relies on the concrete type of the database, which is why we need to wait until the function is called with a specific database to register it.

src/ingredient_cache.rs
ibraheemdev: Theoretically this should be faster because it's only a single load, `OnceLock` has an atomic flag separate from the data.

src/ingredient_cache.rs
ibraheemdev: Hmm you're right. I guess we still need the nonce check when the `inventory` feature is disabled? You shouldn't be able to register any ingredients manually that weren't already registered when `inventory` is enabled.

src/database_impl.rs
ibraheemdev: Didn't realize this was public API.

src/zalsa.rs
ibraheemdev: This doesn't work because the `inventory` crate is an optional dependency.

src/ingredient_cache.rs
MichaReiser: Hmm, this does mean that our inventory feature isn't strictly addeditve because you won't be able to use the manual registration functions when using inventory (or you can, they're just ignored)

src/ingredient_cache.rs
```diff
@@ -4,26 +4,33 @@ use std::num::NonZeroU32;
...
 ///
 /// If ingredient initialization is database dependent, e.g. for registering
 /// view casters, [`IngredientCache`] should be used instead.
+#[cfg(feature = "inventory")]
 pub struct GlobalIngredientCache<I>
 where
     I: Ingredient,
 {
-    ingredient_index: AtomicU32,
+    ingredient_index: std::sync::atomic::AtomicU32,
```
MichaReiser: What's the reason for using the `std::sync::atomic` here over our version in `crate::sync`?

MichaReiser (approved):
Nice. This is awesome. 

I wonder if this also fixes the non-determinism when replaying multithreaded tests. 

Requesting review from David/Lukas to get a thumbs up/down from the r-a side

Do you know how far off the miri changes are? Also, have you tested if the ty playground wroks with this new salsa version?

src/ingredient_cache.rs
ibraheemdev: With the inventory feature you can't create an ingredient except that it's automatically registered.

davidbarsky (approved):
given that miri supports this + this is a substantial speedup, i think we're okay with landing this. if we run into issues, well, we know where to report.

PR #936, opened 2025-07-17
Deduplicate tracked struct IDs on memos

ibraheemdev:
This is a pretty minor memory usage saving as most queries don't create tracked structs anyways (at least in ty), but still seemed worth doing.

### DIFF
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -10,7 +10,9 @@ use crate::key::DatabaseKeyIndex;
 use crate::runtime::Stamp;
 use crate::sync::atomic::AtomicBool;
 use crate::tracked_struct::{Disambiguator, DisambiguatorMap, IdentityHash, IdentityMap};
-use crate::zalsa_local::{QueryEdge, QueryOrigin, QueryRevisions, QueryRevisionsExtra};
+use crate::zalsa_local::{
+    FullQueryRevisions, QueryEdge, QueryOrigin, QueryRevisions, QueryRevisionsExtra,
+};
 use crate::{Accumulator, IngredientIndex, Revision};
 
 #[derive(Debug)]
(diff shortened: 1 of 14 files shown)


src/function/execute.rs
```diff
@@ -223,12 +229,14 @@ where
...
-                    opt_last_provisional = Some(self.insert_memo(
-                        zalsa,
-                        id,
-                        Memo::new(Some(new_value), zalsa.current_revision(), revisions),
-                        memo_ingredient_index,
-                    ));
+                    let memo = Memo::new(
+                        Some(new_value),
+                        zalsa.current_revision(),
+                        // TODO: Why is this correct?
```
ibraheemdev, line 235: I think this is always correct, but I don't know enough about how cycles work to know why aren't calling `diff_outputs` before creating this memo. Could there be old values in `tracked_struct_ids` here? If so, we are now counting them as output dependencies, whereas previously I think they would just waste memory.

ibraheemdev:
Hmm I did not expect that performance regression.

src/active_query.rs
```diff
@@ -26,13 +28,19 @@ pub(crate) struct ActiveQuery {
...
     /// * invocations of `specify`
     /// * accumulators pushed to
     input_outputs: FxIndexSet<QueryEdge>,
 
+    /// Tracked structs created by this query.
+    ///
+    /// This is stored separate from `input_outputs` and it is not actually
+    /// stored on the resulting memo, as tracked struct IDs are already stored
+    /// in the `IdentityMap`.
+    tracked_outputs: FxIndexSet<DatabaseKeyIndex>,
```
MichaReiser, line 42: Isn't a problem with moving this out of `input_outputs` that we no longer retain the ordering in which inputs/outputs where created. I'd have to think this through that `deep_verify_memo` is sensitive to re-orders. But maybe that only applies to inputs?

https://github.com/MichaReiser/salsa/blob/d28d66bf1390037f38abfe9cfcf9cd1ee4eb6f74/src/function/maybe_changed_after.rs#L438-L443

specifically https://github.com/MichaReiser/salsa/blob/d28d66bf1390037f38abfe9cfcf9cd1ee4eb6f74/src/function/maybe_changed_after.rs#L460-L467

ibraheemdev:
This seems quite complex for the minor memory usage savings, so I'll close this for now.

PR #937, opened 2025-07-18
Remove bounds and type checks from `IngredientCache`

ibraheemdev:
@Veykril might need to verify my logic here, but I think this is correct given either the `inventory` feature or the nonce checks, as `IngredientCache` always loads the index directly from the database.

The safety arguments for unchecked access are less clear for other uses of `lookup_ingredient`. The most notable one is in `DatabaseKeyIndex::maybe_changed_after`, where I think it's actually possible that a function records a dependency on an ingredient from a different database, in which case we should panic. It would be nice if we could guarantee that ingredient indices are stable even with manual registration, but that sort of defeats the purpose.

### DIFF
--- a/components/salsa-macro-rules/src/setup_accumulator_impl.rs
+++ b/components/salsa-macro-rules/src/setup_accumulator_impl.rs
@@ -34,9 +34,13 @@ macro_rules! setup_accumulator_impl {
                 static $CACHE: $zalsa::IngredientCache<$zalsa_struct::IngredientImpl<$Struct>> =
                     $zalsa::IngredientCache::new();
 
-                $CACHE.get_or_create(zalsa, || {
-                    zalsa.lookup_jar_by_type::<$zalsa_struct::JarImpl<$Struct>>()
-                })
+                // SAFETY: `lookup_jar_by_type` returns a valid ingredient index, and the only
+                // ingredient created by our jar is the struct ingredient.
+                unsafe {
+                    $CACHE.get_or_create(zalsa, || {
+                        zalsa.lookup_jar_by_type::<$zalsa_struct::JarImpl<$Struct>>()
+                    })
+                }
             }
 
             impl $zalsa::Accumulator for $Struct {
--- a/src/ingredient.rs
+++ b/src/ingredient.rs
@@ -177,7 +177,8 @@ pub trait Ingredient: Any + std::fmt::Debug + Send + Sync {
 }
 
 impl dyn Ingredient {
-    /// Equivalent to the `downcast` methods on `any`.
+    /// Equivalent to the `downcast` method on `Any`.
+    ///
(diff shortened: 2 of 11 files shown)


Veykril:
Sounds reasonable to me?
> where I think it's actually possible that a function records a dependency on an ingredient from a different database

Would be interesting if we could create a weird test for this (though not necessary for this PR of course)

components/salsa-macro-rules/src/setup_accumulator_impl.rs
```diff
@@ -34,9 +34,13 @@ macro_rules! setup_accumulator_impl {
                 static $CACHE: $zalsa::IngredientCache<$zalsa_struct::IngredientImpl<$Struct>> =
                     $zalsa::IngredientCache::new();
 
-                $CACHE.get_or_create(zalsa, || {
-                    zalsa.lookup_jar_by_type::<$zalsa_struct::JarImpl<$Struct>>()
-                })
+                // SAFETY: `lookup_jar_by_type` returns a valid ingredient index, and the only
+                // ingredient created by our jar is the struct ingredient.
+                unsafe {
```
MichaReiser, line 39: I think it's technically possible that a salsa user implements their own `Ingredient`. Could a malicious `Ingredient` implementation invalidate the SAFETY assumption?

src/ingredient.rs
```diff
@@ -192,7 +193,27 @@ impl dyn Ingredient {
         unsafe { transmute_data_ptr(self) }
     }
 
-    /// Equivalent to the `downcast` methods on `any`.
+    /// Equivalent to the `downcast` methods on `Any`.
+    ///
+    /// Because we do not have dyn-upcasting support, we need this workaround.
```
MichaReiser, line 198: We could consider bumping the MSRV.

davidbarsky: bumping msrv seems reasonable; rust-analyzer shouldn't have an issue with that (we're generally current-1).

ibraheemdev: Upcasting was stabilized in 1.86, which is quite recent. FWIW this is an old comment, so I'm not sure it makes sense to do in this PR (but I'm not opposed to it).

components/salsa-macro-rules/src/setup_accumulator_impl.rs
ibraheemdev: No, this just accesses the ingredient created by this jar.

src/ingredient.rs
MichaReiser: > FWIW this is an old comment, so I'm not sure it makes sense to do in this PR (but I'm not opposed to it).

Agree, I think it's better to do in a separate PR. I only wanted to point out that bumping the MSRV is a possiblity

PR #940, opened 2025-07-19
Optimize page access

ibraheemdev:
The codegen for this function is pretty terrible currently because of all the debug machinery.


MichaReiser (approved):

MichaReiser (commented):
Maybe add a comment explaining why it doesn't use assert

ibraheemdev:
Huh, I moved `type_name` to the vtable and apparently this is a regression now, even though it's only called in the panic path.

PR #941, opened 2025-07-20
Avoid dynamic dispatch to access memo tables

ibraheemdev:
We know the concrete type of the input `Id`, so we shouldn't need dynamic dispatch unless it is a supertype.

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_struct.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_struct.rs
@@ -231,6 +231,16 @@ macro_rules! setup_tracked_struct {
                         $zalsa::None
                     }
                 }
+
+                #[inline]
+                unsafe fn memo_table(
+                    zalsa: &$zalsa::Zalsa,
+                    id: $zalsa::Id,
+                    current_revision: $zalsa::Revision,
+                ) -> $zalsa::MemoTableWithTypes<'_> {
+                    // SAFETY: Guaranteed by caller.
+                    unsafe { zalsa.table().memos::<$zalsa_struct::Value<$Configuration>>(id, current_revision) }
+                }
             }
 
             impl $zalsa::TrackedStructInDb for $Struct<'_> {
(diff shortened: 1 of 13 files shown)


components/salsa-macro-rules/src/setup_tracked_struct.rs
```diff
@@ -231,6 +231,16 @@ macro_rules! setup_tracked_struct {
...
                     }
                 }
+
+                #[inline]
+                unsafe fn memo_table(
+                    zalsa: &$zalsa::Zalsa,
+                    id: $zalsa::Id,
+                    current_revision: $zalsa::Revision,
+                ) -> $zalsa::MemoTableWithTypes<'_> {
+                    // SAFETY: Guaranteed by caller.
```
MichaReiser, line 241: Can we extend the SAFETY documentation on the `memo_table` definition. It only mentions the constraints on `current_revision` but I don't think this is enough to guareantee the memo's type?

src/table.rs
```diff
@@ -220,17 +220,42 @@ impl Table {
...
-    /// Get the memo table associated with `id`
+    /// Get the memo table associated with `id` for the concrete type `T`.
     ///
-    /// # Safety condition
+    /// # Safety
     ///
-    /// The parameter `current_revision` MUST be the current revision
-    /// of the owner of database owning this table.
-    pub(crate) unsafe fn memos(
+    /// The parameter `current_revision` must be the current revision of the owner of database
```
MichaReiser:
```suggestion
    /// The parameter `current_revision` must be the current revision of the database owning this table.
```

components/salsa-macro-rules/src/setup_tracked_struct.rs
ibraheemdev: It's not a safety requirement for the type to match, the `memo_table` function uses `assert_type`, which will panic. It's possible for the user to pass an `Id` created by a different database, which makes it unsound to assume the type.

PR #942, opened 2025-07-20
Outline all tracing events

ibraheemdev:
It doesn't make sense for these to be inlined as they're typically disabled anyways.

### DIFF
--- a/src/database_impl.rs
+++ b/src/database_impl.rs
@@ -16,7 +16,7 @@ impl Default for DatabaseImpl {
             // Default behavior: tracing debug log the event.
             storage: Storage::new(if tracing::enabled!(Level::DEBUG) {
                 Some(Box::new(|event| {
-                    tracing::debug!("salsa_event({:?})", event)
+                    crate::tracing::debug!("salsa_event({:?})", event)
                 }))
             } else {
                 None
--- a/src/function/backdate.rs
+++ b/src/function/backdate.rs
@@ -34,7 +34,7 @@ where
             if revisions.durability >= old_memo.revisions.durability
                 && C::values_equal(old_value, value)
             {
-                tracing::debug!(
+                crate::tracing::debug!(
                     "{index:?} value is equal, back-dating to {:?}",
                     old_memo.revisions.changed_at,
                 );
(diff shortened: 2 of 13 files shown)


MichaReiser (approved):
An alternative to this could be something like https://github.com/bevyengine/bevy/pull/7639

So that the logs can be removed statically at compile time. It may require us to fine tune some of the log levels used by salsa

davidbarsky:
> > Personally, I don't recommend it. We made a mistake using cargo's features to compile out tracing; we should have used RUSTFLAGS instead.
> 
> Can you say more?

Cargo features are additive and therefore unify across the crate graph, while `RUSTFLAGS` do not. The latter can only be set by the _root_ Cargo package and _do not_ propagate across a `cargo publish`. This means that the existing compile-time filtering features in `tracing` can be enabled at distance by another dependency in your crate graph.

> How would a RUSTFLAGS approach work?

Take a look at [Tokio's usage](https://docs.rs/tokio/latest/tokio/#unstable-features); I think the examples are pretty solid.

ibraheemdev:
I realized that the `is_enabled` macro does a lot of work for checking fine-grained event filters, compared to checking `LevelFilter::current`, which just loads a global atomic. The downside of a compile-time setting is of course that you have to recompile salsa just to enable logging, which is particularly difficult for users trying to submit extra context for a bug report. I think it's fine to have a runtime check, especially if it's just a load and branch.

MichaReiser:
> FWIW Tracing also has max_level_* feature flags which should eliminate the branches if we check the STATIC_MAX_LEVEL constant FWIW, though that of course applies globally not just to salsa.

We use this in ty for `TRACE` level logs but Salsa has many debug/info level traces with which this won't help :(

ibraheemdev:
Alternatively, we could make the assumption that the logging filters do not change across calls to salsa (e.g. you aren't using scoped subscribers for different queries), at least within the same thread, and cache the check.

ibraheemdev:
I reverted the last commit, so this just uses `tracing::enabled!` now. This should still be a clear improvement and hopefully can be optimized more on the tracing side.

MichaReiser:
Oh no... you just demonstrated that there's still a salsa concurrency bug

```
    thread '<unnamed>' panicked at /home/runner/work/salsa/salsa/src/storage.rs:249:23:
      left: 2
     right: 0
```

PR #945, opened 2025-07-26
remove backdate field option, use maybe_update instead

puuuuh:


puuuuh:
> Whats the motivation for this change?

More straightforward code, especially in macro_rules.

PR #946, opened 2025-07-30
Gate accumulator feature behind a feature flag

Veykril:
Mainly wanna see if this has a perf impact, but the other rational is that rust-analyzer simply does not use this feature (and from past discussions we don't really see how we could use it given it defeats incrementality).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,12 +55,10 @@ jobs:
         run: cargo clippy --workspace --all-targets -- -D warnings
       - name: Test
         run: cargo nextest run --workspace --all-targets --no-fail-fast
-      - name: Test Manual Registration
+      - name: Test Manual Registration / no-default-features
         run: cargo nextest run --workspace --tests --no-fail-fast --no-default-features --features macros
       - name: Test docs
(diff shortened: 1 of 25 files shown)


MichaReiser:
> and from past discussions we don't really see how we could use it given it defeats incrementality

I'm not sure I understand this. Or is this about the expensive tree traversal to collect all values?

Veykril:
Maybe I am misremembering that part.

But either way they seem quite expensive. All we would need them for in rust-analyzer is diagnostics I believe, and there hand rolling things is likely a lot cheaper than using accumulators. We don't need to support accumulating same types across different queries (at least for now), and all that indirection + type erasing seems not so ideal compared to just collecting things into a vec manually (which is what we do for diagnostics atm)

Veykril:
Before codspeed overwrites it, 
image
these are the bench results with accumulator disabled

ibraheemdev:
This seems reasonable given that we don't use accumulators in ty, but I'll leave it up to @MichaReiser who has more context.

MichaReiser (approved):
I didn't review the PR line by line but the gating does make sense to me

PR #947, opened 2025-07-30
Do manual trait upcasting instead of downcasting

Veykril:


src/runtime.rs
```diff
@@ -54,67 +55,6 @@ pub(crate) enum BlockResult<'me> {
     Cycle { same_thread: bool },
 }
 
-pub struct Running<'me>(Box<BlockedOnInner<'me>>);
```
MichaReiser, line 57: I added the `Box` to keep the struct size small (the case that a thread is blocked should be rare and the allocation is probably cheap in comparison to the actual waiting)

Veykril: I am aware, this retains the size by stashing the state away in `DependencyGraph` instead of allocating it anew, given while all of this happening the mutex of `DependencyGraph` is held throughout, so there is only ever a single `Running` instance for a runtime.

But you are right in that this is a scenario that seldom happens (two threads racing on a query), was mainly curious how this would look like.

Veykril:
meh, std/shuttle's mutexguard is not `Send` :/

Also looks like the allocation changes cause a bunch of noise in the benches (the spans changed there should be unrelated to this change)

PR #950, opened 2025-08-01
Revert "remove borrow checks from `ZalsaLocal`"

ibraheemdev:
This is showing up as a 7% regression on master?


Veykril:
That is odd ...

ibraheemdev:
The benchmarks are super noisy now and jump +/-10% spuriously, so it's unlikely this is actually an improvement.

PR #951, opened 2025-08-02
fix: Do not unnecessarily require `Debug` on fields for interned structs

Veykril:


components/salsa-macro-rules/src/setup_interned_struct.rs
```diff
@@ -256,22 +256,58 @@ macro_rules! setup_interned_struct {
...
-                        f.finish()
-                    }).unwrap_or_else(|| {
-                        f.debug_tuple(stringify!($Struct))
-                            .field(&$zalsa::AsId::as_id(&this))
-                            .finish()
-                    })
+            // Duplication can be dropped here once we no longer allow the `no_lifetime` hack
+            $zalsa::macro_if! {
+                iftt ($($db_lt_arg)?) {
+                    impl $Struct<'_> {
```
ibraheemdev, line 264: Shouldn't `impl<$($db_lt_arg)?> $Struct<$($db_lt_arg)?>` work? Why does this need a separate `impl` block at all?

Veykril: That causes issues with the higher ranked where clauses due to lifetime shadowing unfortunately

Issue #952, opened 2025-08-02
Multithreaded `maybe_changed_after` with fixpoint iteration.


MichaReiser:
I believe there's a bug in salsa related to multithreaded validation after a revision change here:

https://github.com/MichaReiser/salsa/blob/f303b6db56f6c95501c2f27c8ee8823983094568/src/function/maybe_changed_after.rs#L515-L529

This function assumes that the entire cycle is on the same thread. However, that's not guaranteed and `maybe_changed_after` isn't allowed to return if the cycle head is running on another thread. 

I think `maybe_changed_after` requries similar treatment to `fetch`:

https://github.com/MichaReiser/salsa/blob/f303b6db56f6c95501c2f27c8ee8823983094568/src/function/fetch.rs#L249-L261

PR #954, opened 2025-08-02
allow reuse of cached provisional memos within the same cycle iteration during `maybe_changed_after`

MichaReiser:
This is the same as https://github.com/salsa-rs/salsa/pull/786 but during `maybe_changed_after`. 

`validate_same_iteration` over pessimistically always returns `false` if a cycle head is currently being checked with `maybe_changed_after` because `maybe_changed_after` never pushes the current query on the stack. 

This PR addresses the issue by comparing the cycle head's iteration with the latest memo iteration for that cycle head. This is the same as we already did for a cycle that spans multiple threads.

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -379,7 +379,7 @@ where
                             // However, we've to be careful to skip over fixpoint initial values:
                             // If the head is the memo we're trying to validate, always return `None`
                             // to force a re-execution of the query. This is necessary because the query
-                            // has obvioulsy not completed its iteration yet.
+                            // has obviously not completed its iteration yet.
                             //
                             // This should be rare but the `cycle_panic` test fails on some platforms (mainly GitHub actions)
                             // without this check. What happens there is that:


src/function/maybe_changed_after.rs
```diff
@@ -377,7 +377,7 @@ where
                             let wait_result = ingredient
                                 .wait_for(zalsa, cycle_head.database_key_index.key_index());
 
-                            if !wait_result.is_cycle_with_other_thread() {
+                            if !wait_result.is_cycle() {
```
MichaReiser, line 407: Trying to remember why I explicitly added `is_cycle_with_other__thread` in [`30f6eb4` (#882)](https://github.com/salsa-rs/salsa/pull/882/commits/30f6eb4b62fa0491cff41eaff2d5a07aadcb27b2)

It suggests that it is to fix a panic but there's no failing test :( Maybe it only failed on CI?

MichaReiser: Nice, it reproduces on CI (but why only on CI?)

MichaReiser: I initially restricted the logic here to only apply when this leads to a cross-thread cycle because the `cycle_panic` test started to hang on Github actions. 

Restricting this logic to multi-threaded cycle does fix the test-hang but it's over restrictive. The alternative fix that I added now (the if below the huge comment) accomplishes the same without making any assumptions about on how many threads the cycle runs.

An alternative fix is to change `maybe_changed_after` to push the current query in `deep_verify_memo`. While I think that this would have the nice added benefit that our query stack traces are more accurate (they also show queries running `maybe_changed_after), it does have the significant downside that pushing a new query isn't free and we would have to pay that cost for all queries, not just queries participating in cycles. 

That's why I opted for this fix instead.

tests/cycle.rs
```diff
@@ -1025,3 +1025,42 @@ fn repeat_provisional_query() {
...
+    // `validate_same_iteration` incorrectly returns `false`.
+    db.assert_logs(expect![[r#"
+        [
+            "salsa_event(WillExecute { database_key: min_panic(Id(2)) })",
+            "salsa_event(WillExecute { database_key: min_panic(Id(1)) })",
+            "salsa_event(WillExecute { database_key: min_iterate(Id(0)) })",
+            "salsa_event(WillIterateCycle { database_key: min_iterate(Id(0)), iteration_count: IterationCount(1), fell_back: false })",
+            "salsa_event(WillExecute { database_key: min_panic(Id(1)) })",
+            "salsa_event(WillExecute { database_key: min_panic(Id(2)) })",
+        ]"#]]);
```
MichaReiser, line 1065: The output before the fix:

```
            "salsa_event(WillExecute { database_key: min_panic(Id(1)) })",
            "salsa_event(WillExecute { database_key: min_panic(Id(2)) })",
            "salsa_event(WillExecute { database_key: min_panic(Id(2)) })", <- unnecesary
            "salsa_event(WillExecute { database_key: min_panic(Id(2)) })", <- unnecesary
            "salsa_event(WillIterateCycle { database_key: min_iterate(Id(0)), iteration_count: IterationCount(1), fell_back: false })",
            "salsa_event(WillExecute { database_key: min_panic(Id(1)) })",
            "salsa_event(WillExecute { database_key: min_panic(Id(2)) })",
```

MichaReiser:
No idea what's up with codspeed. This code doesn't exercise any fixpoint iteration logic

src/function/maybe_changed_after.rs
```diff
@@ -369,15 +369,42 @@ where
...
-                            // check if it has the same iteration count.
+                            // If the cycle head isn't on our stack because:
+                            //
+                            // * another thread holds the lock on the cycle head (but it waits for the current query to complete)
+                            // * we're in `maybe_changed_after` because `maybe_changed_after` doesn't modify the cycle stack
+                            //
+                            // check if the latest memo has the same iteration count.
+
+                            // However, we've to be careful to skip over fixpoint initial values:
+                            // If the head is the memoy we're trying to validate, always return `None`
```
Veykril:
```suggestion
                            // If the head is the memo we're trying to validate, always return `None`
```

src/function/maybe_changed_after.rs
```diff
@@ -369,15 +369,42 @@ where
...
+                            //
+                            // * another thread holds the lock on the cycle head (but it waits for the current query to complete)
+                            // * we're in `maybe_changed_after` because `maybe_changed_after` doesn't modify the cycle stack
+                            //
+                            // check if the latest memo has the same iteration count.
+
+                            // However, we've to be careful to skip over fixpoint initial values:
+                            // If the head is the memo we're trying to validate, always return `None`
+                            // to force a re-execution of the query. This is necessary because the query
+                            // has obvioulsy not completed its iteration yet.
```
carljm:
```suggestion
                            // has obviously not completed its iteration yet.
```

carljm (approved):
Looks good, thank you!

MichaReiser:
I believe there are a few more bugs in `maybe_changed_after` related to fixpoint but I'll tackle them in separate PRs:

1. I don't think it's safe to return `Unchanged` if `validate_same_iteration` returns `true`. We simply don't know yet if the value has changed.
2. We should never return `Unchanged` for a provisional value (here https://github.com/MichaReiser/salsa/blob/c2f4827b512b82842dbc84b1ccc4367500e301ed/src/function/maybe_changed_after.rs#L191-L193)
3. We need to update `maybe_changed_after_cold` to do the same retry logic as for `fetch` to correctly handle multi threaded validation, see https://github.com/salsa-rs/salsa/issues/952

PR #955, opened 2025-08-02
Extract the cycle branches from `fetch` and `maybe_changed_after`

MichaReiser:
and mark them as cold. I also reordered some other match statements so that the most likely branches come first.


src/function/maybe_changed_after.rs
```diff
@@ -557,6 +545,33 @@ where
                     accumulated: inputs,
                 }
             }
+
+            QueryOriginRef::Assigned(_) => {
```
carljm, line 577: This change is just to put the more frequent match arm first?

carljm (approved):
Looks good!

src/function/fetch.rs
```diff
@@ -272,4 +213,76 @@ where
 
         Some(memo)
     }
+
+    #[cold]
```
ibraheemdev, line 217: I'd also add `#[inline(never)]`, because LLVM tends to be aggressive with inlining even with `#[cold]`.

ibraheemdev (approved):

src/function/maybe_changed_after.rs
MichaReiser: Yeah. I find this also helps with readability because I can reason about the common case first

PR #958, opened 2025-08-03
Replace manual `WaitGroup` impl with `crossbeam`'s

Veykril:
Slight code reduction, though this does mean cancellation incurs an arc allocation and mutex allocation (as we need to recreate the wait group).


MichaReiser:
Is there any other benefit than code size reduction. It seems codspeed doesn't like it at all.

Veykril:
No there isn't. I was curious about the perf impact, I agree this isn't great. Checking if we can actually simplify our impl now

Veykril:
Hm, we could get rid of the additional `Arc` as well and re-use the `Zalsa` arc but that makes the code a lot more ugly (thanks to having to deal with drop order), so this is the best I can come up with

MichaReiser:
@ibraheemdev could you take a look at this PR?

src/storage.rs
```diff
@@ -168,12 +156,14 @@ impl<Db: Database> Storage<Db> {
...
-            clones = self.handle.coordinate.cvar.wait(clones);
-        }
-        // The ref count on the `Arc` should now be 1
-        let zalsa = Arc::get_mut(&mut self.handle.zalsa_impl).unwrap();
+        let mut coordinate_lock = self.handle.coordinate.coordinate_lock.lock();
+        let zalsa = loop {
+            if let Some(zalsa) = Arc::get_mut(&mut self.handle.zalsa_impl) {
+                // SAFETY: Polonius when ... https://github.com/rust-lang/rfcs/blob/master/text/2094-nll.md#problem-case-3-conditional-control-flow-across-functions
+                break unsafe { mem::transmute::<&mut Zalsa, &mut Zalsa>(zalsa) };
+            }
```
ibraheemdev:
```suggestion
            if Arc::strong_count(&self.handle.zalsa_impl) == 1 {
                // SAFETY: The strong count is 1, and we never create any weak pointers,
                // so we have a unique reference.
                break unsafe { &mut *(Arc::as_ptr(&self.handle.zalsa_impl) as *mut Zalsa) };
            }
```

This should avoid the Polonius issue and is a bit cheaper. There is a little question about writing through `Arc::as_ptr`, but it is valid in the current implementation and nothing is really blocking it being guaranteed (see https://github.com/rust-lang/rust/pull/104337).

ChayimFriedman2:
This introduced a potential to deadlock. I'm still investigating how exactly, but my guess is that the atomic ordering is not correct (it's not at least in one thing, but that doesn't fix the deadlock).

Issue #960, opened 2025-08-04
`mark_validated_output` can blow up for queries that create many tracked structs


MichaReiser:
Here's a trace from ty: https://share.firefox.dev/4mkOjl2

Notice how `mark_validated_output` takes about 27% of the runtime (which is 15s ;)) 

The worst about this is, that I think, this is not even needed anymore because the tracked struct's `mark_validated_output` method is empty! 

https://github.com/MichaReiser/salsa/blob/c2f4827b512b82842dbc84b1ccc4367500e301ed/src/tracked_struct.rs#L826-L835

What I believe is now the only reason for tracking tracked structs as output is so that we can diff them in `diff_outputs`. 

https://github.com/MichaReiser/salsa/blob/c2f4827b512b82842dbc84b1ccc4367500e301ed/src/function/diff_outputs.rs#L29-L35

We should investigate whether `diff_outputs` can use `tracked_struct_ids` instead. 

Removing the tracked structs from the outputs reduce the walltime from 15s to 2s (there's more that's going wrong here but this seem a significant improvement on its own)

The remaining two seconds come from the code here

https://github.com/MichaReiser/salsa/blob/c2f4827b512b82842dbc84b1ccc4367500e301ed/src/function/execute.rs#L268-L272

In fact, commenting out this line on its own fixes the entire runtime blow-up.

MichaReiser:
Okay, I know what the issue is. The problem is that the way we currently pass `cycle_heads` to `maybe_changed_after` not only propagates cycle heads upwards (which is what we want) but also to sibling queries. That means, we fail to mark queries as verified that aren't participating in a cycle and we re-validate the same queries over and over again. Some logs from ty:

```
2025-08-04T15:58:29.610181Z  WARN salsa::function::maybe_changed_after: Can't mark semantic_index(Id(2)) as verified because it's part of a cycle
2025-08-04T15:58:29.610187Z  WARN salsa::function::maybe_changed_after: Can't mark source_text(Id(2)) as verified because it's part of a cycle
2025-08-04T15:58:29.610830Z  WARN salsa::function::maybe_changed_after: Can't mark parsed_module(Id(2)) as verified because it's part of a cycle
2025-08-04T15:58:29.610835Z  WARN salsa::function::maybe_changed_after: Can't mark source_text(Id(2)) as verified because it's part of a cycle
[...]

PR #961, opened 2025-08-05
Fix `maybe_changed_after` runnaway for fixpoint queries

MichaReiser:
This PR fixes a few bugs related to `maybe_changed_after` and fixpoint iteration.

## Runnaway 

The most important fix is for a runaway issue where `maybe_changed_after` revalidates the same queries repeatedly, see https://github.com/salsa-rs/salsa/issues/960.

Given a query:

```
a -> b -> a
       -> c
```

C doesn't participate in the cycle, but we didn't mark it as verified before this PR because the outer query `a is participating in a not-yet-resolved cycle. 

This PR fixes this by tracking the cycles per query (each subtree tracks the cycles independently). I added `cycle_sibling_interference` as a regression test. 

## `validate_same_iteration`

`maybe_changed_after` internally calls `deep_verify_memo`, which, on master, calls `maybe_validate_provisional`, which calls `validate_same_iteration`. `validate_same_iteration` returns `true` for a provisional query if the cycle heads between its cycle head entries and the latest available memos for those cycle heads have the same iteration count. This is important in `fetch`, where we want to reuse cached query results during a fixpoint iteration. 

However, we shouldn't return `Unchanged` in `maybe_changed_after` for a memo that's not finalized but `validate_same_iteration` returns `true` because we simply don't know yet if that value has changed or not (we need to wait for its cycle heads to resolve).

This PR moves the `validate_same_iteration` check out of `deep_verify_memo` and instead duplicates it inside `fetch_cold` and `maybe_changed_after_cold`, where only `fetch_cold` returns early when `validate_same_iteration` returns `true`. 

I tried to write a regression test for this but I failed. However, we have an incremental integration test in ty that starts failing without this change (because a query now returns an outdated result).

## Re-execution and provisional memos

If there are no pending cycles in `maybe_changed_after` and there's an old memo, Salsa executes the query again and then compares the `last_changed_at` with the `revision` passed to `maybe_changed_after` to decide if the value has changed. 

This PR adds an additional check to only return `unchanged` if the value isn't provisional. 

I don't know if this is a 100% necessary and I failed to write an integration test for it (and no test in ty fails). So it might be that some other combination of checks prevent this but I can't say for sure. That's why I added the check.

### DIFF
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -148,28 +148,7 @@ where
                     // initial provisional value from there.
                     let memo = self
                         .get_memo_from_table_for(zalsa, id, memo_ingredient_index)
-                        .map(|memo| {
-                            if memo.may_be_provisional() {
-                                memo
-                            } else {
-                                // It's possible that we don't have a provisional value if we hit the cycle
-                                // in `maybe_changed_after`. There, we only return Changed/Unchanged but never insert
-                                // a memo.
-                                let revisions =
-                                    QueryRevisions::fixpoint_initial(database_key_index);
-                                let initial_value = C::cycle_initial(db, C::id_to_input(zalsa, id));
-                                self.insert_memo(
-                                    zalsa,
-                                    id,
-                                    Memo::new(
-                                        Some(initial_value),
-                                        zalsa.current_revision(),
-                                        revisions,
-                                    ),
-                                    memo_ingredient_index,
-                                )
-                            }
-                        })
+                        .filter(|memo| memo.verified_at.load() == zalsa.current_revision())
                         .unwrap_or_else(|| {
                             unreachable!(
                                 "{database_key_index:#?} is a cycle head, \
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -180,7 +180,6 @@ where
             return Some(VerifyResult::unchanged());
         }
 
-        // Check if the inputs are still valid. We can just compare `changed_at`.
         let deep_verify = self.deep_verify_memo(
             db,
             zalsa,
@@ -191,6 +190,7 @@ where
         );
 
         if deep_verify.is_unchanged() {
+            // Check if the inputs are still valid. We can just compare `changed_at`.
             return Some(if old_memo.revisions.changed_at > revision {
                 VerifyResult::changed()
             } else {


src/function/maybe_changed_after.rs
```diff
@@ -155,54 +229,67 @@ where
...
                 .push_query(database_key_index, IterationCount::initial());
             let memo = self.execute(db, active_query, Some(old_memo));
             let changed_at = memo.revisions.changed_at;
 
-            return Some(if changed_at > revision {
-                VerifyResult::Changed
+            // Always assume that a provisional value has changed (we simply don't know yet
+            // and we need to iterate some outer cycle to determine it, which we can't do here).
+            return Some(if changed_at > revision || memo.may_be_provisional() {
+                deep_verify.into_changed()
```
MichaReiser: Should this copy over the cycle heads from `memo`?

tests/cycle.rs
```diff
@@ -994,6 +994,79 @@ fn cycle_unchanged_nested_intertwined() {
...
+    // - b re-executes again during cycle iteration
+    db.assert_logs(expect![[r#"
+        [
+            "salsa_event(DidValidateMemoizedValue { database_key: min_panic(Id(2)) })",
+            "salsa_event(WillExecute { database_key: min_iterate(Id(0)) })",
+            "salsa_event(WillExecute { database_key: min_iterate(Id(1)) })",
+            "salsa_event(WillExecute { database_key: min_panic(Id(3)) })",
+            "salsa_event(WillIterateCycle { database_key: min_iterate(Id(0)), iteration_count: IterationCount(1), fell_back: false })",
+            "salsa_event(WillExecute { database_key: min_iterate(Id(1)) })",
+        ]"#]]);
```
MichaReiser, line 1067: The output before this change was:

```
[
    "salsa_event(WillExecute { database_key: min_iterate(Id(1)) })",
    "salsa_event(WillExecute { database_key: min_panic(Id(3)) })",
    "salsa_event(WillIterateCycle { database_key: min_iterate(Id(0)), iteration_count: IterationCount(1), fell_back: false })",
    "salsa_event(WillExecute { database_key: min_iterate(Id(1)) })",
]
```

Note how `c` (`min_panic(Id(2))`) doesn't get verified as part of `a`'s `deep_verify_memo`. Instead, it only gets verified once `a` re-runs.

src/function/maybe_changed_after.rs
```diff
@@ -138,56 +161,90 @@ where
...
+        //   * depends on `c` (input)
+        //   * creates interned `x` (input)
+        // * query `c`
+        //   * depends on `a` (input)
+        //   * depeonds on `d` (input/changed)
+        //   * reads `x`
+        //
+        // When `maybe_changed_after` runs on `a`, it recurses into `b` and then `c`. Now, `c` will recurse into `a` again
+        // where we return `Unchanged` but add `a` to the cycle heads. `maybe_changed_after` now continues checking `d`
+        // which returns `changed` so we'll reach the point here. Now, the reason we can't `execute` c
```
davidbarsky: what's "the point here"? this part of the function, or...?

src/function/maybe_changed_after.rs
```diff
@@ -138,56 +161,90 @@ where
...
+        // is because `a` hasn't re-interned `x` at this point (we started recursing into `c`).
+        //
+        // We can't run into this situation with normal queries because a dependency can't depent on values from
```
davidbarsky:
```suggestion
        // We can't run into this situation with normal queries because a dependency can't depend on values from
```

src/function/maybe_changed_after.rs
```diff
@@ -138,56 +161,90 @@ where
...
+        // the outer query (it's a tree and not a graph). However, it's possible with cycles because
```
davidbarsky:
```suggestion
        // the outer query—queries form a tree, not a graph. However, this is possible with cycles because
```

src/function/maybe_changed_after.rs
```diff
@@ -138,56 +161,90 @@ where
...
+        // an interned or tracked struct like `x` can flow into a dependent query as part of a later iteration and
+        // those values come later in `input_outputs` than the first call to the query that ultimately causes the cycle.
```
davidbarsky: Some phrasing nits/clarity, since I had a some difficulty understanding this *and* that this code seems kinda complex 😄 

```suggestion
        // an interned or tracked struct like `x` can flow into a dependent query as part of a subsequent iteration.
        // The values that appear in subsequent cycles become populated in `input_outputs` rather than the first call to the query that caused the cycle to begin with.
```

Also, what's `input_outputs`? A Command+F on this page didn't turn it up.

src/function/maybe_changed_after.rs
```diff
@@ -138,56 +161,90 @@ where
...
+        //
+        // However, executing the query here is only safe if there's no "unverified"
+        // cycle. Let's assume we have:
+        // * query `a`
+        //   * depends on `b` (input)
+        //   * depends on `c` (input)
+        //   * creates interned `x` (input)
+        // * query `c`
+        //   * depends on `a` (input)
+        //   * depeonds on `d` (input/changed)
```
davidbarsky:
```suggestion
        //   * depends on `d` (input/changed)
```

src/function/maybe_changed_after.rs
```diff
@@ -138,56 +161,90 @@ where
...
             let active_query = db
                 .zalsa_local()
                 .push_query(database_key_index, IterationCount::initial());
             let memo = self.execute(db, active_query, Some(old_memo));
             let changed_at = memo.revisions.changed_at;
 
-            return Some(if changed_at > revision {
-                VerifyResult::Changed
+            // Always assume that a provisional value has changed (we simply don't know yet
+            // and we need to iterate some outer cycle to determine it, which we can't do here).
```
davidbarsky:
```suggestion
            // Always assume that a provisional value has changed. 
            // 
            // We don't know if a provisional value has actually changed. To determine whether a provisional
            // value has changed, we need to iterate some outer cycle, which cannot be done here.
```

src/function/maybe_changed_after.rs
```diff
@@ -490,13 +544,28 @@ where
                 for &edge in edges {
                     match edge.kind() {
                         QueryEdgeKind::Input(dependency_index) => {
-                            match dependency_index.maybe_changed_after(
+                            debug_assert!(child_cycle_heads.is_empty());
+
+                            // Pass fresh cycle heads to the child query to avoid
+                            // that cycles introduced by siblings prevent queries from being verified.
```
davidbarsky: I'm really able to parse this sentence. consider maybe rephrasing?

davidbarsky (commented):
a few nits/asks for clarification

src/function/maybe_changed_after.rs
MichaReiser: I reverted most of that comment because I wasn't able to prove that this is the specific scenario

src/function/maybe_changed_after.rs
```diff
@@ -625,3 +677,72 @@ impl ShallowUpdate {
...
+/// calling [`maybe_changed_after`]. The [`maybe_changed_after`] then collects all cycle heads
+/// encountered while verifying this ingredient and its subtree.
+///
+/// Note that `heads` only contains the cycle heads up to the point where [`maybe_changed_after`]
+/// returned [`VerifyResult::Changed`]. Cycles that only manifest when verifying later dependencies
+/// aren't included.
+///
+/// [`maybe_changed_after`]: crate::ingredient::Ingredient::maybe_changed_after
+#[derive(Debug, Default)]
+pub struct MaybeChangeAfterCycleHeads {
```
carljm: naming nit -- the missing `d` is inconsistent with the existing name `maybe_changed_after`, and gives it a different meaning
```suggestion
pub struct MaybeChangedAfterCycleHeads {
```

Alternatively I think a name like `ValidationCycleHeads` could also work

src/function/execute.rs
```diff
@@ -148,12 +148,35 @@ where
...
+                                    id,
+                                    Memo::new(
+                                        Some(initial_value),
+                                        zalsa.current_revision(),
+                                        revisions,
+                                    ),
+                                    memo_ingredient_index,
+                                )
+                            }
+                        })
```
carljm: I didn't see this case discussed in the PR description (unless I misunderstood the part that was describing it. Is there a test that exercises this case? Or where did you see it occur?

Without this, were we hitting the unreachable below, or were we pulling an old memo from a previous revision? If the latter, should we be checking the revision number on the memo too (instead of just checking `may_be_provisional`)?

src/function/maybe_changed_after.rs
```diff
@@ -138,21 +161,41 @@ where
...
+                // See https://github.com/salsa-rs/salsa/pull/961
+                false,
+            )
         {
+            self.update_shallow(zalsa, database_key_index, old_memo, can_shallow_update);
+
+            return Some(VerifyResult::unchanged());
+        }
+
+        // Check if the inputs are still valid. We can just compare `changed_at`.
```
carljm: This comment seems obsolete (or at least mis-located)? Deep-verify is more than "just comparing `changed_at`"

carljm (approved):
Looks good, thank you!

src/function/maybe_changed_after.rs
MichaReiser: I went with `VerifyCycleHeads` so that it matches `VerifyResult`

src/function/execute.rs
MichaReiser: I think you're right that this isn't necessary anymore. It only was when I changed the check whether we can call `execute` in `maybe_changed_after` to only test for own cycles. I'll leave this fix for now. It seems easy enough to change this later.

But I'll add a check that the memo is from the same revision so that we never pick up a value from a previous revision.

PR #962, opened 2025-08-05
Remove allocation lock

ibraheemdev:
I believe we don't need the lock as pages are local to a given thread. I also moved the `type_name` field to the vtable, because it's only used in the cold path. I've noticed that any changes to the `Page` can make codspeed angry, so I'll try running the benchmarks on ty as well.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -60,9 +60,11 @@ struct SlotVTable {
     /// [`Slot`] methods
     memos: SlotMemosFnRaw,
     memos_mut: SlotMemosMutFnRaw,
+    /// The type name of what is stored as entries in data.
+    type_name: fn() -> &'static str,
     /// A drop impl to call when the own page drops
-    /// SAFETY: The caller is required to supply a correct data pointer to a `Box<PageDataEntry<T>>` and initialized length,
-    /// and correct memo types.
+    /// SAFETY: The caller is required to supply a valid pointer to a `Box<PageDataEntry<T>>`, and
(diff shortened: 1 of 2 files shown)


Veykril (approved):

src/table.rs
```diff
@@ -347,15 +344,14 @@ impl<'db, T: Slot> PageView<'db, T> {
         // SAFETY: `index` is also guaranteed to be in bounds as per the check above.
         let entry = unsafe { &*data.as_ptr().add(index) };
 
-        // SAFETY: We acquired the allocation lock, so we have unique access to the UnsafeCell
-        // interior
+        // SAFETY: The caller guarantees we are the unique writer, and readers will not attempt to
+        // access this index until we have updated the length.
```
MichaReiser, line 348: What I understand from this comment is that the table access must be exlusive to one thread when `allocate` is called. This is different from how I read the comment on `allocate` where I assumed that it's only about that concurrent calls to `allocate` but a concurrent `allcoate` and some read would be fine. 

Should we also mark all read methods as unsafe if the entire table access must be exclusive (no concurrent writes and reads)?

MichaReiser (approved):
Nice. Did you see any improvements in ty's benchmark?

src/table.rs
Veykril: concurrent reads to a write are fine, we only bump the allocated index after a slot has been initialized, so readers cannot observe a slot that is in the middle of being allocated

PR #963, opened 2025-08-06
test: add parallel maybe changed after test

MichaReiser:
This PR adds a parallel test for `maybe_changed_after`. 

It demonstrates that what I raised in https://github.com/salsa-rs/salsa/issues/952 isn't an issue.

Let's say we have `a -> b -> a` and `a` runs on thread `1` and thread `2` runs `maybe_changed_after` for `b`. 

`b` calls `maybe_changed_after(a)` which returns `unchanged` with a cycle head of `a`. `b` propagates that cycle head. 
Now, `fetch` only skips `execute` for queries that have no cycle heads. Given that this isn't the case, `b` will be executed. 

This can lead to some queries being re-executed when it isn't aboslutely necessary but that's probably fine.

Closes https://github.com/salsa-rs/salsa/issues/952


tests/parallel/cycle_nested_deep_conditional_changed.rs
```diff
@@ -0,0 +1,151 @@
+//! Test a deeply nested-cycle scenario where cycles have changing query dependencies.
+//!
+//! The trick is that different threads call into the same cycle from different entry queries and
+//! the cycle heads change over different iterations
+//!
+//! * Thread 1: `a` -> b -> c
```
carljm:
```suggestion
//! * Thread 1: `a` -> `b` -> `c`
```

PR #967, opened 2025-08-08
Initial persistent caching prototype

ibraheemdev:
The overall design is laid out [in this document](https://hackmd.io/@ibraheemdev/S17Bxowwxg). All the new functionality is gated under the `persistence` feature. However, this PR has a couple of limitations:

- All transitive dependencies are forced to be serializable.
- There is no easy way to update inputs after they are deserialized.
- Query hashing is not implemented.
- Accumulators and supertypes are not serializable.

### DIFF
--- a/src/database.rs
+++ b/src/database.rs
@@ -156,7 +156,7 @@ pub fn current_revision<Db: ?Sized + Database>(db: &Db) -> Revision {
 mod persistence {
     use crate::plumbing::Ingredient;
     use crate::zalsa::Zalsa;
-    use crate::{Database, Runtime};
+    use crate::{Database, IngredientIndex, Runtime};
 
     use std::fmt;
 
@@ -205,21 +205,14 @@ mod persistence {
                 .filter(|ingredient| ingredient.should_serialize(zalsa))
                 .collect::<Vec<_>>();
 
-            ingredients.sort_by(|a, b| {
-                // Ensure structs are serialized before tracked functions, as deserializing a
-                // memo requires its input struct to have been deserialized.
-                //
-                // We also further sort by debug name, to maintain a consistent ordering across
-                // builds.
-                a.jar_kind()
-                    .cmp(&b.jar_kind())
-                    .then(a.debug_name().cmp(b.debug_name()))
-            });
+            // Ensure structs are serialized before tracked functions, as deserializing a
+            // memo requires its input struct to have been deserialized.
+            ingredients.sort_by_key(|ingredient| ingredient.jar_kind());
 
             let mut map = serializer.serialize_map(Some(ingredients.len()))?;
             for ingredient in ingredients {
                 map.serialize_entry(
-                    &ingredient.debug_name(),
+                    &ingredient.ingredient_index().as_u32(),
                     &SerializeIngredient(ingredient, zalsa),
                 )?;
             }
@@ -332,21 +325,17 @@ mod persistence {
         {
             let DeserializeIngredients(zalsa) = self;
 
-            while let Some(name) = access.next_key::<&str>()? {
-                // TODO: Serialize the ingredient index directly.
-                let ingredient = zalsa
-                    .ingredients()
-                    .find(|ingredient| ingredient.debug_name() == name)
-                    .unwrap();
+            while let Some(index) = access.next_key::<u32>()? {
+                let index = IngredientIndex::new(index);
 
                 // Remove the ingredient temporarily, to avoid holding an overlapping mutable borrow
                 // to the ingredient as well as the database.
-                let mut ingredient = zalsa.take_ingredient(ingredient.ingredient_index());
+                let mut ingredient = zalsa.take_ingredient(index);
 
                 // Deserialize the ingredient.
                 access.next_value_seed(DeserializeIngredient(&mut *ingredient, zalsa))?;
 
-                zalsa.replace_ingredient(ingredient.ingredient_index(), ingredient);
+                zalsa.replace_ingredient(index, ingredient);
             }
 
             Ok(())
--- a/src/zalsa.rs
+++ b/src/zalsa.rs
@@ -208,7 +208,10 @@ impl Zalsa {
         let mut jars = jars;
 
         // Ensure structs are initialized before tracked functions.
-        jars.sort_by_key(|jar| jar.kind);
+        //
+        // We also further sort by debug name, to maintain a consistent ordering across
+        // builds.
+        jars.sort_by(|a, b| a.kind.cmp(&b.kind).then(a.type_name().cmp(b.type_name())));
 
         for jar in jars {
             zalsa.insert_jar(jar);
@@ -490,6 +493,7 @@ impl Zalsa {
 pub struct ErasedJar {
     kind: JarKind,
     type_id: fn() -> TypeId,
+    type_name: fn() -> &'static str,
     id_struct_type_id: fn() -> TypeId,
     create_ingredients: fn(&mut Zalsa, IngredientIndex) -> Vec<Box<dyn Ingredient>>,
 }
@@ -514,9 +518,14 @@ impl ErasedJar {
         Self {
             kind: I::KIND,
             type_id: TypeId::of::<I::Jar>,
+            type_name: std::any::type_name::<I::Jar>,
             create_ingredients: <I::Jar>::create_ingredients,
             id_struct_type_id: <I::Jar>::id_struct_type_id,
         }
+    }
+
+    pub fn type_name(&self) -> &'static str {
+        (self.type_name)()
     }
 }
 
--- a/tests/memory-usage.rs
+++ b/tests/memory-usage.rs
@@ -90,21 +90,21 @@ fn test() {
                 ),
             },
             IngredientInfo {
-                debug_name: "MyTracked",
-                count: 4,
-                size_of_metadata: 128,
-                size_of_fields: 96,
-                heap_size_of_fields: Some(
-                    300,
-                ),
-            },
-            IngredientInfo {
                 debug_name: "MyInterned",
                 count: 3,
                 size_of_metadata: 168,
                 size_of_fields: 72,
                 heap_size_of_fields: Some(
                     450,
+                ),
+            },
+            IngredientInfo {
+                debug_name: "MyTracked",
+                count: 4,
+                size_of_metadata: 128,
+                size_of_fields: 96,
+                heap_size_of_fields: Some(
+                    300,
                 ),
             },
             IngredientInfo {
--- a/tests/persistence.rs
+++ b/tests/persistence.rs
@@ -55,7 +55,7 @@ fn everything() {
             ]
           },
           "ingredients": {
-            "MyInput": {
+            "0": {
               "1": {
                 "durabilities": [
                   "Low"
@@ -104,7 +104,7 @@ fn everything() {
             ]
           },
           "ingredients": {
-            "MyInput": {
+            "0": {
               "1": {
                 "durabilities": [
                   "Low"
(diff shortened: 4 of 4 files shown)


components/salsa-macros/src/salsa_struct.rs
```diff
@@ -414,6 +414,36 @@ where
...
+        &self,
+    ) -> syn::Result<impl Iterator<Item = proc_macro2::TokenStream> + '_> {
+        let path = match &self.args.serde {
+            None => None,
+            Some(SerializeOptions::Custom {
+                serialize_fn: Some(serialize_fn),
+                deserialize_fn: Some(deserialize_fn),
+            }) => Some(quote! { (#serialize_fn, #deserialize_fn) }),
+            // Fallback to the `{Serialize, Deserialize}` implementations tuples if only one of the
+            // two is provided.
```
ibraheemdev: There is a little bit of weirdness here in that all of the following are accepted:
```rust
#[salsa::tracked(serialize)]
#[salsa::tracked(deserialize)]
#[salsa::tracked(serialize, deserialize)]
#[salsa::tracked(serialize = ...)]
#[salsa::tracked(deserialize = ...)]
#[salsa::tracked(serialize = ..., deserialize = ...)]
struct ...;
```
But I'm not sure disallowing any of those is better. With the derive-macro API, we should be able to get rid of this anyways and just use `serde::{Serialize, Deserialize}` directly.

src/database.rs
```diff
@@ -152,6 +154,230 @@ pub fn current_revision<Db: ?Sized + Database>(db: &Db) -> Revision {
...
+        }
+
+        fn visit_map<M>(self, mut access: M) -> Result<Self::Value, M::Error>
+        where
+            M: serde::de::MapAccess<'de>,
+        {
+            let DeserializeIngredients(zalsa) = self;
+
+            while let Some(name) = access.next_key::<&str>()? {
+                // TODO: Serialize the ingredient index directly.
```
ibraheemdev: Technically the debug name is not a unique identifier, but I found it more helpful for debugging than the ingredient index. It could also just be a field on each ingredient.

components/salsa-macro-rules/src/setup_input_struct.rs
```diff
@@ -53,6 +53,12 @@ macro_rules! setup_input_struct {
         // The function used to implement `C::heap_size`.
         heap_size_fn: $($heap_size_fn:path)?,
 
+        // If `true`, a `serde_fn` has been provided.
+        serializable: $serializable:tt,
```
MichaReiser: Should we call this persistable? `serializable` suggests that it only indicates whether a type as `serialize` but not necessarily `deserialize`.

src/cycle.rs
```diff
@@ -95,13 +95,15 @@ pub enum CycleRecoveryStrategy {
 /// would be the cycle head. It returns an "initial value" when the cycle is encountered (if
 /// fixpoint iteration is enabled for that query), and then is responsible for re-iterating the
 /// cycle until it converges.
-#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
```
MichaReiser: We should feature gate the `serde` dependency for now (either behind `serde` or a specific `persistable` feature that isn't enabled by default.

src/database.rs
```diff
@@ -152,6 +154,230 @@ pub fn current_revision<Db: ?Sized + Database>(db: &Db) -> Revision {
...
+            let SerializeIngredients(zalsa) = self;
+
+            let mut ingredients = zalsa
+                .ingredients()
+                .filter(|ingredient| ingredient.should_serialize(zalsa))
+                .collect::<Vec<_>>();
+
+            ingredients.sort_by(|a, b| {
+                // Ensure structs are serialized before tracked functions, as deserializing a
+                // memo requires its input struct to have been deserialized.
```
MichaReiser: It's fine deferring this to later but that will mean that we always serialize all tracked structs even if some query instances are never serialized (that created those tracked structs)

src/database.rs
```diff
@@ -152,6 +154,230 @@ pub fn current_revision<Db: ?Sized + Database>(db: &Db) -> Revision {
...
+        {
+            let mut result = None;
+            let mut serializer = Some(serializer);
+
+            // SAFETY: `<dyn Database>::as_serialize` take `&mut self`.
+            unsafe {
+                self.0.serialize(self.1, &mut |serialize| {
+                    let serializer = serializer.take().expect(
+                        "`Ingredient::serialize` must invoke the serialization callback only once",
+                    );
```
MichaReiser, line 239: This is a clever approach to avoid using thread locals.

src/database.rs
MichaReiser: A separate (maybe even optinal) field would make sense to me.

src/function.rs
```diff
@@ -385,6 +408,60 @@ where
...
+                .memo_ingredient_indices
+                .get(instance_key.ingredient_index());
+
+            let memo = self.get_memo_from_table_for(
+                zalsa,
+                instance_key.key_index(),
+                memo_ingredient_index,
+            );
+
+            if memo.is_some_and(|memo| memo.value.is_some()) {
```
MichaReiser: We should probably also skip over any memos that aren't finalized (it was built as part of some fix point query but isn't part of the final result). You can detect them with `maybe_provisional(...) == true` or `validate_may_be_provisional(...) == false` (with `allow_non_finalized` set to `false`). 

A second and related question is whether we should persist memos that haven't been verified in the most recent revision. This would allow us to prune old query results but it has the downside that we might end up pruning results only because they weren't queried in a while. I'm okay deferring this to later but it might be worth adding a TODO comment here.

src/function.rs
```diff
@@ -397,3 +474,152 @@ where
             .finish()
     }
 }
+
+#[cfg(not(feature = "shuttle"))]
```
MichaReiser: What's the reason that this is turned off for shuttle?

src/function/memo.rs
```diff
@@ -340,6 +340,94 @@ where
...
+mod persistence {
+    use crate::function::memo::Memo;
+    use crate::function::Configuration;
+    use crate::revision::AtomicRevision;
+    use crate::zalsa_local::QueryRevisions;
+
+    use serde::ser::SerializeStruct;
+    use serde::Deserialize;
+
+    impl<C> serde::Serialize for Memo<'_, C>
```
MichaReiser, line 360: Would it be possible to derive the serialize instead? I'm a bit worried that we'll forget to add fields here

src/input.rs
```diff
@@ -201,12 +231,22 @@ impl<C: Configuration> IngredientImpl<C> {
...
     }
 
-    #[cfg(feature = "salsa_unstable")]
     /// Returns all data corresponding to the input struct.
     pub fn entries<'db>(&'db self, zalsa: &'db Zalsa) -> impl Iterator<Item = &'db Value<C>> {
         zalsa.table().slots_of::<Value<C>>()
     }
 
+    /// Returns the IDs of all input structs of this type.
+    pub fn instances<'db>(
```
MichaReiser: I guess there's some overlap between `entries` and `instances` and the distinction between the two isn't very clear from their names. `entries` should probably be `values` and `instances` is closer to `keys()`? 

It would be nice if we could avoid having both but obviously, not a big deal

src/interned.rs
```diff
@@ -605,6 +623,40 @@ where
             }),
         });
 
+        // Insert the newly allocated ID.
+        self.insert_id(id, zalsa, shard, hash, value);
+
+        let index = self.database_key_index(id);
```
MichaReiser, line 627: It's not clear to me how this change is related to persistant caching. Should this be a separate PR/fix?

src/table.rs
```diff
@@ -202,6 +211,50 @@ impl Table {
...
+                page.assert_type::<T>();
+            }
+
+            None => {
+                // Create dummy pages until we reach the page we want.
+                while self.page_count() < page_idx.as_usize() {
+                    // We make sure not to claim any intermediary pages for ourselves, as they may
+                    // be required by a different ingredient when it is deserialized.
+                    self.push_uninit_page();
+                }
```
MichaReiser, line 252: How expensive is this in terms of memory usage to push those dummy pages? Do they allocate all entries (but empty)?

MichaReiser: Edit: I just saw below that we don't allocate any data.

tests/serialize.rs
```diff
@@ -0,0 +1,386 @@
...
+          }
+        }"#]];
+
+    expected.assert_eq(&serialized);
+
+    let mut db = common::EventLoggerDatabase::default();
+    <dyn salsa::Database>::deserialize(
+        &mut db,
+        &mut serde_json::Deserializer::from_str(&serialized),
+    )
```
MichaReiser: Does `deserialize` bump the revision to `2` to force a validation of all queries?

MichaReiser (approved):
This is great. I didn't read line by line and instead focused on the overall design. 

What's your plan with this PR? Do you want to land it behind a `persistence` feature or do you want to keep iterating in a separate branch?

src/function/memo.rs
ibraheemdev: I changed the implementation to force an exhaustive check of the `Memo` fields. It's a little difficult to derive here because the fields don't implement `Serialize` directly.

src/interned.rs
ibraheemdev: I just split out some of the functionality into the `insert_id` method so we can reuse it when deserializing, there should be no semantic change.

tests/serialize.rs
ibraheemdev: Hmm no, should it?

components/salsa-macros/src/salsa_struct.rs
ibraheemdev: I moved this into a `persist` attribute to make things more consistent:
```rust
#[salsa::tracked(persist)]
#[salsa::tracked(persist(serialize = ...))]
#[salsa::tracked(persist(deserialize = ...))]
#[salsa::tracked(persist(serialize = ..., deserialize = ...))]
struct ...;
```

We don't accept the `serialize` and `deserialize` arguments on tracked functions because both the input and output types need to implement `serde::{Serialize, Deserialize}`, and the arguments would be ambiguous.

ibraheemdev:
I gated everything under a `persistence` feature. We should rework CI to use something like [cargo-hack](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--feature-powerset) to test all the new feature permutations, but I'll do that in a separate PR.

src/database.rs
ibraheemdev: I removed the debug name for now, it's unfortunately very annoying to add extra fields because serde's macros aren't compatible with `DeserializeSeed` (which we need for some fields, because we deserialize into the table directly).

tests/serialize.rs
MichaReiser: I think there are a few things to consider here

* memos and other ingredients that weren't verified in the last revision (current when deserializing) still need to be dirty. The easiest here might be to only serialize ingredients that are verified in the current revision and we revisit this later
* we need to bump the revision if any input changes during deserialization. Now, we don't support this yet so it's fine to ignore for now.

ibraheemdev: > memos and other ingredients that weren't verified in the last revision (current when deserializing) still need to be dirty. The easiest here might be to only serialize ingredients that are verified in the current revision and we revisit this later

We serialize and deserialize the current revision of the database, so I don't think this is a problem. Any ingredients that weren't verified in the last revision will have to be verified after deserialization.

PR #969, opened 2025-08-09
refactor: Remove tracked structs from query outputs

MichaReiser:
We currently store tracked structs twice:

* In `QueryRevisions::origin` (as an output): This is mainly for historic reasons because tracked structs required to be verified (`mark_validated_output`), but that's no longer the case since we're using generational ids. 
* In `QueryRevisionExtras::tracked_struct_ids`: To get the tracked struct ID for a given tracked struct in a later revision (we want stable ids for better incrementality).

This PR removes tracked struct ids from `origin`. It requires tracking in the `IdentityMap` whether a tracked struct
was created in this revision (or if it's an entry from a previous revision but the tracked struct wasn't recreated)
so that we can remove now stale tracked structs in `diff_outputs`. 

This should help reduce memory usage and potentially even improve performance because 
it removes the `active_query.add_output` call for tracked structs and `deep_verify_memo` 
no longer has to iterate over tracked structs.


MichaReiser:
This looks promising :)

MichaReiser:
The results on [ty's benchmarks](https://codspeed.io/astral-sh/ruff/branches/micha%2Fsalsa-tracked-structs?runnerMode=Instrumentation) aren't as impressive but still a solid 1-3% performance win and a reduction in memory usage.

src/function/execute.rs
```diff
@@ -39,7 +40,7 @@ where
         });
         let memo_ingredient_index = self.memo_ingredient_index(zalsa, id);
 
-        let (new_value, mut revisions) = match C::CYCLE_STRATEGY {
+        let (new_value, mut completd_query) = match C::CYCLE_STRATEGY {
```
```suggestion
        let (new_value, mut completed_query) = match C::CYCLE_STRATEGY {
```

src/tracked_struct.rs
```diff
@@ -220,43 +221,65 @@ pub struct IdentityHash {
...
+
+                Some(std::mem::replace(&mut tracked.id, id))
+            }
         }
     }
 
-    pub(crate) fn get(&self, key: &Identity) -> Option<Id> {
+    /// Re-uses an existing identity if it already exists in this
+    /// [`IdentitiyMap`]. Returns the existing id or `None` if
+    /// no id for the given identity exists.
```
ibraheemdev:
```suggestion
    /// Reuses an existing identity if it already exists in this
    /// [`IdentityMap`]. Returns the existing ID, or `None` if
    /// no ID exists for the given identity.
```

src/tracked_struct.rs
```diff
@@ -267,9 +290,46 @@ impl IdentityMap {
         self.table.clear()
     }
 
-    pub(crate) fn into_thin_vec(self) -> ThinVec<(Identity, Id)> {
-        self.table.into_iter().collect()
+    pub(crate) fn did_create(&self, key: DatabaseKeyIndex) -> bool {
+        self.table.iter().any(|entry| {
+            entry.id == key.key_index()
+                && entry.identity.ingredient_index() == key.ingredient_index()
+        })
```
ibraheemdev: Why don't we have to check `entry.create` here?

src/function/diff_outputs.rs
```diff
@@ -20,12 +20,24 @@ where
...
+                .then(a.key_index().cmp(&b.key_index()))
+        });
+
+        for removed_tracked in removed_tracked_structs {
+            Self::report_stale_output(zalsa, key, removed_tracked);
+        }
+
         // Iterate over the outputs of the `old_memo` and put them into a hashset
         //
         // Ignore key_generation here, because we use the same tracked struct allocation for
```
ibraheemdev: I don't think this is relevant anymore

src/tracked_struct.rs
MichaReiser: Oh, we should. Nice catch

davidbarsky:
The memory usage, for posterity: https://github.com/astral-sh/ruff/pull/19843#issuecomment-3171905321

MichaReiser:
Uhhm, why?

MichaReiser:
Hmm, I think this introduced a subtle regression regarding fixpoint iteration. 

**Before**: Salsa seeded the `tracked_struct_ids` in every iteration with the values from the previous revision and the values from the previous iteration. 

**Now**: Salsa seeds the `tracked_struct_ids` initially with the values from the previous iteration. Later iterations seed `tracked_struct_ids` with the values from the previous iteration but any tracked struct id that wasn't created in the first iteration isn't copied over anymore because the new `QueryRevisions` removes unused tracked struct ids eagerly and not only when `diff_outputs` is called (which we only ran between the memo from the last revision with the memo from the new revision). 

Now, this is sort of a pre-existing issue for queries that aren't cycle heads because they already had the behavior that we now observe for cycle heads too.

PR #975, opened 2025-08-12
Flatten unserializable query dependencies

ibraheemdev:
Follow up to https://github.com/salsa-rs/salsa/pull/967. This PR removes the requirement that all transitive query dependencies are persistable by flattening them down to their base inputs.


src/function.rs
```diff
@@ -288,6 +288,28 @@ where
         self.maybe_changed_after(db, input, revision, cycle_heads)
     }
 
+    fn minimum_serialized_edges(&self, zalsa: &Zalsa, edge: QueryEdge) -> Vec<QueryEdge> {
```
MichaReiser: Should we change this to a collector pattern to avoid allocating that many `Vec`s?
```suggestion
    fn minimum_serialized_edges(&self, zalsa: &Zalsa, edge: QueryEdge, output: &mut Vec<QueryEdge>) {
```

src/function/memo.rs
```diff
@@ -359,12 +359,36 @@ mod persistence {
...
-    use crate::zalsa_local::QueryRevisions;
+    use crate::zalsa_local::persistence::MappedQueryRevisions;
+    use crate::zalsa_local::{QueryOrigin, QueryRevisions};
 
     use serde::ser::SerializeStruct;
     use serde::Deserialize;
 
-    impl<C> serde::Serialize for Memo<'_, C>
+    /// A reference to the fields of a [`Memo`], with its [`QueryRevisions`] transformed.
+    pub(crate) struct MappedMemo<'memo, 'db, C: Configuration> {
```
MichaReiser, line 369: That could also allow us to replace `CompletedQuery` if we wanted to

src/interned.rs
```diff
@@ -896,6 +897,20 @@ where
         VerifyResult::unchanged()
     }
 
+    fn minimum_serialized_edges(&self, _zalsa: &Zalsa, edge: QueryEdge) -> Vec<QueryEdge> {
+        if C::PERSIST {
```
MichaReiser, line 907: Interesting: I initially assumed that an implementation flattening the dependency at runtime would use the same logic as what we need for persistence, but I can see now that separate methods make sense.

src/tracked_struct.rs
```diff
@@ -937,8 +938,13 @@ where
...
     ) -> VerifyResult {
-        // Any change to a tracked struct results in a new ID generation.
-        VerifyResult::unchanged()
+        // Any change to a tracked struct results in a new ID generation, so there
+        // are no direct dependencies on the struct, only on its tracked fields.
+        panic!("nothing should ever depend on a tracked struct directly")
+    }
+
+    fn minimum_serialized_edges(&self, _zalsa: &Zalsa, _edge: QueryEdge) -> Vec<QueryEdge> {
+        panic!("nothing should ever depend on a tracked struct directly")
```
MichaReiser, line 960: I agree, it's true that `TrackedStruct`s are never inputs, thanks to generational IDs. 

`TrackedStruct`s are still considered outputs; it's just that they're no longer tracked on the query's origin. I assume we don't need to serialize them here because we serialize the tracked struct map separately and we only do so if the creating query gets serialized too. 


I think this does raise an interesting question for ty because most tracked structs are created as part of the `semantic_index` query which we don't intend on caching. However, those tracked structs are heavily used inside `Type`s and a changing ID can lead to invalidating large chunks of the type checking queries. I think the reason why this shouldn't be as much of a problem for ty is because all tracked structs are guaranteed to have unique fields, so that we should rarely rely on their creation order.

src/function.rs
```diff
@@ -288,6 +288,28 @@ where
...
+        let Some(memo) =
+            self.get_memo_from_table_for(zalsa, input, self.memo_ingredient_index(zalsa, input))
+        else {
+            return Vec::new();
+        };
+
+        let origin = memo.revisions.origin.as_ref();
+
+        // Collect the minimum dependency tree.
+        origin
```
MichaReiser: Shouldn't this include `edge` if the ingredient is serializable?

MichaReiser (approved):
nice

src/tracked_struct.rs
ibraheemdev: I added a comment that addresses this, it's possible that we could flatten the identity map if this was an issue.

src/function.rs
ibraheemdev: This method is only called on the outer query if `edge` is not persistable. For any transitive queries, the logic for whether or not to include transitive dependency edges is contained within `collect_minimum_serialized_edges` itself, e.g. the implementation for interned structs checks whether or not it will be persisted. We specifically don't want to include transitive queries directly, though, unless they are persistable.

MichaReiser: I see. Should this function return a set? I assume that many queries will have overlapping dependencies or is this done a level higher up?

ibraheemdev: That was being done a level up, but now it collects directly into the set.

PR #979, opened 2025-08-14
fix: Delete not re-created tracked structs after fixpoint iteration

MichaReiser:
This PR fixes a regression that I introduced with https://github.com/salsa-rs/salsa/pull/969

https://github.com/salsa-rs/salsa/pull/969 moves the tracked structs out from `QueryRevisions::origin` and instead tracks the "alive" tracked ids directly in `IdentityMap`. Completing the `IdentityMap` then returns to vecs: 

1. The alive tracked struct ids and their identity
2. The not re-created tracked struct ids

Unlike before https://github.com/salsa-rs/salsa/pull/969, the not re-created tracked struct ids are identified eagerly when completing the query. This isn't an issue for non-cyclic queries but it resulted in leaking if a cyclic query didn't create all tracked structs in the first iteration of the second revision. 

This is fixed in this PR. I added a regression test and verified that the logs match the behavior before we merged https://github.com/salsa-rs/salsa/pull/969.

### DIFF
--- a/examples/calc/db.rs
+++ b/examples/calc/db.rs
@@ -48,6 +48,7 @@ impl CalcDatabaseImpl {
     }
 
     #[cfg(test)]
+    #[allow(unused)]
     pub fn take_logs(&self) -> Vec<String> {
         let mut logs = self.logs.lock().unwrap();
         if let Some(logs) = &mut *logs {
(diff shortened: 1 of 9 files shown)


MichaReiser:
@carljm Could you take a look at this?

PR #981, opened 2025-08-18
fix: Runaway for unchanged queries participating in cycle

MichaReiser:
This PR fixes a run-away (or even infinite loop) in `maybe_changed_after` for queries participating in a cycle. 

For normal queries, `maybe_changed_after` is guaranteed to be only called exactly once (if they have an old value) because it always updates verified at:

* After a shallow verification
* When `deep_verify_memo` returns unchanged
* After executing the query (which replaces the current memo with a new value)

Unfortunately, this isn't the case for queries participating in a cycle (that aren't a cycle head) because

* `deep_verify_memo` returns `unchanged` without bumping the `verified_at` because the result depends on the outcome of the cycle head's verification
* `deep_verify_memo` returns `changed` but we skip executing the query because the cycle head hasn't been fully verified at this point and this query might depend on an interned or tracked struct that only gets created later during the cycle execution.

This PR fixes the run-away by caching the `maybe_changed_after` results for queries participating in a cycle to avoid re-validating the same subtrees over and over again.

However, I then noticed that we keep re-executing some queries because `validate_same_iteration` is disabled for `maybe_changed_after`. This PR allows `validate_same_iteration` but I made it more strict to only return true if the memo and its cycle heads are from the same revision (verified at). The same is true for `validate_provisional`. 

this resolves https://github.com/astral-sh/ty/issues/1024


tests/cycle.rs
```diff
@@ -1137,3 +1137,245 @@ fn repeat_provisional_query_incremental() {
...
+    assert_eq!(head(&db, input), 2);
+
+    // `DidValidateInternedValue { key: Interned(Id(400)), revision: R2 }` should only be logged
+    // once per `maybe_changed_after` root-call (e.g. validating `head` shouldn't validate `query_hot` multiple times).
+    //
+    // This is important to avoid a run-away situation where a query is called many times within a cycle and
+    // Salsa would end up recusively validating the hot query over and over again.
+    db.assert_logs(expect![[r#"
+        [
+            "salsa_event(DidValidateInternedValue { key: Interned(Id(400)), revision: R2 })",
```
MichaReiser, line 1356: Before, there were 3 or even 4 `DidValidateInternedValue` calls because each `query_x` calls `maybe_changed_after(query_hot)` which in turn called `deep_verify_memo(query_hot)`


As you can see, there are still multiple `deep_verify_memo(query_hot)` calls when re-executing `head` because salsa tries to narrow down which child query needs to re-run.

tests/cycle.rs
```diff
@@ -1137,3 +1137,245 @@ fn repeat_provisional_query_incremental() {
...
+            "salsa_event(WillExecute { database_key: head(Id(0)) })",
+            "salsa_event(DidValidateInternedValue { key: Interned(Id(400)), revision: R2 })",
+            "salsa_event(WillExecute { database_key: query_a(Id(0)) })",
+            "salsa_event(DidValidateInternedValue { key: Interned(Id(400)), revision: R2 })",
+            "salsa_event(WillExecute { database_key: query_hot(Id(0)) })",
+            "salsa_event(WillExecute { database_key: query_hot(Id(0)) })",
```
MichaReiser: Uhm, why are we executing this query multiple times?

src/function/maybe_changed_after.rs
```diff
@@ -424,62 +464,74 @@ where
...
-                        .or_else(|| {
-                            // If the cycle head isn't on our stack because:
-                            //
-                            // * another thread holds the lock on the cycle head (but it waits for the current query to complete)
-                            // * we're in `maybe_changed_after` because `maybe_changed_after` doesn't modify the cycle stack
-                            //
-                            // check if the latest memo has the same iteration count.
+                    // If the memo was verified in the current revision, check if it has the same iteration count as any query on the stack.
+                    // We skip this check for memo's from past revisions because we want to re-execute them if the cycle heads re-execute.
+                    if current_revision == verified_at {
```
MichaReiser, line 476: The main change here is that we skip the stack if `verified_at` isn't the current revision because we want all cycle participating queries to re-execute if they're from a past revision and their cycle head is re-executing

tests/cycle.rs
MichaReiser: This is now fixed

carljm (approved):
Awesome!

PR #982, opened 2025-08-18
Persistent caching fixes

ibraheemdev:
Various persistent caching bug fixes/improvements. See the individual commits for details.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -905,16 +905,16 @@ where
         serialized_edges: &mut FxIndexSet<QueryEdge>,
         _visited_edges: &mut FxHashSet<QueryEdge>,
     ) {
-        if C::PERSIST {
+        if C::PERSIST && C::REVISIONS != IMMORTAL {
             // If the interned struct is being persisted, it may be reachable through transitive queries.
             // Additionally, interned struct dependencies are impure in that garbage collection can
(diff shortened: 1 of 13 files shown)


src/interned.rs
```diff
@@ -905,11 +905,12 @@ where
...
         _visited_edges: &mut FxHashSet<QueryEdge>,
     ) {
-        if C::PERSIST {
+        if C::PERSIST && C::REVISIONS != IMMORTAL {
             // If the interned struct is being persisted, it may be reachable through transitive queries.
             // Additionally, interned struct dependencies are impure in that garbage collection can
             // invalidate a dependency without a base input necessarily being updated. Thus, we must
-            // preserve the transitive dependency on the interned struct.
+            // preserve the transitive dependency on the interned struct, if garbage collection is
+            // enabled.
```
MichaReiser, line 913: Can you extend the comment why serializing the edge isn't necessary if garbage collection is disabled. Is it because we serialize all interned values of that ingredient?

ibraheemdev: If garbage collection is enabled, the base inputs don't cover the dependency because the interned struct can be invalidated without an input changing. If garbage collection is disabled, this is no longer true, so the input edges are sufficient, which I think the comment explains.

PR #987, opened 2025-08-22
Refactor `entries` API

ibraheemdev:
This makes it easier to collect the actual struct entries from the database without having to go through `FromId`, which is necessary to access structs after deserialization.


src/interned.rs
```diff
@@ -840,11 +837,43 @@ where
...
+impl<'db, C> ValueEntry<'db, C>
+where
+    C: Configuration,
+{
+    /// Returns the `DatabaseKeyIndex` for this entry.
+    pub fn key(&self) -> DatabaseKeyIndex {
+        self.key
+    }
+
+    /// Returns the tracked struct.
```
MichaReiser: Interned struct

src/input.rs
```diff
@@ -257,6 +257,35 @@ impl<C: Configuration> IngredientImpl<C> {
...
+impl<'db, C> ValueEntry<'db, C>
+where
+    C: Configuration,
+{
+    /// Returns the `DatabaseKeyIndex` for this entry.
+    pub fn key(&self) -> DatabaseKeyIndex {
+        self.key
+    }
+
+    /// Returns the tracked struct.
```
MichaReiser: Input struct

src/input.rs
```diff
@@ -257,6 +257,35 @@ impl<C: Configuration> IngredientImpl<C> {
     }
 }
 
+/// An input struct entry.
+pub struct ValueEntry<'db, C>
```
MichaReiser: Nit: maybe StructEntry bit I'm fine with any name

PR #993, opened 2025-09-24
fix: Cleanup provisional cycle head memos when query panics

MichaReiser:
We've seen in https://github.com/astral-sh/ruff/pull/20477 that Salsa can hang (it gets stuck in a `fetch_with_retry` and `Memo::retry_provisional` cycle) if a cycle head query panics. 

I invested some time to see if there's an easy way to identify a Memo of a panicked query, but without success. That's why I decided to explore if we can clean up any provisional memo if a cycle head panics, and that turns out to be much easier.

```rust
/// Replaces any inserted memo with a fixpoint initial memo without a value if the current thread panics.
///
/// A regular query doesn't insert any memo if it panics and the query
/// simply gets re-executed if any later called query depends on the panicked query (and will panic again unless the query isn't deterministic).
///
/// Unfortunately, this isn't the case for cycle heads because Salsa first inserts the fixpoint initial memo and later inserts
/// provisional memos for every iteration. Detecting whether a query has previously panicked
/// in `fetch` (e.g., `validate_same_iteration`) and requires re-execution is probably possible but not very straightforward
/// and it's easy to get it wrong, which results in infinite loops where `Memo::provisional_retry` keeps retrying to get the latest `Memo`
/// but `fetch` doesn't re-execute the query for reasons.
///
/// Specifically, a Memo can linger after a panic, which is then incorrectly returned
/// by `fetch_cold_cycle` because it passes the `shallow_verified_memo` check instead of inserting
/// a new fix point initial value if that happens.
///
/// We could insert a fixpoint initial value here, but it seems unnecessary.
```

I also used this PR as an opportunity to add some more logging which should help debug similar issues more easily in the future.

### DIFF
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -299,22 +299,22 @@ where
     }
 }
 
-/// Replaces any inserted memo with a fixpoint initial memo without a value if the current thread panicks.
+/// Replaces any inserted memo with a fixpoint initial memo without a value if the current thread panics.
 ///
-/// A regular query doesn't insert any memo if it panicks and the query
+/// A regular query doesn't insert any memo if it panics and the query
 /// simply gets re-executed if any later called query depends on the panicked query (and will panic again unless the query isn't deterministic).
 ///
 /// Unfortunately, this isn't the case for cycle heads because Salsa first inserts the fixpoint initial memo and later inserts
 /// provisional memos for every iteration. Detecting whether a query has previously panicked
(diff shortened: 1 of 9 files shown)


carljm:
Any idea why this is improving several benchmarks? Seems surprising. Is that just noisy benchmarks? That also seems surprising that it would be consistent improvement across four different ones.

MichaReiser:
> Any idea why this is improving several benchmarks? Seems surprising. Is that just noisy benchmarks? That also seems surprising that it would be consistent improvement across four different ones.

Nope. Maybe the first change that builds with Rust 1.90?

PR #999, opened 2025-10-06
Run fixpoint per strongly connected component

MichaReiser:
This PR rewrites how Salsa runs fixpoint iterations when multiple cylce heads participate in a cycle (when there are multiple nested cycles). 

## Motivation

The main motivation for this is that we see a runaway in ty for code like this:

```py
self.x1 = 0
self.x2 = 0

self.y1 = self.y2 + self.y3
self.y2 = self.y1
self.y3 = self.y1 + self.y2

self.z1 = self.y1 + self.z2
self.z2 = self.z1
```

Or as a graph

```mermaid
flowchart TD
    x1
    x2
    
    y1 --> y2
	y1 --> y3
    y2 --> y1
	y3 --> y1
	y3 --> y2

    z1 --> y1 
    z1 --> z2
    z2 --> z1
```

Today, we'll run separate fixpoint iterations for `y1 <-> y2`, `y3 <-> y1`, and `z1 <-> z2`. 

The downside of this is that Salsa runs the nested fixpoint iterations `y1<->y3` to convergence for every `y1 <-> y2` iteration. If there are more deeply nested cycles, then each of those inner cycles is run to completion for every `y1 <-> y2` and `y1 <-> y3` iteration, making the number of iterations dependent on the nesting of queries rather than how quickly the problem converges.

To avoid this, the idea is to identify the strongly connected components in the dependency graph. These are:

* `x1` (no dependencies)
* `x2` (no dependencies)
* `y1`, `y2`, `y3`
* `z1`, `z2` (only depends on `y1` in one direction but not both)

Then, run a fixpoint per strongly connected component. That means, we only run 2 fixpoint iterations for the example above. This ensures that the runtime is bound by how quickly the problem converges and not by how deeply nested our queries are (which creates a tension between granular queries and performance).

## Single-threaded implementation

Let's ignore multi-threading first as this only complicates things ;) 

The change in itself isn't that complicated:

* `execute_maybe_iterate`: If the query is part of an outer cycle, then don't remove the cycle head from `cycle_heads`, set a flag on the memo indicating whether this head has converged, and return the provisional value. The query will iterate as part of the outer cycle
* The outer cycle is the query that, when claiming it, results in a cycle (this tells us it's higher up on the stack)
* If the query is an outer cycle, then:
  * Test if itself and all its inner cycle heads (can be retrieved from `memo.cycle_heads`) have converged
  * If not, update the iteration count on each cycle head and iterate again
  * Otherwise, mark all the cycle heads as finalized. 

That's roughly it :)

## Multithreading

Where things get complicated is multithreading. I tried to [keep the existing locking](https://github.com/salsa-rs/salsa/pull/995) (where we wait on cycle heads in some places), but I wasn't able to make it work (there were spurious errors till the end). That's why I ended up rewriting cycle head synchronization from scratch. 

What makes cycles tricky is that multiple threads can enter the same cycle simultaneously, each starting from a different cycle head that ultimately resolves to the same outer cycle. The new implementation makes this even trickier because inner cycles now complete immediately without iterating, which results in them releasing their locks. That allows other threads to claim those locks again, which can result in them competing for the same locks forever. 

This PR extends Salsa's `DependencyGraph` with a mechanism to transfer the ownership of a query to another query. Let's say we have `a -> b -> a` and `b -> c -> b`. The inner cycle head can transfer its lock to the query `a`. `b` then remains locked until `c` completes (successfully or panicking). Only the thread owning `a` (because it's the thread holding that lock or it's another thread on which `a` is blocked on) is allowed to reclaim `b`. 

This ensures that the locks of all cycle heads participating in a query ultimately converge to be owned by a single query/thread. We need to do this for every query that supports cycle handling, even if it isn't a cycle head, because the query *could become* a cycle head in later iterations (e.g., due to changes in the cycle-entry point). 

I also found this approach much easier to reason about than the old approach of "just wait for all cycle heads". 

However, there's a downside. It requires acquiring the global `DependencyGraph` lock and tracking dependencies whenever claiming or releasing any query with cycle handle support that participates in a cycle. We incur this cost even when all queries participating in the cycle run on the same thread; this is also the main reason why `converge-diverge`'s performance regresses. I'm not sure if there's a solution to avoid this.

TODO: 

* [x] Documentation
* [x] Investigate perf regressions
* [x] Review if there are any changes necessary to `maybe_changed_after`
* [x] Investigate why `cycle_nested_deep_conditional_changed::the_test` sometimes hangs or panics
* [x] Add a test for a panic in a nested cycle

### DIFF
--- a/tests/parallel/cycle_nested_deep_panic.rs
+++ b/tests/parallel/cycle_nested_deep_panic.rs
@@ -107,7 +107,9 @@ fn run() {
 
 #[test_log::test]
 fn the_test() {
-    for _ in 0..200 {
+    let count = if cfg!(miri) { 1 } else { 200 };
+
+    for _ in 0..count {
         run()
     }
 }


src/function/maybe_changed_after.rs
```diff
@@ -141,7 +144,7 @@ where
     ) -> Option<VerifyResult> {
         let database_key_index = self.database_key_index(key_index);
 
-        let _claim_guard = match self.sync_table.try_claim(zalsa, key_index) {
+        let _claim_guard = match self.sync_table.try_claim(zalsa, key_index, false) {
```
MichaReiser: Should this be `true` or `false`???

src/function/maybe_changed_after.rs
```diff
@@ -553,6 +535,12 @@ where
         cycle_heads: &mut VerifyCycleHeads,
         can_shallow_update: ShallowUpdate,
     ) -> VerifyResult {
+        // If the value is from the same revision but is still provisional, consider it changed
+        // because we're now in a new iteration.
```
MichaReiser, line 546: I always found the `deep_verify_memo` logs for provisional values confusing. Exiting early here ensures they don't show up

src/function/execute.rs
```diff
@@ -133,25 +141,52 @@ where
...
-        let mut iteration_count = IterationCount::initial();
-        let mut active_query = zalsa_local.push_query(database_key_index, iteration_count);
 
         // Our provisional value from the previous iteration, when doing fixpoint iteration.
-        // Initially it's set to None, because the initial provisional value is created lazily,
-        // only when a cycle is actually encountered.
-        let mut opt_last_provisional: Option<&Memo<'db, C>> = None;
+        // This is different from `opt_old_memo` which might be from a different revision.
+        let mut last_provisional_memo: Option<&Memo<'db, C>> = None;
+        // TODO: Can we seed those somehow?
```
MichaReiser, line 159: This is https://github.com/salsa-rs/salsa/pull/980

src/function/fetch.rs
```diff
@@ -132,21 +141,21 @@ where
...
                 blocked_on.block_on(zalsa);
 
-                let memo = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
+                if C::CYCLE_STRATEGY == CycleRecoveryStrategy::FallbackImmediate {
+                    let memo = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
 
-                if let Some(memo) = memo {
-                    // This isn't strictly necessary, but if this is a provisional memo for an inner cycle,
-                    // await all outer cycle heads to give the thread driving it a chance to complete
-                    // (we don't want multiple threads competing for the queries participating in the same cycle).
```
MichaReiser, line 145: This is required for fallback immediate in multi threaded context. I didn't spend any time understanding why

src/function/execute.rs
```diff
@@ -133,25 +141,52 @@ where
...
+                let memo_iteration_count = old_memo.revisions.iteration();
+
+                // The `DependencyGraph` locking propagates panics when another thread is blocked on a panicking query.
+                // However, the locking doesn't handle the case where a thread fetches the result of a panicking cycle head query  **after** all locks were released.
+                // That's what we do here. We could consider re-executing the entire cycle but:
+                // a) It's tricky to ensure that all queries participating in the cycle will re-execute
+                //    (we can't rely on `iteration_count` being updated for nested cycles because the nested cycles may have completed successfully).
+                // b) It's guaranteed that this query will panic again anyway.
+                // That's why we simply propagate the panic here. It simplifies our lives and it also avoids duplicate panic messages.
+                if old_memo.value.is_none() {
```
carljm, line 177: If `value.is_none()` always means "panicked" (which is what it looks like is implied here), would it make sense to reflect that more clearly in method naming, e.g. add a `memo.panicked()` or similar?

src/function/execute.rs
```diff
@@ -160,118 +195,267 @@ where
...
+            // t1: a -> b
+            // t2: c -> b (blocks on t1)
+            // t1: a -> b -> c (cycle, returns fixpoint initial with c(0) in heads)
+            // t1: a -> b (completes b, b has c(0) in its cycle heads, releases `b`, which resumes `t2`, and `retry_provisional` blocks on `c` (t2))
+            // t2: c -> a (cycle, returns fixpoint initial for a with a(0) in heads)
+            // t2: completes c, `provisional_retry` blocks on `a` (t2)
+            // t1: a (completes `b` with `c` in heads)
+            //
+            // Note how `a` only depends on `c` but not `a`. This is because `a` only saw the initial value of `c` and wasn't updated when `c` completed.
+            // That's why we need to resolve the cycle heads recursively to `cycle_heads` contains all cycle heads at the moment this query completed.
```
carljm:
```suggestion
            // That's why we need to resolve the cycle heads recursively so `cycle_heads` contains all cycle heads at the moment this query completed.
```

src/cycle.rs
```diff
@@ -188,15 +323,42 @@ impl CycleHeads {
         self.0.reserve(other.0.len());
 
         for head in other {
-            if let Some(existing) = self
-                .0
-                .iter()
-                .find(|candidate| candidate.database_key_index == head.database_key_index)
-            {
-                assert_eq!(existing.iteration_count, head.iteration_count);
```
ibraheemdev, line 196: Is there a reason we no longer need this assertion?

src/function/execute.rs
```diff
@@ -133,25 +141,52 @@ where
...
+
+                // The `DependencyGraph` locking propagates panics when another thread is blocked on a panicking query.
+                // However, the locking doesn't handle the case where a thread fetches the result of a panicking cycle head query  **after** all locks were released.
+                // That's what we do here. We could consider re-executing the entire cycle but:
+                // a) It's tricky to ensure that all queries participating in the cycle will re-execute
+                //    (we can't rely on `iteration_count` being updated for nested cycles because the nested cycles may have completed successfully).
+                // b) It's guaranteed that this query will panic again anyway.
+                // That's why we simply propagate the panic here. It simplifies our lives and it also avoids duplicate panic messages.
+                if old_memo.value.is_none() {
+                    ::tracing::warn!("Propagating panic for cycle head that panicked in an earlier execution in that revision");
```
ibraheemdev: Is there a reason we're using the `tracing` crate directly instead of our wrapper (and in other places as well)? Is it because this is already a cold path?

src/cycle.rs
```diff
@@ -131,45 +168,126 @@ impl IterationCount {
...
-        let Some(found) = found else { return false };
-        self.0.swap_remove(found);
-        true
+    /// Removes all cycle heads except `except` by marking them as removed.
+    ///
+    /// Note that the heads aren't actually removed. They're only marked as removed and will be
+    /// skipped when iterating. This is because we might not have a mutable reference.
+    pub(crate) fn remove_all_except(&self, except: DatabaseKeyIndex) {
+        for head in self.0.iter() {
+            if head.database_key_index == except {
```
ibraheemdev, line 278: Can we use `iter_not_eq` here?

src/function/sync.rs
```diff
@@ -20,17 +24,36 @@ pub(crate) enum ClaimResult<'a> {
...
+        /// [`SyncTable::try_claim`] with [`Reentrant::Allow`].
+        inner: bool,
+    },
     /// Successfully claimed the query.
     Claimed(ClaimGuard<'a>),
 }
 
 pub(crate) struct SyncState {
-    id: ThreadId,
+    /// The thread id that is owning this query (actively executing it or iterating it as part of a larger cycle).
```
ibraheemdev:
```suggestion
    /// The thread id that currently owns this query (actively executing it or iterating it as part of a larger cycle).
```

src/function/sync.rs
```diff
@@ -20,17 +24,36 @@ pub(crate) enum ClaimResult<'a> {
...
+    /// Whether any other query has transferred its lock ownership to this query.
+    /// This is only an optimization so that the expensive unblocking of transferred queries
+    /// can be skipped if `false`. This field might be `true` in cases where queries *were* transferred
+    /// to this query, but have since then been transferred to another query (in a later iteration).
+    is_transfer_target: bool,
+
+    /// Whether this query has been claimed by the query that currently owns it.
+    ///
+    /// If `a` has been transferred to `b` and the stack for t1 is `b -> a`, then `a` can be claimed
+    /// and `claimed_transferred` is set to `true`. However, t2 won't be able to claim `a` because
```
ibraheemdev: Is `claimed_transferred` a typo here?

src/function/sync.rs
```diff
@@ -62,22 +105,112 @@ impl SyncTable {
...
+            state.anyone_waiting = true;
+            state.is_transfer_target = true;
+
+            state.id
+        })
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum SyncOwnerId {
```
ibraheemdev: Nit: Maybe just call this `SyncOwner`, since it doesn't necessarily hold an ID.

src/function/sync.rs
```diff
@@ -122,3 +369,60 @@ impl std::fmt::Debug for SyncTable {
...
+    }
+}
+
+/// Controls whether this thread can claim a query that transferred its ownership to a query
+/// this thread currently holds the lock for.
+///
+/// For example: if query `a` transferred its ownership to query `b`, and this thread holds
+/// the lock for `b`, then this thread can also claim `a` — but only when using [`Self::Allow`].
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub(crate) enum Reentrant {
```
ibraheemdev: Nit: This should probably be named `Reentrancy`.

src/runtime/dependency_graph.rs
```diff
@@ -138,6 +145,381 @@ impl DependencyGraph {
         // notify the thread.
         edge.notify();
     }
+
+    /// Invoked when the query `database_key` completes and it owns the locks of other queries
+    /// (the queries transferred their locks to `database_key`).
+    pub(super) fn unblock_runtimes_blocked_on_transferred_queries_owned_by(
```
ibraheemdev, line 151: This is quite the name :sweat_smile:

src/runtime/dependency_graph.rs
```diff
@@ -138,6 +145,381 @@ impl DependencyGraph {
...
+                    .get_mut(&old_owner)
+                    .unwrap()
+                    .remove(&query);
+
+                entry.insert((new_owner_thread, new_owner));
+
+                // If we have `c -> a -> d` and we now insert a mapping `d -> c`, rewrite the mapping to
+                // `d -> c -> a` to avoid cycles.
+                //
+                // Or, starting with `e -> c -> a -> d -> b` insert `d -> c`. We need to respine the tree to
```
ibraheemdev: 'respine' looks like a typo here.

src/runtime/dependency_graph.rs
```diff
@@ -138,6 +145,381 @@ impl DependencyGraph {
...
+        /// that the thread is blocked on (key into `query_dependents`) and the second element
+        /// is the index in the list of blocked threads (index into the `query_dependents` value) for that query.
+        fn find_blocked_thread(
+            me: &DependencyGraph,
+            query: DatabaseKeyIndex,
+            new_owner_id: ThreadId,
+        ) -> Option<(DatabaseKeyIndex, usize)> {
+            if let Some(blocked_threads) = me.query_dependents.get(&query) {
+                for (i, id) in blocked_threads.iter().copied().enumerate() {
+                    if id == new_owner_id || me.edges.depends_on(new_owner_id, id) {
```
ibraheemdev, line 350: I assume this is non-deterministic if there are multiple threads blocked on `new_owner_id`?

src/runtime/dependency_graph.rs
```diff
@@ -165,7 +547,6 @@ mod edge {
 
         /// Signalled whenever a query with dependents completes.
         /// Allows those dependents to check if they are ready to unblock.
-        // condvar: unsafe<'stack_frame> Pin<&'stack_frame Condvar>,
```
ibraheemdev, line 168: Looks like this was mistakenly removed.

src/function/sync.rs
```diff
@@ -62,22 +105,112 @@ impl SyncTable {
...
+    /// Marks `key_index` as a transfer target.
+    ///
+    /// Returns the `SyncOwnerId` of the thread that currently owns this query.
+    ///
+    /// Note: The result of this method will immediately become stale unless the thread owning `key_index`
+    /// is currently blocked on this thread (claiming `key_index` from this thread results in a cycle).
+    pub(super) fn mark_as_transfer_target(&self, key_index: Id) -> Option<SyncOwnerId> {
+        let mut syncs = self.syncs.lock();
+        syncs.get_mut(&key_index).map(|state| {
+            state.anyone_waiting = true;
```
ibraheemdev, line 195: Why do we unconditionally set `anyone_waiting` here?

src/function/sync.rs
ibraheemdev:
```suggestion
    /// and `claimed_twice` is set to `true`. However, t2 won't be able to claim `a` because
```

src/function/execute.rs
```diff
@@ -360,3 +544,44 @@ impl<C: Configuration> Drop for ClearCycleHeadIfPanicking<'_, C> {
...
+    // Using the outer most over the inner most should reduce the need
+    // for transitive transfers.
+    // SAFETY: We don't call into with_query_stack recursively
+    if let Some(same_thread) = unsafe {
+        zalsa_local.with_query_stack_unchecked(|stack| {
+            stack
+                .iter()
+                .find(|active_query| {
+                    cycle_heads.contains(&active_query.database_key_index)
+                        && active_query.database_key_index != current_key
```
ibraheemdev, line 568: Is this enough to identify strongly connected components? Do we never need to prefer the outermost query for deeply nested cycles?

src/function/execute.rs
```diff
@@ -133,25 +141,52 @@ where
...
+        let mut iteration_count = IterationCount::initial();
+
+        if let Some(old_memo) = opt_old_memo {
+            if old_memo.verified_at.load() == zalsa.current_revision()
+                && old_memo.cycle_heads().contains(&database_key_index)
+            {
+                let memo_iteration_count = old_memo.revisions.iteration();
+
+                // The `DependencyGraph` locking propagates panics when another thread is blocked on a panicking query.
+                // However, the locking doesn't handle the case where a thread fetches the result of a panicking cycle head query  **after** all locks were released.
```
```suggestion
                // However, the locking doesn't handle the case where a thread fetches the result of a panicking cycle head query **after** all locks were released.
```

ibraheemdev (approved):
I didn't review everything too closely, but I left a couple nits and questions.

src/function/execute.rs
MichaReiser: I tried to add `warn` to our own tracing but Rust was mad at me for `warn` being a reserved keyword or something. I then decided not to bother because we only use `warn` in very cold paths.

src/runtime/dependency_graph.rs
MichaReiser: It wasn't clear to me what it documents? Do you think this should be part of the field comment?

src/runtime/dependency_graph.rs
MichaReiser: I don't think so. There should always only be one query that currently blocks `new_owner_id` (it can't be blocked on multiple threads). (or it wouldn't be a tree)

src/cycle.rs
MichaReiser: I moved the assertion into `insert` (line 351)

src/cycle.rs
MichaReiser: We could, but it might be slightly more expensive because it requires reading `removed`, when we can also just always set it to `true`.

src/function/execute.rs
MichaReiser: Finding the outer-most query is somewhat difficult when it stretches across multiple threads. At some point, I had an implementation that queried the `DependencyGraph` to find the outermost cycle. However, it turned out that finding the outermost cycle was fairly expensive. Just finding "an outer" query is cheaper, and we can leave it to "that outer" query to find the "next outer query". Ultimately, we'll reach the outermost query. 

Things are much easier if all queries are on the stack, which is what we handle here. We start from the outermost active query and try to find the first query that is a cycle head. This gives us a "local maximum" for what the outermost query is.

src/function/sync.rs
MichaReiser: I can add a comment. It's an optimization so that `ClaimGuard::release` can always early exit if `anyone_waiting` is false without having to check `is_transfer_target` too.

src/runtime/dependency_graph.rs
ibraheemdev: Yeah it would be fine as a field comment.

src/function/execute.rs
MichaReiser: It's a bit tricky. `value.is_none` only indicates that this is a panicked memo if the memo is also from the same zalsa revision, contains its own cycle head, and has no value. That's literally repeating all checks from above. That's why I'd prefer leaving it as is because it keeps the logic: Is this a provisional value that we can reuse, in a single place

PR #1007, opened 2025-10-17
Introduce a `CancellationToken` for cancelling specific computations

Veykril:
Introduces a `CancellationToken` that can be used to cancel a specific database computation opposed to cancelling the whole runtime. This is traced by storing a second cancellation state on `ZalsaLocal` itself opposed to the runtime cancel flag.

When a thread gets cancelled, it will unwind as usual but with a different payload than pending write cancellation, threads blocked on such a cancelled thread will instead of propagating the cancellation run the computation they are blocked on themselves.

The cancellation state of the database gets reset when we exit the database TLS.

### DIFF
--- a/src/attach.rs
+++ b/src/attach.rs
@@ -40,6 +40,9 @@ impl Attached {
         Db: ?Sized + Database,
     {
         struct DbGuard<'s> {
+            /// The database that *we* attached on scope entry.
+            ///
+            /// `None` if one was already attached by a parent scope.
             state: Option<&'s Attached>,
         }
 
@@ -47,6 +50,7 @@ impl Attached {
             #[inline]
             fn new(attached: &'s Attached, db: &dyn Database) -> Self {
                 match attached.database.get() {
+                    // A database is already attached, make sure it's the same as the new one.
                     Some(current_db) => {
                         let new_db = NonNull::from(db);
                         if !std::ptr::addr_eq(current_db.as_ptr(), new_db.as_ptr()) {
@@ -54,8 +58,8 @@ impl Attached {
                         }
                         Self { state: None }
                     }
+                    // No database is attached, attach the new one.
                     None => {
-                        // Otherwise, set the database.
                         attached.database.set(Some(NonNull::from(db)));
                         Self {
                             state: Some(attached),
@@ -88,7 +92,13 @@ impl Attached {
         Db: ?Sized + Database,
     {
         struct DbGuard<'s> {
+            /// The database that *we* attached on scope entry.
+            ///
+            /// `None` if one was already attached by a parent scope.
             state: Option<&'s Attached>,
+            /// The previously attached database that we replaced, if any.
+            ///
+            /// We need to make sure to rollback and activate it again when we exit the scope.
             prev: Option<NonNull<dyn Database>>,
         }
 
@@ -97,21 +107,24 @@ impl Attached {
             fn new(attached: &'s Attached, db: &dyn Database) -> Self {
                 let db = NonNull::from(db);
                 match attached.database.replace(Some(db)) {
+                    // A database was already attached by a parent scope.
                     Some(prev) => {
                         if std::ptr::eq(db.as_ptr(), prev.as_ptr()) {
+                            // and it was the same as ours, so we did not change anything.
                             Self {
                                 state: None,
                                 prev: None,
                             }
                         } else {
+                            // and it was the a different one from ours, record the state changes.
                             Self {
                                 state: Some(attached),
                                 prev: Some(prev),
                             }
                         }
                     }
+                    // No database is attached, attach the new one.
                     None => {
-                        // Otherwise, set the database.
                         attached.database.set(Some(db));
                         Self {
                             state: Some(attached),
--- a/src/function/memo.rs
+++ b/src/function/memo.rs
@@ -181,7 +181,7 @@ impl<'db, C: Configuration> Memo<'db, C> {
                     TryClaimHeadsResult::Running(running) => {
                         all_cycles = false;
                         if !running.block_on(zalsa) {
-                            // We cannot handle local cancellations in fixpoints
+                            // We cannot really handle local cancellations reliably here
                             // so we treat it as a general cancellation / panic.
                             //
                             // We shouldn't hit this though as we disable local cancellation


Veykril:
Looks like allocator noise from the new arc alloc

Veykril:
That would be the ideal I think

src/cancelled.rs
```diff
@@ -10,15 +12,25 @@ use std::panic::{self, UnwindSafe};
...
 
+#[derive(Debug)]
+pub struct UncancelGuard(pub(crate) CancellationToken);
+
+impl Drop for UncancelGuard {
+    fn drop(&mut self) {
+        self.0.uncancel();
+    }
+}
+
```
Veykril: Not too fond of this way to reset the guard. I think we can make this work in (un)attach instead

src/cancelled.rs
```diff
@@ -45,6 +46,7 @@ impl Cancelled {
 impl std::fmt::Display for Cancelled {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let why = match self {
+            Cancelled::Cancelled => "canellation request",
```
MichaReiser: Nit: Maybe `Cancelled::Local`?

src/database.rs
```diff
@@ -59,14 +60,19 @@ pub trait Database: Send + ZalsaDatabase + AsDynDatabase {
...
     /// will block until that snapshot is dropped -- if that snapshot
     /// is owned by the current thread, this could trigger deadlock.
     fn trigger_cancellation(&mut self) {
         let _ = self.zalsa_mut();
     }
 
+    /// Retrives a [`CancellationToken`] for the current database.
+    fn cancellation_token(&self) -> CancellationToken {
+        self.zalsa_local().cancellation_token()
+    }
```
MichaReiser: The `database` terminology is a bit misleading here. The way I think about the database is that there's only a single database, even if multiple owned instances refer to it after cloning it. 

While using the terminology thread is sort of inaccurate too because the database can move across threads, I think it would give me a better sense of what's going on.

src/function/fetch.rs
```diff
@@ -105,32 +105,39 @@ where
...
+        let claim_guard = loop {
+            match self
+                .sync_table
+                .try_claim(zalsa, zalsa_local, id, Reentrancy::Allow)
+            {
+                ClaimResult::Claimed(guard) => break guard,
+                ClaimResult::Running(blocked_on) => {
+                    if !blocked_on.block_on(zalsa) {
+                        continue;
+                    }
```
MichaReiser: Could we just return `None` in that case, the outer `fetch` loop should then retry.

src/function/maybe_changed_after.rs
```diff
@@ -141,21 +141,24 @@ where
...
-                return None;
-            }
-            ClaimResult::Cycle { .. } => {
-                return Some(self.maybe_changed_after_cold_cycle(
-                    zalsa_local,
-                    database_key_index,
-                    cycle_heads,
-                ))
+        let claim_guard = loop {
+            match self
```
MichaReiser, line 145: Same here: Could we simply exit with `None` if `try_claim` failed and let the otuer loop retry

src/runtime.rs
```diff
@@ -121,7 +122,11 @@ struct BlockedOnInner<'me> {
 
 impl Running<'_> {
     /// Blocks on the other thread to complete the computation.
-    pub(crate) fn block_on(self, zalsa: &Zalsa) {
+    ///
+    /// Returns `true` if the computation was successful, and `false` if the other thread was cancelled.
+    #[must_use]
+    #[cold]
```
MichaReiser: Does the `cold` here improve performance?

src/zalsa.rs
```diff
@@ -301,8 +301,11 @@ impl Zalsa {
     #[inline]
     pub(crate) fn unwind_if_revision_cancelled(&self, zalsa_local: &ZalsaLocal) {
         self.event(&|| crate::Event::new(crate::EventKind::WillCheckCancellation));
+        if zalsa_local.is_cancelled() {
+            zalsa_local.unwind_cancelled();
+        }
```
MichaReiser: I wonder what the impact of this is, given that this is in a very hot function.

src/attach.rs
```diff
@@ -85,25 +88,50 @@ impl Attached {
         Db: ?Sized + Database,
     {
         struct DbGuard<'s> {
-            state: &'s Attached,
+            state: Option<&'s Attached>,
```
MichaReiser, line 98: What's the reason to make this optional? It's not evident to me how it's related to the change (but it probably it is, just not evident to me)

Veykril: We need to track when the database actually changes which this helps in doing, since when it changes thats a signal for it having exited its outermost scope allowing us to reset the cancellation state.

src/database.rs
Veykril: Maybe just database handle?

src/runtime.rs
Veykril: Unsure, wanted see if it had an impact. Would be best to check as a standalone PR I think. Will remove it for now given that.

src/zalsa.rs
Veykril: its hard to tell given the new arc allocation throws over the code speed results.

MichaReiser: I upgraded ty and the benchmarks were neutral

.github/workflows/test.yml
```diff
@@ -91,7 +91,7 @@ jobs:
       - name: Test with Miri
         run: cargo miri nextest run --no-fail-fast --tests
         env:
-          MIRIFLAGS: -Zmiri-disable-isolation -Zmiri-retag-fields
```
Veykril, line 94: This is a default now

src/function/memo.rs
```diff
@@ -180,7 +180,14 @@ impl<'db, C: Configuration> Memo<'db, C> {
                     }
                     TryClaimHeadsResult::Running(running) => {
                         all_cycles = false;
-                        running.block_on(zalsa);
+                        if !running.block_on(zalsa) {
+                            // We cannot handle local cancellations in fixpoints
+                            // so we treat it as a general cancellation / panic.
+                            //
+                            // We shouldn't hit this though as we disable local cancellation
+                            // in cycles.
```
MichaReiser: `block_on_heads` isn't used by fixpoint anymore. It's only used by the `FallbackImmediate` cycle recovery strategy

src/function/sync.rs
```diff
@@ -446,7 +469,11 @@ impl<'me> ClaimGuard<'me> {
 impl Drop for ClaimGuard<'_> {
     fn drop(&mut self) {
         if thread::panicking() {
-            self.release_panicking();
+            if self.zalsa_local.should_trigger_local_cancellation() {
+                self.release_cancelled();
+            } else {
+                self.release_panicking();
+            }
```
MichaReiser: Let's parameterize `release_panicking` with a `WaitResult` instead so that we keep this function minimal

src/function/sync.rs
```diff
@@ -295,6 +306,7 @@ pub(crate) struct ClaimGuard<'me> {
     zalsa: &'me Zalsa,
     sync_table: &'me SyncTable,
     mode: ReleaseMode,
+    zalsa_local: &'me ZalsaLocal,
```
MichaReiser, line 309: We could consider exposing `zalsa_local` on `ClaimGuard` and use the new method in functions that take a `ClaimGuard` and `ZalsaLocal` today

tests/dataflow.rs
```diff
@@ -25,7 +25,7 @@ struct Definition {
 #[derive(Eq, PartialEq, Clone, Debug, salsa::Update)]
 enum Type {
     Bottom,
-    Values(Box<[usize]>),
+    Values(BTreeSet<usize>),
```
MichaReiser: What's the reason for this change?

src/attach.rs
MichaReiser: I would find it useful if we document the different states that are possible (what does `state: Some`, `prev: None` mean etc)

src/attach.rs
```diff
@@ -85,25 +88,50 @@ impl Attached {
...
 
         impl Drop for DbGuard<'_> {
             #[inline]
             fn drop(&mut self) {
-                self.state.database.set(self.prev);
+                // Reset database to null if we did anything in `DbGuard::new`.
+                if let Some(attached) = self.state {
+                    if let Some(prev) = attached.database.replace(self.prev) {
+                        // SAFETY: `prev` is a valid pointer to a database.
+                        unsafe { prev.as_ref().zalsa_local().uncancel() };
```
MichaReiser, line 145: What if `prev` was cancelled too? Which is possible because `CancellationToken` is `Clone`. 

Should we instead simply consider that db as "dead". Meaning, you have to call `Clone` to get a new Db afterwards (What you do is probably the opposite, you clone the Db before passing it to a request handler, and the db is later dropped)

MichaReiser (commented):
It probably makes sense to have this in Salsa but is an implementation outside Salsa considerably more work?

tests/dataflow.rs
Veykril: Reduce re-allocations which speeds up the miri run of the test a bit. Was gonna lift that out in a separate PR but forgot (will kick it out of this one again)

src/attach.rs
Veykril: If `prev` was cancelled too then the query that initially attached it will reset the state once it exits its scope. We always put the databases back that we replaced in attached when the scope ends due to this.

src/function/memo.rs
Veykril: Is `FallbackImmediate` even necessary anymore? iirc you've removed the requirement of fixpoint having to converge right?

MichaReiser: Whether it's still necessary depends on r-a. We aren't using it in ty

> iirc you've removed the requirement of fixpoint having to converge right?

No, fixpoint queries still have to converge. I only changed how we run nested queries.

Veykril: Ah i mixed it up with https://github.com/salsa-rs/salsa/pull/991

src/attach.rs
MichaReiser: Okay, I think I got confused by the variable being named `prev`. I assumed it's `DbGuard.prev` when it is the database stored in `attached.database` (which isn't prev but the current db up to when we started dropping).

I think there's still a risk that we uncancel if we have something like `a -> b -> a` and `a` gets cancelled. I understand that we call `a.uncancel()` when we drop the guard for `a -> b -> a` (the last db). 


Should we make this a method on `attached` instead, saying that it resets the database state `(e.g. `.attached.exit`). This also avoids duplicating the same logic.

Veykril: Right, I believe this should be fine though. This could basically cause cancellations to be .. well cancelled. But only if you interleave a bunch of different databases (which ideally we wouldn't even allow imo, but is kind of a necessary evil given the lack of speculative execution).

PR #1014, opened 2025-10-24
Fix cache invalidation when cycle head becomes non-head

MichaReiser:
This PR fixes a `assertion `left == right` failed: Can't merge cycle heads`, but panicked with `ClassLiteral < 'db >::is_typed_dict_(Id(3007)): execute: too many cycle iterations` instead.` panic in ty. 

This panic is normally due to:

1. It's a multithreading issue where one thread returns its result too early, before the outer cycle completes. 
2. Salsa doesn't re-execute a query when it should (`validate_same_iteration` incorrectly returns `false`)

I can rule out 1. because I can reproduce the issue when running ty single-threaded. So it's two. 

The smallest example I've been able to reduce it to is https://github.com/astral-sh/ruff/pull/21059/files#diff-601b44c1045125b03854a01be732f5bbda7b9b70673cbdc983bf9acc952c0a0d which still produced ~300MB of log files. Now, I didn't trace through the log file entirely (I also asked claude and it was very confused and always wanted to remove the assertion because that's obviously the correct fix to make the panic go away...), but my understanding of what's happening (and the fix confirms) is.:

* There's a nested cycle with the outermost query A
* B participates in the cycle and is a cycle head in the first few iterations
* B becomes a non-cycle head in a later iteration
* There's a query `C` that has `B` as its cycle head

The crucial point is that `B` switches from being a cycle head to being a regular cycle participant. The issue with that is that `A` doesn't update `B`'s `iteration_count `when the iteration completes because it only does that for cycle heads (and collecting all queries participating in a query would be sort of expensive?). 

When we now pull `C` in a later iteration, `validate_same_iteration` iterates over all its cycle heads (`B`), and check if the iteration count still matches. Which is the case because `A` didn't update `B`'s iteration count. 

The solution in this PR is that `validate_same_iteration` should also check whether the most recent memo for a query's cycle head is still a cycle head. If it isn't, then the query is obviously outdated and needs to be re-run. 

## Testing

I tried very hard to write a test, I also employed Claude (with different models) and, while we all produced a lot of test code, none of us managed to write a test that triggers this condition successfully. 

That's why I opted to write a regression test in ty instead https://github.com/astral-sh/ruff/pull/21059

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -521,7 +521,22 @@ where
                     }
 
                     // Check if the memo is still a cycle head and hasn't changed
-                    // to a normal cycle participant.
+                    // to a normal cycle participant. This is to force re-execution in
+                    // a scenario like this:
+                    //
+                    // * There's a nested cycle with the outermost query A
+                    // * B participates in the cycle and is a cycle head in the first few iterations
+                    // * B becomes a non-cycle head in a later iteration
+                    // * There's a query `C` that has `B` as its cycle head
+                    //
+                    // The crucial point is that `B` switches from being a cycle head to being a regular cycle participant.
+                    // The issue with that is that `A` doesn't update `B`'s `iteration_count `when the iteration completes
+                    // because it only does that for cycle heads (and collecting all queries participating in a query would be sort of expensive?).
+                    //
+                    // When we now pull `C` in a later iteration, `validate_same_iteration` iterates over all its cycle heads (`B`),
+                    // and check if the iteration count still matches. Which is the case because `A` didn't update `B`'s iteration count.
+                    //
+                    // That's why we also check if `B` is still a cycle head in the current iteration.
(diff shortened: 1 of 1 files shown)


src/cycle.rs
```diff
@@ -468,15 +469,25 @@ pub(crate) fn empty_cycle_heads() -> &'static CycleHeads {
...
 }
 
-#[derive(Debug, PartialEq, Eq)]
-pub enum ProvisionalStatus {
+#[derive(Debug)]
+pub enum ProvisionalStatus<'db> {
     Provisional {
         iteration: IterationCount,
         verified_at: Revision,
+        cycle_heads: &'db CycleHeads,
```
MichaReiser, line 477: I decided to merge `Ingredient::cycle_heads` and `Ingredient::provisional_status` into `provisional_status` so that `validate_same_iteration` doesn't require an additional method call

PR #1015, opened 2025-10-27
Expose the input query Id with `cycle_initial`

mtshiba:
The `cycle_fn` function receives a `salsa::Id` (from #1012), but in this PR we'll make it so that the `cycle_initial` function can also receive it.
This PR contains breaking API changes.

For context on this PR, see https://github.com/astral-sh/ruff/pull/20566#issuecomment-3449992521.

### DIFF
--- a/benches/dataflow.rs
+++ b/benches/dataflow.rs
@@ -70,7 +70,7 @@ fn infer_definition<'db>(db: &'db dyn Db, def: Definition) -> Type {
     }
 }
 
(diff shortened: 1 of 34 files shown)


MichaReiser:
Hmm, the panics from the failed merge queue test run look concerning. I doubt they're related to this change but they suggest that query results can "leak" before the cycle completes. But I haven't been able to reproduce this locally

PR #1018, opened 2025-10-29
Track cycle function dependencies as part of the cyclic query

MichaReiser:
This fixes a bug in our fixpoint handling: the dependencies introduced by the cycle recovery function were recorded as dependencies of the outer function rather than the query with cycle handling. 

This PR also requires that a cycle recovery function not introduce any new cycles, since it's unclear how to handle that.

I have the impression that we should probably to the same for `cycle_initial` but this is a bit trickier because we'd have to merge the `input/outputs` in `execute_maybe_iterate`. I also think it matters less because the dependencies are added to *some query within the cycle* and we always need to execute the cycle as a whole if any input it depends on changed. That's why I think it's probably fine if the dependencies are tracked by another cycle participant.

## Test plan

I added a regression test, yay :) 

Except... it revealed a bug with deep verify memo where we failed to remove the head when we early-returned. Luckily, this was easy to fix.

### DIFF
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -271,9 +271,9 @@ where
                     .provisional_status(zalsa, head.database_key_index.key_index())
                     .expect("cycle head memo must have been created during the execution");
 
-                // A cycle head isn't allowed to be final because that would mean that this query
-                // dependents on the initial value of B, but B finalized without
-                // completing this query that also participates in the cycle.
+                // A query should only ever depend on other heads that are provisional.
+                // If this invariant is violated, it means that this query participates in a cycle,
+                // but it wasn't executed in the last iteration of said cycle.
                 assert!(provisional_status.is_provisional());
 
                 for nested_head in provisional_status.cycle_heads() {


tests/cycle_recovery_dependencies.rs
```diff
@@ -0,0 +1,82 @@
...
+            "WillCheckCancellation",
+            "WillIterateCycle { database_key: query(Id(0)), iteration_count: IterationCount(2) }",
+            "WillCheckCancellation",
+            "WillIterateCycle { database_key: query(Id(0)), iteration_count: IterationCount(3) }",
+            "WillCheckCancellation",
+            "WillIterateCycle { database_key: query(Id(0)), iteration_count: IterationCount(4) }",
+            "WillCheckCancellation",
+            "WillIterateCycle { database_key: query(Id(0)), iteration_count: IterationCount(5) }",
+            "WillCheckCancellation",
+            "DidValidateMemoizedValue { database_key: entry(Id(0)) }",
```
MichaReiser, line 80: Before the fix, this incorrectly re-executed `entry` but not the cycle (which is what depends on the value).

With the fix in `execute_maybe_iterate`, it did re-execute the cycle, but it also re-executed `entry`, which is incorrect because `query` returns the same value (it can be backdated). Now, the behavior is what we want. `entry` does not get re-executed but the cycle is

src/function/execute.rs
```diff
@@ -262,6 +271,11 @@ where
                     .provisional_status(zalsa, head.database_key_index.key_index())
                     .expect("cycle head memo must have been created during the execution");
 
+                // A cycle head isn't allowed to be final because that would mean that this query
+                // dependents on the initial value of B, but B finalized without
+                // completing this query that also participates in the cycle.
```
carljm: I find this comment hard to parse. What is `B`, and how does it relate to the cycle head that we are currently looking at? How do we know that this cycle head depends on the initial value of `B`?

src/function/execute.rs
```diff
@@ -298,6 +312,8 @@ where
                     claim_guard.set_release_mode(ReleaseMode::SelfOnly);
                 }
 
+                let mut completed_query = active_query.pop();
+                *completed_query.revisions.verified_final.get_mut() = false;
```
carljm, line 316: Not sure I understand in what case this would be `true` and we need to reset it to `false`.

carljm (approved):

src/function/execute.rs
MichaReiser: `QueryRevisions::new` defaults to `true` if cycle heads are empty. The cycle heads are always empty when calling `pop` because of the preceding `query_guard.take_cycle_heads`

PR #1020, opened 2025-10-31
Update compile fail snapshots to match new rust stable output

MichaReiser:


carljm (approved):

MichaReiser:
I don't get it why the tests on stable are failing but they just pass fine on my machine (with Rust 1.91....)

PR #1021, opened 2025-10-31
Call `cycle_fn` for every iteration

MichaReiser:
This PR changes when salsa calls `cycle_fn`. 

Before: Salsa only called `cycle_fn` when the last provisional and current value were different (this cycle head hasn't converged)

Now: Salsa will call `cycle_fn` for every iteration

This feels more consistent for a cycle with nested cycles because, before, a query could have converged but kept being executed
because one of the other heads in this cycle hasn't converged yet. 

Always calling the `cycle_fn` also has the benefit that users can use it to finalize the query result. 

This also makes `CycleRecoveryAction` redundant because there's no real difference between `Iterate` and `Fallback`. That's why this PR changes the `cycle_fn` signature to take an owned `value` and return `Self::Output`.

### DIFF
--- a/benches/dataflow.rs
+++ b/benches/dataflow.rs
@@ -6,7 +6,7 @@ use std::collections::BTreeSet;
 use std::iter::IntoIterator;
 
 use codspeed_criterion_compat::{criterion_group, criterion_main, BatchSize, Criterion};
-use salsa::{CycleRecoveryAction, Database as Db, Setter};
+use salsa::{Database as Db, Setter};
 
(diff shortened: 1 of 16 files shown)


carljm (approved):
Thank you!

MichaReiser:
I'd appreciate a thumbs up from @mtshiba on this change before merging, considering that it's a breaking change. I, unfortunately, can't request review from you

mtshiba:
LGTM!

PR #1025, opened 2025-11-18
Fully qualify std Result type

alilleybrinker:
The prior macro expansion could produce errors if the macros were called in a context where `Result` is redefined, for example in a crate with its own `Result` type which pre-fills the error type. This replaces existing `Result` uses with `std::result::Result` to avoid the compilation error in that case.

### DIFF
--- a/benches/dataflow.rs
+++ b/benches/dataflow.rs
@@ -76,13 +76,12 @@ fn def_cycle_initial(_db: &dyn Db, _id: salsa::Id, _def: Definition) -> Type {
 
 fn def_cycle_recover(
     _db: &dyn Db,
-    _id: salsa::Id,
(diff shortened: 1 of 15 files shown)


components/salsa-macro-rules/src/setup_input_struct.rs
```diff
@@ -123,7 +123,7 @@ macro_rules! setup_input_struct {
                 fn serialize<S: $zalsa::serde::Serializer>(
                     fields: &Self::Fields,
                     serializer: S,
-                ) -> Result<S::Ok, S::Error> {
+                ) -> std::result::Result<S::Ok, S::Error> {
```
MichaReiser: Do we have to use `::std` over `std` in case there's a local `std` use or module?

PR #1027, opened 2025-11-20
support interned enums

### LINKED ISSUE #1026: enum interning
Add support for using `#[salsa::interned]` on enums. 

```rust
#[salsa::interned]
enum TestEnum {
	A(String),
	B(Vec<String>)
}
```

Internally, salsa creates a regular interned struct with a single field, the enum value.

gvozdvmozgu:
close #1026

### DIFF
--- a/components/salsa-macros/src/interned.rs
+++ b/components/salsa-macros/src/interned.rs
@@ -1,4 +1,8 @@
 use proc_macro2::TokenStream;
+use quote::ToTokens;
+use syn::spanned::Spanned;
+use syn::visit::Visit;
+use syn::visit_mut::VisitMut;
 
 use crate::hygiene::Hygiene;
 use crate::options::{AllowedOptions, AllowedPersistOptions, Options};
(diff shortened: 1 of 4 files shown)


components/salsa-macros/src/interned.rs
```diff
@@ -144,10 +175,12 @@ impl Macro {
         let Configuration = self.hygiene.ident("Configuration");
         let CACHE = self.hygiene.ident("CACHE");
         let Db = self.hygiene.ident("Db");
+        let additional_items = &self.additional_items;
 
         Ok(crate::debug::dump_tokens(
             struct_ident,
             quote! {
+                #(#additional_items)*
```
MichaReiser, line 183: I'm not a 100% sure whether this is possible but I think it should be. 

Could we move the synthesized struct definition within the `const _: () => {` block. It has the advantage that we can auto generate a name for the interned struct

https://github.com/MichaReiser/salsa/blob/0c64a1a99aeb0986b951eb7b1367dac5b1fbb32f/components/salsa-macro-rules/src/setup_interned_struct.rs#L101-L103

tests/interned-enum.rs
MichaReiser, line 1: Can you add or extend an existing persistence test to show that persistence works with interned enums too

https://github.com/MichaReiser/salsa/blob/0c64a1a99aeb0986b951eb7b1367dac5b1fbb32f/tests/persistence.rs#L4

MichaReiser (commented):
Thank you for working on this.

PR #1028, opened 2025-11-20
pass `Cycle` to the cycle recovery function

mtshiba:
Currently, the cycle recovery function receives the following arguments:

* db: &dyn Db
* id: Id,
* last_provisional_value: &Output,
* value: Output,
* iteration: u32,
* (input: Input, ...)

[After some consideration](https://github.com/astral-sh/ruff/pull/20566#issuecomment-3532767625), we realized that we needed to expose the IDs of all cycle heads, which meant an additional argument was added, and the current approach of passing all information about the cycle to the cycle recovery function one by one would no longer be desirable. Cycle information that can be passed independently of the query function signature should be passed as a single context structure.

In this PR, we will expose this context as `Cycle` and pass it to the cycle recovery function.

### DIFF
--- a/src/cycle.rs
+++ b/src/cycle.rs
@@ -469,16 +469,29 @@ impl Iterator for CycleHeadIdsIterator<'_> {
 }
 
 /// The context that the cycle recovery function receives when a query cycle occurs.
-#[derive(Clone)]
 pub struct Cycle<'a> {
+    pub(crate) head_ids: CycleHeadIdsIterator<'a>,
+    pub(crate) id: Id,
+    pub(crate) iteration: u32,
+}
+
+impl Cycle<'_> {
     /// An iterator that outputs the [`Id`]s of the current cycle heads.
-    /// This will always contain the [`Id`] of the query being processed by the current cycle recovery function,
-    /// but its position will depend on the query execution order.
-    pub head_ids: CycleHeadIdsIterator<'a>,
+    /// This always contains the [`Id`] of the current query but it can contain additional cycle head [`Id`]s
+    /// if this query is nested in an outer cycle or if it has nested cycles.
+    pub fn head_ids(&self) -> CycleHeadIdsIterator<'_> {
+        self.head_ids.clone()
+    }
+
     /// The [`Id`] of the query that the current cycle recovery function is processing.
-    pub id: Id,
+    pub fn id(&self) -> Id {
+        self.id
+    }
+
     /// The counter of the current fixed point iteration.
-    pub iteration: u32,
+    pub fn iteration(&self) -> u32 {
+        self.iteration
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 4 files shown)


src/cycle.rs
```diff
@@ -448,6 +455,32 @@ pub(crate) fn empty_cycle_heads() -> &'static CycleHeads {
...
+    type Item = crate::Id;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next().map(|head| head.database_key_index.key_index())
+    }
+}
+
+/// The context that the cycle recovery function receives when a query cycle occurs.
+#[derive(Clone)]
+pub struct Cycle<'a> {
```
mtshiba, line 472: After actually using it, I decided that it's better not to put `last_provisional_value` inside `Cycle`.

For example, suppose a query function outputs a structure with an internal structure. When the cycle recovery function performs appropriate cycle handling for the structure, it will also need to perform cycle handling for the internal structure. In this case, it would be meaningless if the context passed included the `last_provisional_value` of the outer structure.

cf. https://github.com/mtshiba/ruff/blob/46982fb439176dee7e51e1c62ff01a252c727425/crates/ty_python_semantic/src/place.rs#L693-L714

MichaReiser: Agree, I also came to that conclusion but for a slightly different reason. I was mainly concerned that the generic of `T` could be annoying. I also like that `Cycle` now stores the metadata about the cycle. All data that's specific to the query (inputs, computed values) are passed as separate arguments.

src/cycle.rs
```diff
@@ -448,6 +453,34 @@ pub(crate) fn empty_cycle_heads() -> &'static CycleHeads {
...
+    /// An iterator that outputs the [`Id`]s of the current cycle heads.
+    /// This will always contain the [`Id`] of the query being processed by the current cycle recovery function,
+    /// but its position will depend on the query execution order.
+    pub head_ids: CycleHeadIdsIterator<'a>,
+    /// The [`Id`] of the query that the current cycle recovery function is processing.
+    pub id: Id,
+    /// The counter of the current fixed point iteration.
+    pub iteration: u32,
```
MichaReiser: I suggest making all those private and instead providing public accessor functions only

src/cycle.rs
```diff
@@ -448,6 +453,34 @@ pub(crate) fn empty_cycle_heads() -> &'static CycleHeads {
...
+            .map(|head| head.database_key_index.key_index())
+    }
+}
+
+/// The context that the cycle recovery function receives when a query cycle occurs.
+#[derive(Clone)]
+pub struct Cycle<'a> {
+    /// An iterator that outputs the [`Id`]s of the current cycle heads.
+    /// This will always contain the [`Id`] of the query being processed by the current cycle recovery function,
+    /// but its position will depend on the query execution order.
```
MichaReiser: nit: 

This always contains the id of the current query but it can contain additional cycle head ids if this query is nested in an outer cycle or if it has nested cycles.

MichaReiser (commented):
This looks good to me, I only have a few nit comments

Issue #968, opened 2025-08-08
Document macros


axelkar:
The attribute and derive macros are not currently documented in the Rustdoc: https://docs.rs/salsa/latest/salsa/#attributes

It'd be useful to have a reference for the macros. For example, there's no documentation stating what `#[salsa::db]` does to traits, except a [short line in `salsa::Database`](https://docs.rs/salsa/latest/salsa/trait.Database.html) describing that it can create subtraits.

MichaReiser:
Yes, having some documentation on the macros and what attributes they support would be great. I'd happily accept a PR for this.

PR #1029, opened 2025-11-21
Partially document proc-macros

axelkar:


axelkar:
Partially resolves https://github.com/salsa-rs/salsa/issues/968

MichaReiser:
Thank you. Most macros should now be documented. If there's something missing or you think is unclear, then let's start off a new PR.

PR #1030, opened 2025-11-27
Fix debug impls section and other mistakes in the tutorial

antoniosarosi:
I was going through the [tutorial](https://salsa-rs.netlify.app/tutorial) and found some things that are outdated making some sections confusing. Particulary:

1. The last section of ["Defining the parser: memoized functions and inputs"](https://salsa-rs.netlify.app/tutorial/parser), named ["The `returns(ref)` annotation"](https://salsa-rs.netlify.app/tutorial/parser#the-returnsref-annotation), appears to be outdated because it starts with:

> You may have noticed that `parse_statements` is tagged with `#[salsa::tracked(returns(ref))]`

wich is not true. The definition of `parse_statements` in the example code is:

```rust
#[salsa::tracked]
pub fn parse_statements(db: &dyn crate::Db, source: SourceProgram) -> Program<'_>
```

2. The section [Defining the parser: debug impls and testing](https://salsa-rs.netlify.app/tutorial/debug#defining-the-parser-debug-impls-and-testing) talks about `DebugWithDb` which seems to have been removed in favor of `salsa::Database::attach`. I included an explanation of how that works instead.

Rest of fixes are typos.


book/src/tutorial/parser.md
```diff
@@ -56,13 +56,3 @@ The **second** parameter to a tracked function is **always** some kind of Salsa
...
-
-### The `returns(ref)` annotation
-
-You may have noticed that `parse_statements` is tagged with `#[salsa::tracked(returns(ref))]`.
-Ordinarily, when you call a tracked function, the result you get back is cloned out of the database.
-The `returns(ref)` attribute means that a reference into the database is returned instead.
-So, when called, `parse_statements` will return an `&Vec<Statement>` rather than cloning the `Vec`.
-This is useful as a performance optimization.
-(You may recall the `returns(ref)` annotation from the [ir](./ir.md) section of the tutorial,
-where it was placed on struct fields, with roughly the same meaning.)
```
MichaReiser, line 68: It would be nice if we move this block to the `ir` section and explain the `returns(ref)` (and deref etc) there

MichaReiser (commented):
Thank you. This is great. I've one small suggestion

book/src/tutorial/parser.md
antoniosarosi: I think it's better to explain it in the `parser.md` file because `ir` only talks about structs (input, tracked, interned) whereas `parser` section talks about tracked functions, I think adding `returns(ref)` there makes more sense.

antoniosarosi: It's just explaining it in different words, I got confused by "`parse_statements` is tagged with `returns(ref)`" but it could say "if we tag `parse_statements` with `returns(ref)`, then this happens".

MichaReiser: The reason I suggested moving it to `ir` is because many structs in the ir module have fields attributed with `returns(ref)`

antoniosarosi: Yeah that could make sense. I added an explanation of `returns` for both functions and structs in the `parser` section, it could be split into `returns` for structs in `ir` and then `returns` for functions in `parser`.

antoniosarosi:
@MichaReiser I added separate explanations for both `returns` on struct fields and `returns` on functions. I don't know if they are entirely accurate though.

I also added an example that made it easier for me to understand how the combinatorics of `returns` works when used in structs and functions. I can remove it if it doesn't make sense to have it there.

book/src/tutorial/ir.md
```diff
@@ -177,3 +177,190 @@ whenever anything in a function body changes, we consider the entire function bo
...
+
+```
+number: Number(42)
+```
+
+Type of `number`:
+
+```rust
+let number: Number = input.number(&db);
+```
```
MichaReiser: I find this section too long. I think what I would do instead is use an existing definition from the `ir` file or we define a made up struct and use the simple bullet list above to explain the different return types:

* `clone`: Reading the field returns a clone (requires `Clone`), e.g. `Option<String>`
* `copy`: Reading the field returns a copy (requires `Copy`), e.g. `Option<String>`
* `ref`: Reading the field returns a reference, e.g. `&Option<String>`
* `deref`: Reading the field returns the dereferenced field (requires `SalsaAsDeref`), e.g. `Option<&str>`
* `asref`: Reading the ... (requires `SalsaAsRef`), e.g. `Option<&String>`

book/src/tutorial/parser.md
```diff
@@ -57,12 +57,257 @@ Tracked functions may take other arguments as well, though our examples here do
...
+struct fields.
+
+- `salsa::tracked(returns(clone))` (**the default**): Invokes `Clone::clone` on function's return type.
+- `salsa::tracked(returns(ref))`: Returns a reference to the functions return type: `&Type` .
+- `salsa::tracked(returns(deref))`: Invokes `Deref::deref` on the return type.
+- `salsa::tracked(returns(copy))`: Returns an owned copy of the value.
+
+We will use a modified version of the example from the [IR](./ir.md#the-returns-attribute-for-struct-fields)
+chapter to explain how the `returns` annotation works for function return types
+instead of struct fields.
```
MichaReiser: I would skip that entire section. The semantics are exactly the same and pointing to the ir documentation should be enough

antoniosarosi: @MichaReiser I changed it. Still using the made up struct but only bullet list explanation for each `returns`.

MichaReiser:
The section should now have been updated

PR #1033, opened 2025-12-03
implement `Update` for `OrderMap` and `OrderSet`

oconnor663:
There are places in `ty` where we need `IndexMap` because it's `salsa::Update`, and other places where we need `OrderMap` because it's `Hash`. @ibraheemdev has suggested that if we `impl Update for OrderMap` we could just use `OrderMap` across the board.

This is my first Salsa PR, so apologies for any obvious mistakes. In particular, I didn't see any tests for the existing implementations, so I haven't added any for the new ones?


Cargo.toml
```diff
@@ -19,6 +19,7 @@ hashbrown = "0.15"
 hashlink = "0.10"
 indexmap = "2"
 intrusive-collections = "0.9.7"
+ordermap = "1"
```
Veykril, line 22: This should be a feature gated dependency at the very least

oconnor663: I was going off how `indexmap` was handled, but now that I look again I guess that's because it's used internally and not purely imported for trait impls. I'll put up a followup PR.

MichaReiser: Whoops, sorry, I missed this. Thanks @Veykril

oconnor663: https://github.com/salsa-rs/salsa/pull/1034

PR #1050, opened 2025-12-30
Collect cycle heads transitively

MichaReiser:
I started looking into fixing the `maybe_changed_after` unsoundness when I noticed that I don't fully understand how it can happen that `outer_cycle` is `None` for a query that has cycle heads but doesn't depend on itself. It's essential that there's always an outer cycle in that case that finalizes the cycle participant once it converges. 

My first change was to make Salsa panic if we encounter the situation that `outer_cycle` is `None` (and the query has cycle heads but doesn't depend on itself), but we then started seeing frequent panics in ty. So I dug a bit deeper and found that we need to transitively collect the cycle heads (only going one level deep isn't enough). 

I added a regression test that demonstrates why transitively collecting all cycle heads is necessary. 

Knowing the outer cycle is not only necessary so that we can transfer the head's lock to the proper outer cycle, it's also necessary so that the memo's considered stale when the iteration count of the outer-most cycle changes (`validate_same_iteration`).

### DIFF
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -620,7 +620,7 @@ fn complete_cycle_participant(
     iteration_count: IterationCount,
 ) -> CompletedQuery {
     // For as long as this query participates in any cycle, don't release its lock, instead
-    // transfer it to the outermost cycle head (if any). This prevents any other thread
+    // transfer it to the outermost cycle head. This prevents any other thread
     // from claiming this query (all cycle heads are potential entry points to the same cycle),
     // which would result in them competing for the same locks (we want the locks to converge to a single cycle head).
     claim_guard.set_release_mode(ReleaseMode::TransferTo(outer_cycle));


src/function/execute.rs
```diff
@@ -503,64 +506,107 @@ fn outer_cycle(
...
+            depends_on_self |= nested_depends_on_self;
         }
+
+        (max_iteration_count, depends_on_self)
     }
 
-    for (head_key, iteration_count) in missing_heads {
-        max_iteration_count = max_iteration_count.max(iteration_count);
-        depends_on_self |= head_key == database_key_index;
+    let mut missing_heads: SmallVec<[(DatabaseKeyIndex, IterationCount); 4]> = SmallVec::new();
```
ibraheemdev, line 591: I assume the choice of 4 is ~arbitrary here?

src/function/execute.rs
```diff
@@ -570,18 +616,14 @@ fn complete_cycle_participant(
     active_query: ActiveQueryGuard,
     claim_guard: &mut ClaimGuard,
     cycle_heads: CycleHeads,
-    outer_cycle: Option<DatabaseKeyIndex>,
+    outer_cycle: DatabaseKeyIndex,
     iteration_count: IterationCount,
 ) -> CompletedQuery {
     // For as long as this query participates in any cycle, don't release its lock, instead
     // transfer it to the outermost cycle head (if any). This prevents any other thread
```
```suggestion
    // transfer it to the outermost cycle head. This prevents any other thread
```

src/function/execute.rs
MichaReiser: Yes

PR #1052, opened 2026-01-05
Add `DidFinalizeCycle` event, add more tests for cyclic nested queries

MichaReiser:
This adds a new `DidFinalizeCycle` event that Salsa emits when the outermost cycle head is finalized. 

This PR also adds two new tests:

1. A test where the nested cycle converges before the outer cycle converges (it can be finalized before the outer cycle
2. A test where the nested cycle never converges, but the outer cycle only depends on it in the first iterations.

I haven't yet decided on what the expectation is for both those cases, but I thought it's worth adding tests

### DIFF
--- a/src/event.rs
+++ b/src/event.rs
@@ -59,7 +59,7 @@ pub enum EventKind {
         database_key: DatabaseKeyIndex,
     },
 
-    /// Salsa starts a new fixpoint iteration for the cycle with `database_key` as it's
+    /// Salsa starts a new fixpoint iteration for the cycle with `database_key` as its
     /// outermost cycle.
     WillIterateCycle {
         /// The database-key for the cycle head. Implements `Debug`.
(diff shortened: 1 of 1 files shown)


src/event.rs
```diff
@@ -59,12 +59,21 @@ pub enum EventKind {
         database_key: DatabaseKeyIndex,
     },
 
+    /// Salsa starts a new fixpoint iteration for the cycle with `database_key` as it's
```
ibraheemdev:
```suggestion
    /// Salsa starts a new fixpoint iteration for the cycle with `database_key` as its
```

src/event.rs
```diff
@@ -59,12 +59,21 @@ pub enum EventKind {
...
+    /// outermost cycle.
     WillIterateCycle {
         /// The database-key for the cycle head. Implements `Debug`.
         database_key: DatabaseKeyIndex,
         iteration_count: IterationCount,
     },
 
+    /// Salsa completed a fixpoint iteration for the cycle with `database_key` as it's
```
ibraheemdev:
```suggestion
    /// Salsa completed a fixpoint iteration for the cycle with `database_key` as its
```

PR #1056, opened 2026-01-09
Document cycle_fallback in the book

Wilfred:
This was added in #801 but wasn't previously documented.

### DIFF
--- a/book/src/cycles.md
+++ b/book/src/cycles.md
@@ -65,4 +65,5 @@ fn cycle_fallback(_db: &dyn KnobsDatabase, _id: salsa::Id) -> u32 {
 }
 ```
 
-Note that `cycle_fallback` can [result in non-determinism](https://github.com/salsa-rs/salsa/pull/798#issuecomment-2812855285). Prefer using `cycle_fn` with a fixpoint where possible.
+Unlike fixpoint iteration, queries attributed with `cycle_fallback` also use their fallback value if 
+they participate in a cycle. This is to ensure the query result doesn't depend on the query execution order ([details](https://github.com/salsa-rs/salsa/pull/798#issuecomment-2812855285)).


book/src/cycles.md
```diff
@@ -42,6 +46,21 @@ When the `cycle_fn` receives a newly computed value, you can implement a strateg
...
 Also, in fixed-point iteration, it is advantageous to be able to identify which cycle head seeded a value. By embedding a `salsa::Id` (2nd parameter) in the initial value as a "cycle marker", the recovery function can detect self-originated recursion.
 
-## Calling Salsa queries from within `cycle_fn` or `cycle_initial`
+### Calling Salsa queries from within `cycle_fn` or `cycle_initial`
 
 It is permitted to call other Salsa queries from within the `cycle_fn` and `cycle_initial` functions. However, if these functions re-enter the same cycle, this can lead to unpredictable results. Take care which queries are called from within cycle-recovery functions, and avoid triggering further cycles.
+
+## Fallback Values
+
+You can use `cycle_result` to specify a fallback value if Salsa detects a cycle. Salsa will ignore all the other queries and use this value directly.
```
MichaReiser: To me it's not entirely clear what `ignore all other queries` means in this context. Are you trying to say that salsa discards the in-progress query result and instead returns fallback.

I think it might also be worth mentioning here that this can result in non-determinism because Salsa never enforces that all queries converge. But I don't fully remember the details here. But I think there was some discussion around it on the PR that introduced the feature

Wilfred: @MichaReiser "ignore all other queries" is how it's described in #801 FWIW, and per https://github.com/salsa-rs/salsa/pull/801#issuecomment-2817189581 I think the final version of the PR is deterministic.

> Are you trying to say that salsa discards the in-progress query result and instead returns fallback.

I think so? That's my understanding from a skim of the PR, but happy to tweak wording if you prefer something else :)

MichaReiser: I think the confusing part is the `ignore all other queries`. This makes it sound as if it throws away the result of any query called as part of the cyclic function. However, this is not the case. The only result that it throws away is the result of the cycle head (the query with the `cycle_fallback`), and instead, replaces the value with the fallback value. 

Maybe an easier way to describe this is that a query with an `cycle_result` attribute always runs to completion, but its resulting value will be replaced with fallback value if it hits a cycle (instead of whatever value it ended up computing). 

The non-determinism is described in this comment https://github.com/salsa-rs/salsa/pull/798#issuecomment-2812855285 (I think it's fine linking to that comment but it's worth pointing out that this is a limitation of this approach over proper fixpoint)

Wilfred: @MichaReiser just to confirm, that's a different PR -- #798 vs #801 -- does #801 also have non-determinism?

MichaReiser: Yes, they both have the same non-determinism

book/src/cycles.md
```diff
@@ -42,6 +46,24 @@ When the `cycle_fn` receives a newly computed value, you can implement a strateg
...
+fn query(db: &dyn salsa::Database) -> u32 {
+    // ...
+}
+
+fn cycle_fallback(_db: &dyn KnobsDatabase, _id: salsa::Id) -> u32 {
+    42
+}
+```
+
+Unlike fixpoint iteration, queries attributed with `cycle_fallback` also use their fallback value if
```
MichaReiser, line 68: I learned more about `cycle_fallback` and am now convinced that it doesn't have any non-deterministic behavior. I updated the section here

Issue #1023, opened 2025-11-03
Flaky cycle fixpoint test


MichaReiser:
There's a good chance that running `cargo nextest run --test parallel cycle_nested_deep_conditional_changed` fails.

Trace of one such run https://gist.github.com/MichaReiser/ab218daba30ed41904f0f1f4d2f338d3

Issue #909, opened 2025-06-11
Flatten query dependencies


MichaReiser:
We discussed this in our sync meeting today. The basic idea is to allow queries to opt-in to a more coarse grained tracking of dependencies. In this mode, Salsa wouldn't track all dependencies but instead only track the base inputs a query (transitively) depends on. 

This should, in many cases, reduce the overall dependencies, unless a query depends on many inputs, in which case it probably shouldn't opt-in to this mode. For example, a query that computes information for a single file would only depend on that input file instead of all files.

davidbarsky:
cc: @ChayimFriedman2, as I think this might be of interest to you.

ibraheemdev:
We talked a bit today about combining this idea with bloom-filter-like bucketing to avoid input explosion. This was in the context of persistent caching, but it could be useful at runtime as well.

Veykril:
I mean, do we still want to support accumulated values? Is anyone using them?

MichaReiser:
> I mean, do we still want to support accumulated values? Is anyone using them?

Niko uses them in the data compiler but we agreed that we should either remove them or implement them in a different way.

PR #1061, opened 2026-01-10
Fix out-of-order verification of cycle head dependencies

MichaReiser:
## Problem

A foundational assumption of `deep_verify_memo` is that it walks query dependencies in execution order (as seen by the query being validated). This PR fixes a bug related to cyclic queries where this invariant was violated because of how Salsa tracks dependencies:

* Salsa only stores the `DatabaseKeyIndex` but not the `IterationCount` when adding a read dependency. The result is that Salsa adds a single dependency even if a query reads a `x_1` and `x_2` where `_i` signifies the iteration.
* We only store the last memo, making it impossible to retrieve a query's dependencies in iteration `i`. 

The result is that `deep_verify_edges` verifies dependencies before the structs on which the query results are stored, which is unsound.

Here an example where this happened before (this is a new test case):

```rust
#[salsa::tracked(cycle_initial=cycle_initial)]
fn query_a<'db>(db: &'db dyn salsa::Database) -> Interned<'db> {
    let interned = query_b(db);
    let value = interned.value(db);

    if value < 10 {
        Interned::new(db, value + 1)
    } else {
        interned
    }
}

#[salsa::tracked]
fn query_b<'db>(db: &'db dyn Database) -> Interned<'db> {
    let interned = query_a(db);
    query_x(db, interned);
    interned
}

#[salsa::tracked]
fn query_x<'db>(_db: &'db dyn Database, _i: Interned<'db>) {}

fn cycle_initial(db: &dyn Database, _id: Id) -> Interned<'_> {
    Interned::new(db, 0)
}

#[salsa::interned]
struct Interned {
    value: u32,
}
```

Collected input/outputs by query:

* `query_a`
    * `query_b`
    * `Interned::new(1)`
        <hr>
    * `Interned::new(2)`
        <hr>
    * `Interned::new(3)`
        <hr>
    * ...
    * `Interned::new(10)`
* `query_b`
    * `Interned::new(0)`
    * `query_a`
    * `query_x(Interned(0))`
        <hr>
    * `query_x(Interned(1))`
        <hr>
    * `query_x(Interned(2))`
        <hr>
    * ...
    * `query_x(Interned(10))`
* `query_x`: Not important

Traversal order in `maybe_changed_after`:

1. `query_a`
    1. `query_b`
        1. `Interned::new(0)`
        3. `query_a` (returns unchanged, fixpoint initial)
        4. `query_x(0)`
        5. `query_x(1)`
        6. `query_x(2)`
        7. `query_x(...)`
        8. `query_x(10)`
        1. `Interned::new(1)`
    2. `Interned::new(2)`
    3. `Interned::new(...)`
    4. `Interned::new(10)`

Note how Salsa verifies `query_x(1)`, and `query_x(...)` before verifying that the interned values haven't changed. 

The issue is that Salsa sees a dependency on `query_b` (without any iteration!), but it then doesn't just validate the dependencies of `query_b` in iteration 0, it validates all its dependencies, even if they are from later iterations. Only once that's done does salsa verify the remaining dependencies from `query_a`, even though they were created before some of the dependencies that were just verified as part of `query_b`. 

## Solution

This PR flattens the dependencies of all queries with cycle handling that participate in a cycle after each execution (each iteration) so that the cycle head depends only on inputs, tracked structs, interned values, and finalized queries. This ensures that the query's dependencies form a tree, not a graph (no cycles). 

Not only does this fix the bug mentioned above, but it also greatly simplifies `maybe_changed_after`, as it no longer needs to handle cyclic dependencies. This allows us to remove almost all fixpoint-related code from `maybe_changed_after` (and friends).

## Downsides

There are a few downsides to this approach:

* It inherently breaks accumulated values. Collecting the accumulated values requires walking the query's dependency tree. This is fundamentally at odds with this new approach because we intentionally trim the dependency tree. I didn't consider this a blocker because trimming dependency trees (specifically by flattening) is something that we want to introduce in other parts of Salsa, to reduce memory usage (https://github.com/salsa-rs/salsa/issues/909). The other reason is that accumulated values were already "broken," in the sense that they returned too many values when a cycle head depended on the accumulating function in some iterations but not all, but the result would contain the accumulated values from all iterations.
* Flattening the dependencies for each cycle head participating in a cycle isn't cheap, as the benchmarks show. I initially worked on a version that only required the outermost cycle head's dependencies. This implementation only regressed performance by 8-9%. However, it required a much more complicated design because `deep_verify_memo` for a nested head needed to delegate to the outermost cycle head. This, in turn, required tracking more state to verify if the outermost cycle head is still "the same" as when the inner head was executed (a cycle finalized in the field), and it's also much more difficult to reason about in general. It also didn't work 😆 Specifically, it didn't correctly handle the case where an inner cycle converges before its outer cycle. The issue is that the inner cycle counted on the outer most cycle to do the dependency flattening. But the inner cycle now suddenly needs to know its own flattened dependencies across all previous iterations. We could probably make this approach work, but it would require that the inner cycle walks its dependencies from the previous iteration, and replaces its dependency on the outer-most query with its flattened dependencies. However, it's not clear to me how to detect the outer-most cycle without storing it in the Memo. Ultimately, this didn't seem worthwhile, given there's no meaningful performance regression on ty (one benchmark shows a 1% regression)
* The memory overhead can be substantial for large cycles if they have many dependencies, because most dependencies are duplicated across all heads participating in the same cycle. It would be nice if there were some structural sharing to reduce the overhead, but I consider this a follow-up. I also think that there are much bigger opportunities in Salsa to reduce memory usage (immortal durability, within the same revision LRU).

## Alternative solutions

I still think the ideal is for Salsa to store multiple memos for queries participating in fixpoint, and for the iteration count to be part of the `QueryEdge`. This would remove much of the special casing required for fixpoint iteration and fit more naturally into Salsa's overall design. It probably would also set the foundation for speculative execution, where an iteration is "one speculative execution", but we could use other keys. 

I decided not to explore this further because there are some big open questions, where it was unclear to me how to resolve them in the limited time I had available:

* How can we store multiple memos?
* Should the iteration count be part of the `DatabaseKeyIndex` or separate? I think it must be a single key; the caller of `fetch` doesn't know what the "latest" iteration of a given query is. But it always wants the most recent value.
* How can we include the iteration count in the `QueryEdge` without increasing its size?
* How does locking work? Validating different iterations of the same query in `maybe_changed_after` shouldn't result in a cycle, but acquiring the lock for the key without an iteration should. 

## How to support `diff_outputs`?

I don't think there's an open issue for it, but our fixpoint handling is prone to removing outputs and tracked structs too early for queries participating in a cycle. Salsa diffs the outputs immediately after completing an iteration. However, some outputs (including tracked structs) might only have been created in later iterations. Salsa deleting them in the first iteration is overly aggressive (and it used to be unsound https://github.com/salsa-rs/salsa/pull/980). To fix this, it would be very helpful if there were a hook to finalize all cycle participants, rather than our current lazy validation. 

My initial approach to this PR used eager validation, but that didn't handle the case where a nested cycle converges before the outer cycle. So I switched to flattening all cycle head dependencies; however, that also means that walking the dependency tree during finalization no longer contains the full dependency tree, making eager finalization impossible (or we'd have to retain the query dependencies in a different way). But I think there's still a solution to this problem, but we have to defer `diff_outputs` to the next revision.

I don't have this fully figured out but my thinking is along these lines:

* Store the old-memo's revision metadata on the memo when completing the first iteration and carry that information forward across all later iterations.
* When completing a query, check if this is the initial iteration and the `old_memo` has `prev_revision_metadata` set. If so, diff the `prev_revision_metadata` with the `old_memo.input_outputs` and delete any no-longer-created outputs. 

## Notes for reviewer

This hopefully fixes https://github.com/salsa-rs/salsa/issues/1023

For some more context, see the write up in https://hackmd.io/XCIIyYjlQkuVg33THEr4iQ

This PR is based on top of https://github.com/salsa-rs/salsa/pull/1063 because it removes the need to special-case fallback-immediate.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -124,7 +124,7 @@ jobs:
 
   benchmarks:
     # https://github.com/CodSpeedHQ/action/issues/126
-    if: github.event_name != 'merge_group'
+    if: github.repository == 'salsa-rs/salsa' &&github.event_name != 'merge_group'
     name: Benchmarks
     runs-on: ubuntu-latest
     steps:
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -142,17 +142,17 @@ where
             {
                 ClaimResult::Claimed(guard) => guard,
                 ClaimResult::Running(blocked_on) => {
-                    _ =blocked_on.block_on(zalsa);
-                return None;
-            }
-            ClaimResult::Cycle { .. } => {
-                return Some(maybe_changed_after_cold_cycle(
-                    zalsa_local,
-                    database_key_index,
-                    C::CYCLE_STRATEGY,
-                ))
-            }
-        };
+                    let _ = blocked_on.block_on(zalsa);
+                    return None;
+                }
+                ClaimResult::Cycle { .. } => {
+                    return Some(maybe_changed_after_cold_cycle(
+                        zalsa_local,
+                        database_key_index,
+                        C::CYCLE_STRATEGY,
+                    ))
+                }
+            };
         // Load the current memo, if any.
         let Some(old_memo) = self.get_memo_from_table_for(zalsa, key_index, memo_ingredient_index)
         else {


src/function/accumulated.rs
```diff
@@ -79,12 +79,7 @@ where
                 stack.reserve(edges.len());
             }
 
-            stack.extend(
-                origin
-                    .inputs()
-                    .filter_map(|input| TryInto::<DatabaseKeyIndex>::try_into(input).ok())
```
MichaReiser, line 85: I don't know why we used `TryInto` here. `input` is a `DatabaseKeyIndex`, so this is a no-op conversion

src/function/backdate.rs
```diff
@@ -22,7 +22,7 @@ where
         // right now whether backdating could be made safe for queries participating in queries.
         // TODO: Write a test that demonstrates that backdating queries participating in a cycle isn't safe
         // OR write many tests showing that it is (and fixing the case where it didn't correctly account for today).
-        if !revisions.cycle_heads().is_empty() {
+        if !revisions.cycle_heads().is_empty() || old_memo.may_be_provisional() {
```
MichaReiser, line 25: We can't backdate if `old_memo` is the memo from the last iteration (or a previous revision but never finalized). To support backdating cyclic queries, we'd have to back up some of the previous-revision memo's metadata before overriding the memo with the first iteration's result. 

For now, simply skip this path. The downside is that this requires running a few extra queries, but the blast radius should still be contained, assuming those queries aren't cyclic (so that they can backdate).

tests/cycle.rs
```diff
@@ -1065,9 +1068,9 @@ fn cycle_sibling_interference() {
     db.assert_logs(expect![[r#"
         [
             "salsa_event(DidValidateMemoizedValue { database_key: min_panic(Id(2)) })",
+            "salsa_event(WillExecute { database_key: min_panic(Id(3)) })",
```
MichaReiser, line 1071: `d` and `c` are both queries that don't participate in a cycle. They aren't flattened into `a`. `a`'s flattened dependencies are:

* `input(a)`
* `input(b)`
* `c`
* `d`

Salsa now walks `a`s dependency edges in `deep_verify_edges`, starting with `input(a)`, followed by `input(b)` and `c` that all haven't changed. `d` has a changed input. Salsa re-executes it to see if it can be backdated, but it can't, which is why we later re-execute ' a `.

This test does suggest to me that the old implementation is too eager when it comes to re-executing cyclic queries, because it immediately starts executing the outer most cyclic query, even if `d` is later able to backdate.

tests/cycle.rs
```diff
@@ -1129,17 +1132,17 @@ fn repeat_provisional_query_incremental() {
...
     // * Once before iterating
     // * Once as part of iterating
     //
     // If it runs more than once before iterating, than this suggests that
     // `validate_same_iteration` incorrectly returns `false`.
     db.assert_logs(expect![[r#"
         [
-            "salsa_event(WillExecute { database_key: min_panic(Id(2)) })",
-            "salsa_event(WillExecute { database_key: min_panic(Id(1)) })",
             "salsa_event(WillExecute { database_key: min_iterate(Id(0)) })",
```
MichaReiser, line 1143: This now is also more correct. Before, we re-executed `c` first, even though it is a cycle participant. Now, we correctly detect that an input going into `a`'s cycle changed, and immediately start re-executing the entire cycle.

tests/cycle.rs
```diff
@@ -1347,9 +1350,7 @@ fn repeat_query_participating_in_cycle2() {
         [
             "salsa_event(DidValidateInternedValue { key: Interned(Id(400)), revision: R2 })",
             "salsa_event(WillExecute { database_key: head(Id(0)) })",
-            "salsa_event(DidValidateInternedValue { key: Interned(Id(400)), revision: R2 })",
```
MichaReiser, line 1350: We now avoid checking the same interned value multiple times, because we immediately see that the entire cycle needs to re-execute

tests/cycle_output.rs
```diff
@@ -174,9 +174,9 @@ fn revalidate_with_change_after_output_read() {
             "salsa_event(DidSetCancellationFlag)",
             "salsa_event(DidValidateMemoizedValue { database_key: read_value(Id(400)) })",
             "salsa_event(DidValidateInternedValue { key: query_d::interned_arguments(Id(800)), revision: R2 })",
+            "salsa_event(WillExecute { database_key: query_d(Id(800)) })",
```
MichaReiser, line 177: This makes sense to me and is similar to another change we've seen above. 

`query_d` doesn't participate in the cycle, so it doesn't get flattened out. 

`query_b`'s dependencies are:

* `Output(0)` (tracked new)
* `read_value(Output(0))` 
* `query_d` (has changed)
* `ab.value`
* `Output(1)` (next iteration)
* `read_value(Output(1))`
* `Output(2)` (next iteration)
* `read_value(Output(2))`
* `Output(3)` (next iteration)
* `read_value(Output(3))`

Salsa detects that `query_d`'s input changed, it executes `query_d` but can't backdate it because the result is different. In turn, the `query_b` cycle changed, so that salsa re-executes it

src/function/execute.rs
```diff
@@ -785,3 +770,61 @@ fn assert_no_new_cycle_heads(
...
+
+    #[cfg(feature = "accumulator")]
+    {
+        assert!(
+            head.accumulated_inputs.load().is_empty(),
+            "Fixpoint iteration doesn't support accumulated values because it doesn't preserve the original query dependency tree."
+        )
+    }
+
+    // Don't insert `self` here. This is important to ensure that we copy over the
```
ibraheemdev: What is `self` in this context?

src/zalsa_local.rs
```diff
@@ -949,6 +929,47 @@ impl QueryOrigin {
         origin
     }
 
+    /// Sets the `input_outputs` of this query orivin if it's derived or derived untracked.
```
ibraheemdev:
```suggestion
    /// Sets the `input_outputs` of this query origin if it's derived or derived untracked.
```

src/ingredient.rs
```diff
@@ -188,6 +182,25 @@ pub trait Ingredient: Any + std::fmt::Debug + Send + Sync {
         unreachable!("finalize_cycle_head should only be called on cycle heads and only functions can be cycle heads");
     }
 
+    /// Flattens the dependencies of a query with cycle handling that participates in a cycle.
+    ///
+    /// This query recursively walks the dependency graph of `id` and flattens input dependencies
+    /// on provisial queries into `flattened_input_outputs`. Outputs aren't flattened because
+    /// outputs are owned by the creating query and not the cycle head.
+    ///
+    /// Flattening the dependencies is necessary because the memo's depdency graph only captures
```
ibraheemdev:
```suggestion
    /// Flattening the dependencies is necessary because the memo's dependency graph only captures
```

PR #1063, opened 2026-01-11
Merge `FallbackImmediate` and `Fixpoint` code paths

MichaReiser:
## Summary

Spinned out of https://github.com/salsa-rs/salsa/pull/1059#discussion_r2678737797

This PR removes most special casing of `FallbackImmediate` and, instead, integrates it into the normal `Fixpoint` code. Not only should this fix the caching issues that `FallbackImmediate` has today, it also reduces overall complexity and makes it easier to make changes to how we handle cycles in Salsa (no need to update two completely different implementations).

This PR preserves the existing `FallbackImmediate` behavior where all heads participating in a query use their fallback value (and not just the cycle heads). I now believe that this behavior is essential for deterministic outputs, as pointed out by Carl in https://github.com/salsa-rs/salsa/pull/798#issuecomment-2812855285

See https://github.com/salsa-rs/salsa/blob/b6398ecdfb41de7250d21bb91183a9f8bfdc5827/src/function/execute.rs#L96-L107

This PR is based on top of https://github.com/salsa-rs/salsa/pull/1062

One difference worth mentioning is that we now overapproximate the dependencies for queries with `cycle_result`, keeping all dependencies rather than just those up to the initial cycle. This is fine as it is, and I'm not aware of an easy way to remember when we hit the cycle and which dependencies we've seen up to this point.

### DIFF
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -65,14 +65,21 @@ where
                 );
                 (new_value, active_query.pop())
             }
-            CycleRecoveryStrategy::FallbackImmediate | CycleRecoveryStrategy::Fixpoint => self
-                .execute_maybe_iterate(
+            CycleRecoveryStrategy::FallbackImmediate | CycleRecoveryStrategy::Fixpoint => {
+                let zalsa_local = claim_guard.zalsa_local();
+                let was_disabled = zalsa_local.set_cancellation_disabled(true);
+
+                let res = self.execute_maybe_iterate(
                     db,
                     opt_old_memo,
                     &mut claim_guard,
-                    zalsa_local,
                     memo_ingredient_index,
-                ),
+                );
+
+                zalsa_local.set_cancellation_disabled(was_disabled);
+
+                res
+            }
         };
 
         if let Some(old_memo) = opt_old_memo {
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -220,7 +220,8 @@ where
                 };
 
                 // For FallbackImmediate, use the fallback value instead of the computed value
-                // for all cycle participants.
+                // for all cycle participants. This ensures that the results don't depend on the query call order, see
+                // https://github.com/salsa-rs/salsa/pull/798#issuecomment-2812855285.
                 let new_value = if C::CYCLE_STRATEGY == CycleRecoveryStrategy::FallbackImmediate {
                     C::cycle_initial(db, id, C::id_to_input(zalsa, id))
                 } else {
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -7,7 +7,6 @@ use crate::function::sync::ReleaseMode;
 use crate::function::{ClaimGuard, Configuration, IngredientImpl};
 use crate::ingredient::WaitForResult;
 use crate::plumbing::ZalsaLocal;
-use crate::sync::atomic::{AtomicBool, Ordering};
 use crate::sync::thread;
 use crate::tracked_struct::Identity;
 use crate::zalsa::{MemoIngredientIndex, Zalsa};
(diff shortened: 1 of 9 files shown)


src/function/execute.rs
```diff
@@ -311,26 +275,35 @@ where
...
-            let cycle = Cycle {
-                head_ids: cycle_heads.ids(),
-                id,
-                iteration: iteration_count.as_u32(),
+            // For FallbackImmediate, the value always converges immediately (we use the
+            // fallback directly). We still iterate if metadata hasn't converged.
+            // For Fixpoint, ask the recovery function what value to use and check convergence.
+            let value_converged = if C::CYCLE_STRATEGY == CycleRecoveryStrategy::FallbackImmediate {
+                // Use the fallback value instead of the computed value.
+                new_value = C::cycle_initial(db, id, C::id_to_input(zalsa, id));
```
MichaReiser, line 297: I opted for calling `C::cycle_initial` here again because it is easier than `Cloning` (would require a new method on `C` that `panics` for all other recovery strategies or we make `CycleRecoverStrategy` a trait). But I (or someone else ;)) can look into cloning if that's desired or, what would even be better, adding an API that allows us to "take" a value from a `Memo` (which would also allow us to pass an owned value to `recover_from_cycle`)

src/function/fetch.rs
```diff
@@ -191,35 +180,6 @@ where
         database_key_index: DatabaseKeyIndex,
         memo_ingredient_index: MemoIngredientIndex,
     ) -> &'db Memo<'db, C> {
-        // check if there's a provisional value for this query
-        // Note we don't `validate_may_be_provisional` the memo here as we want to reuse an
-        // existing provisional memo if it exists
-        let memo_guard = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index);
```
MichaReiser, line 202: I moved this into the cycle handling branch to help the compiler eliminate it for queries without cycle handling

tests/cycle_fallback_immediate.rs
```diff
@@ -24,18 +22,12 @@ fn simple() {
 
 #[salsa::tracked(cycle_result=two_queries_cycle_result)]
 fn two_queries1(db: &dyn salsa::Database) -> i32 {
-    two_queries2(db);
-    0
+    two_queries2(db) + 1
 }
 
-#[salsa::tracked]
+#[salsa::tracked(cycle_result=two_queries_cycle_result)]
```
MichaReiser, line 28: The test works regardless of whether we add cycle handling here but adding cycle handling is "more correct" because `two_queries2` requires cycle handling if it is called before `two_queries1`

src/function/execute.rs
MichaReiser: Having a way to take the value would also be nice in `fetch_cold_cycle` where we need to call `remove_all_except` (requiring `Atomic`s) to remove the cycle heads from the previous iteration because we can't create a new memo with the same value but slightly different metadata. 

I tried to implement this once but it touches on Salsa parts that I'm not at all familiar with, which is why I ended up going for atomics

carljm (commented):
Looks good to me, thank you.

ChayimFriedman2:
We have reports (https://github.com/rust-lang/rust-analyzer/issues/21610#issuecomment-3887777107) that this PR made rust-analyzer constantly use high CPU and emit Salsa warnings.

Veykril:
Looking into this right now, I think the cause might be the local cancellation support, as that gets turned into propagated panics in fallbacks which I think triggers the poisoning

Veykril:
AH yea I think that is kind of it? From what I can tell, panic poisoning relies on the revision being bumped to be cleared, which does not happen for local cancellation (or in fact if one just calls into garbage collection only), which means we will start throwing this warning.

MichaReiser:
> AH yea I think that is kind of it? From what I can tell, panic poisoning relies on the revision being bumped to be cleared, which does not happen for local cancellation 

Yes. cleaning up intermediate results requires on a revision bump (at least how it's implemented today, maybe someone else is able to come up with something better). 

> which does not happen for local cancellation

Right, didn't we disable local cancellation within cycles?

Veykril:
Well, I think so, but I wonder if there is a bug in there somewhere maybe. Actually, rust-analyzer also calls `trigger_lru_eviction` directly which could cause the issue if it then cancels within a fix point cycle...

Veykril:
https://github.com/salsa-rs/salsa/pull/1085 confirms this

PR #1080, opened 2026-03-30
Detect overlapping supertype variants via leaf type IDs

Veykril:
This can cause hard to debug issues and is never useful so we should recognize this. Ideally at compile time but comparing type ids at compile time is not yet possible

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,7 @@ rustc-hash = "2"
 smallvec = { version = "1", features = ["const_new"] }
 thin-vec = { version = "0.2.14" }
 tracing = { version = "0.1", default-features = false, features = ["std"] }
+typeid = "1.0"
 
 # Automatic ingredient registration.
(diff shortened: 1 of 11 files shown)


MichaReiser:
Could we use https://docs.rs/static_assertions/latest/static_assertions/macro.assert_type_ne_all.html or does it perform structural equality?

Veykril:
We can't do that unfortuantely, we need to transitively check all contained variants which we can't do at macro expansion time. That is

```rust
#[derive(Supertype)]
enum E {
    A(A),
    B(B),
}

#[derive(Supertype)]
enum B {
    A(A)
}
```
would be needed to be detected here. Supertype effectively flattens this variant tree. Thats what we want to guard against here, as that means that salsa will turn a `E(B::A(A))` into an `E::A(A)`, so variants lose their meaning if they overlap

MichaReiser (approved):
Sorry that it took me so long to review. This looks good and it's certainly useful to have this check in place.

PR #1084, opened 2026-04-03
Fix tracked impl db lifetime for as_deref return mode

Veykril:


MichaReiser (approved):

MichaReiser:
Took me a while to understand what this PR is about, but what I understand is that it sets the right lifetime for:

```rust
 - <AsDerefInput as InnerTrait_>::as_deref_method_(this, db)
  + <AsDerefInput as InnerTrait_<'db>>::as_deref_method_(this, db)
```

Veykril:
Right, sorry for the lack of commit message this was a drive by while being busy. But yea, basically the macro needs to emit lifetimes explicitly when invoked in associated item position due to elision rules

PR #1085, opened 2026-04-16
Add tests for cancellation in fixpoints

Veykril:
`revision_bump_clears_poisoned_memo` replicates the issue discussed in https://github.com/salsa-rs/salsa/pull/1063

### DIFF
--- a/tests/parallel/lru_eviction_cancels_cycle.rs
+++ b/tests/parallel/lru_eviction_cancels_cycle.rs
@@ -1,0 +1,240 @@
+// Shuttle doesn't like panics inside of its runtime.
+#![cfg(not(feature = "shuttle"))]
+
+use std::panic::catch_unwind;
+
+use salsa::{Cancelled, Database};
+
+use crate::setup::{Knobs, KnobsDatabase};
+use crate::sync::thread;
+
+#[salsa::input]
+struct Input {
+    value: u32,
+}
+
+/// A query with cycle recovery that will be interrupted by LRU eviction.
+/// This uses `cycle_initial` which gives it `FallbackImmediate` cycle recovery strategy,
+/// meaning it goes through `execute_maybe_iterate` and has `PoisonProvisionalIfPanicking`.
+#[salsa::tracked(cycle_initial = cycle_initial)]
+fn cycle_query(db: &dyn KnobsDatabase, input: Input) -> u32 {
+    // Signal that we've started the cycle query
+    db.signal(1);
+    // Wait for signal that the cancellation flag has been set
+    db.wait_for(2);
+    // This fetch will check for cancellation and throw PendingWrite
+    inner_query(db, input)
+}
+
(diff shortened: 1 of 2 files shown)


tests/parallel/lru_eviction_cancels_cycle.rs
```diff
@@ -0,0 +1,240 @@
...
+    drop(db_waiter);
+
+    // Thread 2: Trigger LRU eviction, which will:
+    // 1. Set the cancellation flag (this signals stage 2, letting thread 1 continue)
+    // 2. Wait for thread 1 to drop its snapshot (thread 1 will check cancellation and panic)
+    // 3. NOT bump the revision
+    let t2 = thread::spawn({
+        let mut db_writer = db_writer;
+        move || {
+            db_writer.trigger_lru_eviction();
```
MichaReiser, line 75: I guess we have the same problem when someone calls `trigger_cancellation`, since it doesn't bump a new revision either.

Veykril: Yea, both of those have the same issue. We don't really need to merge this right now, just put this up so I don't lose it. So yea, unsure what we wanna do here, either we need to remove these functions (which would be a shame, their point is that they do not bump the revision), or we need to figure out how to handle poisoning differently. In r-a we'll avoid calling these for now as the gain from their use is marginal.

MichaReiser:
What makes cycle recovery tricky is that there's no guarantee that queries will be executed in the same order after they've been cancelled. E.g., if we have `a -> b -> a ' and the query gets canceled, there's a chance that the execution plan now is executed `b -> a -> b`. 

A possibility is to walk the entire dependency tree and nuke all memos.

Veykril:
You mean the plan changes the next time someone enters the cycle? What would be the issue with it changing?

MichaReiser:
> You mean the plan changes the next time someone enters the cycle? What would be the issue with it changing?

Yes, because the user calls a different root query. 


The issue with it is that it requires invalidating all provisional memoized values that participated in the query because the iteration count isn't sufficient to distinguish a provisional memoized value that was computed when entering from `a` from one computed when entering from `b` (and the values from iteration 2 could be different dependend on what the root query is)

MichaReiser:
I believe we're now also seeing instances of this in ty (https://github.com/astral-sh/ty/issues/3339#issuecomment-4350549894)

If I remember correctly, why we saw a hang previously was that a query that now should re-execute never did because `validate_same_iteration` always returned true.

https://github.com/MichaReiser/salsa/blob/d1c1ee10a464a7644c5b27bf7350b56fd771997e/src/function/maybe_changed_after.rs#L606-L625

But maybe this is no longer an issue, now that we rewrote cycle handling.

MichaReiser:
I added the tests as part of https://github.com/salsa-rs/salsa/pull/1100

PR #1090, opened 2026-05-03
Do not use `#[allow(non_local_definitions)]`

### LINKED ISSUE #607: `#[tracked]` emitted `accumulated` uses `non_local_definitions`
Meaning rust-analyzer is unable to figure out what `accumulated` is, which is bad if we want to use salsa in our codebase. The trick used here is mainly to get around lack of defsite hygiene, as otherwise a user couldn't create a tracked function named `accumulated`. Given salsa already forbids certain field names i feel like it would be alright to forbid the `accumulated` function name as well. That simplifies a couple of things for the `tracked` macro in general I'd say.

ChayimFriedman2:
It was a bad idea from the beginning, but now rust-analyzer started reporting trait errors, and since it cannot find those impls it report errors on any tracked fn, which is... not fun.

Closes https://github.com/salsa-rs/salsa/issues/607

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -95,13 +95,43 @@ macro_rules! setup_tracked_fn {
         ) -> ::salsa::plumbing::return_mode_ty!(($return_mode, __), $db_lt, $output_ty) {
             use ::salsa::plumbing as $zalsa;
 
+            $zalsa::attach($db, || {
+                let (zalsa, zalsa_local) = $db.zalsas();
+                let result = $zalsa::macro_if! {
+                    if $needs_interner {
+                        {
+                            let key = $fn_name::intern_ingredient_(zalsa).intern_id(zalsa, zalsa_local, ($($input_id),*), |_, data| data);
+                            $fn_name::fn_ingredient_($db, zalsa).fetch($db, zalsa, zalsa_local, key)
+                        }
+                    } else {
+                        {
+                            $fn_name::fn_ingredient_($db, zalsa).fetch($db, zalsa, zalsa_local, $zalsa::AsId::as_id(&($($input_id),*)))
+                        }
+                    }
+                };
+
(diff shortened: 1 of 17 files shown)


tests/cycle.rs
```diff
@@ -170,7 +170,7 @@ where
 
 /// Query minimum value of inputs, with cycle recovery.
 #[salsa::tracked(cycle_fn=cycle_recover, cycle_initial=min_initial)]
-fn min_iterate<'db>(db: &'db dyn Db, inputs: Inputs) -> Value {
+fn min_iterate(db: &dyn Db, inputs: Inputs) -> Value {
```
Veykril, line 173: I take it something now warns on this hence these change?

Veykril (commented):
Finally! I was not aware the diff for this could be this small even, that's great

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -95,13 +95,43 @@ macro_rules! setup_tracked_fn {
...
+                $zalsa::return_mode_expression!(($return_mode, __), $output_ty, result,)
+            })
+        }
+
+        // The module needs be last in the macro expansion in order to make the tracked
+        // function's ident be identified as a function, not a struct, during semantic highlighting.
+        // for more details, see https://github.com/salsa-rs/salsa/pull/612.
+        #[doc(hidden)]
+        #[allow(non_camel_case_types)]
+        $vis struct $fn_name {
```
Veykril, line 122: Drive by nit as I just noticed that, couldn't this be an empty enum since we only use it for the namespacing? 🤔

Veykril (approved):

tests/cycle.rs
ChayimFriedman2: Yes that's Clippy (the second commit), the macro probably confused it before.

ChayimFriedman2:
Can we get a release with this please? The rust-analyzer errors are really jarring.

Veykril:
merged https://github.com/salsa-rs/salsa/pull/1078

PR #1092, opened 2026-05-05
Convert CRLF-terminated sources to LF-terminated

musicinmybrain:
This is just a minor bit of housekeeping.

Before this PR, these four sources are CRLF-terminated (DOS/Windows style), while all of the other sources in the repository are LF-terminated (UNIX-style). After this PR, these four sources are also LF-terminated.

This is simply the result of `dos2unix src/function/eviction.rs tests/parallel/cancellation_token_cycle_nested.rs tests/parallel/cancellation_token_multi_blocked.rs tests/supertype_overlap.rs`.


Veykril (approved):

Veykril:
Thanks!

PR #1095, opened 2026-05-19
Fix cycle head finalized at an unconverged value when its recursive dependency drops mid-fixpoint

RyanJamesStewart:
## Motivation

The fixpoint loop in `execute_maybe_iterate` (`src/function/execute.rs`) has one exit that does not enforce the convergence contract documented in `src/cycle.rs`:

> If the cycle head ever observes that the value returned by `cycle_fn` is the same as the provisional value from the previous iteration, this cycle has converged.

Every exit from the loop checks `C::values_equal(&new_value, last_provisional_value)` before treating a value as final, except the empty-cycle-heads arm. When a query that was mid-fixpoint (`!iteration_count.is_initial()`) computes a value whose cycle-head set is now empty because it conditionally dropped its recursive dependency on a late iteration (the case `cycle.rs` flags via `CycleHead::removed`), the loop breaks and memoizes that freshly computed value with no convergence comparison, and without calling the user `cycle_fn` for that iterate.

The consequence is that a cycle head can be finalized at a value the recurrence never converged to.

## Reproduction

Added `tests/cycle_conditional_drop_unconverged.rs`. `query_x` is a self-cycle head with a monotone bounded recurrence `inner + 1` capped at `cap`, whose unique fixpoint is `V(cap)`. While its recursive self-dependency is live it iterates normally. On a chosen iteration before the recurrence reaches `cap` it stops recursing and returns an off-trajectory value `V(99)`, dropping its self-dependency so its cycle-head set is empty.

On `main` (HEAD `7e77c49`) the test fails:

```
  left: V(99)
 right: V(10)
```

`V(99)` is memoized as the final result even though the recurrence has a well-defined fixpoint `V(10)` it had not reached, the last provisional value was far from `V(99)`, and `cycle_fn` was never consulted about `V(99)`.

## Solution

On the empty-cycle-heads arm, when the query was mid-fixpoint, compare `new_value` against the last provisional value using the same `C::values_equal` predicate every other exit uses. If equal, the cycle has genuinely converged and the loop breaks as before. If not equal, the value is treated as the next provisional and the loop iterates again, so the cycle either reaches a real fixpoint or hits the existing `MAX_ITERATIONS` backstop, which is the documented contract for a non-converging `cycle_fn`. The initial-iteration path (the query was never a cycle head) is unchanged.

This is scoped to the one unguarded exit. It does not change the convergence predicate, the iteration cap, or any other path.

## Tests

The new regression test passes with the change. The full `cycle*` integration suite, the library tests, and the full test suite (with `--features inventory`) all pass with the change applied. `cargo fmt --check` is clean on the modified files.

## Provenance

This came out of a personal project exercising incremental fixpoint evaluation, where the convergence behavior of `cycle_fn` under conditionally dropped recursive dependencies was being studied against the formal least-fixpoint semantics. This is a downstream incremental-computation use case where conditionally dropped recursive dependencies are common, so an unconverged value memoized on this path can be silently consumed downstream rather than surfacing as an error; the attached test reproduces that case directly. The narrow framing here (one unguarded loop exit, a constructed failing test, no change to the convergence predicate) is deliberate: the broader question of whether `values_equal` adjacent-iterate equality is the right convergence test in general is a separate design discussion and is not part of this change.

### DIFF
--- a/tests/cycle_conditional_drop_unconverged.rs
+++ b/tests/cycle_conditional_drop_unconverged.rs
@@ -16,14 +16,28 @@
 //! loop breaks and memoizes that freshly computed value with no convergence
 //! comparison and without calling the user's `cycle_fn` for that iterate.
 //!
+//! On the impurity. The bug is specifically about a recursive dependency that
+//! is dropped on a *particular iterate* by something that is **not part of
+//! the recurrence's fixpoint**. That asymmetry is essential: if the drop
+//! trigger is folded into the cycle (e.g. a second tracked query that reads
+//! the recurrence's provisional value), the mutually-recursive system gains a
+//! genuine self-consistent fixpoint at the dropped value, Salsa correctly
+//! converges to it, and there is no bug to reproduce. A trigger that varies
+//! across iterations of the same query in the same revision *without* being
+//! part of the fixpoint is, by definition, an untracked read. So this test
+//! keeps the iteration-driven drop but declares the impurity the
+//! Salsa-sanctioned way, via `db.report_untracked_read()` (the same idiom
+//! `tests/cycle.rs` uses), rather than hiding it: cache invalidation stays
+//! sound and the construction is honest about what it is.
+//!
 //! Construction. `query_x` is a self-cycle head with a monotone, bounded
-//! recurrence `inner + 1` capped at `cap`, whose unique fixpoint is `cap`.
+//! recurrence `inner + 1` capped at `cap`, whose unique fixpoint is `V(cap)`.
 //! While its recursive self-dependency is live it iterates normally (its
-//! `cycle_fn` is consulted each iteration). On a chosen iteration *before*
-//! the recurrence has reached `cap` it stops recursing and returns an
-//! unrelated value (`V(99)`), dropping its self-dependency so its cycle-head
-//! set becomes empty. On HEAD that iteration exits via the early-break and
-//! `V(99)` is memoized as the final result, even though the recurrence has a
+//! `cycle_fn` is consulted each iteration). On a chosen entry *before* the
+//! recurrence has reached `cap` it stops recursing and returns an unrelated
+//! value (`V(99)`), dropping its self-dependency so its cycle-head set
+//! becomes empty. On HEAD that iterate exits via the early-break and `V(99)`
+//! is memoized as the final result, even though the recurrence has a
 //! well-defined fixpoint `V(cap)` it had not yet reached, the last
 //! provisional value was far from `V(99)`, and the user's `cycle_fn` was
 //! never asked about `V(99)`.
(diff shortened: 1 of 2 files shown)


tests/cycle_conditional_drop_unconverged.rs
```diff
@@ -0,0 +1,109 @@
...
+    cap: u32,
+    /// 1-based entry count of `query_x` on which it drops the recursive
+    /// self-dependency. Chosen below the iterations needed to reach `cap`, so
+    /// the drop happens while the recurrence is provably not yet converged.
+    drop_at: u32,
+}
+
+thread_local! {
+    static X_RUNS: Cell<u32> = const { Cell::new(0) };
+}
```
MichaReiser, line 66: Can you come up with a test that doesn't require global state. It should normally be possible by making the query depend on the result of another query instead. But I suspect that this won't be possible here, because you then wont hit the `cycle_heads.is_empty` branch. 

The reason why this is important is that Salsa fundamentally assume that query functions are pure, and, if they're not, that you call `report_untracked_read` to make this clear. This is important for cache invalidation to work properly.

I know that agents love to write salsa tests that use global state. You may need to hint it a bit towards not doing so.

tests/cycle_conditional_drop_unconverged.rs
```diff
@@ -0,0 +1,109 @@
...
+
+fn x_initial(_db: &dyn Db, _id: salsa::Id, _input: Input) -> V {
+    V(0)
+}
+
+fn x_recover(_db: &dyn Db, _c: &salsa::Cycle, _last: &V, value: V, _i: Input) -> V {
+    // Drive `query_x` toward its own fixpoint normally (identity recovery).
+    value
+}
```
MichaReiser: You should not need this recovery function, I believe this is the default

RyanJamesStewart:
Both points addressed in the latest commit.

Recovery function: removed. It was identity recovery, which is the default, so the attribute now relies on the default.

Global state in the test: your hedge was correct, and it is structural rather than incidental. The bug class is "the empty-cycle-heads exit memoizes a value not equal to the prior provisional". If the drop decision is part of the tracked-query graph (any shape: a single self-query, two queries, N queries), then "drop or not" is a function of the current provisional values, so at any stable point of the joint system it returns one answer. The cases are exhaustive:

- At the fixpoint the answer is "drop": the dropped value is the fixpoint, salsa converges correctly, no bug.
- At the fixpoint the answer is "do not drop": the empty-heads exit is never taken in steady state, so there is nothing to reproduce on that exit.
- No fixpoint exists: MAX_ITERATIONS backstop, a different path.

There is no case where the drop fires for some iterations and then stops, because the decision is a function of current state, not iteration history. The only public surface a tracked query could use to observe progress is current_revision, and that does not tick within a single fixpoint resolution (iteration count never leaves execute_maybe_iterate), so it cannot reintroduce history dependence. A per-iterate trigger that is not part of the fixpoint is therefore, by definition, an untracked read.

So the test keeps the iteration-driven drop but declares the impurity via db.report_untracked_read(), which is the construct the cycle suite already models as supported: the Input::UntrackedRead variant and the self_untracked_panic test. Cache invalidation stays sound and the test is explicit about being an externally-perturbed query, which is the case this code path exists to handle.

Verification:

| ref | result |
|---|---|
| pre-fix parent | fails, V(99) vs V(10) |
| with the fix applied | passes |
| full cycle suite with the fix | unaffected |

cargo fmt --check is clean on the modified file.

MichaReiser:
Thanks. I'm not sure it's worth adding this complexity only to support fixpoint queries with untracked reads. It seems a rather weird case to use an untracked read to gate its own dependency

RyanJamesStewart:
That's fair — I won't push it. The case is narrow enough that I don't think it's worth the added complexity either, given it only arises when an untracked read gates a query's own recursive dependency. Closing this out; thanks for taking the time to look it over.

PR #1099, opened 2026-05-26
fix: accumulated values lost when reused tracked fn skips re-execution

### LINKED ISSUE #923: Accumulator misses accumulated values when behind tracked fn
In 3c230f8 I modified the existing test 'accumulate-from-tracked-fn.rs' to demonstrate. After setting `l1` to 11, the value accumulated by `compute_single` on l0 is missing.

SAY-5:
When a tracked fn's inputs haven't changed, `maybe_changed_after_cold` fast-paths through `deep_verify_memo` and returns the `VerifyResult` directly. That result carries the accumulated flag from the fn's *edge inputs* only; if the fn itself pushed accumulated values, those are ignored. Callers upstream then see `Empty` and skip traversing this subtree, so accumulated values are silently dropped.

Fix: after `deep_verify` succeeds, check whether the memo's own `AccumulatedMap` is non-empty and return `Any` in that case, matching what the re-execute path already does (lines 229-232).

Fixes #923.

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -190,8 +190,23 @@ where
             return Some(if old_memo.revisions.changed_at > revision {
                 VerifyResult::changed()
             } else {
-                // Returns unchanged but propagates the accumulated values
-                deep_verify
+                // Propagate accumulated values from inputs *and* from this memo itself.
+                // Without the own-accumulated check, a caller querying accumulated values
+                // would see `Empty` and skip traversing into this subtree even though
+                // this memo directly pushed accumulated values.
+                VerifyResult::unchanged_with_accumulated(
+                    #[cfg(feature = "accumulator")]
+                    {
+                        let VerifyResult::Unchanged { accumulated } = deep_verify else {
+                            unreachable!()
+                        };
+                        if old_memo.revisions.accumulated().is_some() {
+                            InputAccumulatedValues::Any
+                        } else {
+                            accumulated
+                        }
+                    },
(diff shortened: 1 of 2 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -190,8 +190,23 @@ where
...
+                    #[cfg(feature = "accumulator")]
+                    {
+                        let VerifyResult::Unchanged { accumulated } = deep_verify else {
+                            unreachable!()
+                        };
+                        if old_memo.revisions.accumulated().is_some() {
+                            InputAccumulatedValues::Any
+                        } else {
+                            accumulated
+                        }
```
MichaReiser: Let's avoid the unreachable here
```suggestion
if let VerifyResult::Unchanged { accumulated } = deep_verify {
// Check if the inputs are still valid. We can just compare `changed_at`.
return Some(if old_memo.revisions.changed_at > revision {
VerifyResult::changed()
} else {
                // Propagate accumulated values from inputs *and* from this memo itself.
                // Without the own-accumulated check, a caller querying accumulated values
                // would see `Empty` and skip traversing into this subtree even though
                // this memo directly pushed accumulated values.
                VerifyResult::unchanged_with_accumulated(
                    #[cfg(feature = "accumulator")]
                    {
                        if old_memo.revisions.accumulated().is_some() {
                            InputAccumulatedValues::Any
                        } else {
                            accumulated
                        }
```

MichaReiser (approved):

PR #1101, opened 2026-06-01
Reduce metadata per memo by packing `QueryRevisionsExtra` into `QueryOrigin`

MichaReiser:
## Summary

Store uncommon query revision state indirectly so common memos use eight fewer bytes.

Today, `QueryRevisionExtra` stores a thin pointer to the extra data (`Box<Option<QueryRevisionInner>>`). We use a thin pointer because most memos don't have any extra data.

```
QueryRevisions (32 bytes)
+--------------------------------+
| changed_at                     |
| origin                         |----> Box<[edges]>
| extra                          |----> Box<QueryRevisionsExtraInner>
| durability / flags / padding   |
+--------------------------------+
```

This PR takes this one step further by packing the `extra` data into `origin` (already packed) and renaming the field to `origin_and_extra`. 

```
QueryRevisions (24 bytes)
+--------------------------------+
| changed_at                     |
| origin_and_extra               |
| durability / flags             |
+-------------|------------------+
              |
       no extra: Box<[edges]>

              or

              +----> Box<IndirectOriginAndExtra>
                     +---------------------------+
                     | direct origin ------------|---> Box<[edges]>
                     | QueryRevisionsExtraInner  |
                     +---------------------------+
```

As you can see in the ASCII art, revisions without an extra now only store a single `Box<[edges]>`. The thin pointer to the extra data is entirely removed. 

For memos with extra, Salsa now creates a thin pointer to a `Box<IndirectOriginAndExtra>` which holds a slice to the edges and the extra metadata. 

This change doesn't come for free. Accessing edges for memos with extras now requires one additional dereference. It also introduces a fair bit of code, but it's mostly mechanical. 

We can push this further by avoiding the extra dereference by using a HeaderWithSlice type, see https://github.com/salsa-rs/salsa/pull/1112

### DIFF
--- a/src/zalsa_local.rs
+++ b/src/zalsa_local.rs
@@ -463,7 +463,6 @@ impl std::panic::RefUnwindSafe for ZalsaLocal {}
 
 /// Summarizes "all the inputs that a query used" and "all the outputs it has written to".
 #[derive(Debug)]
-#[cfg_attr(feature = "persistence", derive(serde::Serialize, serde::Deserialize))]
 // #[derive(Clone)] cloning this is expensive, so we don't derive
 pub(crate) struct QueryRevisions {
     /// The most revision in which some input changed.
@@ -472,8 +471,8 @@ pub(crate) struct QueryRevisions {
     /// Minimum durability of the inputs to this query.
     pub(crate) durability: Durability,
 
-    /// How was this query computed?
-    pub(crate) origin: QueryOrigin,
+    /// How this query was computed and any optional revision data.
+    pub(crate) origin_and_extra: OriginAndExtra,
 
     /// [`InputAccumulatedValues::Empty`] if any input read during the query's execution
     /// has any direct or indirect accumulated values.
@@ -481,7 +480,6 @@ pub(crate) struct QueryRevisions {
     /// Note that this field could be in `QueryRevisionsExtra` as it is only relevant
     /// for accumulators, but we get it for free anyways due to padding.
     #[cfg(feature = "accumulator")]
-    #[cfg_attr(feature = "persistence", serde(skip))] // TODO: Support serializing accumulators
     pub(super) accumulated_inputs: AtomicInputAccumulatedValues,
 
     /// Are the `cycle_heads` verified to not be provisional anymore?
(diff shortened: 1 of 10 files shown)


MichaReiser:
This looks interesting from a memory usage perspective and the perf regression is fairly low. 

The gist of it is to pack the `QueryRevisionExtra` Option into `QueryOrigin`'s packed representation. This saves 8 bytes per memo

src/zalsa_local.rs
```diff
@@ -960,18 +943,98 @@ union QueryOriginData {
...
 
-/// SAFETY: The `input_outputs` pointer is owned and not accessed or shared concurrently.
-unsafe impl Send for QueryOriginData {}
-/// SAFETY: Same as above.
-unsafe impl Sync for QueryOriginData {}
+/// SAFETY: The active union field either owns the `input_outputs` allocation or points to an
+/// `IndirectQueryRevisionsData`. Both pointees are `Send`.
+unsafe impl Send for QueryRevisionsDataUnion {}
+/// SAFETY: Same as above, and shared access to the pointee is `Sync`.
+unsafe impl Sync for QueryRevisionsDataUnion {}
```
Veykril: Drive by nit, could we do the following
```suggestion
/// SAFETY: The active union field either owns the `input_outputs` allocation or points to an
/// `IndirectQueryRevisionsData`. Both pointees are `Send`.
unsafe impl Send for QueryRevisionsDataUnion where Id: Send, Box<IndirectQueryRevisionsData>: Send, Box<QueryEdge>: Send {}
/// SAFETY: Same as above, and shared access to the pointee is `Sync`.
unsafe impl Sync for QueryRevisionsDataUnion where Id: Sync, Box<IndirectQueryRevisionsData>: Sync, Box<QueryEdge>: Sync {}
```
That still gives us the compiler checks in a sense (with the exception of changing the union having potential for going out of sync here.

src/zalsa_local.rs
```diff
@@ -912,19 +892,19 @@ enum QueryOriginKind {
     DerivedUntracked = 0b10,
 }
```
Veykril, line 860: I wonder if shuffling these around could make some of the perf numbers look better again (like giving `Derived` the `0` slot as its likely the most common operation).

Veykril (approved):
This sounds good to me!

PR #1103, opened 2026-06-02
Add packed `QueryOrigin`

MichaReiser:
## Summary

This PR introduces a new packed `QueryOrigin` that uses 8 bytes to represent a query edge instead of 12 for as long as all edges

* `ingredient index <= 4095`
* `generation <= 1,048,575`
* it's an output (only used by specify)

Which should be true for most edges. The implementation uses the same 12 byte representation as we use today if any edge has an ingredient index or a generation larger than those limits. 

This PR is entirely written by codex and i mainly want to get some initial feedback on if this is an optimization we want or if we'd prefer https://github.com/salsa-rs/salsa/compare/master...MichaReiser:salsa:micha/query-edge-compact-boxed?expand=1 which implements a similar optimization but where the packing happens at the `QueryEdge` level (edges with larger ingredient index or generations are boxed). 

I'm leaning towards this approach because the characteristics between the packed and non packed remain mostly the same. That is, it doesn't introduce extra allocations. It's just that the allocated query edges is somewhat bigger. 

This reduces ty's memory usage by about 10%

### DIFF
--- a/src/zalsa_local.rs
+++ b/src/zalsa_local.rs
@@ -988,6 +988,8 @@ pub struct QueryOrigin {
     metadata: u32,
 }
 
+const _: [(); std::mem::size_of::<QueryOrigin>()] = [(); 13];
+
 /// The data portion of `PackedQueryOrigin`.
 union QueryOriginData {
     /// Query edges for the derived variants.
@@ -1113,6 +1115,7 @@ impl QueryOrigin {
     /// Create a query origin of type `QueryOriginKind::Assigned`, with the given key.
     pub fn assigned(key: DatabaseKeyIndex) -> QueryOrigin {
         QueryOrigin {
+            // Assigned origins do not store edges, so the layout bit is unused.
             tag: QueryOriginTag::new(QueryOriginKind::Assigned, QueryEdgeLayout::Packed),
             metadata: key.ingredient_index().as_u32(),
             data: QueryOriginData {
@@ -1254,6 +1257,8 @@ pub struct QueryEdge {
     ingredient: IngredientIndex,
 }
 
+const _: [(); std::mem::size_of::<QueryEdge>()] = [(); 12];
+
 impl QueryEdge {
     /// Create an input query edge with the given index.
     pub const fn input(key: DatabaseKeyIndex) -> QueryEdge {
@@ -1268,9 +1273,13 @@ impl QueryEdge {
 
     /// Create an output query edge with the given index.
     pub const fn output(key: DatabaseKeyIndex) -> QueryEdge {
-        let mut edge = Self::input(key);
-        edge.ingredient = edge.ingredient.with_tag(true);
-        edge
+        let id = key.key_index();
+
+        QueryEdge {
+            index: id.index(),
+            generation: id.generation(),
+            ingredient: key.ingredient_index().with_tag(true),
+        }
     }
 
     /// Return the key of this query edge.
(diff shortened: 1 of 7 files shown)


MichaReiser:
Hmm, the performance results are a bit surprising

MichaReiser:
Okay, this looks pretty reasonable

MichaReiser:
CC: @Veykril

MichaReiser:
This version is perf-neutral on ty https://github.com/astral-sh/ruff/pull/25545

AlexWaygood:
> This version is perf-neutral on ty [astral-sh/ruff#25545](https://github.com/astral-sh/ruff/pull/25545)

the codspeed report looks good, but there are still some big timing changes reported by ecosystem-analyzer FYI: https://github.com/astral-sh/ruff/pull/25545#issuecomment-4602018431 (full report at https://61f6224b.ty-ecosystem-ext.pages.dev/timing)

MichaReiser:
> This version is perf-neutral on ty https://github.com/astral-sh/ruff/pull/25545

> the codspeed report looks good, but there are still some big timing changes reported by ecosystem-analyzer FYI: https://github.com/astral-sh/ruff/pull/25545#issuecomment-4602018431 (full report at [61f6224b.ty-ecosystem-ext.pages.dev/timing](https://61f6224b.ty-ecosystem-ext.pages.dev/timing))

I saw that. I'm a bit skeptical on the results. I have to do a closer review but I don't think there's anything in here that justifies a 2x cost

MichaReiser:
Unlike what I thought, this was a real regression in the ecosystem check but only on x86.

src/zalsa_local.rs
```diff
@@ -967,78 +1021,99 @@ unsafe impl Sync for QueryOriginData {}
...
+
+        let (layout, input_outputs) = 'edges: {
+            for edge in input_outputs.by_ref() {
+                let Some(edge) = PackedQueryEdge::new(edge) else {
+                    let input_outputs = packed_input_outputs
+                        .into_iter()
+                        .map(PackedQueryEdge::edge)
+                        .chain(std::iter::once(edge))
+                        .chain(input_outputs)
+                        .collect::<Box<[QueryEdge]>>();
```
Veykril: `chain` does not implement `ExactSizeIterator` (as it could in theory overflow the length), I think that might affect the collect iirc in being less optimal. So it might be better for us to pre-allocated the box here (as a vec) ourselves first and extend into that, then finally turning it into a boxed slice.

src/zalsa_local.rs
```diff
@@ -1162,44 +1245,62 @@ impl std::fmt::Debug for QueryOrigin {
 /// the size of the type. Notably, this type is 12 bytes as opposed to the 16 byte
 /// `QueryEdgeKind`, which is meaningful as inputs and outputs are stored contiguously.
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
-#[cfg_attr(feature = "persistence", derive(serde::Serialize, serde::Deserialize))]
-#[cfg_attr(feature = "persistence", serde(transparent))]
 pub struct QueryEdge {
-    key: DatabaseKeyIndex,
+    // Store the key parts directly rather than as a `DatabaseKeyIndex`. Packed origins
+    // unpack many transient `QueryEdge`s while flattening cycles; keeping the fields
+    // split lets `kind`, `Hash`, and `Eq` operate on the decoded words directly.
```
Veykril: Curious but this makes it sound like `#[repr(transparent)]` could have the same effect no?

Veykril (approved):

MichaReiser:
I plan to clean this up

src/zalsa_local.rs
MichaReiser: The issue is that going from packed to `DatabaseKeyIndex` requires

```
// Clear the tag to restore the original index.
DatabaseKeyIndex::new(self.ingredient.with_tag(false), self.id())
```

and Id is a `NonZero`, also requiring some computations

which turned out to be slow on x86 when called in a hot `Hash` or `Eq`?

PR #1104, opened 2026-06-02
Add packed `QueryEdge`

MichaReiser:
Same as https://github.com/salsa-rs/salsa/pull/1103/changes but at the query edge level


MichaReiser:
This seems slower than the alternative, closing

PR #1108, opened 2026-06-04
Skip dependencies on non-reusable interned values

MichaReiser:
## Summary

Today, Salsa only supports GCing interned values with durability Low (I don't remember the details why). The way we guarantee that interned values are re-created is that the creating query has a read dependency on the created interned value. During validation, the creating query is considered changed and is re-executed if any of the interned value it creates got collected. We don't add dependencies when reading the interned value (we use generations for that). 

Now, we do know that interned values with durability Medium or High can never be garbage collected. That means, it's unnecessary for us to add a read dependency from the creating query, the interned value never needs to be recreated.

This is what this PR implements. For medium and high durability interned values, skip adding the read dependency to the constructing query. 

This saves about 277MB (or 4%) memory when running ty on homeassisstant (https://github.com/astral-sh/ruff/pull/25627). 

## Persistent caching

We may need the edge in the future when we want to run a GC during persistent caching. However, given that persistent caching isn't used today, this seems a worthwhile memory win.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -187,6 +187,11 @@ struct ValueShared {
 impl ValueShared {
     /// Returns `true` if this value slot can be reused when interning, and should be added to the LRU.
     fn is_reusable<C: Configuration>(&self) -> bool {
+        Self::is_reusable_with_durability::<C>(self.durability)
+    }
+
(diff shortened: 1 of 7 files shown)


ibraheemdev (approved):
Very nice!

PR #1110, opened 2026-06-07
LRU benches

Veykril:
I am planning on replacing the LRU implementation we have today with something more scan resistant.

### DIFF
--- a/benches/lru.rs
+++ b/benches/lru.rs
@@ -1,0 +1,156 @@
+use std::hint::black_box;
+
+use codspeed_criterion_compat::{
+    BatchSize, BenchmarkId, Criterion, criterion_group, criterion_main,
+};
+use salsa::Database as _;
+
+const HOT_ITEMS: usize = 1_024;
+const EVICTION_ITEMS: usize = 4_096;
+const EVICTION_CAPACITY: usize = 512;
+const SCATTERED_HOT_ITEMS: usize = EVICTION_CAPACITY * 2;
+
+#[salsa::input]
+struct Item {
+    value: usize,
+}
+
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone)]
+struct Value(usize);
+
+impl Drop for Value {
+    fn drop(&mut self) {
+        black_box(self);
+    }
+}
+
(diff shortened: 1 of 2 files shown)


benches/lru.rs
```diff
@@ -0,0 +1,123 @@
...
+    group.bench_function(
+        BenchmarkId::new("working_set_over_capacity", EVICTION_ITEMS),
+        |b| {
+            b.iter_batched_ref(
+                || {
+                    let mut db = salsa::DatabaseImpl::new();
+                    lru_value::set_lru_capacity(&mut db, EVICTION_CAPACITY);
+                    let items = new_items(&db, EVICTION_ITEMS);
+                    let expected = prewarm(&db, &items);
+                    let hot_items = &items[EVICTION_ITEMS - EVICTION_CAPACITY..];
```
MichaReiser, line 113: This selection seems optimal for our current LRU design where the most recently accessed items survive, even if they weren't necessarily the most used items within a revision (e.g. in a type checker, the ASTs of the last checked files survive over files, even if they're entirely insubstantial to the program because they're never imported). 

Should this benchmark instead select `hot_items` that exceed `EVICTION_CAPACITY` and more evenly split them across  EVICTION_ITEMS?

MichaReiser (approved):
Nice. It's funny, I also looked into the function LRU recently. Specifically, the central lock is a huge problem for ty. Enabling LRU for a hot query easily results in a 43% walltime reduction. It may make sense to also add a concurrent hot benchmark that covers this as well. 

The approaches I explored and compared perf on ty are:

* Shard the lock (no global LRU)
* Allow later promotions (after the initial execution) optimistic. Meaning, it uses `try_lock` over `lock`. This can mean that we fail to promote a later read if it allows us to avoid waiting for a lock.
* I also experimented with a revision based LRU. It didn't perform as well, but that might also have been because of a poor implementation

benches/lru.rs
Veykril: I've added another bench for that

Veykril:
I'll merge this since you are showing interest in changing the lru here so it seems useful even if the benches are somewhat primitive right now

PR #1111, opened 2026-06-08
Revamp tracked attribute for methods and impls to better handle lifetimes

Veykril:


components/salsa-macros/src/tracked_impl.rs
```diff
@@ -114,11 +104,60 @@ impl Macro {
...
+            .into_iter()
+            .map(|ty| ChangeLt::all_to(&body_db_lt).in_type(&ty))
+            .collect();
+        let output_ty = ChangeLt::all_to(&body_db_lt).in_type(&output_ty);
+        let skipped_inputs = if self_token.is_some() { 2 } else { 1 };
+
+        if args.self_ty.is_none() {
+            args.self_ty = Some(body_self_ty.clone());
+        }
+        if self.return_mode_uses_db_lifetime(&args) && args.non_update_types.is_none() {
```
MichaReiser: Is it intentional that we always add `non_update_types` here when the return mode uses a db lifetime? Doesn't this now skip `Update` checks for returned references and such?

MichaReiser (commented):
Codex found a few issues with this

**Findings**
- **[P1] Do not silently enable `unsafe(non_update_types)`.** `/Users/micha/.codex/worktrees/9e79/salsa/components/salsa-macros/src/tracked_impl.rs:121` automatically disables the `Update` check for every `ref`/`deref` return mode. That option is explicitly unsafe because lifetime-bearing outputs may contain database references that become dangling across revisions. The source method should require an explicit `unsafe(non_update_types)` acknowledgment rather than having the generated implementation assume its contract.
- **[P1] `ChangeLt::all_to` also rewrites unrelated lifetimes.** `/Users/micha/.codex/worktrees/9e79/salsa/components/salsa-macros/src/tracked_impl.rs:108` and `/Users/micha/.codex/worktrees/9e79/salsa/components/salsa-macros/src/tracked_impl.rs:469` replace *every* lifetime, including `'static` and higher-ranked bound lifetimes. I confirmed a regression: this compiles on the base but fails on the PR because the generated return becomes `&'db str` and cannot satisfy the public `&'static str` signature:
  ```rust
  #[salsa::tracked(returns(copy), unsafe(non_update_types))]
  fn value(db: &dyn salsa::Database) -> &'static str {
      "test"
  }
  ```
  The transformation should replace only the identified database lifetime and elided `'_`, not arbitrary lifetimes. `ChangeLt` likely needs a targeted `from -> to` operation rather than `from: None`.
- **[P2] The macro suppresses legitimate user warnings.** `/Users/micha/.codex/worktrees/9e79/salsa/components/salsa-macros/src/tracked_impl.rs:134` adds `#[allow(unused_variables)]` to the cloned function containing the user’s original body. Consequently, unused database or input parameters stop producing warnings. Any allowance should be attached only to generated scaffolding, not the function carrying user-authored code.

**Validation**
- PR head reviewed: `e55ba299`.
- `cargo test --test compile_pass --all-features` passes.
- The base-vs-PR `'static` regression was reproduced directly.
- Selected runtime tests under `--all-features` hit the existing Shuttle-outside-runtime abort, unrelated to this PR.
- Worktree restored to its original `f8162cf7` state.

components/salsa-macros/src/tracked_impl.rs
Veykril: No, that was unintentional

Veykril:
Alright, added a couple more broken cases and fixed those as well after addressing the review.

MichaReiser:
Hmm, this change seems tricky:

```rust
#[derive(Copy, Clone, Debug, PartialEq, Eq, salsa::Update)]
struct Things<'db>(std::marker::PhantomData<&'db ()>);

#[salsa::interned]
struct Other {
    value: bool,
}
```


1. Lifetime in function body

```rust
#[salsa::tracked]
impl<'a> Things<'a> {
    #[salsa::tracked]
    fn body_annotation(db: &'a dyn salsa::Database) -> u32 {
        let _: &'a dyn salsa::Database = db;
        0
    }
}
```

The PR changes the generated signature from `'a` to `'db`, but does not change the annotation inside the body. 

Maybe it's best if we don't rename the lifetime variable?

2. Lifetime conflict

Similar issue here:

```rust
#[salsa::tracked]
impl<'a> Things<'a> {
    #[salsa::tracked(unsafe(non_update_types))]
    fn hrtb(
        db: &'a dyn salsa::Database,
    ) -> for<'db> fn(&'a (), &'db ()) {
        _ = db;
        unimplemented!()
    }
}
```

The transformation replaces the outer `'a` with its hardcoded target `'db`. Inside `for<'db>`, that target is already bound, so the transformed type captures the wrong lifetime:

```rust
// Intended
for<'db> fn(&'a (), &'db ())

// Effectively generated
for<'db> fn(&'db (), &'db ())
```



3. Elided lifetimes

I don't fully understand this Codex finding:

```rust
#[salsa::tracked]
impl<'db> Things<'db> {
    #[salsa::tracked]
    fn named_return(
        db: &'db dyn salsa::Database,
        other: Other<'_>,
    ) -> Things<'db> {
        _ = (db, other);
        unimplemented!()
    }
}
```

The important detail is that `Other<'_>` introduces an anonymous lifetime independent of the impl’s `'db`. Conceptually, the public signature behaves like:


```rust
fn named_return<'other>(
    db: &'db dyn salsa::Database,
    other: Other<'other>,
) -> Things<'db>
```

Expansion:

```rust
impl<'db> Things<'db> {
    fn named_return(
        db: &'db dyn salsa::Database,
        other: Other<'_>, // Still an independent lifetime.
    ) -> Things<'db> {
        trait InnerTrait<'db> {
            fn inner(
                db: &'db dyn salsa::Database,
                other: Other<'db>, // Changed from '_ to 'db.
            ) -> Things<'db>;
        }

        impl<'db> InnerTrait<'db> for Things<'db> {
            fn inner(
                db: &'db dyn salsa::Database,
                other: Other<'db>,
            ) -> Things<'db> {
                _ = (db, other);
                unimplemented!()
            }
        }

        #[salsa::tracked(self_ty = Things<'db>)]
        fn inner<'db>(
            db: &'db dyn salsa::Database,
            other: Other<'db>,
        ) -> Things<'db> {
            <Things<'db> as InnerTrait<'db>>::inner(db, other)
        }

        // `other` is Other<'other>, but `inner` requires Other<'db>.
        inner(db, other)
    }
}
```

The internal input becomes `Other<'db>`, But the outer signature is only normalized when [tracked_impl.rs (line 207)](/Users/micha/code/.worktrees/fbf7/salsa/components/salsa-macros/src/tracked_impl.rs:207) sees a lifetime-dependent return mode or an elided lifetime in the return type:

Veykril:
> Maybe it's best if we don't rename the lifetime variable?

Yea good point, we should just figure out what the lifetime name is of the db (or conjure one that doesn't conflict if its elided)


The codex finding im not sure about whether thats quite an issue? You can't have lifetime dependent inputs outputs other than the database lifetime. Exception being higher ranked trait objects and function pointers, that one we can address I guess

Veykril:
Appreciate the thorougness, I started this with thinking it was gonna be a simple fix but its clearly spiraling into fixing up the entire macro to work correctly on assoc functions 😅

MichaReiser (approved):
Codex found a minor regression. I'm not convinced this is worth fixing:

```rust
use std::{marker::PhantomData, sync::Arc};

#[derive(Copy, Clone, Debug, PartialEq, Eq, salsa::Update)]
struct Things<'db> {
    marker: PhantomData<&'db ()>,
}

trait Marker<'a>: Send + Sync {}

struct MarkerImpl;

impl<'a> Marker<'a> for MarkerImpl {}

#[salsa::tracked]
impl Things<'_> {
    #[salsa::tracked(no_eq, unsafe(non_update_types))]
    fn query(
        db: &dyn salsa::Database,
    ) -> Arc<dyn Marker<'_>> {
        _ = db;
        Arc::new(MarkerImpl)
    }
}

fn main() {}
```

```
error[E0637]: `'_` cannot be used here
  --> ... -> Arc<dyn Marker<'_>>
                              ^^ `'_` is a reserved lifetime name
```


The issue is that `_` is not replaced.


Codex came up with this diff which reduces line count and it also claims to fix the issue above (more robust over all?)


```diff
 pub(crate) struct ChangeLt {
     to: syn::Lifetime,
-    /// Depth of higher-ranked binders we are currently nested within. While this
-    /// is non-zero, lifetimes must not be rewritten.
-    binder_depth: usize,
 }

 impl ChangeLt {
     pub fn elided_to(db_lt: &syn::Lifetime) -> Self {
-        ChangeLt {
-            to: db_lt.clone(),
-            binder_depth: 0,
-        }
+        ChangeLt { to: db_lt.clone() }
     }

 impl syn::visit_mut::VisitMut for ChangeLt {
     fn visit_lifetime_mut(&mut self, i: &mut syn::Lifetime) {
-        if self.binder_depth == 0 && i.ident == "_" {
+        if i.ident == "_" {
             *i = self.to.clone();
         }
     }

-    fn visit_type_bare_fn_mut(&mut self, i: &mut syn::TypeBareFn) {
-        self.binder_depth += 1;
-        syn::visit_mut::visit_type_bare_fn_mut(self, i);
-        self.binder_depth -= 1;
-    }
+    // Function-pointer lifetimes are independently bound.
+    fn visit_type_bare_fn_mut(&mut self, _: &mut syn::TypeBareFn) {}

     fn visit_trait_bound_mut(&mut self, i: &mut syn::TraitBound) {
-        self.binder_depth += 1;
-        syn::visit_mut::visit_trait_bound_mut(self, i);
-        self.binder_depth -= 1;
+        // Only `for<...>` introduces a higher-ranked binder.
+        if i.lifetimes.is_none() {
+            syn::visit_mut::visit_trait_bound_mut(self, i);
+        }
     }
 }

 pub(crate) fn uses_elided_lifetime(ty: &syn::Type) -> bool {
-    let mut finder = ElidedLifetimeFinder {
-        found: false,
-        binder_depth: 0,
-    };
+    let mut finder = ElidedLifetimeFinder { found: false };
     finder.visit_type(ty);
     finder.found
 }

 struct ElidedLifetimeFinder {
     found: bool,
-    binder_depth: usize,
 }

 impl<'ast> syn::visit::Visit<'ast> for ElidedLifetimeFinder {
     fn visit_lifetime(&mut self, i: &'ast syn::Lifetime) {
-        if self.binder_depth == 0 && i.ident == "_" {
+        if i.ident == "_" {
             self.found = true;
         }
     }

-    fn visit_type_bare_fn(&mut self, i: &'ast syn::TypeBareFn) {
-        self.binder_depth += 1;
-        syn::visit::visit_type_bare_fn(self, i);
-        self.binder_depth -= 1;
-    }
+    fn visit_type_bare_fn(&mut self, _: &'ast syn::TypeBareFn) {}

     fn visit_trait_bound(&mut self, i: &'ast syn::TraitBound) {
-        self.binder_depth += 1;
-        syn::visit::visit_trait_bound(self, i);
-        self.binder_depth -= 1;
+        if i.lifetimes.is_none() {
+            syn::visit::visit_trait_bound(self, i);
+        }
     }
 }
```

Veykril:
I re-used your (well, Codex)'s simplification there. I also added a triomphe feature flag for the update impls as r-a heavily uses that over std Arc. With this patch all impl uses of tracked now compile in https://github.com/rust-lang/rust-analyzer/pull/22548

PR #1113, opened 2026-06-09
[experiment] Fewer indices and generations

MichaReiser:
Shrink `DatabaseKeyIndex` to 8 bytes by limiting the size of ingredient index and generation to:

* Ingredient Index: 0..=2,047. ty only uses ~240 ingredients
* generations: 0..=1,048,575. It takes 1.21 days before we reach the limit if the same interned value is GCed 10times per second. In which case we leak the interned slot (not the value)

These values seem like plenty and maybe it's not worth worrying about larger ingredient indices and higher generations until we hit a use case where it matters? 

The advantage of reducing `DatabaseKeyIndex` over my `PackedQueryEdges` is that it reduces memory while adding dependencies to active queries, but also everywhere else where `DatabaseKeyIndex` is stored (accumulated values, cycle heads, ...)

@Veykril do you know how many ingredients r-a has? Do you feel concerned about those limits?

The memory usage improvements for ty are minimal compared to https://github.com/salsa-rs/salsa/pull/1101 (base: https://github.com/astral-sh/ruff/pull/25773, this PR https://github.com/astral-sh/ruff/pull/25790). Performance wins are also minimal. So maybe this is not worth it and my packed query edges is a pretty decent compromise

### DIFF
--- a/src/id.rs
+++ b/src/id.rs
@@ -26,6 +26,7 @@ pub struct Id {
 impl Id {
     pub const MAX_U32: u32 = u32::MAX - 0xFF;
     pub const MAX_USIZE: usize = Self::MAX_U32 as usize;
+    pub const MAX_GENERATION: u32 = 0xF_FFFF;
 
     /// Create a `salsa::Id` from a u32 value, without a generation. This
     /// value should be less than [`Self::MAX_U32`].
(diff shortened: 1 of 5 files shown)


MichaReiser:
I think what we have is a perfectly fine compromise

PR #1124, opened 2026-06-10
Remove FLATTEN_MAPS thread local, store it in ZalsaLocal instead

Veykril:


MichaReiser:
Can you say more what the benefits of this change are? Is it that the memory gets released when dropping the ZalsaLocal instance over retaining it forever? 

I'd have to measure this change on ty but I'm a little concerned about a typical multithreaded use case where we create a new rayon task for each file, and each task uses its own Db clone (its own `ZalsaLocal`)

Veykril:
I generally favor having less thread locals, given we already have *the* thread local state stashed away in ZalsaLocal. Generally I wanted to check the effect of the benching numbers here. You do bring up a good point regarding the fact that ZalsaLocal's tend to get dropped and re-created frequently though (sounds like another thing we might want to have a benchmark for in fact).

Gonna put this in draft again until I can give that some more thought.

Issue #1032, opened 2025-12-03
Enabling LRU for tracked functions leads to lock-congestion


MichaReiser:
We tried to enable LRU for some tracked functions in [ty](https://github.com/astral-sh/ruff/pull/21749#issuecomment-3605760158) but it resulted in a 25% performance reduction when running ty multi-threaded. The root cause is that the LRU linked list uses a single `Mutex`, which becomes a bottleneck for very hot queries, because all threads end up blocking on it.

We should explore whether we can shard the LRU lock in a similar way to how we do for interned structs. This should dramatically reduce lock congestion.

PR #1133, opened 2026-06-13
Shard tracked-function LRU locks

charliermarsh:
## Summary

Tracked functions with LRU enabled currently serialize every cache hit through one mutex. Hot queries can therefore spend more time contending on recency bookkeeping than executing, especially when many worker threads repeatedly access different keys.

This shards recency tracking across up to 32 independently locked linked sets. Shards are allocated only when LRU is enabled, their count is bounded by the configured capacity, and capacity is divided across them during eviction. Each shard also remembers its most recently used key so consecutive accesses can avoid locking entirely.

The policy remains exact within each shard but becomes an approximation of a single global LRU across shards. Runtime capacity changes redistribute the tracked keys, and disabling LRU drops the shard storage.

In Ruff's multi-threaded Home Assistant type-checking workload, this reduced CPU time by 4.8% across 20 paired runs, winning all 20. A follow-up run after the final dynamic-allocation cleanup reduced CPU time by 6.0% across 10 pairs. Ruff diagnostics were byte-identical.

The downstream [Ruff benchmark draft](https://github.com/astral-sh/ruff/pull/25949) reports a 4.84% overall CodSpeed improvement with no regressions. Its significant changes are memory improvements of 4.1% to 6.1%, and Ruff's project memory report shows total memory reductions of 3.7% to 9.5% across flake8, trio, sphinx, and prefect.

This addresses the lock congestion described in #1032.

### DIFF
--- a/src/function/eviction/lru.rs
+++ b/src/function/eviction/lru.rs
@@ -8,23 +8,55 @@ use std::num::NonZeroUsize;
 use crate::Id;
 use crate::hash::FxLinkedHashSet;
 use crate::sync::Mutex;
+use crate::sync::atomic::{AtomicU64, Ordering};
 
 use super::{EvictionPolicy, HasCapacity};
+
+const MAX_SHARD_COUNT: usize = 32;
 
 /// Least Recently Used eviction policy.
 ///
 /// When the number of memoized values exceeds the configured capacity,
 /// the least recently accessed values are evicted at the start of each
 /// new revision.
+///
+/// Recency is tracked independently across shards. This makes the policy an
+/// approximation of a global LRU, but prevents unrelated keys from contending
+/// on a single lock during parallel query execution.
 pub struct Lru {
     capacity: Option<NonZeroUsize>,
+    shards: Box<[LruShard]>,
+}
+
+struct LruShard {
+    last_used: AtomicU64,
     set: Mutex<FxLinkedHashSet<Id>>,
+}
+
+impl Default for LruShard {
+    fn default() -> Self {
+        Self {
+            last_used: AtomicU64::new(u64::MAX),
+            set: Mutex::default(),
+        }
+    }
 }
 
 impl Lru {
     #[inline(never)]
     fn insert(&self, id: Id) {
-        self.set.lock().insert(id);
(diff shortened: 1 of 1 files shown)


MichaReiser:
I have a local draft for this as well. I think we also need to align with @Veykril on the design because, as far as I understand, this changes from a global LRU to a per-shard LRU. 

I also found that what already helps a lot (without changing the design) is to use `try_lock` when reading an already cached value. This means we may fail to promote a value, but achieves much higher throughput

Veykril:
I am fine with changing whatever here right now fwiw. I was looking into trying out a different approach entirely, but changing the current setup doesn't really block that experimentation. I do think we want to land some benches before either way though (havent gotten to cleaning up my PR there yet). Generally I think we want to trade "perfect LRU" tracking away in favor of performance improvements though, that is approximate LRI instead if it yields better throughput.

MichaReiser:
>  Generally I think we want to trade "perfect LRU" tracking away in favor of performance improvements though, that is approximate LRI instead if it yields better throughput.

Agree, which also seemed to work fine with interned values.

src/function/eviction/lru.rs
```diff
@@ -8,59 +8,165 @@ use std::num::NonZeroUsize;
...
+            set: Mutex::default(),
+        }
+    }
+}
+
 impl Lru {
     #[inline(never)]
     fn insert(&self, id: Id) {
-        self.set.lock().insert(id);
+        let shard = &self.shards[id.index() as usize % self.shards.len()];
```
MichaReiser, line 48: Can we use the same hashing as we use in interned value here. That is. can we hash to get a more even distribution

Veykril:
Also I believe nothing really stops us from allowing different LRU schemes at some point if there is going to be value in that (I kind of doubt it, but you never know)

charliermarsh:
Gah, sorry, this was opened by Codex as part of a broader run. (I allowed it to open drafts in Ruff to get timings from CI, but didn’t intend for it to open PRs elsewhere.) Really sorry, bad form!

MichaReiser:
It's okay. It helped get some consensus real quick :)

charliermarsh:
That's what I'm here for!

PR #1134, opened 2026-06-14
chore: use GitHub runners for CodSpeed benchmarks

MichaReiser:
Use GitHub-hosted runners for CodSpeed benchmarks while keeping the existing simulation and memory modes.

We ran out of macro runner credits and I didn't get the impression that the macro runners helped to reduce flakiness.


MichaReiser:
We can always switch back if the benchmarks remain too noisy (but we may need to figure out how to stay within budget or find a sponsor)

PR #1137, opened 2026-06-15
feat: use generational tracked function eviction

MichaReiser:
Replace capacity-driven tracked-function LRU bookkeeping with an off-memo generational collector. First-use values receive 16 revisions of compact probation; reused values promote through warm, hot, and tenured tiers, with revision-driven aging and incremental due-cohort processing.
Home Assistant with semantic-index LRU enabled remains effectively flat against the prior scheduler: 3.627 ± 0.146 s versus 3.613 ± 0.102 s cold, approximately 3.842 GB RSS for both, and 5.079 s versus 4.999 s over 65 edits; main was 4.177 ± 0.283 s cold.
Testing: Added collector coverage, ran the full default Salsa suite and all-feature checks, and benchmarked Home Assistant wall time and RSS.

### DIFF
--- a/book/src/tuning.md
+++ b/book/src/tuning.md
@@ -1,10 +1,9 @@
 # Tuning Salsa
 
-## Cache Eviction (LRU)
+## Cache Eviction
 
-Salsa supports Least Recently Used (LRU) cache eviction for tracked functions.
-By default, memoized values are never evicted (unbounded cache). You can enable
-LRU eviction by specifying a capacity at compile time:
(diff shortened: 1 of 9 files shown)


MichaReiser:
I need to update the benchmarks to cover both execution AND collection for them to be meaningful here

MichaReiser:
This looks fairly promising. I have to take a closer look with what codex came up here. It's also nice that the upper bound is no longer an absolute limit. It only influences when GC runs.

PR #1142, opened 2026-06-17
refactor: Introduce `MemoHeader` to reduce monomorphization

MichaReiser:
Moves configuration-independent memo validation and cycle metadata into `MemoHeader` to avoid monomorphization. 

This does not change the size of `Memo` because `MemoHeader` has no padding (it's always 32 bytes).

**Release Binary size**

Metric | Base | Branch | Difference
-- | -- | -- | --
Stripped binary | 23,189,920 B | 22,430,096 B | −759,824 B / −3.28%
Unstripped binary | 26,479,232 B | 25,674,128 B | −805,104 B / −3.04%
Mach-O __text | 16,110,760 B | 15,403,432 B | −707,328 B / −4.39%

**LLVM IR**

Scope | Base | Branch | Difference
-- | -- | -- | --
ty_python_semantic | 12,484,695 lines | 11,096,129 lines | −1,388,566 / −11.12%
Entire ty build | 35,139,488 lines | 33,380,017 lines | −1,759,471 / −5.01%

Testing: Passed local validation and GitHub CI.

### DIFF
--- a/src/function/fetch.rs
+++ b/src/function/fetch.rs
@@ -135,14 +135,7 @@ where
             if old_memo.value.is_some()
                 && old_memo
                     .header
-                    .maybe_changed_after_cold(
-                        db.into(),
-                        &claim_guard,
-                        zalsa.current_revision(),
-                        C::CYCLE_STRATEGY,
-                        true,
-                    )
-                    .is_some_and(|result| result.is_unchanged())
+                    .verify_memo(db.into(), &claim_guard, C::CYCLE_STRATEGY, true)
             {
                 // SAFETY: memo is present in memo_map and we have verified that it is
                 // still valid for the current revision.
--- a/src/function/fetch.rs
+++ b/src/function/fetch.rs
@@ -137,9 +137,7 @@ where
                     .header
                     .maybe_changed_after_cold(
                         db.into(),
-                        zalsa,
-                        zalsa_local,
-                        database_key_index,
+                        &claim_guard,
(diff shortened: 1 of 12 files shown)


src/function/fetch.rs
```diff
@@ -128,31 +132,21 @@ where
...
-                if can_shallow_update.yes()
-                    && self.validate_may_be_provisional(
-                        zalsa,
-                        zalsa_local,
-                        database_key_index,
-                        old_memo,
+            if old_memo.value.is_some()
+                && old_memo
+                    .header
+                    .maybe_changed_after_cold(
```
ibraheemdev: It looks like `maybe_change_after_cold` does an additional check that `self.revisions.changed_at > revision` to return `unchanged`, not sure if this is consequential.

ibraheemdev (approved):
A further opportunity here is to mark `Memo` as `repr(C)`, allowing you to safely treat a `*const Memo<C>` as `*const MemoHeader`, and [avoid the monomorphization entirely in some of the helper methods, such as `provisional_status`](https://github.com/ibraheemdev/salsa/blob/9a7aed6be3ee48ec0f05b6360f0e7914b1d0209f/src/function.rs#L559), which still rely on looking up the concrete memo type in this PR. I was able to get much larger wins on that branch by extracting some of our query execution logic into state machines as well, but I think this PR is a good first step, and it makes sense to get most of the boilerplate out of the way before making more targeted changes.

PR #1146, opened 2026-06-18
Fix default interned GC configuration

ibraheemdev:
It looks like this has been incorrect since it was introduced, meaning the default interned GC configuration is very aggressive. Note that the interned tests hardcode `revisions = 3` so this doesn't affect any of the snapshots, the `cfg(test)` override was mostly meant to make it more likely to catch bugs for users of Salsa.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -37,10 +37,10 @@ pub trait Configuration: Sized + 'static {
     const PERSIST: bool;
 
     // The minimum number of revisions that must pass before a stale value is garbage collected.
-    #[cfg(test)]
+    #[cfg(not(test))]
     const REVISIONS: NonZeroUsize = NonZeroUsize::new(3).unwrap();
 
-    #[cfg(not(test))] // More aggressive garbage collection by default when testing.
+    #[cfg(test)] // More aggressive garbage collection by default when testing.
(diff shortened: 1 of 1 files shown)


ibraheemdev:
cc @MichaReiser it would be interesting to see if this affects our LSP benchmarks/memory usage.

Veykril (approved):

MichaReiser:
I asked codex to do some instrumentation on ty. It's conclusion is that this benefits ty


| Metric (mean of 2 runs) | Current default (1 revision) | This PR (3 revisions) | Change |
|---|---:|---:|---:|
| Cold-check time | 45.16s | 45.25s | +0.2% (noise) |
| Cold-check CPU | 48.59s | 48.50s | −0.2% (noise) |
| Nine edits: total check time | 35.37s | 32.66s | **−7.7%** |
| Nine edits: total CPU | 36.17s | 33.63s | **−7.0%** |
| Mean CPU per edit | 4.02s | 3.74s | **−7.0%** |
| Peak RSS | 4,149.5 MiB | 4,149.3 MiB | effectively unchanged |
| RSS after restoring the file | 3,557.1 MiB | 3,569.3 MiB | +12.2 MiB (+0.34%) |

PR #1158, opened 2026-06-20
fix: release tracked struct lock after panic

MichaReiser:
A panic while updating tracked fields could leave the struct permanently write-locked. Restore its previous update revision during unwinding so a later execution can retry.

Testing: Added a regression test and ran the related tracked-struct tests.

### DIFF
--- a/src/tracked_struct.rs
+++ b/src/tracked_struct.rs
@@ -428,30 +428,6 @@ where
     memos: MemoTable,
 }
 
-#[must_use]
-struct WriteGuard<'a> {
-    updated_at: &'a OptionalAtomicRevision,
-    restore_to: Option<Revision>,
-    id: Id,
-}
-
-impl WriteGuard<'_> {
-    fn release(mut self, revision: Revision) {
-        self.restore_to = Some(revision);
-    }
-}
-
-impl Drop for WriteGuard<'_> {
-    fn drop(&mut self) {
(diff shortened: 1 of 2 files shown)


Veykril (approved):

src/tracked_struct.rs
MichaReiser, line 740: Hnmm, there's an other issue here. `update_fields` could succeed for the first field but then panic. We would need to poison the tracked struct. I'm not sure this is worth it. But arguably, what this PR produces is not strictly better than always remaining locked.

MichaReiser:
I don't know if the complexity is justified here

Veykril (approved):

Veykril:
Do we reap the poisoned tracked structs in `drain` later on?

MichaReiser:
I think I'm just gonna require that `Update` is not allowed to panic

Veykril:
I am fine with that as well

MichaReiser:
Unfortunately, that's not sufficient because of our `PartialEq` fallback. 

I think I'm leaving it as is. This does not trigger UB. Salsa will crash on the next run with an inaccurate panic, but that's it. There are other cases where a panic in user-code can lead to inconsistent result. E.g. a query that called `specify` but later panicked can leave around stale results. Happy to fix if someone actually runs into this and has a use case for Salsa handling it more gracefully :)

PR #1160, opened 2026-06-20
fix: make interned slot reuse unwind-safe

MichaReiser:
A panic while assembling a reused interned value could orphan its slot after removing the old key and LRU entry. Run user-controlled work before mutation and finish the structural commit before callbacks or drops can unwind.

Testing: Added a regression test and ran the interned-revision, interned-struct, and LRU tests.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -512,14 +512,21 @@ where
                 continue;
             };
 
-            // Mark the slot as reused.
-            *value_shared = ValueShared {
-                id: new_id,
-                durability,
-                last_interned_at,
-            };
+            // Assemble and hash the replacement before mutating the existing slot. Both operations
+            // can invoke user code and panic.
+            // SAFETY: We call `from_internal_data` to restore the correct lifetime before access.
+            let new_fields = unsafe { self.to_internal_data(assemble(new_id, key)) };
 
-            let index = self.database_key_index(value_shared.id);
+            // SAFETY: We hold the lock for the shard containing the value.
+            let old_hash = self.hasher.hash_one(unsafe { &*value.fields.get() });
+
+            // Ensure inserting the replacement cannot invoke user hashing after the slot has been
+            // modified.
+            // SAFETY: We hold the lock for the shard containing every value passed to `hasher`.
+            let hasher = |id: &_| unsafe { self.value_hash(*id, zalsa) };
+            shard.key_map.reserve(1, hasher);
+
+            let index = self.database_key_index(new_id);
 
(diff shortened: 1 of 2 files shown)


src/interned.rs
```diff
@@ -512,14 +512,21 @@ where
...
+            let new_fields = unsafe { self.to_internal_data(assemble(new_id, key)) };
+
+            // SAFETY: We hold the lock for the shard containing the value.
+            let old_hash = self.hasher.hash_one(unsafe { &*value.fields.get() });
+
+            // Ensure inserting the replacement cannot invoke user hashing after the slot has been
+            // modified.
+            // SAFETY: We hold the lock for the shard containing every value passed to `hasher`.
+            let hasher = |id: &_| unsafe { self.value_hash(*id, zalsa) };
+            shard.key_map.reserve(1, hasher);
```
ibraheemdev: hashbrown internally calls `reserve` before inserting the new value in `insert_unique`, though I guess this is technically not guaranteed. That said, I would prefer if we just avoid keep the `old_fields` mutable reference live during insertion, instead of adding this extra call.

src/interned.rs
```diff
@@ -528,51 +535,54 @@ where
...
-            *old_fields = unsafe { self.to_internal_data(assemble(new_id, key)) };
-
-            // SAFETY: We hold the lock for the shard containing the value.
-            let hasher = |id: &_| unsafe { self.value_hash(*id, zalsa) };
+            // SAFETY: We hold the lock for the shard containing the value, and the
+            // value has not been interned in the current revision, so no references to
+            // it can exist. This mutable reference must be created after `reserve`, whose
+            // hasher can read the fields of this value.
+            let old_fields = unsafe { &mut *value.fields.get() };
+            let old_fields = std::mem::replace(old_fields, new_fields);
```
ibraheemdev:
```suggestion
            // SAFETY: We hold the lock for the shard containing the value, and the
            // value has not been interned in the current revision, so no references to
            // it can exist.
            let old_fields = unsafe { std::mem::replace(&mut *value.fields.get(), new_fields) };
```

src/interned.rs
```diff
@@ -582,15 +592,16 @@ where
             //
             // SAFETY: The memo table belongs to a value that we allocated, so it has the
             // correct type.
-            unsafe { self.clear_memos(zalsa, memo_table, new_id) };
+            unsafe { self.clear_memos(zalsa, memo_table, old_id) };
```
ibraheemdev, line 591: This is a surprising bug.

ibraheemdev (approved):

PR #1162, opened 2026-06-20
fix: discard tracked outputs overwritten by specify

MichaReiser:
`specify` could replace a derived memo without discarding tracked structs owned by the old execution. Carry those tracked IDs into output diffing and use the function key when reporting their removal.

Testing: Added regression coverage and ran the inventory-enabled test suite.


Veykril:
I actually wonder whether we would want to rip out specify. I do not know of anyone actually using it (and iirc there was an alternative design to it that niko wanted to explore at some point anyways?)

Veykril (approved):

MichaReiser:
I recently saw that we started using it (too my horror)

Veykril:
Oh dear

PR #1167, opened 2026-06-20
fix: preserve custom interned ID auto traits

MichaReiser:
Interned handles now derive Send and Sync from their custom ID type instead of overriding those auto traits unsafely.

I think this is mainly historical from when the id wrapped a `NonNull<interned::Value<...>>`


Veykril (commented):
I am not sure this change matters? You can't move values across threads etc due to this, since the database already requires these bounds anyways. If anything this might make things more awkward in a certain few scenarios.

MichaReiser:
You mean because of the `&'db lifetime?

PR #1171, opened 2026-06-20
fix: reject unregistered inventory ingredients

MichaReiser:
Inventory-backed ingredient caches assume stable indices across databases. Reject builder-only jars during database construction while continuing to allow redundant registration of globally inventoried jars, preserving the unchecked lookup fast path.

Testing: Added inventory registration coverage and ran the full test suite and clippy.


MichaReiser:
I'm fine with this for now

PR #1174, opened 2026-06-21
perf: avoid type checks in memo and field lookups

MichaReiser:
Memo and field lookups repeated TypeId checks even though the database-bound ingredient fixes the accessed type. Bind each ingredient to its database context and acquire thread-local state only for operations that need it.

Fixed hot memo hits drop from 233 to 228 instructions (-2.15%), untracked tracked-struct field bodies drop from 59 to 58, CodSpeed reports a 6.76% aggregate improvement, and ty's 29-case suite improves by 0.145% weighted overall.

Testing: Full target, manual-registration, lint, and performance suites passed.

### DIFF
--- a/components/salsa-macro-rules/src/setup_input_struct.rs
+++ b/components/salsa-macro-rules/src/setup_input_struct.rs
@@ -152,6 +152,16 @@ macro_rules! setup_input_struct {
                     Self::ingredient_(db.zalsa())
                 }
 
+                #[inline]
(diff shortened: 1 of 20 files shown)


MichaReiser:
This is cool but I don't think codex's reasoning here is safe. The issue is that `zalsa` could be smuggled in. So we'd have to make `fetch` unsafe.

PR #1177, opened 2026-06-21
perf: defer database attachment for cached queries

MichaReiser:
Defer database attachment until a tracked query takes a cold path, so cached top-level and nested queries avoid repeated TLS guard work.

Event callbacks, tracing, cancellation cleanup, and query execution still run with the database attached; mismatched databases continue to panic.

Testing: Passed the full test suite and all-features checks.


MichaReiser:
huh

PR #1178, opened 2026-06-21
perf: allocate memo table slots lazily

MichaReiser:
Allocate memo-table slots on first insertion so Salsa values that are never queried avoid an empty slot allocation.

Salsa stores the memo results on the tracked struct in an associated memo table. Today, these memo tables are allocated eagerly. Meaning, creating any salsa struct immediately allocates a Memo table with N entries (where N are the queries parametrized by the struct). This is wasteful if the struct is never used to store any memo result. This, to some extend, defeats Salsa's lazyness, because answering an LSP request might only need to run queries on a few of all the created salsa structs (e.g. a semantic index that creates a struct for every definition in a scope, but the LSP then only runs the queries for a handful of definitions).

This shows a 4-5% peak memory reduction in ty's simulated benchmark, and performance is neutral (+- a few percent across projects), see https://github.com/astral-sh/ruff/pull/26202

A `ty check` CLI measurement showed roughly a 2-3% RSS reduction

### DIFF
--- a/src/table/memo.rs
+++ b/src/table/memo.rs
@@ -102,7 +102,8 @@ impl LazyMemoEntries {
             return None;
         }
 
-        Some(&self.initialize()[index])
+        let memos = self.as_slice().unwrap_or_else(|| self.initialize());
+        Some(&memos[index])
     }
 
     #[inline]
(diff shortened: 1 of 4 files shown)


MichaReiser:
@Veykril curious to hear your take on this.

Veykril:
On paper this sounds reasonable to me. (have not looked at the impl yet)

src/table/memo.rs
```diff
@@ -73,6 +74,119 @@ struct MemoEntry {
...
+
+    #[inline]
+    fn get(&self, index: usize) -> Option<&MemoEntry> {
+        self.as_slice()?.get(index)
+    }
+
+    #[inline]
+    fn get_or_init(&self, index: usize) -> Option<&MemoEntry> {
+        if index >= self.len {
+            return None;
```
Veykril, line 102: Don't we need to re-allocate here (assuming no inventory feature enabled)?

Veykril: Ah nevermind I guess thats what `clear` is for (if have a resize). My memory isn't fresh on all of this anymore

Veykril (approved):

src/table/memo.rs
MichaReiser: I don't think we support registering new ingredients after creating the storage anymore.

PR #1179, opened 2026-06-21
perf: reduce table page size

MichaReiser:
Salsa table pages currently reserve 1,024 slots even when only a few are used. Reduce both their physical capacity and ID stride to 128, keeping IDs densely packed with eight slot bits.

Each salsa struct ingredient has its own tables in which it arena allocates the structs. Today, we allocate pages of 1024 entries, this PR reduces it to 128, to reduce excess capacity, at the cost of a few extra allocations. We also use `ZalsaLocal` "local" pages, meaning, a run with 32 `ZalsaLocal` allocates structs into 32 different pages, which results in even more excess capacity.

I also experimented with only reducing the size to 256. 128 gives another 2-3% peak memory reduction, without a meaningful perf hit.

We could consider making the page size configurable, or even dynamic, but I'd prefer to leave this to a separate PR

Codspeed is a bit over dramatic when it comes to the memory improvement. It suggests that ty's memory usage drops by ~10-18% (https://github.com/astral-sh/ruff/pull/26207) without a change in runtime (+-1-3%). However, this number is an overstatement because we only allocate a virtual page (we use `MaybeInit`). However, I ran ty over multiple projects and there's a meaningful RSS win, which grows with the number of threads.

Project | Threads | RSS 1024 MiB | RSS 128 MiB | RSS win MiB | RSS win
-- | -- | -- | -- | -- | --
Flake8 | 1 | 103.9 | 100.1 | 3.8 | 3.7%
Flake8 | 64 | 513.5 | 432.5 | 81.0 | 15.8%
Trio | 1 | 191.6 | 182.3 | 9.3 | 4.9%
Trio | 64 | 671.2 | 530.4 | 140.8 | 21.0%
Sphinx | 1 | 383.0 | 379.5 | 3.5 | 0.9%
Sphinx | 64 | 829.8 | 709.1 | 120.7 | 14.5%
Prefect | 1 | 1,104.6 | 1,119.1 | −14.5 | −1.3%
Prefect | 64 | 2,844.4 | 2,747.8 | 96.6 | 3.4%

This is mainly because we avoid pages with only few items:

> - A page size of 128 removes about **89–90% of unused inline slot capacity** relative to 1024 in every tested project and thread configuration.
> - At 64 threads, aggregate fill improved from **4.9% → 32.5%** on Flake8, **8.6% → 48.7%** on Trio, **17.9% → 69.4%** on Sphinx, and **43.3% → 88.7%** on Prefect.

Given this, I believe that 128 is the better default. A next step is to explore if we can change the page size per ingredient. 

> - Structs differ substantially. High-volume ingredients such as `Definition` fill 1024-slot pages well even with 64 workers. Sparse-per-worker ingredients such as `FileModule`, `File`, `ModuleNameIngredient`, `StaticClassLiteral`, and `IntersectionType` leave most of every 1024-slot page empty. Slot width also matters: an unfilled `IntersectionType` slot costs 168 bytes, versus 72 bytes for some interned-argument slots.


Veykril (approved):
This is probably a good tradeoff for the general case, though I do think we want to look into having finer per query grained control here ultimately. Some memos will be massive so you want small pages, some will be tiny where you likely want big pages

Veykril:
Ah you talked about that yourself at the end of the description :)

MichaReiser:
I also thought about whether we should adjust the page size automatically based on the value's size. That could give us even better defaults without extending our macros (we could still decide to make it configurable, but it should reduce the need for it)

PR #1180, opened 2026-06-22
fix: preserve query revision for non-reusable interned values

MichaReiser:
Non-reusable interned values can omit their dependency edge, but their allocation or reuse revision must still contribute to the creating query. Otherwise, recreating a reusable key as a stable key can move changed_at backwards and trigger a backdating panic.


MichaReiser:
I'll merge this, just in case we make a release because the current Salsa version is kind of broken :)

PR #1182, opened 2026-06-23
fix: preserve current query values when specifying

MichaReiser:
`specify` now preserves a value that was already produced in the current revision, preventing current derived memos and their tracked outputs from being discarded. It participates in query synchronization and re-records ownership when the same active query specifies unconditionally.

Testing: Added regression coverage and ran the full inventory-enabled test suite.


MichaReiser:
I merge this because it fixes a latent bug in ty and I want to cut a release. I'll open a follow up PR if there's any feedback after merge

PR #1188, opened 2026-06-24
perf: Introduce detailed-trace feature

MichaReiser:
## Summary

The `debug` and `trace` level calls add a fair amount of instructions to Salsa's hot path. Given that these traces are mainly useful to us maintainers when debugging a Salsa bug, it seems unjustified to have every application pay that cost. 

This PR introduces a new detailed-trace feature (off by default) that removes those tracing calls entirely. 

The remaining tracing usages are info and warning level messages:

* Log the execution of a query
* When starting a fixpoint iteration
* When starting the next fixpoint iteration
* When a query panicked

A next step could be to add a `tracing` or `trace` feature, so that upstream tools that don't want any salsa logging can opt-out fully.

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -258,11 +258,10 @@ impl MemoHeader {
         cycle_recovery_strategy: CycleRecoveryStrategy,
         has_value: bool,
     ) -> Option<VerifyResult> {
-        let database_key_index = claim_guard.database_key_index();
-
         crate::tracing::debug!(
             "{database_key_index:?}: maybe_changed_after_cold, successful claim, \
                 revision = {revision:?}, old_memo = {old_memo:#?}",
+            database_key_index = claim_guard.database_key_index(),
             old_memo = self.tracing_debug(has_value)
         );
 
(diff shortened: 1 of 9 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -259,6 +259,8 @@ impl MemoHeader {
         has_value: bool,
     ) -> Option<VerifyResult> {
         let database_key_index = claim_guard.database_key_index();
+        #[cfg(not(feature = "detailed-trace"))]
+        let _ = database_key_index;
```
Veykril:
```suggestion
        #[cfg(not(feature = "detailed-trace"))]
        let database_key_index = claim_guard.database_key_index();
```

src/tracing.rs
Veykril, line 1: Imo we should wrap the macro expansion in 

```rust
if cfg!(feature = "detailed-trace") {
      ....
}
```

instead, then we wont have to sprinkle cfg attributes all over the place

src/function/maybe_changed_after.rs
Veykril: given my previous comment, then we can also inline this into the macro i think?

src/tracing.rs
MichaReiser: Oh nice, that's precisely what I was looking for

MichaReiser:
Thanks Lukas. This is so much cleaner

Veykril (approved):

PR #1195, opened 2026-06-25
chore: migrate benchmarks to Divan and CodSpeed macro runners

MichaReiser:
Run simulation and memory benchmarks on CodSpeed macro runners again, while keeping compilation and Cargo cache work on an ARM64 GitHub runner. It turns out, it does help with flakiness. CodSpeed was also so kind to grant us some extra free minutes and we're in the process of switching to Pro. 

I also use this PR as an opportunity to move the allocation-heavy LRU to walltime mode. Codspeed simulation benchmarks aren't made for allocation (sys call heavy) workloads.

### DIFF
--- a/benches/accumulator.rs
+++ b/benches/accumulator.rs
@@ -1,7 +1,10 @@
 use std::hint::black_box;
 
-use codspeed_criterion_compat::{BatchSize, Criterion, criterion_group, criterion_main};
 use salsa::Accumulator;
+
+fn main() {
+    divan::main();
+}
 
 #[salsa::input]
 struct Input {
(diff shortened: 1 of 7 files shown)


benches/accumulator.rs
```diff
@@ -42,31 +45,25 @@ fn infer_expression<'db>(db: &'db dyn salsa::Database, expression: Expression<'d
     number
 }
 
-fn accumulator(criterion: &mut Criterion) {
-    criterion.bench_function("accumulator", |b| {
-        b.iter_batched_ref(
-            || {
-                let db = salsa::DatabaseImpl::new();
+#[divan::bench(name = "benches::accumulator::accumulator")]
```
MichaReiser, line 48: The `name` is necessary to preserve our existing benchmark names

PR #1206, opened 2026-06-28
perf: reduce memo validation monomorphization

MichaReiser:
Move configuration-independent memo validation into a type-erased cold path, reducing ty's generated LLVM IR by 22,813 lines. I plan on using the erased memo for more changes, which should justify the introduced complexity. But I thought this change is worth splitting out, to ease reviewing.

### DIFF
--- a/src/function/maybe_changed_after.rs
+++ b/src/function/maybe_changed_after.rs
@@ -2,13 +2,14 @@
 use crate::accumulator::accumulated_map::InputAccumulatedValues;
 use crate::cycle::{CycleHeads, CycleRecoveryStrategy, ProvisionalStatus};
 use crate::database::RawDatabase;
-use crate::function::memo::{ErasedMemo, MemoHeader, TryClaimCycleHeadsIter, TryClaimHeadsResult};
+use crate::function::memo::{
+    ErasedMemo, FunctionMemoTable, MemoHeader, TryClaimCycleHeadsIter, TryClaimHeadsResult,
+};
 use crate::function::sync::{ClaimGuard, ClaimResult};
 use crate::function::{Configuration, IngredientImpl, Reentrancy, SyncTable};
 use std::sync::atomic::Ordering;
 
 use crate::key::DatabaseKeyIndex;
-use crate::table::memo::MemoTableWithTypes;
 use crate::zalsa::{MemoIngredientIndex, Zalsa, ZalsaDatabase};
 use crate::zalsa_local::{QueryEdgeKind, QueryEdges, QueryOriginRef, QueryRevisions, ZalsaLocal};
 use crate::{Id, Revision};
@@ -94,11 +95,10 @@ where
     ) -> VerifyResult {
         let (zalsa, zalsa_local) = db.zalsas();
         let memo_ingredient_index = self.memo_ingredient_index(zalsa, id);
+        let database_key_index = self.database_key_index(id);
         zalsa.unwind_if_revision_cancelled(zalsa_local);
 
         loop {
-            let database_key_index = self.database_key_index(id);
-
             crate::tracing::debug!(
                 "{database_key_index:?}: maybe_changed_after(revision = {revision:?})"
             );
@@ -133,11 +133,11 @@ where
         }
     }
 
-    fn maybe_changed_after_cold<'db>(
-        &'db self,
-        zalsa: &'db Zalsa,
-        zalsa_local: &'db ZalsaLocal,
-        db: &'db C::DbView,
+    fn maybe_changed_after_cold(
+        &self,
+        zalsa: &Zalsa,
+        zalsa_local: &ZalsaLocal,
+        db: &C::DbView,
         database_key_index: DatabaseKeyIndex,
         revision: Revision,
         memo_ingredient_index: MemoIngredientIndex,
@@ -157,7 +157,7 @@ where
             zalsa: &'db Zalsa,
             zalsa_local: &'db ZalsaLocal,
             db: RawDatabase<'db>,
-            memo_table: MemoTableWithTypes<'db>,
+            memo_table: FunctionMemoTable<'db>,
             database_key_index: DatabaseKeyIndex,
             revision: Revision,
             memo_ingredient_index: MemoIngredientIndex,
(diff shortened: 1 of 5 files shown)


src/function/maybe_changed_after.rs
```diff
@@ -112,95 +114,163 @@ where
...
+            // that reset cannot occur and the loaded allocation remains immovable and valid for
+            // shared access for `'db`, even if another thread replaces the table entry during this
+            // revision.
+            let Some(old_memo) = (unsafe { memo_table.get_erased(memo_ingredient_index) }) else {
+                return ColdResult::Verified(VerifyResult::changed());
+            };
+
+            let old_header = old_memo.header();
+
+            crate::tracing::debug!(
```
MichaReiser, line 193: Inlined from memo_header.maybe_changed_after_cold`

src/function/maybe_changed_after.rs
```diff
@@ -112,95 +114,157 @@ where
                 zalsa,
                 database_key_index,
                 revision,
+                #[cfg(feature = "detailed-trace")]
```
MichaReiser, line 117: This makes it more explicit that this argument is only used for tracing.

src/function/maybe_changed_after.rs
```diff
@@ -112,95 +114,157 @@ where
...
-                // We failed to claim, have to retry.
+                return result;
             }
         }
     }
 
-    #[inline(never)]
-    fn maybe_changed_after_cold<'db>(
-        &'db self,
+    fn maybe_changed_after_cold(
```
MichaReiser, line 136: I already verified that the behavior is identical and is only about moving code around.

src/function/maybe_changed_after.rs
```diff
@@ -112,95 +114,157 @@ where
...
-        // Load the current memo, if any.
-        let Some(old_memo) = self.get_memo_from_table_for(zalsa, key_index, memo_ingredient_index)
-        else {
-            return Some(VerifyResult::changed());
-        };
 
-        if let Some(result) = old_memo.header.maybe_changed_after_cold(
+            // Load the current memo after claiming the query because it may have changed while
+            // this query was blocked on another thread.
+            let Some(old_memo) = memo_table.get_erased(memo_ingredient_index) else {
```
MichaReiser: This is the main change. We now call `get_erased` here

src/function/maybe_changed_after.rs
```diff
@@ -112,95 +114,157 @@ where
...
         ) {
-            return Some(result);
-        }
+            ColdResult::Retry => None,
+            ColdResult::Verified(result) => Some(result),
+            ColdResult::Reexecute {
+                claim_guard,
+                old_memo,
+            } => {
+                let old_memo = old_memo.downcast::<C>();
```
MichaReiser, line 251: And we downcast to `Memo<C>` here

src/function/memo.rs
```diff
@@ -74,15 +86,104 @@ impl<C: Configuration> IngredientImpl<C> {
...
+    data: NonNull<DummyMemo>,
+
+    /// Coerces `data` to a trait object using the vtable for its concrete memo type.
+    to_dyn_fn: ToDynMemo,
+
+    /// The concrete memo type, used to assert that downcasts match the registered memo type.
+    type_id: TypeId,
+
+    /// Binds shared access to the allocation lifetime.
+    _lifetime: PhantomData<&'db ()>,
```
Veykril, line 105: nit
```suggestion
    _lifetime: PhantomData<fn() -> &'db ()>,
```
using fn ptrs is usually better as it won't affect auto traits as much.

src/function/memo.rs
```diff
@@ -483,11 +606,16 @@ mod _memory_usage {
...
+    // Required by `ErasedMemo::header`.
+    const _: () = assert!(std::mem::offset_of!(super::Memo<DummyConfiguration>, header) == 0);
+
+    // Memos are stored a lot, make sure their size doesn't randomly increase.
     const _: [(); std::mem::size_of::<super::MemoHeader>()] =
         [(); std::mem::size_of::<[usize; 4]>()];
     const _: [(); std::mem::size_of::<super::Memo<DummyConfiguration>>()] =
         [(); std::mem::size_of::<[usize; 5]>()];
+    const _: [(); std::mem::size_of::<super::ErasedMemo<'static>>()] =
+        [(); std::mem::size_of::<[usize; 4]>()];
```
Veykril, line 584: Can we go even further and assert field offset equality via `offset_of!` usage?

src/function/maybe_changed_after.rs
```diff
@@ -112,95 +115,156 @@ where
...
-                // We failed to claim, have to retry.
+                return result;
             }
         }
     }
 
-    #[inline(never)]
-    fn maybe_changed_after_cold<'db>(
-        &'db self,
+    fn maybe_changed_after_cold(
```
ibraheemdev, line 136: Was removing the `inline(never)` intentional here?

src/function/memo.rs
```diff
@@ -264,6 +369,24 @@ where
     }
 }
 
+/// A function memo table whose allocations remain valid for `'db`, including after replacement.
+pub(super) struct FunctionMemoTable<'db> {
```
ibraheemdev: What's the point of this wrapper type? Is there some additional invariant being upheld here?

src/function/memo.rs
```diff
@@ -74,15 +86,104 @@ impl<C: Configuration> IngredientImpl<C> {
     }
 }
 
+/// A memoized query result.
+///
+/// # Layout
+///
+/// [`ErasedMemo`] retains a pointer to the complete `Memo` allocation so that it can recover the
```
ibraheemdev: Very minor nit, a lot of these comments refer to the idea "complete" allocation. It's not clear to me what exactly that means, or if that is just LLM-speak for "valid" (for reads/writes).

ibraheemdev (approved):

src/function/maybe_changed_after.rs
MichaReiser: Yes, we could add it to the `inner` function. We want to avoid that LLVM now compiles this to two function calls:

* a call to `maybe_changed_after_cold` that's generic over `C`
* a call to the non-generic `inner` function

Let me measure if adding `inline(never)` to inner changes the instruction count.

src/function/memo.rs
MichaReiser: Yes, it has a stricter lifetime than `MemoTableWithTypes`. I renamed the lifetime to make this clearer. I also renamed the type to `MemoSlot<'a>`. It's a reference to a memo slot, and the reference returned by `get_erased` is guaranteed to stay alive `'a`. `MemoTableWithTypes` can't guarantee this because you can call `insert` multiple times, which replaces the memo in that slot. It's only the `FunctionIngredient` that defers the deletion of the memos, guaranteeing that `'a` (the memo) remains dereferencable for `'db`. 

Note, `MemoSlot` does not guarantee you that it returns the same `Memo` for `'a`. Calling `get_erased` multiple times with `insert` calls in between can return different memos. It only guarantees that the memos return each have a lifetime of `'a`.

src/function/memo.rs
MichaReiser: Thanks for pointing this out. I updated the comments

PR #1207, opened 2026-06-28
perf: reduce ingredient cache monomorphization

MichaReiser:
Move ingredient cache initialization behind a shared type-erased slow path while preserving statically dispatched lookups for manual multi-database use.

The key here is to remove the `closure`, and instead pass the type and index explicitly.

This reduces ty's total LLVM IR by 42,584 lines (1.03%) without regressing the multi-database benchmark.

Testing: Passed the Salsa test suites, Clippy, formatting, MSRV checks, and targeted performance measurements.

### DIFF
--- a/components/salsa-macro-rules/src/setup_accumulator_impl.rs
+++ b/components/salsa-macro-rules/src/setup_accumulator_impl.rs
@@ -36,11 +36,7 @@ macro_rules! setup_accumulator_impl {
 
                 // SAFETY: `lookup_jar_by_type` returns a valid ingredient index, and the only
                 // ingredient created by our jar is the struct ingredient.
-                unsafe {
-                    $CACHE.get_or_create(zalsa, || {
-                        zalsa.lookup_jar_by_type::<$zalsa_struct::JarImpl<$Struct>>()
-                    })
-                }
+                unsafe { $CACHE.get_or_create::<$zalsa_struct::JarImpl<$Struct>, 0>(zalsa) }
             }
 
             impl $zalsa::Accumulator for $Struct {
--- a/src/ingredient_cache.rs
+++ b/src/ingredient_cache.rs
@@ -3,10 +3,11 @@ pub use imp::IngredientCache;
 #[cfg(feature = "inventory")]
 mod imp {
     use crate::IngredientIndex;
-    use crate::plumbing::Ingredient;
+    use crate::plumbing::{Ingredient, Jar};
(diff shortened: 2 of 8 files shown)


src/ingredient_cache.rs
```diff
@@ -202,36 +216,36 @@ mod imp {
...
-                Ordering::Release,
-                Ordering::Relaxed,
-            );
-
-            // Use our locally computed index regardless of which one was cached.
-            index
-        }
+    #[cold]
+    #[inline(never)]
+    fn get_or_create_index_slow(
```
MichaReiser, line 231: This function is now shared between all Ingredients instead of having one implementation per ingredient

src/ingredient_cache.rs
```diff
@@ -78,36 +84,43 @@ mod imp {
...
-            // It doesn't matter if we overwrite any stores, as `create_index` should
-            // always return the same index when the `inventory` feature is enabled.
-            self.ingredient_index
-                .store(ingredient_index.as_u32(), Ordering::Release);
-
-            ingredient_index
-        }
+    #[cold]
+    #[inline(never)]
+    fn get_or_create_index_slow(
```
MichaReiser, line 91: This function is now shared between all Ingredients instead of having one implementation per ingredient

MichaReiser:
I think this is fairly uncontroversial. I'll merge this without review. I'm happy to follow up with feedback if there's any (or revert the PR if I misjudged)

components/salsa-macro-rules/src/setup_accumulator_impl.rs
```diff
@@ -36,11 +36,7 @@ macro_rules! setup_accumulator_impl {
 
                 // SAFETY: `lookup_jar_by_type` returns a valid ingredient index, and the only
```
ada4a, line 37: I'm trying to update r-a to use salsa 0.28, and it has a similar call at https://github.com/rust-lang/rust-analyzer/blob/0069617d8971879f47679f559fd6a103f029ce9e/crates/span/src/hygiene.rs#L136-L142, with a similar safety comment. But I think both the r-a comment and the comments in this PR need to be updated, given that `lookup_jar_by_type` is no longer called explicitly.

Veykril:
```
// SAFETY: The ingredient at offset 0 in `JarImpl<$Struct>` has type
// `IngredientImpl<$Struct>`.
```
or similar I assume would be fine

PR #1215, opened 2026-06-29
docs: update Salsa book for current APIs

MichaReiser:
Update stale Salsa book examples and explanations to match the current tracked APIs and runtime model.
Keep the public book targets referenced by #1213 intact while replacing dead links and obsolete examples.
Testing: mdBook link checking and focused Salsa tests pass.


MichaReiser:
Not perfect, but less outdated :)

PR #1216, opened 2026-06-30
feat: return references by default

### LINKED ISSUE #719: Flip default for `return_ref`
This has come up in https://github.com/salsa-rs/salsa/pull/718. 

It's very easy to forget a `return_ref` on a non-copy type. On the other hand, returning a ref for a `copy` type may be undesired but has less performance consequences. 

Should we change the default for `return_ref` to be true and instead introduce a `return_cloned`.

Related: Should we introduce a `return_deref`? There have been a few occasions where I had an `Option<T>` and I ended up writing a manual accessor to get the deref behavior. That might also suggest that the syntax should be `return(cloned)`, `return(deref)` `return(ref`)?

MichaReiser:
Changes tracked functions and Salsa struct field getters to return references by default, 

I don't know what the appetite for this is, but this closes https://github.com/salsa-rs/salsa/issues/719

IMO, we should either make the change and accept that it's breaking or decide that we stick with `clone`. At least for ty, this defaulting to `ref` has been a repeating issue, and I occasionally go through all queries to find cases where we forgot the returns annotation. I also had Codex migrate ty, and it did so without much effort. So I don't expect this to be as painful as it used to be.

One caveat. There's one remaining place where Salsa always clones (and which may be surprising). And that's for many argument queries. The arguments are cloned out of the interned value. This does not change my perspective that `returns(ref)` is the better default.

### DIFF
--- a/benches/accumulator.rs
+++ b/benches/accumulator.rs
@@ -22,7 +22,7 @@ struct Expression<'db> {
     number: usize,
 }
 
-#[salsa::tracked(returns(clone))]
+#[salsa::tracked(returns(deref))]
 #[inline(never)]
 fn root(db: &dyn salsa::Database, input: Input) -> Vec<usize> {
     (0..input.expressions(db))
--- a/benches/accumulator.rs
+++ b/benches/accumulator.rs
@@ -8,6 +8,7 @@ fn main() {
 
 #[salsa::input]
 struct Input {
+    #[returns(copy)]
     expressions: usize,
 }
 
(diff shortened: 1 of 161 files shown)


Veykril:
I think we had an issue for this even right? Either way in full agreement

MichaReiser:
Haha yes. You're too fast. I was still in the progress of cleaning up the codex generated summary and linking to the relevant issue.

MichaReiser:
Given the support from r-a, I'm inclined to land this. Reverting before the next release is easy, in case someone later objects.

Veykril (approved):

PR #1217, opened 2026-06-30
feat: replace `Update` with `SalsaValue` and `PartialEq`

### LINKED ISSUE #592: Rename `Update` to `StoreInSalsaDb` or something like that
The name `Update` is highly unintuitive. What this value really means is "safe to store in the salsa database".

### LINKED ISSUE #681: Unsoundness: Tracked-fields with `no_eq` can store references
```rust
#[salsa::tracked]
pub(crate) struct Expression<'db> {

    #[no_eq]
    pub(crate) test: &'db str,

}
```

This compiles when it should not. We still need to enforce the `Update` trait for `test` or, have a dedicated trait (because implementing `Update` is a bit non-sensical)

MichaReiser:
This is another breaking change that we've discussed but never pushed forward. 

This PR renames the `Update` trait and removes its `maybe_update` function. 

* `maybe_update` allowed to update a value in-place and it returns whether the new value is equal to the old value. However, this never enabled us to reuse allocations between revisions because, at this point, we already have the fully materialized new value. This PR removes `maybe_update` and uses `PartialEq` instead. Types that don't support `PartialEq` can use `no_eq`. Using `PartialEq` has the benefit that `SalsaValue` needs to be derived in much fewer places than before. Using `PartialEq` also has the benefit that manually implementing `SalsaValue` no longer requires the always the same boilerplate for `maybe_update` (which is also somewhat scary to write because it involves pointers). Removing the `maybe_update` constraint also allows us to fix https://github.com/salsa-rs/salsa/issues/681, because it decouples the lifetime concern from whether a type can be compared.
* The `SalsaValue`'s only remaining responsibility is to forbid `&'db` references. 

I also used this as an opportunity to improve some of the missing `SalsaValue` or not `'static` diagnostics, so that they point to the lifetime or correct field (and hint users toward the derive macro).

Closes https://github.com/salsa-rs/salsa/issues/592
Closes https://github.com/salsa-rs/salsa/issues/681

I'm sorry that Git doesn't recognize the rename from `update` to `salsa_value`. It makes this rather painful to review :(

Tagging @nikomatsakis, not necessarily to review the code changes. More if you have opinions about this somewhat fundamental change.

I had Codex review this PR and it then found an existing missing Safety contract. That is, `Configuration::Output` doesn't guarantee that `Output` implements `SalsaValue`. The same applies for interned and tracked struct `Configuration`. Now, this constraint is satisfied for all automatically derived `Configuration` implementations, and I assume it is also uphold by r-a's custom implementaiton.

**Edit**: I had a version that made `SalsaValue` have an associated `Output` type which is the structs type with a `'db` lifetime. I ultimately decided to back out of this change, and opted for making the `Configuration` traits unsafe. Mainly because we didn't really use much of what the `GAT` offered and it made the derive macro a fair bit more complicated (about 400 lines of additional code in total). What follows below is the explanation for the GAT based approach. 

Codex first marked all the `Configuration` traits as unsafe (tracked, interned, and tracked functions). While there's not much harm to it, given that it's primarily the macro adding implementations, I pushed a little harder to encode the constraint in the type system instead.

The solution I (and Codex) came up with are inspired by yoke. That is, we add an associated type to `SalsaValue`, `Output<'db>` which is the type of `self` with the `'db` lifetime. This allows us to name both the `'static` and `'db` type in code. It also allows us to remove the `'db` lifetime from `Memo` and perform the lifetime erasure at the `Configuration` boundary instead. 

I'm not sure it's worth it. I think it would also be okay to mark the `Configuration` traits as unsafe, document the `Fields`/`Output` constraint and leave it at that. I'm curious what other think on this.

## Test plan

I upgraded rust-analyzer and ty. It requires changes to manual `Configuration` implementations, replacing `Update` with `SalsaValue`, and replacing `non_maybe_update` attributes. But the upgrade was otherwise successful (and clankers are very good at this manual replacement)


MichaReiser:
I'm still trying to figure out if it's even worth keeping `Update`. What is it good for? Two separate traits is also somewhat painful. The worst case is that you now need to derive two traits.

Veykril:
Recalling from my memory its used for two things:
- making sure you can't store bad things in salsa, aka no references
- allowing to do fine grained in-place updates of values instead of wholesale replacements

Obviously we will need the former, the latter is the questionable part I guess?

MichaReiser:
yes, I don't understand the need for the `Second`. Update is arguably even more expensive than `PartialEq`, which often is a simple memcmp. 

I always thought the benefit of `Update` is to avoid unnecessary allocations but where we use it, it's already too late for this.

Veykril:
Yea same thinking. It would be cool if we could save on the allocs, but I think that needs a different mechanism, being able to inspect the previous memo within a query execution itself (I have a branch for this around somewhere I think, cause thats something r-a kind of wants).

Veykril:
So imo we can get rid of the later behavior

MichaReiser:
> Yea same thinking. It would be cool if we could save on the allocs, but I think that needs a different mechanism, being able to inspect the previous memo within a query execution itself (I have a branch for this around somewhere I think, cause thats something r-a kind of wants).


Wow, this sounds exciting. 

I think the other thing that Niko may have had in mind is to do some sort of tracing. But I'm not sure how that would work with the `Update` trait as is. I rather reintroduce it when we have the need for it

MichaReiser:
Ahhh

> Instead, Update was introduced because the intended Salsa 3.0 design wanted to support values containing database-lifetime references. In a later revision, the old value might contain references that were no longer safe to dereference. Calling PartialEq requires creating &old, which could therefore be UB.

src/input.rs
```diff
@@ -37,7 +37,7 @@ pub trait Configuration: Any {
     type Struct: FromId + AsId + 'static + Send + Sync;
 
     /// A (possibly empty) tuple of the fields for this struct.
-    type Fields: Send + Sync;
+    type Fields: 'static + Send + Sync;
```
MichaReiser, line 40: We never enforced that `Input`s didn't contain references. We should. We don't need `SalsaValue` because `Inputs` are a pure storage, Salsa never updates them in-place.

components/salsa-macro-rules/src/setup_interned_struct.rs
```diff
@@ -115,7 +116,36 @@ macro_rules! setup_interned_struct {
...
+                $($assert_fields_are_salsa_values)*
+            }
+            let _ = _assert_fields_are_salsa_values;
+
+            #[derive(Clone, PartialEq, Eq, Hash)]
+            $vis struct $StructDataIdent<$db_lt> {
+                fields: ($($field_ty,)*),
+                // Keeps the database lifetime in the type even if no field uses it.
+                marker: ::std::marker::PhantomData<fn() -> &$db_lt ()>,
+            }
```
MichaReiser: We now need to retain the lifetime for `SalsaValue::Output<'db>`

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -70,7 +73,9 @@ macro_rules! setup_tracked_fn {
         // If true, the input and output values implement `serde::{Serialize, Deserialize}`.
         persist: $persist:tt,
 
-        assert_types_are_update: {$($assert_types_are_update:tt)*},
+        assert_interned_inputs_are_salsa_values: {$($assert_interned_inputs_are_salsa_values:tt)*},
```
MichaReiser, line 73: This now enforces that we deny references in function arguments

```rust
#[salsa::tracked]
fn test<'db>(db: &'db dyn Db, value: &'db str) -> String {
	String::from(vlaue)
}
```

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -128,6 +136,16 @@ macro_rules! setup_tracked_fn {
...
 
+            #[repr(transparent)]
+            struct $Output($output_static_ty);
+
+            // SAFETY: The generated assertion below proves the retained value
+            // can be exposed with the current database lifetime. The unsafe
+            // escape hatch makes this the caller's responsibility instead.
+            unsafe impl<$db_lt> $zalsa::SalsaValue<$db_lt> for $Output {
+                type Output = $output_ty;
+            }
```
MichaReiser: The `Configuration::Output` now requires that the type implements `SalsaValue`. But a static dispatch fallback type that is known to be static does not implement `SalsaValue`. This is why we wrap it here

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -142,6 +160,32 @@ macro_rules! setup_tracked_fn {
...
+                        for $InternedFields<$db_lt>
+                    {
+                        fn hash<H: ::core::hash::Hasher>(&self, hasher: &mut H) {
+                            <($($interned_input_ty),*) as $zalsa::HashEqLike<($($interned_input_ty),*)>>::hash(&self.0, hasher);
+                        }
+
+                        fn eq(&self, data: &($($interned_input_ty),*)) -> bool {
+                            <($($interned_input_ty),*) as $zalsa::HashEqLike<($($interned_input_ty),*)>>::eq(&self.0, data)
+                        }
+                    }
```
MichaReiser: Needed, because we now have the `InternedFields` wrapper

src/salsa_value.rs
```diff
@@ -0,0 +1,229 @@
...
+/// remain safe to retain and use in a newer revision. This includes calling
+/// safe trait methods such as [`PartialEq::eq`] on the old value. In particular,
+/// `SalsaValue` must not be implemented for a database-lifetime reference such
+/// as `&'db T`, which may point to storage changed or freed in a newer revision.
+///
+/// `#[derive(SalsaValue)]` checks this requirement structurally. It cannot
+/// account for additional invariants maintained by unsafe code in the type's
+/// methods.
+///
+/// For example, this derive is invalid even though its fields pass the
```
Veykril: I think the phrasing is a bit off, the derive itself is valid just fine. The undefined behavior happens in the user written `value` method, that is the impl of the trait is valid, the assumption from the user of the invariants are wrong though.

src/salsa_value.rs
```diff
@@ -0,0 +1,229 @@
...
+    std::num::NonZeroI128,
+    std::num::NonZeroIsize,
+    std::num::NonZeroU8,
+    std::num::NonZeroU16,
+    std::num::NonZeroU32,
+    std::num::NonZeroU64,
+    std::num::NonZeroU128,
+    std::num::NonZeroUsize,
+    crate::Id,
+    Box<str>,
```
Veykril: This being here is surprising, I'd expect us to have a generic `T: ?Sized + SalsaValue` impl fox Box, Arc and Rc and also impl `SalsaValue` for `str` directly (as well as `Path` maybe).

src/salsa_value.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+#[cfg(feature = "compact_str")]
+salsa_values!(compact_str::CompactString);
+
+// SAFETY: A genuinely `'static` reference remains valid across revisions.
+unsafe impl<T: ?Sized> SalsaValue for &'static T {}
```
Veykril, line 117:
```suggestion
unsafe impl<T: ?Sized + SalsaValue> SalsaValue for &'static T {}
```
as well as elsewhere, or `&'static &'db Thing` becomes a `SalsaValue` I think?

Veykril: Nevermind actually, `&'static T` implies `T: 'static`

Veykril (commented):
Haven't looked at the macro changes yet

src/salsa_value.rs
MichaReiser: I'm okay removing this entire section (and in related documentation). I agree that the author then didn't understand the `'db` lifetime.

src/salsa_value.rs
MichaReiser: Yes, agree. I'll double check the list here. These might be left-overs from some intermediate states.

Veykril (approved):
I thnk the `#[salsa_value(prove_safe_to_retain_manually)]` helper should be required to be wrapped in `unsafe`, `#[salsa_value(unsafe(prove_safe_to_retain_manually))]` given you could use this to store a reference which is unsound right? Otherwise lgtm

PR #1219, opened 2026-06-30
feat: add `previous()` to access a tracked function's prior memoized value

Veykril:
Introduce a generated `previous()` associated function on every tracked
function, returning a reference to the value memoized for the active key
in the previous revision (or `None` if the function has never completed
for that key, or its prior result is provisional).

This lets a tracked function build its new result on top of its own
prior result while re-executing. Reading the previous value replays that
result's dependencies into the current execution — durability, dependency
edges, untracked-read flag, changed-at revision, and tracked-struct ids —
so incremental validation stays correct and previously-created tracked
structs remain live.

### DIFF
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -469,6 +469,36 @@ macro_rules! setup_tracked_fn {
                     }
                 }
 
+                /// Returns a reference to this function's memoized value from the previous
+                /// revision, or `None` if there is no usable previous value.
+                ///
+                /// This is only meaningful while the function is re-executing for the current
+                /// key: it lets the new computation build on its own prior result. Reading the
+                /// previous value replays that result's dependencies into the current execution,
+                /// so incremental validation stays correct.
+                ///
+                /// Returns `None` when the function has never completed for this key, when the
+                /// previous result is provisional (for example, mid-cycle), or when the memoized
+                /// value has been evicted (for example, due to an LRU capacity limit).
+                ///
+                /// Values accumulated by the previous execution itself are not replayed; a
+                /// function that accumulates must push its values again even when it returns
+                /// its previous result.
+                ///
+                /// # Panics
+                ///
+                /// Panics if called outside of this tracked function's own execution.
+                pub fn previous<$db_lt>(
+                    $db: &$db_lt dyn $Db,
+                ) -> Option<&$db_lt $output_ty> {
+                    use ::salsa::plumbing as $zalsa;
+
+                    $zalsa::attach($db, || {
+                        let (zalsa, zalsa_local) = $db.zalsas();
+                        $Configuration::fn_ingredient_($db, zalsa).previous(zalsa, zalsa_local)
+                    })
+                }
+
                 $zalsa::macro_if! { $is_specifiable =>
                     pub fn specify<$db_lt>(
                         $db: &$db_lt dyn $Db,
--- a/components/salsa-macro-rules/src/setup_tracked_fn.rs
+++ b/components/salsa-macro-rules/src/setup_tracked_fn.rs
@@ -485,6 +485,10 @@ macro_rules! setup_tracked_fn {
                 /// function that accumulates must push its values again even when it returns
                 /// its previous result.
                 ///
+                /// For a specifiable function, the previous value may have been assigned via
+                /// `specify` rather than computed by this function; the two cases are
+                /// indistinguishable here.
+                ///
                 /// # Panics
                 ///
                 /// Panics if called outside of this tracked function's own execution.
--- a/src/active_query.rs
+++ b/src/active_query.rs
@@ -11,7 +11,9 @@ use crate::key::DatabaseKeyIndex;
 use crate::runtime::Stamp;
 use crate::sync::atomic::AtomicBool;
 use crate::tracked_struct::{Disambiguator, DisambiguatorMap, IdentityHash, IdentityMap};
-use crate::zalsa_local::{OriginAndExtra, QueryEdge, QueryRevisions, QueryRevisionsExtra};
+use crate::zalsa_local::{
+    OriginAndExtra, QueryEdge, QueryEdges, QueryRevisions, QueryRevisionsExtra,
+};
 use crate::{
     Id,
     cycle::{CycleHeads, IterationStamp},
@@ -72,7 +74,7 @@ impl ActiveQuery {
         &mut self,
         durability: Durability,
         changed_at: Revision,
-        edges: crate::zalsa_local::QueryEdges<'_>,
+        edges: QueryEdges<'_>,
         untracked_read: bool,
         active_tracked_ids: &[(Identity, Id)],
     ) {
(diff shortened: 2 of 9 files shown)


Veykril:
@MichaReiser this is the feature i was hinting at (I just revived the branch on top of main). There are still quite a bit of unknowns that I have to think about in terms of whether this is sound and what exactly it means to read your previous memo in a query, but opus thinks its safe (and iirc codex which I used to write this back then agreed).

src/active_query.rs
```diff
@@ -93,6 +97,32 @@ impl ActiveQuery {
             .seed(active_tracked_ids.iter().map(|(id, _)| id));
     }
 
+    pub(super) fn add_previous_memo_read(
```
Veykril, line 98: This is *almost* the same as `seed_iteration`, the only difference being us seeding `current_revision` instead of the previous `changed_at` (which I have to verify whether we want to do the same as in seed_iteration actually) and us seeding all edges instead of just the outputs (also not 100% sure on this but I think this is what we want).

Which makes sense, its kind of similar to a fixpoint cycle after all in spirit, just across revisions?

MichaReiser:
Yeah, I think this should probably be fine (you already handle fixpoint queries, which was my main concern :))

So this could e.g. be used for an incremental parse?

Veykril:
Also I haven't given the tests a proper look yet (they are llm generated) and there are likely scenarios that are missing still.

As for why r-a is interested in this, see https://github.com/rust-lang/rust-analyzer/pull/22532. Basically on trivia edits, we don't want to re-execute the proc-macros, instead we'd like to just take the previous result, fix up the spans manually and emit that as the new memo.

So yea, this is useful for more manual incremental recomputation handling. ~~Though on second thought, this alone won't be enough since we need to know the diff of the inputs for that 🤔~~ Nevermind I think I remember, we'd store that info in the input when building it in its given query. Bit iffy on the details, would need to test drive that to know

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -469,6 +469,40 @@ macro_rules! setup_tracked_fn {
...
+                /// previous result is provisional (for example, mid-cycle), or when the memoized
+                /// value has been evicted (for example, due to an LRU capacity limit).
+                ///
+                /// Values accumulated by the previous execution itself are not replayed; a
+                /// function that accumulates must push its values again even when it returns
+                /// its previous result.
+                ///
+                /// For a specifiable function, the previous value may have been assigned via
+                /// `specify` rather than computed by this function; the two cases are
+                /// indistinguishable here.
```
Veykril, line 490: Calling out the specify case, we could recognize this and have the function return something else here (enum distinguishing the cases) but I don't think thats worth it until useful to someone.

Veykril:
I think I've covered everything now (forgot about accumulators which is now fixed)

MichaReiser:
I haven't reviewed the code yet. But I wonder if `previous` should also return the current memo value for fixpoint iteration. It seems equally useful to avoid rebuilding the entire struct (for as long as it's only read-only). `previous` would always be the previous memo: Either from the last iteration or last revision

Veykril:
Can we just return it or would that require more plumbing? Not familiar with how fixpoint deals with its memos

MichaReiser:
Let me have a look at the code. I'm not sure what you mean by just returning it :) 

Overall, the memos are handled mostly the same as regular memos. Except that we replace them after each iteration

components/salsa-macro-rules/src/setup_tracked_fn.rs
MichaReiser: I agree with your reasoning here. The idea of specify is that it's mostly transparent. So handling them as described here makes sense to me

components/salsa-macro-rules/src/setup_tracked_fn.rs
```diff
@@ -469,6 +469,40 @@ macro_rules! setup_tracked_fn {
...
+                ///
+                /// # Panics
+                ///
+                /// Panics if called outside of this tracked function's own execution.
+                pub fn previous<$db_lt>(
+                    $db: &$db_lt dyn $Db,
+                ) -> Option<&$db_lt $output_ty> {
+                    use ::salsa::plumbing as $zalsa;
+
+                    $zalsa::attach($db, || {
```
MichaReiser, line 500: Do we need to attach here? Is it mainly to assert the same-db constraint?

src/function/previous.rs
```diff
@@ -0,0 +1,55 @@
...
+where
+    C: Configuration,
+{
+    /// Returns the value memoized for the active key in the previous revision,
+    /// replaying that result's dependencies into the active query.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the active query is not an execution of this tracked function.
+    pub fn previous<'db>(
```
MichaReiser, line 15: I need to land my monomorphization PR, so that we can move this out of the `C` block :)

src/function/previous.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        if memo.header.may_be_provisional() {
+            return None;
+        }
+
+        let value = memo.value.as_ref()?;
+        let revisions = &memo.header.revisions;
+        zalsa_local.report_previous_memo_read(
+            revisions.durability,
+            revisions.origin().edges(),
```
MichaReiser, line 46: Hmm, I'm not sure this is sound. I have to think this through a bit more carefully.

There's no guarantee that the input edges from a previous revision are still valid. That is, some input edges may no longer exist because they were garbage collected, or were never deserialized after persistent caching. 

Carrying these dependencies forward could also mean that a query now depends on the same IDs that only differ b y generations.

src/function/previous.rs
```diff
@@ -0,0 +1,55 @@
...
+                "cannot access previous memoized value for {} while executing {database_key_index:?}",
+                C::DEBUG_NAME,
+            );
+        }
+
+        let id = database_key_index.key_index();
+        let memo_ingredient_index = self.memo_ingredient_index(zalsa, id);
+        let memo = self.get_memo_from_table_for(zalsa, id, memo_ingredient_index)?;
+
+        if memo.header.may_be_provisional() {
```
MichaReiser, line 38: Maybe it's best to exclude provisional memos for now, even if they're from the previous revision (it's not guaranteed that all memos participating in a cycle are finalized). 

For same-revisional provisionals, we'd have to carry over the cycle heads. Ultimately, this is the same as calling self again (fixpoint!). It's a bit less clear to me what needs to happen when depending on a provisional from a previous revision.

src/function/previous.rs
```diff
@@ -0,0 +1,55 @@
...
+        zalsa_local.report_previous_memo_read(
+            revisions.durability,
+            revisions.origin().edges(),
+            revisions.is_derived_untracked(),
+            revisions.tracked_struct_ids(),
+            #[cfg(feature = "accumulator")]
+            revisions.accumulated_inputs.load(),
+            zalsa.current_revision(),
+        );
+        Some(value)
```
MichaReiser, line 53: What's unclear to me is whether this is safe if `value` contains tracked structs that were created by the active query. You can now read tracked structs that are being updated as your query runs. In general, you can now read values that are potentially outdated, because they haven't been verified yet (or contain interned values that were garbage collected)

MichaReiser (commented):
Probably not what you intended it for, but I have a use case for this in ty to basically check: Has this been computed, if so, call the query, otherwise do something else.

I have to think this through a bit more carefully, but I don't think this is sound for non `'static` values. That is, any value that contains anything with a `'db` lifetime, because you can now read values as they're been updated or that haven't been verified yet (or GCed).


Another issue is that calling `previous` multiple times calls `seed_tracked_struct` multiples times.

MichaReiser:
I think restricting values to `'static` should solve the tracked struct and interned struct issue. I'm not sure how to solve the dependency issue. 

It's not even clear to me what the expected dependencies should be. Depending on all previous dependencies is an over-approximation. I also doesn't protect against missing new dependencies. E.g. a query could clone parts of the result and skip a struct read, which it normally would depend on if `previous` isn't provided. IT also has the issue that it's possible that the query is immediately out dated again because one of its dependency hasn't been verified. 

I sort of think that we may need to iterate over all dependencies and repeat some of `deep_verify_memo`, but in a slightly different way? 

* For interned: If an interned changed, it means it was garbage collected. We need to bail, because later edges could be queries that were stored on this interned struct (which no longer exists). We need to re-execute the query to recreate all interned structs. 
* Inputs: Fine, carry over? 
* Query: Do the same as `deep_verify_memo`. That is, rerun the query. This ensures that `maybe_changed_after` will return `false` right after the outer query completed (which also ensures that any tracked structs and interned structs created by that query are validated)


The part that remains unclear to me are tracked structs. But I'd be fine to limit this feature to return `None` if the query created any tracked structs. The issue with tracked structs is the same as with interned. We don't know if the tracked struct will be collected after the query completes. But collecting them would tear down queries that `previous` dependent on... 

To some extend, this reminds me of fixpoint where `previous` is a very educated guess of what the initial value is. Now, we can't use fixpoint because the results aren't guaranteed to be monotonic and this is also a stricter form where queries are only allowed to depend on their own previous result, other queries aren't allowed to form a cycle.

Veykril:
> I think restricting values to 'static should solve the tracked struct and interned struct issue.

Makes sense but then it becomes somewhat useful for r-a at least. Turns out this feature is a lot more complicated than I had hoped for (but I guess its also expected, given this is a form of user requested cycle handling)

PR #1221, opened 2026-06-30
refactor: generalize eviction policy interface

MichaReiser:
Generalize tracked-function eviction around insertion and usage events plus a revision context. This keeps the existing LRU behavior while allowing policies to inspect memo revisions and evict values without depending on table internals.

This should put most infrastructure (except some macro `eviction_policy(lru)`) in place, for generational and even within-revision garbage collection


MichaReiser:
No longer needed. We only need `record_use`, because the query API can't distinguish between an insert or use, whether it's one or the other depends on the eviction policy's state.

PR #1223, opened 2026-07-01
perf: avoid hashing query edge generations

MichaReiser:
Skip hashing query-edge generations, which are unique per ingredient and index within one query. Equality still distinguishes generations, so an invariant violation only creates a collision.

This removes unnecessary hash work and slightly reduces ty's simulated instruction count (<1%).


MichaReiser:
wow, the benchmarks are now even worse 😆

MichaReiser:
I reached out to CodSpeed. The benchmarks are now all over the place.

src/zalsa_local.rs
```diff
@@ -1682,6 +1682,17 @@ impl QueryEdge {
...
 }
 
+impl std::hash::Hash for QueryEdge {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        // A query should never depend on the same ingredient and index at different generations:
+        // advancing the generation replaces the previous identity. If this assumption is ever
+        // violated, `Eq` still compares the generation, so omitting it here only causes a hash
+        // collision. The ingredient's tag bit includes the edge kind in the hash.
+        state.write_u32(self.index);
+        state.write_u32(self.ingredient.as_u32());
```
Veykril:
```suggestion
    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
        let &Self { index, ingredient, generation: _ } = self;
        // A query should never depend on the same ingredient and index at different generations:
        // advancing the generation replaces the previous identity. If this assumption is ever
        // violated, `Eq` still compares the generation, so omitting it here only causes a hash
        // collision. The ingredient's tag bit includes the edge kind in the hash.
        state.write_u32(index);
        state.write_u32(ingredient.as_u32());
```
Even if the structure won't likely change, just in case (r-a has an assist for this you can invoke on `&self`)

MichaReiser:
@codspeedbot investigate the performance regression

PR #1227, opened 2026-07-02
Inline input field access helper

MichaReiser:
## Summary
- Add an inline hint to the input field access helper

## Notes
Local instruction-count inspection showed that inlining `input::IngredientImpl::field` removes the hot call from the input field read query body. Marking `data` / `data_raw` alone does not help, because the query body still calls `field`; interned field reads were already inlined in the benchmark body without additional hints.

## Verification
- cargo check


MichaReiser:
@codspeedbot analyze this regression

PR #1228, opened 2026-07-02
fix: replace tracked method db-arg index panic with proper compile error

### LINKED ISSUE #1058: `#[salsa::tracked]` panics when applied to a method missing database parameter
```rust
#[salsa::tracked]
impl Foo {
    #[salsa::tracked]
    pub fn foo<'db>(self) {}
}
```

```
error: custom attribute panicked
  --> src/lib.rs:1:1
   |
21 | #[salsa::tracked]
   | ^^^^^^^^^^^^^^^^^
   |
   = help: message: index out of bounds: the len is 0 but the index is 1
```

fuleinist:
﻿## Summary

When a #[salsa::tracked] method inside a #[salsa::tracked] impl has only self and no database parameter, the proc-macro panics with:

`
error: custom attribute panicked
  --> src/lib.rs:1:1
   |
21 | #[salsa::tracked]
   | ^^^^^^^^^^^^^^^^^
   |
   = help: message: index out of bounds: the len is 0 but the index is 1
`

## Fix

Replace the direct index access n_item.sig.inputs[db_input_index] with .iter().nth(db_input_index) + ok_or_else(), emitting a clear compile-time error instead of a panic:

`
error: tracked methods must have a database parameter after self
`

## Root cause

In alidity_check(), when is_method is true, db_input_index is set to 1 (skipping self). But if the method has no database parameter, inputs has only 1 element, causing the OOB panic.

## Testing

- All 122 existing tests pass
- Build verified with cargo check -p salsa-macros and cargo test

Closes #1058

### DIFF
--- a/components/salsa-macros/src/tracked_impl.rs
+++ b/components/salsa-macros/src/tracked_impl.rs
@@ -243,7 +243,7 @@ impl Macro {
             (None, 0, 1)
         };
 
-        let db_arg = fn_item.sig.inputs.iter().nth(db_input_index).ok_or_else(|| {
+        let db_arg = fn_item.sig.inputs.get(db_input_index).ok_or_else(|| {
             syn::Error::new_spanned(
                 &fn_item.sig,
                 "tracked methods must have a database parameter after `self`",
--- a/components/salsa-macros/src/tracked_impl.rs
+++ b/components/salsa-macros/src/tracked_impl.rs
@@ -243,7 +243,13 @@ impl Macro {
             (None, 0, 1)
         };
 
-        let (db_ident, db_ty) = self.check_db_argument(&fn_item.sig.inputs[db_input_index])?;
+        let db_arg = fn_item.sig.inputs.iter().nth(db_input_index).ok_or_else(|| {
+            syn::Error::new_spanned(
+                &fn_item.sig,
+                "tracked methods must have a database parameter after `self`",
+            )
+        })?;
+        let (db_ident, db_ty) = self.check_db_argument(db_arg)?;
 
         let input_ids: Vec<syn::Ident> =
             crate::fn_util::input_ids(&self.hygiene, &fn_item.sig, skipped_inputs);
--- a/components/salsa-macros/src/tracked_impl.rs
+++ b/components/salsa-macros/src/tracked_impl.rs
@@ -235,7 +235,15 @@ impl Macro {
             None => self.conjure_db_lifetime(impl_item, fn_item),
         };
 
-        let is_method = matches!(&fn_item.sig.inputs[0], syn::FnArg::Receiver(_));
+        // Use `first()` so a zero-argument method doesn't panic with
+        // "index out of bounds" here — the downstream `.get(db_input_index)`
+        // guard plus `check_db_argument`/`check_self_argument` already
+        // produce a clean compile error in that case.
+        let is_method = fn_item
+            .sig
+            .inputs
+            .first()
+            .is_some_and(|arg| matches!(arg, syn::FnArg::Receiver(_)));
 
         let (self_token, db_input_index, skipped_inputs) = if is_method {
             (Some(self.check_self_argument(fn_item)?), 1, 2)
--- a/tests/compile-fail/tracked_method_zero_args.stderr
+++ b/tests/compile-fail/tracked_method_zero_args.stderr
@@ -1,0 +1,11 @@
+error: tracked functions must have at least a database argument
+  --> tests/compile-fail/tracked_method_zero_args.rs:19:8
+   |
+19 |     fn no_args() {}
+   |        ^^^^^^^
+
+error: tracked methods must have a database parameter after `self`
+  --> tests/compile-fail/tracked_method_zero_args.rs:19:5
+   |
+19 |     fn no_args() {}
+   |     ^^^^^^^^^^^^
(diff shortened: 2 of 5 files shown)


Veykril:
Thanks! Do you mind adding a compile failure test for this as well? 

CI also wants a reformat

components/salsa-macros/src/tracked_impl.rs
```diff
@@ -243,7 +243,13 @@ impl Macro {
             (None, 0, 1)
         };
 
-        let (db_ident, db_ty) = self.check_db_argument(&fn_item.sig.inputs[db_input_index])?;
+        let db_arg = fn_item.sig.inputs.iter().nth(db_input_index).ok_or_else(|| {
```
Veykril:
```suggestion
        let db_arg = fn_item.sig.inputs.get(db_input_index).ok_or_else(|| {
```

fuleinist: Thanks — applied in cdbc2b6a3fbd4bfe9fb8b40fbccb9f50e2917b30: https://github.com/fuleinist/salsa/commit/cdbc2b6a3fbd4bfe9fb8b40fbccb9f50e2917b30

MichaReiser:
>Thanks! Do you mind adding a compile failure test for this as well?

This looks good once we have the compile failure test in place

components/salsa-macros/src/tracked_impl.rs
MichaReiser: Do we also need to protect against zero argument methods?

MichaReiser: Do we also need to protect against zero argument methods?

fuleinist:
Good catch — yes, the zero-arg case still panics, but earlier than the line I guarded. In `validity_check`:

```rust
let is_method = matches!(&fn_item.sig.inputs[0], syn::FnArg::Receiver(_));
```

When `sig.inputs` is empty this indexes out of bounds before reaching my `.get(db_input_index)` guard. For a zero-arg method `is_method` would be `false`, `db_input_index = 0`, and `check_db_argument(&sig.inputs[0])` would also panic.

Want me to extend the same defensive pattern there? Sketch:

```rust
let (self_token, db_input_index, skipped_inputs) = if !fn_item.sig.inputs.is_empty()
    && matches!(&fn_item.sig.inputs[0], syn::FnArg::Receiver(_))
{
    (Some(self.check_self_argument(fn_item)?), 1, 2)
} else {
    (None, 0, 1)
};

let db_arg = fn_item.sig.inputs.get(db_input_index).ok_or_else(|| {
    syn::Error::new_spanned(
        &fn_item.sig,
        "tracked methods must have a database parameter after `self`",
    )
})?;
let (db_ident, db_ty) = self.check_db_argument(db_arg)?;
```

That gives a clear compile-time error ("tracked method must take a database parameter") for both the self-only case (issue #1058) and the zero-arg case, instead of a proc-macro panic. Happy to push it as a follow-up commit on this branch if you agree.

fuleinist:
Done — pushed as `cc40e8a1`. Summary of what changed since the previous review round:

**Compile-fail regression test** (`tests/compile-fail/tracked_method_missing_db.rs`)
- Exact repro from issue #1058: a `#[salsa::tracked]` method with only `self` inside a tracked impl
- trybuild now locks in the clean compile error, so future refactors of `tracked_impl.rs` can't regress to a proc-macro `index out of bounds` panic
- The .stderr captures both expected errors: the generic "must also be applied to the impl block" from `tracked_fn::check_db_argument` and the new site-specific "tracked methods must have a database parameter after `self`" from this fix

**CI / formatting**
- `cargo fmt -p salsa-macros` is clean (no diff)
- `cargo build -p salsa-macros` and `cargo build -p salsa` both succeed
- `cargo test --test tracked_method` passes (2/2)
- `cargo test --test compile_pass --features inventory,persistence` passes
- `cargo test --test compile_fail --features inventory,persistence` passes — all 41 existing compile-fail cases plus the new `tracked_method_missing_db` case

No changes to the macro itself this round — the previous commit already used `.get()` instead of `.iter().nth()` per Veykril's review. Happy to address any further feedback.

components/salsa-macros/src/tracked_impl.rs
fuleinist: Great catch — confirmed, the zero-arg case was still panicking, but at the earlier 'inputs[0]' indexing site in validity_check, before reaching the .get() guard. Fixed in 0dc0335e9c7853c2ef04b092729c9f94e5e473e1 (use first().is_some_and(...)) + new trybuild test tracked_method_zero_args.rs: https://github.com/salsa-rs/salsa/pull/1228/commits/0dc0335e9c7853c2ef04b092729c9f94e5e473e1

fuleinist: Same fix as above (duplicate comment, replying for the thread record): 0dc0335e9c7853c2ef04b092729c9f94e5e473e1 https://github.com/salsa-rs/salsa/pull/1228/commits/0dc0335e9c7853c2ef04b092729c9f94e5e473e1

tests/compile-fail/tracked_method_zero_args.stderr
```diff
@@ -0,0 +1,11 @@
+error: tracked functions must have at least a database argument
+  --> tests/compile-fail/tracked_method_zero_args.rs:19:8
+   |
+19 |     fn no_args() {}
+   |        ^^^^^^^
+
+error: tracked methods must have a database parameter after `self`
```
MichaReiser, line 7: Nit: there's no `self` in this example, but I think this is fine

MichaReiser (approved):
Thank you

PR #1233, opened 2026-07-04
chore: rework eviction benchmarks

MichaReiser:
Update our LRU benchmarks with a set of benchmarks that we can run across different eviction policies (and compare them on how well they deal with certain access patterns).

I did remove the benchmarks where the cached query set was way over capacity. It was not clear to me how these are valuable when comparing different algorithms, because there are only poor choices and it will always be required to recompute some (if not all) values. I consider this a configuration error (something that would be nice if Salsa could surface somehow)

I added NoEviction as a baseline (I don't fully understand why `disabled_eviction[Lru]` is faster xD)


MichaReiser:
FYI: The weird Codspeed results are a confirmed bug. They're working on a bug fix.

Veykril (approved):

MichaReiser:
Hmm, simulation runs take too long for some of the benchmarks. I had to move them back to walltime.

PR #1234, opened 2026-07-04
perf: cache table page metadata during memo eviction

MichaReiser:
Cache the last resolved table page while evicting memo values, avoiding repeated page, vtable, memo-index, and type lookups for page-local victim streams. It slightly reduces the number of instructions on the eviction path for LRU, but it's a bigger impact for SIEVE where the eviction policy intentionally evicts items in their page order. 

This keeps eviction policies storage-agnostic while preserving arbitrary ID ordering; the isolated sweep prototype improved by roughly 15%.

## Benchmarks

Local benchmarks, because CodSpeed is bugged right now

Benchmark | #1234 on LRU | #1234 on SIEVE (#1226)
-- | -- | --
fast_path | −0.2% | +0.8%
fast_path_and_sweep | +0.6% | +0.4%
fill_and_evict | +0.4% | −0.1%
one_hit_wonders | −0.1% | −2.0%
phase_change | −0.1% | −0.9%
project_check_then_incremental | +1.1% | −0.1%
scan_resistance | −1.4% | −1.5%
Policy-specific instructions | −1.39% | −4.39%

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -793,8 +793,8 @@ mod tests {
     ) -> NonNull<TestMemo> {
         let memo = NonNull::from(Box::leak(Box::new(TestMemo(value))));
         // SAFETY: Test slots ignore the revision, and `memo_index` matches the page's types.
-        let old =
-            unsafe { table.memos::<TestSlot>(id, Revision::start()) }.insert(memo_index, memo);
+        let memos = unsafe { table.memos::<TestSlot>(id, Revision::start()) };
+        let old = memos.insert(memo_index, memo);
         assert!(old.is_none());
         memo
     }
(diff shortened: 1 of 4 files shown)


src/table.rs
```diff
@@ -609,3 +690,141 @@ pub fn split_id(id: Id) -> (PageIndex, SlotIndex) {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::DatabaseKeyIndex;
+    use crate::table::memo::{MemoEntryType, MemoTable};
+    use crate::zalsa::Zalsa;
+
+    #[test]
+    fn memo_cursor_caches_pages_with_ingredient_specific_memo_indices() {
```
MichaReiser, line 702: I'm leaning towards deleting this test

Issue #1237, opened 2026-07-08
Cancellation loses tracked-struct identities, leaking slots on retry


MichaReiser:
A tracked query that creates a tracked struct and is then canceled by a pending write leaves the allocated struct alive, but loses the identity metadata needed to reuse or delete it.

The active frame records the identity in tracked_struct_ids when the struct is allocated ([tracked_struct.rs (line 501)](/tmp/salsa-cancellation-epoch-fix/src/tracked_struct.rs:501)). During cancellation unwinding, `ActiveQueryGuard::drop` pops the frame, and [`ActiveQuery::clear` (line 264)](/tmp/salsa-cancellation-epoch-fix/src/active_query.rs:264) discards that identity map. Ordinary canceled queries do not insert a memo containing the partial output metadata.

Consequently:

* The tracked-struct slot remains live.
* No memo or active query owns its identity.
* An identical retry cannot find it and allocates a second slot.

The following regression test creates one tracked struct, cancels the query without advancing the revision, and then retries with identical inputs. The number of live Tracked entries grows from one to two; it should remain one.

<details><summary>Regression test</summary>

```rust
// Shuttle doesn't like panics inside of its runtime.
#![cfg(all(not(feature = "shuttle"), feature = "inventory"))]

//! A cancelled query must retain enough output metadata to reuse its tracked structs on retry.

use salsa::plumbing::ZalsaDatabase;
use salsa::{Cancelled, Database};

use crate::setup::{Knobs, KnobsDatabase};
use crate::sync::thread;

#[salsa::input]
struct Input {
    #[returns(copy)]
    value: u32,
}

#[salsa::tracked]
struct Tracked<'db> {
    #[returns(copy)]
    value: u32,
}

#[salsa::tracked(returns(copy))]
fn create_tracked(db: &dyn KnobsDatabase, input: Input) -> Tracked<'_> {
    let tracked = Tracked::new(db, input.value(db));

    db.signal(1);
    db.wait_for(2);
    cancellation_point(db, input);

    tracked
}

#[salsa::tracked]
fn cancellation_point(db: &dyn KnobsDatabase, input: Input) {
    input.value(db);
}

#[test]
fn retry_reuses_tracked_struct_created_by_cancelled_execution() {
    let db = Knobs::default();
    let db_writer = db.clone();
    let db_worker = db.clone();
    let db_waiter = db.clone();
    let input = Input::new(&db, 22);

    db.signal_on_did_cancel(2);
    drop(db);

[...]

PR #1238, opened 2026-07-08
fix: restart fixpoint iteration after cancellation

### LINKED ISSUE #3924: `fetch_cold_cycle` should have inserted a provisional memo
### Summary

Pretty regularly when I start VS Code fresh many of my scripts get highlighted red in the "Explorer" pane. Normally I click into one of them, click an offending line, then everything updates and all the red goes away. 

Generally if I make the red go away, then re-start VS-code right away, the red doesn't show up on the re-start.

Image
Image

### Version

0.0.38

MichaReiser:
It turns out, I missed a few places where we have to check for cancellations in https://github.com/salsa-rs/salsa/pull/1100. 

If we have:

* `query_a` -> `query_b` -> `query_a` (initial)
* `query_b` completes, 
* trigger cancellation, `query_a` unwinds before completing 
* Call `query_a` again. It sees a different cancellation count and re-executes
* `fetch_cold` now verifies `query_b`. `validate_provisional` returns true because `query_b` and the provisional memo of `query_a` have the same iteration (and cancellation count). `query_b` returns without re-executing, bypassing the insertion of a new initial value for `query_a`! 
* `query_a` sees that it's a cycle head and tries to read the provisional value. But the value of that memo is `None` -> panic.

This PR adds a check to never return `true` from `validate_provisional` if the memo is from a cancelled execution. 

This PR also adds a check that prevents a call to `seed_iteration` if the memo is from a different cancellation. This can result in leaking tracked structs, but this is an issue that applies equally to normal queries, see https://github.com/salsa-rs/salsa/issues/1237

Fixes [astral-sh/ty#3924](https://github.com/astral-sh/ty/issues/3924)

### DIFF
--- a/src/function/execute.rs
+++ b/src/function/execute.rs
@@ -135,21 +135,31 @@ where
         let mut last_provisional_memo_opt: Option<&Memo<'db, C>> = None;
 
         let mut last_stale_tracked_ids: Vec<(Identity, Id)> = Vec::new();
+        let current_revision = zalsa.current_revision();
         let cancellation_count = zalsa.runtime().cancellation_count();
+        let mut opt_old_memo = opt_old_memo;
         let mut iteration = IterationStamp::initial(cancellation_count);
 
+        // An ordinary query doesn't memoize a cancelled execution. Match that behavior for
+        // fixpoint queries: a memo from an abandoned cancellation epoch in this revision doesn't
+        // seed the retry, while a memo from an older revision remains useful for backdating and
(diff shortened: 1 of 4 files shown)


MichaReiser:
Unfortunately, the benchmarks are broken right now.

carljm (approved):
This looks great, thank you!

Taking your word for it that the benchmarks are broken and this isn't actually a massive regression 😆

MichaReiser:
We'll know when updating salsa 😆

PR #1240, opened 2026-07-09
chore: merge CodSpeed benchmark jobs

MichaReiser:
Build and run CodSpeed benchmarks in one walltime-runner job. This should ensure that all runs build with the same glibc version. 

The original motivation for running build as a separate job was that walltime runners are billed by minute. However, building the salsa benchmarks with a warm cache from master only takes ~30s. Uploading and downloading the build artifacts takes about 20s. We should also move to the pro tier soon, with sponsored runner minutes. So that it's less critical that we stay within the 500 free minutes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,11 +126,12 @@ jobs:
       - name: Test with Shuttle
         run: cargo nextest run --features shuttle --test parallel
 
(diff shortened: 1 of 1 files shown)


MichaReiser:
@codspeedbot what could be the reason that some of the atomic operations now take so much longer?

PR #1204, opened 2026-06-27
refactor: erase configuration from interned storage

MichaReiser:
Erases configuration from interned storage and cold paths to reduce downstream monomorphization while retaining specialized reuse checks on the successful-hit path.

Measured on ty:

* total LLVM lines fall from 3,284,622 to 3,096,046 (5.74%)
* intern-related LLVM lines from 276,433 to 216,077 (21.83%)
* DateType instructions from 674,414,109 to 673,205,080 (0.179%).
* The Salsa `new[InternedInput]` simulation improves from 598 to 597 instructions
* the debug ty binary falls by 3.38%
* the stripped release binary by 1.25%
* the maturin wheel by 0.84%.

PR #1244, opened 2026-07-09
fix: Avoid unnecessary padding for interned values

MichaReiser:
https://github.com/salsa-rs/salsa/pull/1204 regressed the storage requirement for interned structs, because of `ValueHeader`'s 7 byte padding (`durability` is only one byte). Before, Rust could use that padding to store some of `C::Fields`. 

This PR brings this back by keeping `durability`, `memos`, and `shard` on `Value` itself. `ValueHeader` gets replaced by a padding-free `LruEntry`. `LruEntry` stores the metadata that is needed to make a "reuse" decision.

### DIFF
--- a/src/interned.rs
+++ b/src/interned.rs
@@ -119,7 +119,7 @@ struct IngredientShard {
     key_map: hashbrown::HashTable<Id>,
 
     /// An intrusive linked list for LRU.
-    lru: LinkedList<ValueHeaderAdapter>,
+    lru: LinkedList<LruEntryAdapter>,
 }
 
 impl Default for IngredientShard {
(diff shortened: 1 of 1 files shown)


src/interned.rs
```diff
@@ -793,41 +812,41 @@ where
...
-                let header = UnsafeRef::into_raw(header);
+            while let Some(entry) = cursor.as_cursor().clone_pointer() {
+                let entry = UnsafeRef::into_raw(entry);
 
-                // SAFETY: The caller guarantees that `header` points to a live value in this shard
-                // and that we hold the shard lock, which grants exclusive access to `shared`.
-                let value_shared = unsafe { &mut *(*header).shared.get() };
+                // SAFETY: The caller guarantees that `entry` points to a live value in this shard
+                // and that we hold the shard lock, which grants exclusive access to `metadata`.
+                let metadata = unsafe { &mut *(*entry).metadata.get() };
```
MichaReiser, line 828: The code here only requires `metadata`. That's the semantic grouping for what's in `LruEntry`

MichaReiser:
@ibraheemdev it would be great to get this one in, to unblock the next Salsa release, which in turn allows us to pull in a panic bug fix.

ibraheemdev (approved):

MichaReiser:
Thank you

PR #1255, opened 2026-07-17
fix: make immortal interned values cheaper

MichaReiser:
Main benefit; It also reduces the size of each interned value because we avoid storing the unnecessary Lru overhead.

Maps `revisions = usize::MAX` to a compact compile-time no-eviction policy while keeping LRU as the default and preserving the existing macro API.
Removes unused LRU state and bookkeeping for immortal values; simulation improves hot interning by 32.6% and cold interning by 13.6%, with no LRU regression.
Testing: Added behavioral, persistence, layout, and performance coverage; full suites and lint checks pass.

### DIFF
--- a/benches/creation.rs
+++ b/benches/creation.rs
@@ -42,6 +42,22 @@ fn intern_distinct_many(bencher: divan::Bencher) {
         });
 }
 
+/// Interns many distinct values with garbage collection disabled.
+#[divan::bench]
+fn intern_immortal_distinct_many(bencher: divan::Bencher) {
+    bencher
+        .with_inputs(|| {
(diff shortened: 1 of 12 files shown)


ChayimFriedman2:
Does ty have statically immortal interned values? In r-a we use `revisions = usize::MAX` but that's mostly because we haven't completely figured out the story around the GC yet.

If nobody needs this, it does not make sense to optimize this.

MichaReiser:
We currently don't. But it seems wasteful to track all LRU metadata for interned values that have LRU disabled. But we can consider removing `revisions = usize::MAX` altogether.

PR #1260, opened 2026-07-21
Add support for self-ref fields to the interned macro

Veykril:
Completely forgot to add it to the macro since r-a had this hand rolled for 2 years now

### DIFF
--- a/components/salsa-macros/src/interned.rs
+++ b/components/salsa-macros/src/interned.rs
@@ -84,6 +84,8 @@ impl SalsaStructAllowedOptions for InternedStruct {
     const ALLOW_DEFAULT: bool = false;
 
     const ALLOW_MANUAL_RETENTION_PROOF: bool = true;
+
+    const ALLOW_SELF_REF: bool = true;
 }
 
 struct Macro {
@@ -103,16 +105,7 @@ impl Macro {
         let struct_data_ident = format_ident!("{}Data", struct_ident);
         let db_lt = db_lifetime::db_lifetime(&self.struct_item.generics);
         let new_fn = salsa_struct.constructor_name();
-        let field_ids = salsa_struct.field_ids();
-        let field_indices = salsa_struct.field_indices();
         let num_fields = salsa_struct.num_fields();
-        let field_vis = salsa_struct.field_vis();
-        let field_getter_ids = salsa_struct.field_getter_ids();
-        let field_options = salsa_struct.field_options();
-        let field_tys = salsa_struct.field_tys();
-        let field_manual_retention_proofs = salsa_struct.field_manual_retention_proofs();
-        let field_indexed_tys = salsa_struct.field_indexed_tys();
-        let field_unused_attrs = salsa_struct.field_attrs();
         let generate_debug_impl = salsa_struct.generate_debug_impl();
         let has_lifetime = salsa_struct.generate_lifetime();
         let id = salsa_struct.id();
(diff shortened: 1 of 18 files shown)


Veykril:
Turns out this feature was very bugged with the LRU we added (fortunately r-a disabled LRU for this)

Veykril:
Also unsure about the name of this tbh. These late fields are similar to untracked ones in that they do not add to the identity of the interned.

components/salsa-macros/src/lib.rs
```diff
@@ -150,7 +152,8 @@ pub fn db(args: TokenStream, input: TokenStream) -> TokenStream {
 /// - `constructor = IDENT` renames the generated constructor from `new` to `IDENT`.
 /// - `debug` implements [`Debug`] using the field values when a database is attached to the current
 ///   thread. The generated `default_debug_fmt` method can also be called from a manual [`Debug`]
-///   implementation.
+///   implementation. This option cannot be combined with `#[late_init]` fields because they may
+///   introduce cycles in the generated formatting.
```
MichaReiser: It would be nice if those fields would be excluded instead of removing `debug` entirely or that we only print the field name, but without its value

components/salsa-macros/src/lib.rs
```diff
@@ -192,14 +195,20 @@ pub fn db(args: TokenStream, input: TokenStream) -> TokenStream {
 ///   [`salsa::SalsaAsDeref`] to return borrowed forms such as `Option<&T>` and
 ///   `Option<&T::Target>`. Every borrowed result is tied to the database borrow.
 /// - `#[get(IDENT)]` renames the generated getter.
+/// - `#[late_init]` removes the field from the interned key and changes its constructor argument to
+///   `impl FnOnce(Self) -> FieldTy`. Salsa invokes the closure with the newly allocated handle when
+///   the key is first interned; it does not invoke the closure when the key already exists. Salsa
+///   runs the closure while holding an interned-table shard lock. The closure should only compute
+///   from captured data and the supplied handle; accessing the database may deadlock if it attempts
+///   to intern or allocate a value.
```
MichaReiser: Can you say more about r-as use case. I'm a bit worried about accepting an arbitrary handle, as it's very easy to mess up, as the documentation correctly points out (there's also the issue that reading any fields will not be correctly recorded as dependencies if the entry is already cached. 

Could we limit the feature to always make it store `Self`? Or are there cases where the cyclic structure is deeper? If so, how do we avoid over/under-tracking dependencies?

Veykril: r-a specifically needs it for self-referentiality. So yea we could simplify the interface here to only allow that use-case.

MichaReiser: I'd prefer that, if that's sufficient for r-a. And we can generalize it, when someone has a need for it.

Veykril: yea I agree, I already made that change fwiw

MichaReiser: Oh, I didn't notice that. I'll take a look (might not be today, I've been out the last few days and I've too many notifications to catch up to)

components/salsa-macros/src/interned.rs
```diff
@@ -112,11 +114,18 @@ impl Macro {
...
         let generate_debug_impl = salsa_struct.generate_debug_impl();
         let has_lifetime = salsa_struct.generate_lifetime();
         let id = salsa_struct.id();
-        let revisions = salsa_struct.revisions();
+        let revisions = if has_self_ref {
+            quote!(usize::MAX)
+        } else {
+            let revisions = salsa_struct.revisions();
+            quote!(#(#revisions)*)
+        };
```
MichaReiser: What's the reason that we need to disable LRU collection for self referential structs?

components/salsa-macros/src/lib.rs
```diff
@@ -141,6 +141,12 @@ pub fn db(args: TokenStream, input: TokenStream) -> TokenStream {
 /// [`Clone`] + [`Eq`] + [`Hash`] + [`Send`] + [`Sync`]. A field whose type is unconditionally
 /// `'static` is accepted directly; any other field must implement [`salsa::SalsaValue`].
 ///
+/// A field marked `#[self_ref]` does not contribute to identity and becomes an [`Option<T>`]
+/// constructor parameter. [`Some<T>`] stores the supplied value, while [`None`] stores the interned
+/// value being constructed. If the identity is already interned, the constructor returns the
+/// existing value without changing its self-referential fields. Interned types containing these
+/// fields disable slot reuse so that stored references remain valid.
```
MichaReiser: Can you say more why this is necessary? How is it different from any other field that stores another interned struct? I'd expect that the `'db` lifetime prevents you from holding a reference to the interned struct, which also prevents you from reading the field. If the value's different, the interned should get a different id.

After chatting with codex. The main issue seems to be that we exclude the `self_ref` field from hashing, but users can also provide `Some(value)`, where there's no guarantee that `value` is indeed self. What's the reason that r-a needs the `Some(value)` constructor where `value != self`?

components/salsa-macros/src/interned.rs
Veykril: Ah I think this is leftover from the previous iteration. I think LRU ought to be fine for self-refs

components/salsa-macros/src/lib.rs
Veykril: We use this for modeling hygiene which is effectively a graph where the only cycles can be self-referential

See https://github.com/rust-lang/rust-analyzer/blob/master/crates/span/src/hygiene.rs
and the constructor call site https://github.com/rust-lang/rust-analyzer/blob/33815d5957836f6eff67f91aced8420ec5a377d3/crates/hir-expand/src/hygiene.rs#L110-L143

Only allowing the field to actually point to itself would make the feature kind of pointless

MichaReiser: Lol, right. If it's always self, then what's the point of it :)

If the field is `!= self`, shouldn't it then be part of the interning key? Or what's the use case where two interned structs with different `self_ref` (or an instance where `self_ref == self` and one where `self_ref != self`). I think that would fix the LRU issue. I think that also suggest that `Debug` should print the `self_ref` unless it is `self` (in which case we can print `<self>`).

Veykril: Well, it just happens to be the case that for our purposes here, only the non-self ref keys are supposed to contribute to the identity. Or rather, the self-ref fields have no need to contribute as due to the way we construct these the self-ref fields will always be the same for the given non-self ref fields.

I can take a look at the key contribution later, I don't remember whether there was an implementation reason or whether it's just that we didn't need it. Though it would be nice if we could keep the key contribution excluded somehow for that minor perf improvement. I think hygiene accesses can be hot in r-a

MichaReiser: Couldn't you solve this in r-a by having a wrapper type that overrides `Eq` and `Hash` to always be `true`/0?

MichaReiser:
Sorry for the many questions. We can also move this to Zulip if it's easier. How's the `self_ref` different from:

* Storing a `parent: Option<Node<'db>>` on `Node<'db>`
* Having a public accessor method on `Node<'db>` that basically is `self.parent(db).unwrap_or(self)`
* R-a overrides/customizes `Hash`/`Eq` so that `parent` is excluded from hashing?

Veykril:
no worries

> Storing a parent: Option<Node<'db>> on Node<'db>

We used to do this, but it required a lot of annoying branching everywhere iirc (it's been a while). 
If I recall, this also results in some extra database hits / accesses (for something that can be fairly hot).

MichaReiser:
>We used to do this, but it required a lot of annoying branching everywhere iirc (it's been a while).

It should be limited to the one method that does the parent or self lookup. All callsites should remain unchanged. But yes, the compiled code will contain more branching. 

> If I recall, this also results in some extra database hits / accesses (for something that can be fairly hot).

I don't think that should be the case. Both implementations read exactly one field.

Veykril:
I can try cooking up a branch on r-a that tries to go the previous route again (and also read up on why I wanted this change). I can tell you'd like to get rid of the support code here :^) (which is fair if this feature turns out to not be necessary after all)

MichaReiser:
I don't mind the feature on its own, if there are cases where it simplifies downstream code. But if it's something that can be implemented downstream without much complexity, it seems desirable not to add more complexity to salsa

ChayimFriedman2:
Instead of doing all of this, isn't it better to just expose `intern()` that takes a callback somehow?

(Also why doesn't it play well with interneds GC?)

Veykril:
Seems like we don't have a good reason for this anymore and I don't recall why this was necessary before back when I implemented it.

PR #1261, opened 2026-07-22
feat: Add conditional SalsaValue derive proofs

Veykril:
I am not a fan of using the new escape hatch, as usually the reason for having to use it is that the field contains a foreign type you cannot implement the trait for. This implements something I added for the previous `Update` derive that allows changing the required bounds of the emitted impl. That way one can still rely on parts of the typesystem and only escaping specific types in the fields.

### DIFF
--- a/book/src/plumbing/db_lifetime.md
+++ b/book/src/plumbing/db_lifetime.md
@@ -218,14 +218,27 @@ Salsa handles do implement it because access to their data goes back through the
 state.
 
 The derive supports types with at most one lifetime parameter, as well as type and const
-parameters. Generated implementations require generic field types to implement `SalsaValue`;
-types that need different bounds can provide a manual implementation instead.
+parameters. Generated implementations require generic field types to implement `SalsaValue`.
+When an unmodifiable field type cannot implement `SalsaValue`, a conditional proof can replace the
+field check with narrower predicates:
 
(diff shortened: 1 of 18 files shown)


book/src/plumbing/db_lifetime.md
```diff
@@ -218,14 +218,27 @@ Salsa handles do implement it because access to their data goes back through the
...
+
+The predicates are added to the generated implementation, so `QueryValue<T>` only implements
+`SalsaValue` when `T` does. The compiler verifies that premise; the author asserts that the premise
+implies `ForeignContainer<T>` remains valid when Salsa retains it across revisions and rebinds its
+database lifetime.
+
+An unconditional `#[salsa_value(unsafe(prove_safe_to_retain_manually))]` proof skips the structural
+check without adding predicates. It is supported by `derive(SalsaValue)` and directly on tracked and
+interned struct fields. The author must ensure the field is safe for every generic instantiation
+accepted by the enclosing type.
```
MichaReiser: It's not entirely clear to me from the documentation where the new `unsafe(prove: T: SalsaValue))` is supported. Is it only `derive(SalsaValue)` or is it also supported by tracked structs, interned and inputs?

MichaReiser (approved):
This is a great idea.

Veykril:
I added support for tracked and interned as well and touched up the docs

MichaReiser (approved):

PR #1263, opened 2026-07-24
fix: propagate poisoned cycle heads

MichaReiser:
The root cause is a race between:

1. one thread unwinding from a panic and replacing a provisional cycle memo with a poisoned, iteration-0 memo; and
2. another thread still completing an overlapping cycle using that same cycle head at iteration 2.

The second thread then tries to merge the two views of the cycle head:

```text
active cycle:   head @ iteration 2
poisoned memo:  head @ iteration 0
```

Salsa assumes a cycle head has one consistent iteration, so the merge asserts.

The fix detects that the memo was poisoned and propagates the panic instead of treating its reset state as a valid provisional cycle result. This is a logical scheduling race, not a Rust data race.

### DIFF
--- a/tests/parallel/cycle_merge_after_panic.rs
+++ b/tests/parallel/cycle_merge_after_panic.rs
@@ -1,0 +1,122 @@
+//! A poisoned cycle head must not be merged with an earlier provisional iteration.
+//!
+//! This models the overlapping attribute/range cycles from astral-sh/ty#4077. Each scope contains
+//! sixteen expressions in chunks of four, giving the following dependency shape:
+//!
+//! ```text
+//! attribute(scope, index, false)
+//! ├─ attribute(scope, small_index, true)
+//! └─ once the provisional value reaches 3:
+//!    attribute(other_scope, late_index, true)
+//!    └─ range(other_scope, preceding_chunk)
+//!       └─ attribute(other_scope, chunk_expressions, true)
+//! ```
+//!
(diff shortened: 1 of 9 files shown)


tests/parallel/cycle_merge_after_panic.rs
```diff
@@ -0,0 +1,122 @@
...
+            (first + 1).min(FIXPOINT_LIMIT)
+        }
+    }
+}
+
+#[salsa::tracked(
+    returns(copy),
+    cycle_fn = |_, _, _, _, _, _| panic!("range cycle"),
+    cycle_initial = |_, _, _, _| ()
+)]
```
carljm, line 68: In the ty case, we had a query without fixpoint support, rather than a query whose fixpoint support panics. Curious why you chose the latter here. Should we test both?

carljm (approved):
Nice, thank you!

tests/parallel/cycle_merge_after_panic.rs
MichaReiser: I did test both. It doesn't make a difference.

carljm: (I meant "test" as in "automated test", but I don't think it matters.)

MichaReiser: I know. I can add it, but, given that it doesn't make a difference, thought a shorter test is more valuable.

PR #1271, opened 2026-08-01
Remove dead code from `HashEqLike`

Veykril:


Veykril:
Folded into https://github.com/salsa-rs/salsa/pull/1260

PR #1273, opened 2026-08-01
perf: shard tracked query synchronization by key

MichaReiser:
Shard tracked-query synchronization by hashed key so unrelated queries can be claimed and released in parallel. Reuse each query hash for shard selection, synchronization-table lookup, and release. 

We only use one shard per CPU because there's a memory and performance trade-off (applications using less than `available_parallelism` pay the overhead of more hash maps) and using twice or four times as many shards (as we do for interning), didn't show a meaningful improvement. I suspect that this is largely due to our locks being very short held. 

We don't use a `DashMap` because we don't need a `Read/Write` lock, the vast majority of operations require an exclusive lock, making a `Mutex` the right choice.  

Pydantic wall time, mean of eight randomized, interleaved runs:

| Threads | Before | After |
| ---: | ---: | ---: |
| 1 | 2,191 ms | 2,216 ms |
| 2 | 1,356 ms | 1,372 ms |
| 4 | 823 ms | 793 ms |
| 8 | 564 ms | 544 ms |
| 16 | 486 ms | 475 ms |
| 32 | 490 ms | 462 ms |
| 64 | 536 ms | 473 ms |

This was measured on a host with 32 physical and 64 logical core (which is why 64 is slightly slower). There are also some outlier files, which is why the runtime plateaus with more threads. 

At 64 threads, system CPU time falls from 564 ms to 210 ms and voluntary context switches fall from approximately 33,000 to 7,500.

### DIFF
--- a/src/function/sync.rs
+++ b/src/function/sync.rs
@@ -9,7 +9,7 @@ use crate::runtime::{
     BlockOnTransferredOwner, BlockResult, BlockTransferredResult, Running, WaitResult,
 };
 use crate::sync::thread::{self};
-use crate::sync::{Mutex, shard_count};
+use crate::sync::{Mutex, max_parallelism};
 use crate::tracing;
 use crate::zalsa::Zalsa;
 use crate::{Id, IngredientIndex};
@@ -62,8 +62,10 @@ pub(crate) struct SyncState {
 
 impl SyncTable {
     pub(crate) fn new(ingredient: IngredientIndex) -> Self {
+        let shard_count = max_parallelism().next_power_of_two().max(2);
+
         Self {
-            syncs: (0..shard_count()).map(|_| CachePadded::default()).collect(),
+            syncs: (0..shard_count).map(|_| CachePadded::default()).collect(),
             ingredient,
         }
     }
--- a/src/function/sync.rs
+++ b/src/function/sync.rs
@@ -1,5 +1,7 @@
-use rustc_hash::FxHashMap;
+use crossbeam_utils::CachePadded;
+use rustc_hash::{FxBuildHasher, FxHashMap};
 use std::collections::hash_map::OccupiedEntry;
+use std::hash::BuildHasher;
 
 use crate::key::DatabaseKeyIndex;
 use crate::plumbing::ZalsaLocal;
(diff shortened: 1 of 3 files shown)


src/function/sync.rs
```diff
@@ -59,12 +62,26 @@ pub(crate) struct SyncState {
 
 impl SyncTable {
     pub(crate) fn new(ingredient: IngredientIndex) -> Self {
+        let shard_count = max_parallelism().next_power_of_two().max(2);
```
MichaReiser, line 75: I also tested using `max_parallelism() * 4` (which is what dashmap uses and what we use in the interner code). However, the additional shards didn't meaningfully improve performance on ty, and more shards come at a higher memory usage cost (and more hash maps, see codspeed benchmarks)

src/sync.rs
```diff
@@ -1,5 +1,15 @@
 pub use shim::*;
 
+pub(crate) fn max_parallelism() -> usize {
+    static MAX_PARALLELISM: std::sync::OnceLock<usize> = std::sync::OnceLock::new();
+
+    *MAX_PARALLELISM.get_or_init(|| {
+        std::thread::available_parallelism()
+            .map(usize::from)
+            .unwrap_or(1)
```
MichaReiser, line 9: Should we respect `RAYON_NUM_THREADS`?

src/function/sync.rs
```diff
@@ -35,7 +39,10 @@ pub(crate) enum ClaimResult<'a, Guard = ClaimGuard<'a>> {
     },
 }
 
+#[derive(Debug)]
 pub(crate) struct SyncState {
+    key: Id,
```
MichaReiser, line 50: This used to be the `HashMap` key, it is now stored on `SyncState`, so that we can still compare if it's the same id.

Veykril (approved):

src/sync.rs
Veykril: I'd say no, this is unrelated to rayon after all.

src/function/sync.rs
```diff
@@ -303,8 +346,10 @@ pub enum SyncOwner {
 #[must_use]
 pub(crate) struct ClaimGuard<'me> {
```
Veykril, line 352: Our various guard types are getting quite big, I wonder if that is regressing perf overall for us (not necessarily just for this PR but in general)

MichaReiser: That's a good point. I remember that I tried boxing it, but that outweighted the cost. Let me see if we can avoid increasing the size in this PR

PR #1280, opened 2026-08-07
feat: expose revision cancellation receivers

MichaReiser:
Expose a Crossbeam receiver that disconnects when Salsa cancels the current revision.
Allow blocking computations to wait for cancellation without retaining database snapshots or registering callbacks.
Testing: Ran the full Salsa test suite and minimal-feature tests.


MichaReiser:
Hmm, I think what we want here is a broadcast channel (which crossbeam does not support :()

PR #1283, opened 2026-08-10
bump MSRV to 1.88

MichaReiser:
Bump the workspace MSRV from Rust 1.85 to 1.88.

trybuild now requires 1.88; it also gives us dyn trait upcasting and let chains. Rust 1.88 was released in June 2025.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,17 +13,21 @@ concurrency:
 
 jobs:
   test:
-    name: Test
+    name: ${{ matrix.name }}
     strategy:
       matrix:
-        rust:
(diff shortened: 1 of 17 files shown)


MichaReiser:
@Veykril do I remember it correctly that r-a uses something like stable - 2 as MSRV?

Veykril:
Compilation target is effectively latest, we don't mind being aggressive with bumping.

MichaReiser:
Alright, I think we should then merge this to unblock the try-build upgrade and let-chains are nice.

Veykril (approved):

PR #1284, opened 2026-08-10
refactor: replace generic database access with trait upcasting

MichaReiser:
Use stabilized trait upcasting to simplify database APIs and remove obsolete casting shims.

I upgraded ty to see if this requires any downstream changes, and it does. The `cycle` functions may now need type annotations for `db`. I think that's acceptable. 

But I'm happy to go either way (remove the FIXME, use trait upcasting)


MichaReiser:
ty's performance on this is neutral

PR #1287, opened 2026-08-12
perf: Allow toggling events off statically

ChayimFriedman2:
They are typically only used for tests, and removing them statically removes a load+branch and sometimes more (the callback can pessimize compiler optimizations).

This will fail tests as they need this. I'll fix this, but this will prevent seeing the benchmark impact, so I'm first opening the PR without.


ChayimFriedman2:
Seems insignificant. But maybe still worth it given how easy this is? Not sure.

MichaReiser:
I suspect that it's not significant, but the best way to tell is to run some benchmarks on ty/rust-analyzer to see if there's a meaningful difference. If not, I'm inclined to keep as is.

PR #1288, opened 2026-08-13
perf: Do not avoid packing QueryEdges if it has output edges

ChayimFriedman2:
This sacrifices one bit previously reserved for the IngredientIndex. rust-analyzer has many queries with output edges, but only 164 ingredients in total, so 2^11=2048 ingredients is way more than enough.

This saves a few megabytes for rust-analyzer.

### DIFF
--- a/src/zalsa.rs
+++ b/src/zalsa.rs
@@ -91,6 +91,7 @@ impl IngredientIndex {
     ///
     /// The index must be less than or equal to `IngredientIndex::MAX_INDEX`.
     pub(crate) const unsafe fn new_unchecked(v: u32) -> Self {
+        debug_assert!(v <= Self::MAX_INDEX);
         Self(v)
     }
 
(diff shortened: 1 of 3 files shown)


MichaReiser:
I'm interested now, what are the cases where ra heavily uses specify?

One other downside of this is that determining outputs now requires iterating when it was a cheap tag check before

Veykril:
r-a does not use specify at all

MichaReiser:
It's then not clear to me why this results in a memory improvement

ChayimFriedman2:
Aren't outputs also created internds/tracked structs?

ChayimFriedman2:
It seems I was wrong. I indeed wondered why there isn't a bigger saving. Apparently the difference was noise.

MichaReiser:
We used to use outputs for tracked structs but that’s no longer the case.

PR #1289, opened 2026-08-13
perf: If a memo has one packed edge and no extra data, store it inline

ChayimFriedman2:
rust-analyzer has ~25% of the queries fitting into this , and this saves 10 megabytes (in some projects way more - on omicron it saves 70mb, but this is what it saves on rust-analyzer itself).


MichaReiser:
This sounds worth exploring. Thank you. Can you say how you measured this on r-a? 

These are the numbers for ty

Workload | Salsa memos | Allocations eliminated | Recorded memory saved | Total Salsa memory reduction
-- | -- | -- | -- | --
ty benchmark scripts   | 10,694    | 1,626 | 13,008 bytes | 0.056%
Home Assistant helpers | 171,015   | 19,820 | 158,560 bytes | 0.106%
Full Home Assistant    | 3,730,256 | 411,194 | 3,289,552 bytes | 0.128%


and performance is neutral

Scenario | Baseline | PR | Change
-- | -- | -- | --
Home Assistant helpers, normal CLI, 12 runs | 1.6918 s | 1.6921 s | +0.02%
Full Home Assistant, normal CLI, 6 runs | 5.1126 s | 5.1249 s | +0.24%
Full Home Assistant, mutable database, 6 runs | 5.4867 s | 5.4466 s | −0.73%
Home Assistant LSP, initial diagnostics, 20 runs | 212.35 ms | 212.66 ms | +0.15%
Home Assistant LSP, incremental cross-file edit, 20 runs | 72.73 ms | 72.19 ms | −0.74%


We also need to look into a MIRI issue. We store the, what used to be an allocation, as `NonNull<()>`, but the inline key can be 0 (Codex finding)

```rust
#[test]
fn query_origin_inlines_an_all_zero_packed_edge() {
    let input = QueryEdge::input(key(0, 0, 0));
    let origin = OriginAndExtra::derived(
        [input].into_iter(),
        Default::default(),
    );

    assert_eq!(
        origin.origin().edges().iter().collect::<Vec<_>>(),
        vec![input],
    );
}

#[test]
fn query_origin_drops_an_all_zero_packed_edge() {
    let input = QueryEdge::input(key(0, 0, 0));
    let origin = OriginAndExtra::derived(
        [input].into_iter(),
        Default::default(),
    );

    drop(origin);
}
```


A simple fix here could be to not use inline storage for the one all-zeros key.

ChayimFriedman2:
> Can you say how you measured this on r-a?

`analysis-stats <code>`. This command runs all analysis passes on the code. It is not entirely representative of IDE scenarios but the queries impacted here will be populated in IDE scenarios (most of them).

With the right configuration (pinning to one core and enabling FIFO scheduling) noise is almost zero, few megabytes at most. 10mb is beyond it, and 70mb well beyond.

ChayimFriedman2:
The MIRI error is because we read the `NonNull` even when it's not the active field, validating its validity invariant. Avoiding that will remove it, will push soon.

src/zalsa_local.rs
MichaReiser: I believe this needs updating?

ChayimFriedman2: Updated.

MichaReiser:
I still have to review but I also ran the r-a benchmarks and the results align with what we see in ty. This is mainly an improvement to remove allocations. The memory savings itself are minimal in the grand scheme of things:

Workload | Total memos | Allocations eliminated | Eligible | Salsa metadata saved | Peak RSS saved
-- | -- | -- | -- | -- | --
rust-analyzer | 1,210,110 | 310,656 | 25.67% | 2.37 MiB | 9.50 MiB
Omicron | 4,218,745 | 1,027,933 | 24.37% | 7.84 MiB | 34.87 MiB

Codex:

--- 
For **peak process RSS**:

- **rust-analyzer:** 4,384,704 → 4,374,972 KiB, a **0.22% reduction**.
- **Omicron:** 21,766,924 → 21,731,220 KiB, a **0.16% reduction**.

For **Salsa memo metadata specifically**:

- **rust-analyzer:** 114,046,329 → 111,561,081 bytes, a **2.18% reduction**.
- **Omicron:** 432,259,878 → 424,036,414 bytes, a **1.90% reduction**.

So despite eliminating approximately **25% of memo edge allocations**, the end-to-end process memory improvement is only **0.16–0.22%**.

ChayimFriedman2:
How did you measure memory usage? I'm a bit suspicious of your results since my numbers are not only consistent between different runs, they are also the same as we see in CI (https://rust-analyzer.github.io/metrics/).

That said, even with my results 10mb/70mb for r-a/omicron respectively are not a lot, but it's still an improvement. The code change isn't big and we collect those improvements where we can find them.

MichaReiser:
Codex used `ru_maxrss` to get the peak RSS. 

It used the databases memory usage report for the salsa metadata. Here the same results in a nicer table

**rust-analyzer**

Version | Total Salsa memory | Total metadata | Struct metadata | Memo metadata | Struct data | Memo data
-- | -- | -- | -- | -- | -- | --
Before | 238.01 MiB | 139.97 MiB | 31.21 MiB | 108.76 MiB | 23.00 MiB | 75.04 MiB
After | 235.64 MiB | 137.60 MiB | 31.21 MiB | 106.39 MiB | 23.00 MiB | 75.04 MiB
Change | −1.00% | −1.69% | — | −2.18% | — | —

**Omicron**

Version | Total Salsa memory | Total metadata | Struct metadata | Memo metadata | Struct data | Memo data
-- | -- | -- | -- | -- | -- | --
Before | 852.44 MiB | 528.96 MiB | 116.72 MiB | 412.24 MiB | 82.93 MiB | 240.56 MiB
After | 844.60 MiB | 521.11 MiB | 116.72 MiB | 404.39 MiB | 82.93 MiB | 240.56 MiB
Change | −0.92% | −1.48% | — | −1.90% | — | —


So this does reduce memory usage (1% is not nothing), but I don't think it does in as a significant way as the PR summary suggests

ChayimFriedman2:
But what command did it run and in what environment? If you just run the command as-is and once, the variance can be too big to draw a useful conclusion.

MichaReiser:
The memory report changed rust analyzer, because it doesn't have a built in memory report command, unlike ty. I ran it with concurrency disabled. The output of the memory report should be deterministic (single threaded, unless r-a does non deterministic stuff).

I agree that the peak RSS has variance.

ChayimFriedman2:
Oh you measured peak RSS. I measured memory usage at the end (it's printed by the command as reported by the allocator, note that on Windows the reporting is really not accurate), which represents stable memory usage and we care more about.

But yes I agree this is not a lot. We had way bigger savings, even from Salsa. It's not insignificant though, and we're starting to hitting diminishing returns. If you decide to close this, I'm fine with that, but personally I do think this is worth it.

MichaReiser:
Yes, I measured peak RSS, but I also added something along https://github.com/astral-sh/ruff/blob/2668df9b6a47afb53d0c7e64caa9c2c298ae26cf/crates/ty_project/src/db.rs#L211-L278 to r-a to get more deterministic numbers (including breakdowns by metadata/ingredient). 

> But yes I agree this is not a lot. We had way bigger savings, even from Salsa. It's not insignificant though, and we're starting to hitting diminishing returns. If you decide to close this, I'm fine with that, but personally I do think this is worth it.

Yeah, I've to take a closer look to get a better sense of the complexity and whether it limits future optimizations. Otherwise, a 1% metadata reduction is still nice.

MichaReiser:
I plan to do more Salsa work in a week or two and plan to get back to this then. Let me know if you'd appreciate if I take a look sooner.

ChayimFriedman2:
Thanks! I don't need you to look sooner; this is not urgent, and if you'll not work on Salsa anyway then I won't even have merge conflicts.

src/zalsa_local.rs
```diff
@@ -987,13 +983,15 @@ impl QueryOriginTag {
...
-/// | Derived  | Yes   | `SliceWithHeader<QueryRevisionsExtraInner, E>` | Edge count          |
+/// | Derived  | No    | `SliceWithHeader<(), E>`                        | Edge count          |
+/// | Derived  | Yes   | `SliceWithHeader<QueryRevisionsExtraInner, E>`  | Edge count          |
+/// | Derived  | No    | `PackedQueryEdge`                               | Edge count (1)      |
 ///
 /// `E` is [`PackedQueryEdge`] or [`QueryEdge`], as recorded by [`QueryOriginTag`]. Assigned origins
 /// without extra data require no allocation. Assigned origins with extra data use a regular box.
-/// Derived origins always allocate their edge slice, placing the extra data in the same allocation
-/// as the edges when it exists.
+/// Derived origins without extra data and with len<=1 do not allocate, storing a danging pointer for
```
ada4a, line 992:
```suggestion
/// Derived origins without extra data and with len<=1 do not allocate, storing a dangling pointer for
```

ChayimFriedman2:
The Miri failures are in no way relevant. They are probably because of the new Rust version.

PR #1295, opened 2026-08-17
fix: filter attributes on tracked query helpers

MichaReiser:
## Summary

`must_use` failed to compile on associated methods because it isn't a valid attribute on trait methods in an impl block. 

I don't see a clear solution to this. I considered:

Lower the associated method to a method with an inner Salsa query:

```rust
impl Type {
     /// Apply the load rule.
     ///
     /// Reference: <https://www.w3.org/TR/WGSL/#load-rule>
     #[must_use]
     pub fn is_constructible(
         self,
        db: &dyn Db,
     ) {
        #[salsa::tracked(cycle_result = |_, _, _| false)]
        fn is_constructible_inner(db: &dyn Db, this: Type) -> something { ... }
     is_cunstructible_inner(db, this)
}
```

This requires rewriting `self` to `this` in the method body, which can be error prone. This is the approach used by `asyinc-trait`. `asyinc-trait` had multiple issues where their rewriting was incorrect, but it seems they now reached a pretty good spot?

The alternative, which I've taken in this PR, is to special case known attributes. Attributes that define the queries API remain on the query definition, attributes that are specific to the implementation (e.g. `allow`) are moved to the inner function. This approach is inspired by [`mock_all`](https://crates.io/crates/mockall_derive). The obvious downside is that it only works for known attributes. 

I went with the filter approach for now. We can always migrate to the `async-trait` approach if custom attributes are an issue. 

This also handles nested `cfg_attr` without dropping other attributes.


Veykril:
wait where is a trait impl relevant here?

MichaReiser:
For 

```rust
#[salsa::tracked]
impl Input {
    #[must_use]
    #[salsa::tracked(returns(copy))]
    fn doubled(self, db: &dyn salsa::Database) -> u32 {
        self.value(db) * 2
    }
}
```


Salsa generates (before the fix)


```rust
impl Input {
    #[doc = " Return twice the input value."]
    #[must_use]
    fn doubled(self, db: &dyn salsa::Database) -> u32 {
        trait InnerTrait_<'db> {
            fn doubled_(self, db: &'db dyn salsa::Database) -> u32;
        }

        impl<'db> InnerTrait_<'db> for Input {
            #[doc = " Return twice the input value."]
            #[must_use]
            fn doubled_(self, db: &'db dyn salsa::Database) -> u32 {
                self.value(db) * 2
            }
        }

        // Generated cache code omitted.

        doubled_(db, self)
    }
}
```

But the `must_use` on the `InnerTrait_::doubled_` is not allowed

Veykril:
Ah okay now I follow. Yea, rewriting I think would be nicer overall but this is probably good enough (given as you said rewriting also has issues).

Veykril (approved):

PR #1301, opened 2026-08-26
perf: Allow customizing the order of memos to save memory usage

ChayimFriedman2:
The order is inferred based on previous runs, like in PGO.

In rust-analyzer this saves between 16mb and 100mb on the projects I measured. The setup goes like that: the top level crate, `rust-analyzer`, gains a new feature `precomputed-memo-frequencies`. I also included the following code:
```rust
#[macro_export]
macro_rules! precomputed_memo_frequencies_path {
    () => {
        concat!(env!("CARGO_MANIFEST_DIR"), "/../../memo_frequencies.bin")
    };
}

#[cfg(feature = "precomputed-memo-frequencies")]
pub fn memo_frequencies() -> MemoFrequencyStats {
    static MEMO_FREQUENCIES: std::sync::LazyLock<MemoFrequencyStats> =
        std::sync::LazyLock::new(|| {
            let data = include_bytes!(crate::precomputed_memo_frequencies_path!());
            MemoFrequencyStats::deserialize(data)
        });
    MEMO_FREQUENCIES.clone()
}

#[cfg(not(feature = "precomputed-memo-frequencies"))]
pub fn memo_frequencies() -> MemoFrequencyStats {
    MemoFrequencyStats::default()
}
```
Then when creating a database I pass `memo_frequencies()`.

I also included a flag in the CLI command that analyzes a project (`analysis-stats`). When the flag is active it does:
```rust
if let Some(optimal_init_percent) = self.write_memo_frequencies {
    let memo_freq = MemoFrequencyStats::determine_from_db(db, optimal_init_percent / 100.);
    let memo_freq = memo_freq.serialize();
    let path = crate::precomputed_memo_frequencies_path!();
    std::fs::write(path, memo_freq).context("failed to write memo frequency data")?;
    eprintln!(
        "Wrote memo frequencies into `{}`.",
        std::fs::canonicalize(path).unwrap_or_else(|_| PathBuf::from(path)).display(),
    );
}
```

Then, I first compile without the feature, and run `analysis-stats -- write-memo-frequencies <fraction>` on some project. `<fraction>` is determined experimentally, as a compromise between speed (avoiding excess reallocations) and memory usage; I found 90% (0.9) to be a good balance.

After the run finishes, I re-compile with `--features rust-analyzer/precomputed-memo-frequencies`, and now I have an optimized binary.

I'll be very interested to see measurements on `ty` as well; in addition, while this was fun to develop I'm not entirely sure this is worth it - this introduces complex lock-free unsafe code and slows down memo insertion (retrieval is just as fast).


ChayimFriedman2:
The benchmarks are not really effective because they don't precompute the stats; I'll try to find how to do that.

MichaReiser:
This sounds similar in the approach to https://github.com/salsa-rs/salsa/pull/1183. It allows customizing the page size per query. rust analyzer could then "PGO" the page sizes. 

I have codex run this on ty to better understand the overhead for projects not using PGO and the relative memory saving (in % of total data, total salsa meta data)

ChayimFriedman2:
Similar in spirit, maybe (and having automatic PGO there too will be good, we can even combine the methods) but here this is for memo tables, not pages.

MichaReiser:
Performance measurements on ty

  | 1 thread | 4 threads | 8 threads | 32 threads
-- | -- | -- | -- | --
Before | 42.69 s | 11.50 s | 6.38 s | 4.10 s
After, no PGO | 43.23 s | 11.48 s | 6.46 s | 4.21 s
Throughput change | −1.25% | +0.19% | −1.27% | −2.53%
After, PGO | 43.19 s | 11.40 s | 6.37 s | 4.07 s
Throughput change | −1.17% | +0.86% | +0.15% | +0.68%

I think a PGO like feature should be perf neutral for projects not using it.

ChayimFriedman2:
I was interested in memory, not speed.

> I think a PGO like feature should be perf neutral for projects not using it.

Generally I agree, however here it might complicate an already-complicated unsafe code.

MichaReiser:
The memory report was still in the making ;)

--- 

<html><head></head><body><p><strong>Yes—with one qualification:</strong> baseline → PR without PGO measures <strong>all non-PGO changes in the PR</strong>, including the representation and allocation/growth strategy, not representation alone.</p><p>For the matched single-thread Home Assistant run:</p><div><div><div>
Comparison | Accounted memory saved
-- | --
Baseline → PR without PGO | 13.35 MiB
PR without PGO → PR with PGO | 47.41 MiB additional
Baseline → PR with PGO | 60.76 MiB total

</div></div></div><p>So approximately <strong>22% of the combined saving comes without profiling</strong>, and <strong>78% comes from using the PGO information</strong> for this workload.</p><p>These are corrected Salsa accounting figures, not RSS differences.</p></body></html>

--- 

It totals around 7% in metadata savings, which is a fair amount. But this is when using the PGO data for the same project. I have to do another run that uses one shared training data.

Would it make sense to split the PR into a) change the representation and growth strategy and b) the PGO stuff?

ChayimFriedman2:
> Would it make sense to split the PR into a) change the representation and growth strategy and b) the PGO stuff?

It should be fairly easy. Does it make sense? I don't know, the code parts are totally separate and there's little value to one without the other. But I can if you want.

PR #1304, opened 2026-08-28
Allow retuning a tracked fn's LRU capacity by ingredient name

enomado:
The generated `set_lru_capacity` is private to the module that declares the tracked function, and for an *associated* tracked function the macro emits it inside the body of the outer function, where nothing can reach it at all. That leaves no way to retune the queries that actually hold memory in a long-lived process — and it is why rust-analyzer's `update_lru_capacities` has been dead since the salsa transition.

Ingredients are reachable through the registry and `debug_name` gives them addresses, so this adds the missing route there:

- `Ingredient::set_lru_capacity` / `has_tunable_lru_capacity`
- `EvictionPolicy::has_tunable_capacity` (`Lru` overrides it, `Noop` keeps the default)
- `Zalsa::set_lru_capacity_by_name` / `lru_ingredient_names`
- `Database::set_lru_capacity_by_name` / `lru_capacity_names`

`set_capacity` alone could not carry this: it is a no-op on `NoopEviction`, so a caller walking the registry could not tell "capacity applied" from "this query has no LRU". Hence the explicit tunability question, and a **count** of retuned ingredients rather than a `bool` — names are not unique across crates.

### Tests

Two, in `tests/lru.rs`.

The first is a positive control: retuning by name must move the same memos the direct setter moves, checked by live-value count across an eviction — including that the floor is `1`, because `0` disables eviction rather than shrinking the cache to nothing.

The second pins that a typo and an uncapped query both answer zero rather than reading as success.

Verified by mutation: dropping the `set_capacity` call while still returning `true` fails the first test at its positive control and leaves the four pre-existing tests green, so nothing else covers this path.

### Context

Written while chasing a rust-analyzer-based daemon that grew to 14.6 GB. With this in place the capacities can be addressed by `debug_name` from outside the declaring crate, which is what revives `rust-analyzer.lruCapacity` and `lruQueryCapacities`.


src/database.rs
```diff
@@ -47,6 +47,35 @@ pub trait Database: Send + ZalsaDatabase + AsDynDatabase {
...
+    /// caller's configuration, not as a successful no-op.
+    ///
+    /// Note that `capacity == 0` DISABLES eviction for the matched queries; the
+    /// smallest cache that still evicts is `1`.
+    ///
+    /// **WARNING:** Just like an ordinary write, this method triggers
+    /// cancellation. If you invoke it while a snapshot exists, it
+    /// will block until that snapshot is dropped -- if that snapshot
+    /// is owned by the current thread, this could trigger deadlock.
+    fn set_lru_capacity_by_name(&mut self, name: &str, capacity: usize) -> usize {
```
MichaReiser, line 66: I don't think we should use `name` as a stable identifier. We expose name as a debug tool, there's also no guarantee that names are unique.

tests/lru.rs
```diff
@@ -117,6 +117,74 @@ fn lru_can_be_changed_at_runtime() {
     assert_eq!(load_n_potatoes(), 0);
 }
 
+/// The by-name route exists because the generated `set_lru_capacity` is private
+/// to the declaring module, and for an *associated* tracked fn the macro emits it
+/// inside the body of the outer function, where nothing can reach it. Callers
+/// that need to retune such a query have only the ingredient registry.
```
MichaReiser, line 123: I think this is an application problem. If you want to make it public, expose a public function in that module.

src/database.rs
enomado: Agreed — I'll drop the by-name API.

Names are only needed at one boundary: the user-facing config. `rust-analyzer.lruQueryCapacities` is a `FxHashMap<Box<str>, u16>` that a person types, so resolving that string to a query is the application's job, where the names *are* unique and where the key set is documented. What salsa owes is a reachable, typed handle per query; the `match name { .. }` belongs in rust-analyzer, not here.

The only reason I reached for names is the thing in the other thread: for an associated tracked function there is no typed handle to reach, at any visibility. Fix that and this lookup has no reason to exist — I'd rather delete it than argue for it.

tests/lru.rs
enomado: It would be, if there were something to wrap.

`#[salsa::tracked] impl` expands `setup_tracked_assoc_fn_body!` / `setup_tracked_method_body!` **into the method's body**, and those macros emit `$tracked_fn` inside that block. So the generated `$fn_name` struct — and with it `set_lru_capacity`, `specify` and `accumulated` — is an item local to a function body. Nothing in the declaring module can name it, so a `pub fn` wrapper has no callee to forward to.

That is not a corner case. Every LRU-capped query in rust-analyzer today (`70d74f4d13`):

| query | lru | form |
| --- | --- | --- |
| `EditionedFileId::parse` | 128 | method in `#[salsa::tracked] impl` |
| `MacroCallId::parse_macro_expansion` | 512 | method in `#[salsa::tracked] impl` |
| `HirFileId::ast_id_map` | 1024 | method in `#[salsa::tracked] impl` |
| `Body::with_source_map` | 512 | assoc fn in `#[salsa::tracked] impl` |
| `AttrFlags::docs` | 250 | assoc fn in `#[salsa::tracked] impl` |
| `AttrFlags::fields_docs` | 50 | assoc fn in `#[salsa::tracked] impl` |
| `cfgs` | 250 | free fn, declared inside a fn body |
| `fields_cfgs` | 50 | free fn, declared inside a fn body |

Zero module-level free tracked fns. The last two are genuinely an application problem — r-a can hoist them itself. The other six cannot be reached by any application-side change.

And the control that shows this is not really about names: `RootDatabase::update_base_query_lru_capacities` is dead too, and it takes a single `Option<u16>` with no names anywhere. `RA_LRU_CAP` / `--lru-capacity` currently applies to nothing at all.

salsa 0.17 had exactly the typed handle you're asking for, and it is still sitting in r-a's commented-out code:

```rust
// base_db::ParseQuery.in_db_mut(self).set_lru_capacity(lru_capacity);
```

So the change I'd rather make is to restore that — a typed, reachable per-query handle for associated tracked fns — and drop the by-name API entirely. Two smaller things fall out of it:

- the generated `set_lru_capacity` is `fn`, not `pub fn`, unlike `accumulated` and `specify` right next to it in the same `impl`, so even for a free tracked fn the wrapper has to live in the exact declaring module;
- hoisting the generated items out of the method body needs the hygiene ident to include the self type, otherwise two impls with a same-named method in one module collide.

Would you take that shape? Happy to rewrite the PR that way.

PR #1306, opened 2026-09-01
Allow create self-referential interneds without replicating the macro's work

ChayimFriedman2:
The main change is that now `Lookup` takes a `salsa::Id` representing the `Id` for the to-be-created struct. But for that to work, I also made some other changes:

 - The order of `HashEqLike` was reversed. Previously it was `<Owned as HashEqLike<Borrowed>>`, now it's `<Borrowed as HashEqLike<Owned>>`, matching `Lookup`.
 - It's now possible to put `#[no_eq]` on an interned field to exempt them from hashing & equality checks (`#[no_eq]` on tracked structs doesn't change hashing. I think this is a bug we need to fix, but didn't touch it).
 - That in turns mean that we can no longer use a tuple for the fields and have to use a custom struct.

This is a breaking change because of the changes to `Lookup` and `HashEqLike`.

### DIFF
--- a/components/salsa-macro-rules/src/setup_interned_struct.rs
+++ b/components/salsa-macro-rules/src/setup_interned_struct.rs
@@ -54,6 +54,9 @@ macro_rules! setup_interned_struct {
         // Indices for each field from 0..N -- must be unsuffixed (e.g., `0`, `1`).
         field_indices: [$($field_index:tt),*],
 
+        // Indices for fields that participate in `Hash`/`Eq` impls.
+        hash_eq_field_indices: [$($hash_eq_field_index:tt),*],
+
         // Indexed types for each field (T0, T1, ...)
         field_indexed_tys: [$($indexed_ty:ident),*],
 
@@ -96,7 +99,7 @@ macro_rules! setup_interned_struct {
         #[derive(Copy, Clone, PartialEq, Eq, Hash)]
         $vis struct $Struct< $($db_lt_arg)? >(
             $Id,
-            std::marker::PhantomData<fn() -> &$interior_lt ()>
+            ::std::marker::PhantomData<fn() -> &$interior_lt ()>
         );
 
         #[allow(clippy::all)]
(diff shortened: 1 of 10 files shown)


MichaReiser:
Can you say more about how this is different to https://github.com/salsa-rs/salsa/pull/1260 and how it addresses my concerns raised on that PR? I remain concerned about excluding certain fields from the interned lookup, and this is already something that can be done in application code by using a newtype that overrides Eq

ChayimFriedman2:
Oh I didn't remember that PR. It's similar in the goal but has quite different (and IMO simpler, as the diff will tell) implementation, reusing the already-existing `HashEqLike` with minimal changes.

Excluding a field indeed can already be done but it is more convenient this way. The problem is not the exclusion but the self-ref which cannot be done today. And yes this is risky (and probably needs some warning in the docs), but this is essentially what we do currently in r-a, just a lot more clumsily and via internal APIs. So I think there is a real need and we should provide an official way to do that.

MichaReiser:
> Excluding a field indeed can already be done but it is more convenient this way. The problem is not the exclusion but the self-ref which cannot be done today. 

I'm not sure this is true. You can have a `Option<Self>` field where `None` means -> `Self`. I think what isn't possible today is to later change an interned field. That is, Salsa forces you today to make the self-referential or other-instance decision before creating the interned value. And I think this is correct and necessary or you risk interning `self` and the `non-self` variant to the same value

ChayimFriedman2:
Yes you can, yet we don't do that for the reasons @Veykril mentioned in that PR. It isn't possible to create a truly self-ref (where the stored data references the ID) interned today with public API only. And yes this is risky, but like I said we're already taking the risk (re-interning of the same data with different non-compared data is not possible for us in this case), and therefore I think Salsa should provide this option.

MichaReiser:
> It isn't possible to create a truly self-ref (where the stored data references the ID) interned today with public API only. 

It's still not clear why you want that. Create a private field that's an `Option` and a public method. The cost is the same - One field read. The only difference is a possible branch miss prediction. But, given the cost of reading an interned field, I'm not sure how noticeable the difference really is

ChayimFriedman2:
Okay I went with using `Option` in r-a.

PR #1314, opened 2026-09-11
Remove assemble support from interneds

Veykril:
Now that r-a doesn't need this anymore we can simplify this. cc https://github.com/salsa-rs/salsa/pull/1306 github.com/rust-lang/rust-analyzer/pull/23308


.github/workflows/test.yml
```diff
@@ -79,8 +79,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Install Miri
-        uses: dtolnay/rust-toolchain@miri
+        uses: dtolnay/rust-toolchain@master
         id: rust-toolchain
+        with:
+          toolchain: nightly-2026-09-09
```
Veykril, line 85: mirir reports invalid variadics now which causes it to error on parking-lot-core

MichaReiser (approved):

